PR #6, opened 2017-11-29
Fix typo in IAM policy

willglynn:
The README specifies `{"Action": "ec2.ModifyNetworkIntefaceAttribute"}`. This should be `ec2:` instead of `ec2.`, and it should be `Interface` instead of `Inteface`.


liwenwu-amazon:
Thank you

PR #10, opened 2017-12-04
Add preview status to readme and also update version

liwenwu-amazon:


README.md
```diff
@@ -1,6 +1,10 @@
 # amazon-vpc-cni-k8s
 Networking plugin for pod networking in [Kubernetes](https://kubernetes.io/) using [Elastic Network Interfaces](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html) on AWS.
 
+## note
+
+* STATUS: Preview. Do not use for production workloads while in preview
```
dbenhur: Make Status the bold heading.  Let's use more canonical status terms than "Preview".

```
## Status

 **Alpha**  This is an experimental release as part of the [Amazon EKS](https://aws.amazon.com/eks/) Preview.
Interfaces and functionality may change. Expect bugs (and please help us squash them).
DO NOT use for production workloads.
```

main.go
```diff
@@ -24,7 +24,7 @@ import (
 
 const (
 	defaultLogFilePath = "/host/var/log/aws-routed-eni/ipamd.log"
-	version            = "0.1"
+	version            = "1.0.1"
```
dbenhur: We shouldn't say 1.y.z while still in pre-production state; this should be 0.1.1 (patch change from prior).

From [Semver standard](https://semver.org/):
4. Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.
5. Version 1.0.0 defines the public API. The way in which the version number is incremented after this release is dependent on this public API and how it changes.

PR #17, opened 2017-12-13
Adjust veth name generation to be compatible with Calico

caseydavenport:
This PR adjusts the calculation used to generate the host-side veth name such that it aligns with the value expected by Calico.

[See here](https://github.com/projectcalico/libcalico-go/blob/v1.7.3/lib/backend/k8s/conversion.go#L43-L51) for the corresponding calculation in Calico.

This is the groundwork for allowing Calico to be installed as a DaemonSet for network policy enforcement.

Posting this to get feedback as to the approach.

### DIFF
--- a/plugins/routed-eni/cni.go
+++ b/plugins/routed-eni/cni.go
@@ -14,6 +14,8 @@
 package main
 
 import (
+	"crypto/sha1"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"golang.org/x/net/context"
@@ -152,13 +154,9 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 	}
 
 	// build hostVethName
-	// hostVethName := "eni" + args.ContainerID[:min(11, len(args.ContainerID))]
+	// hostVethName := "cali" + sha1(namespace.name)[:11]
 	// Note: the maximum length for linux interface name is 15
-	length := len(args.ContainerID)
-	if length > maxVethNameLen {
-		length = maxVethNameLen
-	}
-	hostVethName := "eni" + args.ContainerID[:length]
+	hostVethName := generateHostVethName(string(k8sArgs.K8S_POD_NAMESPACE), string(k8sArgs.K8S_POD_NAME))
 
 	err = driverClient.SetupNS(hostVethName, args.IfName, args.Netns, addr, int(r.DeviceNumber))
 
@@ -180,6 +178,15 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 	}
 
 	return cniTypes.PrintResult(result, cniVersion)
+}
+
+// generateHostVethName returns a name to be used on the host-side veth device.
+// The veth name is generated such that it aligns with the value expected
+// by Calico for NetworkPolicy enforcement.
+func generateHostVethName(namespace, podname string) string {
+	h := sha1.New()
+	h.Write([]byte(fmt.Sprintf("%s.%s", namespace, podname)))
+	return fmt.Sprintf("cali%s", hex.EncodeToString(h.Sum(nil))[:11])
 }
 
 func cmdDel(args *skel.CmdArgs) error {
--- a/plugins/routed-eni/cni.go
+++ b/plugins/routed-eni/cni.go
@@ -57,6 +57,11 @@ type NetConf struct {
 
 	// Type is the plugin type
 	Type string `json:"type"`
+
+	// VethPrefix is the prefix to use when constructing the host-side
+	// veth device name. It should be no more than four characters, and
+	// defaults to 'eni'.
+	VethPrefix string `json:"vethPrefix"`
 }
 
 // K8sArgs is the valid CNI_ARGS used for Kubernetes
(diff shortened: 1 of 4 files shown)


plugins/routed-eni/cni.go
```diff
@@ -182,6 +180,15 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 	return cniTypes.PrintResult(result, cniVersion)
 }
 
+// generateHostVethName returns a name to be used on the host-side veth device.
+// The veth name is generated such that it aligns with the value expected
+// by Calico for NetworkPolicy enforcement.
+func generateHostVethName(namespace, podname string) string {
+	h := sha1.New()
+	h.Write([]byte(fmt.Sprintf("%s.%s", namespace, podname)))
+	return fmt.Sprintf("cali%s", hex.EncodeToString(h.Sum(nil))[:11])
```
liwenwu-amazon: I think "eni" is better prefix than "cali" because  this is a virtual interface over  AWS VPC ENI (Elastic Network interface).

liwenwu-amazon: From the comment, it looks like you depend on the "cali" prefix to discover interfaces for Network Policy Enforcement.  Is there another mechanism calico could use (configuration or API) to do this discovery that's not dependent on a convention that requires us to name this in a misrepresentative way.

plugins/routed-eni/cni.go
```diff
@@ -182,6 +180,15 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 	return cniTypes.PrintResult(result, cniVersion)
 }
 
+// generateHostVethName returns a name to be used on the host-side veth device.
+// The veth name is generated such that it aligns with the value expected
+// by Calico for NetworkPolicy enforcement.
+func generateHostVethName(namespace, podname string) string {
+	h := sha1.New()
+	h.Write([]byte(fmt.Sprintf("%s.%s", namespace, podname)))
```
liwenwu-amazon: It looks like you're embedding metadata about the pod into the interface name here.  If this is purely to assure name uniqueness, this is fine, but if Calico relies on this data for mapping Network Policy, we should have a stronger contract.  At minimum, documentation about the data embedding requirement and constraints here; but preferably an API to map interface to k8s metadata.

plugins/routed-eni/cni.go
```diff
@@ -152,13 +154,9 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 	}
 
 	// build hostVethName
-	// hostVethName := "eni" + args.ContainerID[:min(11, len(args.ContainerID))]
+	// hostVethName := "cali" + sha1(namespace.name)[:11]
```
liwenwu-amazon: let's just drop this comment

plugins/routed-eni/cni.go
eswarbala: Yes - Ideally the prefix defaults to eni and can be overridden by a config flag/env variable. We can have a seperate cni daemon set file that carries the "cali" prefix as a flag/env.

caseydavenport: I think using `cali` as the prefix provides a number of benefits, namely:

- This is the standard naming convention used across all other Calico deployments, and so it's what users expect.
- Similarly, existing documentation, guides and troubleshooting documentation reference the `cali` prefix, so choosing this prefix will allow all of that material to "just work".

There's another benefit of doing this by default, which is that it allows users to enable Calico for NetworkPolicy at any point in time, not just at cluster creation. This is really valuable for users who might create a cluster without the feature enabled, and then realize they need it later on.

lstoll: This seems to be the inverse of what I'd expect - I don't think of Calico as the defining standard for networking in Kubernetes, but rather one of many options. If Calico is setting out to be an generic add-on to enforce NetworkPolicy on any underlying CNI network, would it not make more sense for it to handle the varying cases that underlying CNI networks might present better? This is something we also faced with implementing NetworkPolicy for our AWS/CNI arrangement. Similar goes for mapping pods to interfaces.

lilida: It doesn't make sense to me to put cali in AWS/CNI. I may run the plugin in AWS without Calico involved but have a different implementation for network policy.

plugins/routed-eni/cni.go
lilida: Are we going to force Calico with this AWS plugin? I hope we are on the direction that different network policy implementation can work with this plugin.

eswarbala: @lilida We definitely want to support multiple deployment scenarios.

plugins/routed-eni/cni.go
eswarbala: @caseydavenport The first two callouts would still stand even with the recommended change using "cali" as the prefix flag/env. Like others have chimed in, its important that this CNI plugin works in a standalone environment for which it doesn't make sense to have "cali" as prefix. 

I understand this would require customers to decide at launch time. That is solvable with Calico supporting a configurable prefix as well.

liwenwu-amazon:
Can you also bump version to "0.1.2" in main.go

gunjan5:
since @caseydavenport is on vacation, I've pulled his commits and applied review feedback in https://github.com/aws/amazon-vpc-cni-k8s/pull/20

PR #20, opened 2018-01-04
Adjust veth name generation to be compatible with Calico

gunjan5:
Pulling in commits from https://github.com/aws/amazon-vpc-cni-k8s/pull/17 and applying review feedback to it since @caseydavenport is on vacation.

CC @liwenwu-amazon @eswarbala @lilida @lstoll

Original description:
This PR adjusts the calculation used to generate the host-side veth name such that it aligns with the value expected by Calico.

See [here](https://github.com/projectcalico/libcalico-go/blob/v1.7.3/lib/backend/k8s/conversion.go#L43-L51) for the corresponding calculation in Calico.

This is the groundwork for allowing Calico to be installed as a DaemonSet for network policy enforcement.

Posting this to get feedback as to the approach

### DIFF
--- a/main.go
+++ b/main.go
@@ -24,7 +24,7 @@ import (
 
 const (
 	defaultLogFilePath = "/host/var/log/aws-routed-eni/ipamd.log"
-	version            = "0.1.1"
+	version            = "0.1.2"
 )
 
 func main() {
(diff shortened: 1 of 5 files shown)


liwenwu-amazon:
Thank you and I have also uploaded a new docker image:  602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:0.1.2 which contains this PR.

Issue #18, opened 2017-12-20
Controller is not aware of the number of available IP addresses


arun-gupta:
Controller is not aware of how many IP addresses are available to be assigned to pods. It tries to assign an IP address for the new pod and then fails reactively. It should have this information proactively.

Created a 2 node `t2.medium` cluster as:

```
kops create cluster \
--name example.cluster.k8s.local \
--zones us-east-1a,us-east-1b,us-east-1c \
--networking amazon-vpc-routed-eni \
--node-size t2.medium
--kubernetes-version 1.8.4 \
--yes
```

Created a Deployment using the configuration file:

```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  replicas: 3 
  template:
    metadata:
      labels:
        app: nginx 
    spec:
      containers:
      - name: nginx 
        image: nginx:1.12.1
        ports: 
        - containerPort: 80
        - containerPort: 443
```

Scaled the replicas:

```
kubectl scale --replicas=30 deployment/nginx-deployment
```
30 pods are expected in the cluster (2 * 3 * 5) but only 27 pods are available. Three pods are always in `ContainerCreating` state.

A similar cluster was created with `m4.2xlarge`. 120 pods (2 * 4 * 15) are expected in the cluster, but only 109 pods are available.

More details about one of the pods that is not getting scheduled:

```
$ kubectl describe pod/nginx-deployment-745df977f7-tndc7
Name:           nginx-deployment-745df977f7-tndc7
Namespace:      default
Node:           ip-172-20-68-116.ec2.internal/172.20.95.9
Start Time:     Wed, 20 Dec 2017 14:32:53 -0800
Labels:         app=nginx
                pod-template-hash=3018953393
Annotations:    kubernetes.io/created-by={"kind":"SerializedReference","apiVersion":"v1","reference":{"kind":"ReplicaSet","namespace":"default","name":"nginx-deployment-745df977f7","uid":"96335f7f-e5d5-11e7-bc3a-0a41...
                kubernetes.io/limit-ranger=LimitRanger plugin set: cpu request for container nginx
Status:         Pending
IP:             
Created By:     ReplicaSet/nginx-deployment-745df977f7
Controlled By:  ReplicaSet/nginx-deployment-745df977f7
Containers:
  nginx:
    Container ID:   
    Image:          nginx:1.12.1
    Image ID:       
    Ports:          80/TCP, 443/TCP
[...]

PR #23, opened 2018-01-15
Enable controller to be aware of number of available IP addresses

liwenwu-amazon:
This PR addresses issue #18 .

It defines **eks.amazon.com/ip** as a [kubernetes' extended resources](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#extended-resources) to represent the number of available of available IP addresses on each node.


mattlandis:
This plugin is usable outside of an eks cluster so the resource should not be scoped to eks.  Maybe vpc-cni.amazon.com/ip or vpc-cni.aws.amazon.com/ip?

aaithal:
I second @mattlandis  comment here. Also, if you have 3611 files in a single commit, it's impossible for anyone to review it effectively.

ofiliz:
I am not sure if this resource would ever make sense outside of EKS. If we name it generically, we risk confusing people who do not need it into thinking they do. This resource exists only because a specific implementation detail of vpc-cni-k8s plugin requires it. Other CNI plugins (at least as of today) do not need it.

Also, the name of the resource should be something like "private-ipv4-address". "ip" is the name of the protocol, and even if it is commonly (and incorrectly) understood to be an address, let's be clear on what type of address it really is.

bmoylan:
We deploy a custom-configured Kubernetes cluster on EC2 and would love to have this feature generically available instead of tied directly to EKS.

liwenwu-amazon:
how about `vpc.amazonaws.com/ip`?

aaithal:
how about `vpc.amazonaws.com/ipv4` to state explicitly that this is for ipv4 addresses?

Gopkg.toml
```diff
@@ -21,13 +21,17 @@
 #  version = "2.4.0"
 
 
+[[constraint]]
+  name = "github.com/aws/amazon-ecs-agent"
```
samuelkarp: Why did you vendor the ECS agent?

Gopkg.lock
```diff
@@ -16,8 +16,8 @@
 [[projects]]
   name = "github.com/aws/amazon-ecs-agent"
   packages = ["agent/utils/ttime","agent/utils/ttime/mocks"]
-  revision = "ea58fddf8462034ff2034aa9788b02341d7702e3"
-  version = "v1.15.2"
+  revision = "ea383f2ff70290029ee69c7089160b10941baeb6"
+  version = "v1.16.1"
 
 [[projects]]
   branch = "master"
```
samuelkarp: Why did you vendor github.com/aws/amazon-ecs-cni-plugins?

Gopkg.toml
liwenwu-amazon: Right now, it uses some code from amazon-ecs-agent and amazon-ecs-cni-plugins. We are planning to move these shared code out into another library repo and only vendor this new library repo.

samuelkarp: What code are you using?  It's worth evaluating whether to factor it out or to just stop using the code; not everything in the ECS agent is code that we'd recommend reusing.

liwenwu-amazon: opened issue #27 to track this.   Thanks

liwenwu-amazon:
Move this to a new PR #31  where it is broken down into multiple commit

PR #28, opened 2018-02-02
Reconcile eni and ip addresses info from instance's metadata service

liwenwu-amazon:
### Summary

* query instance's metadata service and get up-to-date eni and IP addresses information
* reconcile this with local cache

### Testing Done

```
make unit-test
make lint
make vet
```


liwenwu-amazon:
wrong branch

Issue #26, opened 2018-01-29
Kube-Scheduler Support for managing node's available IP addresses


liwenwu-amazon:
# Kube-Scheduler Support for managing node's available IPv4 addresses

## Problem

As described in issue [#18](https://github.com/aws/amazon-vpc-cni-k8s/issues/18), kube-scheduler is not aware of the number of available IPv4 addresses on a node.  Kube-Scheduler can schedule a Pod to run on a node even after the node has exhausted all of its IPv4 addresses.

## Proposal

Here we propose to use [kubernete's extended resources](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#extended-resources) (which is supported in Kubernetes 1.8) to enable kube-scheduler support for managing node's available IPv4 addresses

* We will define a **vpc.amazonaws.com/ipv4** as a node level extended resource
* A pod which is not using **hostNetwork** mode needs to specify following in one of its container:

```
apiVersion: v1
  kind: Pod
  metadata:
    name: my-pod
spec:
  containers:
  - name: my-container
    image: myimage
    resource:
      requests:
        vpc.amazonaws.com/ipv4: 1
      limits:
        vpc.amazonaws.com/ipv4: 1
```

## Solution 

**eni-ip-controller** is a new component which runs in a kubernetes cluster. It watches kubernetes **node** resources. When a new node joins the cluster, **eni-ip-controller** updates API server about node's **vpc.amazonaws.com/ipv4** resource (the number of available IPv4 addresses on the new node). Here is the workflow: 

![eni-ip-jan30](https://user-images.githubusercontent.com/24555154/35600558-5fd48c3a-05e3-11e8-9f55-252e336581d6.png)

Here is an HTTP request that advertises 15 "vpc.amazonaws.com/ipv4" resources on node **k8s-node-1**
```
curl --header "Content-Type: application/json-patch+json" \
--request PATCH \
--data '[{"op": "add", "path": "/status/capacity/vpc.amazonaws.com~1ipv4", "value": "15"}]' \
http://k8s-master:8080/api/v1/nodes/k8s-node-1/status 
```
###  Pod Admission Control

Since "**vpc.amazonaws.com/ipv4**" is NOT a standard resource like "**cpu**" or "**memory**", if a Pod does NOT have "**vpc.amazonaws.com/ipv4**" specified, the Pod can consume an ENI IPv4 resource on a node without accounted in the scheduler.

Here are few options to solve this:

[...]

PR #31, opened 2018-02-02
Enable controller to be aware of number of available IP addresses

liwenwu-amazon:
This PR addresses issue #18 .

It defines vpc.amazonaws.com/ipv4 as a [kubernetes's extended resources](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#extended-resources) to represent the number of available IP addresses on each node.

See #26 for more info


Makefile
```diff
@@ -49,10 +53,14 @@ lint:
...
+	golint controller/*.go
+	golint controller/*/*.go
 
 #go tool vet
 vet:
 	go tool vet ./pkg/awsutils
 	go tool vet ./plugins/routed-eni
 	go tool vet ./pkg/k8sapi
 	go tool vet ./pkg/networkutils
+	go tool vet controller
```
mattlandis: I don't think this is working.  `go vet tool` works on files.  `go vet` works on packages.

controller/eni_ip_controller.go
```diff
@@ -0,0 +1,267 @@
...
+	informer   cache.Controller
+	indexer    cache.Indexer
+	nodeName   string
+	// nodeStoreSynced returns true if the node store has been synced at least once.
+	// Added as a member to the struct to allow injection for testing.
+	nodeStoreSynced cache.InformerSynced
+}
+
+// NewENIIPController creates ENIIPController
+func NewENIIPController(client kubernetes.Interface, vpcIP vpcipresource.Interface, nodeName string) *ENIIPController {
```
mattlandis: Take a look at your coverage.  There is a lot of code in this file not covered by test.

SaranBalaji90:
This PR has amazon-ecs-agent code and some irrelevant aws packages imported - can you do something similiar to  https://github.com/aws/amazon-vpc-cni-k8s/pull/43 and send out new review?

controller/eni_ip_controller.go
```diff
@@ -0,0 +1,267 @@
...
+
+	ipLimit, err := vpcipresource.GetVPICIPLimit(instanceType)
+
+	if err != nil {
+		log.Errorf("Failed to get vpc ip limit %v", err)
+		//TODO handle err
+		return nil
+	}
+
+	if strings.Compare(key, c.nodeName) == 0 {
```
nckturner: This seems wrong.  If it is a running tally of IPs, then we should be doing it for all pods that are already running on the node.  But its called IP limit, so that means to me it should be a limit, and should have nothing to do with any pods currently running on the instance.

incognick:
Any progress on this?  Definitely need this incorporated so the scheduler is aware of IPs as a resource. I keep running into nodes with not enough available IPs and the scheduler does not try to move the pod to a node with available IPs.

incognick:
@liwenwu-amazon ^^

controller/eni_ip_controller.go
liwenwu-amazon: This assumes eni-controller pod is deployed before any other workload Pods and all other Pods have specified using resource of ***vpc.amazonaws.com/ipv4*** resources.

barambani:
+1

umatrangolo:
Any update about this? 

We have serious problems trying to setup a production quality EKS cluster because of this.

mattlandis:
We are working on a solution to this problem.  It will be sent out in another PR (no ETA yet but I would say soonish).  Going to close this one.

Issue #27, opened 2018-01-31
Cleanup vendoring packages


liwenwu-amazon:
Right now, `github.com/aws/amazon-vpc-cni-k8s` uses some code from `github.com/aws/amazon-ecs-agent` and `github.com/aws/amazon-ecs-cni-plugin`.  These shared code needs to be put into a new repo. And `github.com/aws/amazon-vpc-cni-k8s1`should vendor these shared code from new repo

samuelkarp:
What code are you using? It's worth evaluating whether to factor it out or to just stop using the code; not everything in the ECS agent is code that we'd recommend reusing.

PR #32, opened 2018-02-05
Cleanup vendor package

liwenwu-amazon:
### Summary
This PR  is intended for  issue #27  and removes vendor dependency on
 * github.com/aws/amazon-ecs-cni-plugins/
 * github.com/aws/amazon-ecs-agent

### Test Done
```
make static
make lint
make vet
make unit-test
```


pkg/utils/logger/logger.go
```diff
@@ -22,8 +22,8 @@ import (
 )
 
 const (
-	envLogLevel    = "ECS_CNI_LOGLEVEL"
-	envLogFilePath = "ECS_CNI_LOG_FILE"
+	envLogLevel    = "EKS_CNI_LOGLEVEL"
```
aaithal: this should definitely not have `EKS` in the name. A suggestion here would be `AWS_VPC_CNI_K8S_LOGLEVEL`

liwenwu-amazon: Done

mattlandis:
@liwenwu-amazon I am OK merging this if you create an issue to clean up the tests that use ttime to not mock time.  It does the job of cleaning up the vendor directory, but leaves the technical debt it picked up from the dependencies it was pulling in.

liwenwu-amazon:
#77  and move this change to master branch

Issue #36, opened 2018-02-27
Optimize allocation of ENI


SaranBalaji90:
Ipamd allocates ENI based on IP Pool size. But it doesn't consider instance based limits specified in  https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html. Therefore even when the available IP Pool size is reaching its maximum value on a specific instance class, Ipamd tries to allocate ENI continuously.  This loop runs every 5 seconds once and invokes 3 EC2 APIs (Create, Attach and Delete)

2018-02-27T13:58:33Z [DEBUG] IP pool stats: total=15, used=10, c.currentMaxAddrsPerENI =6, c.maxAddrsPerENI = 6
2018-02-27T13:58:33Z [INFO] Created a new eni: eni-a7b80f83
2018-02-27T13:58:33Z [DEBUG] Trying to tag newly created eni: keys=k8s-eni-key, value=i-074705deafbe739c0
2018-02-27T13:58:34Z [DEBUG] Tag the newly created eni with arn: arn:aws:ec2:us-west-2:908176817140:network-interface/eni-a7b80f83
2018-02-27T13:58:34Z [DEBUG] Discovered device number is used: 1
2018-02-27T13:58:34Z [DEBUG] Discovered device number is used: 0
2018-02-27T13:58:34Z [DEBUG] Discovered device number is used: 2
2018-02-27T13:58:34Z [DEBUG] Found a free device number: 3
2018-02-27T13:58:34Z [INFO] Exceeded instance eni attachment limit: 3
2018-02-27T13:58:34Z [ERROR] Failed to attach eni eni-a7b80f83: AttachmentLimitExceeded: Interface count 4 exceeds the limit for t2.medium
	status code: 400, request id: 42da274d-53a5-43bb-90a4-eaa4cc5a94a2
2018-02-27T13:58:34Z [DEBUG] Trying to delete eni: eni-a7b80f83
2018-02-27T13:58:34Z [INFO] Successfully deleted eni: eni-a7b80f83
2018-02-27T13:58:34Z [ERROR] Failed to increase pool size due to not able to allocate ENI allocate eni: error attaching eni: failed to attach eni: AttachmentLimitExceeded: Interface count 4 exceeds the limit for t2.medium
	status code: 400, request id: 42da274d-53a5-43bb-90a4-eaa4cc5a94a2
[...]

PR #38, opened 2018-02-28
Avoid calling EC2 to allocate ENIs when instance's ENIs limit reached

liwenwu-amazon:
To address #36 ,  this PR does following:

- discover the maximum number of ENIs can be attached to the instance
- avoid calling EC2 to allocate ENIs when the number of ENIs attached to instance reaches the limit

### Tests done
- create a cluster which contains 3 worker nodes that uses t2-medium type
- create 40 pods on the cluster
- verify L-IPAMD will create 3 ENIs and will not  trying to allocate 4th ENI after that
- delete 40 pods from the cluster
- verify L-IPAMD will free extra ENIs
- then add back 40 pods 
- verify L-IPAMD will create up to 3 ENIs for each node and will NOT trying to allocate 4th ENI after that

### DIFF
--- a/ipamd/ipamd.go
+++ b/ipamd/ipamd.go
@@ -15,6 +15,7 @@ package ipamd
 
 import (
 	"net"
+	"strings"
 	"time"
 
 	log "github.com/cihub/seelog"
@@ -49,6 +50,10 @@ type IPAMContext struct {
 
 	currentMaxAddrsPerENI int
 	maxAddrsPerENI        int
+	// maxENI indicate the maximum number of ENIs can be attached to the instance
+	// It is initialized to 0 and it is set to current number of ENIs attached
+	// when ipamD receives AttachmentLimitExceeded error
+	maxENI int
 }
 
 // New retrieves IP address usage information from Instance MetaData service and Kubelet
@@ -164,10 +169,24 @@ func (c *IPAMContext) decreaseIPPool() {
 	c.awsClient.FreeENI(eni)
 }
 
+func isAttachmentLimitExceededError(err error) bool {
+	return strings.Contains(err.Error(), "AttachmentLimitExceeded")
+}
+
 func (c *IPAMContext) increaseIPPool() {
+
+	if (c.maxENI > 0) && (c.maxENI == c.dataStore.GetENIs()) {
+		log.Debugf("Skipping increase IPPOOL due to max ENI already attached to the instance : %d", c.maxENI)
+		return
+	}
 	eni, err := c.awsClient.AllocENI()
 	if err != nil {
 		log.Errorf("Failed to increase pool size due to not able to allocate ENI %v", err)
+
+		if isAttachmentLimitExceededError(err) {
+			c.maxENI = c.dataStore.GetENIs()
+			log.Infof("Discovered the instance max ENI allowed is: %d", c.maxENI)
+		}
 		// TODO need to add health stats
 		return
 	}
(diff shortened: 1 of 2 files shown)


ipamd/ipamd.go
```diff
@@ -164,10 +169,24 @@ func (c *IPAMContext) decreaseIPPool() {
 	c.awsClient.FreeENI(eni)
 }
 
+func isAttachmentLimitExceededError(err error) bool {
+	return strings.Contains(err.Error(), "AttachmentLimitExceeded")
+}
+
 func (c *IPAMContext) increaseIPPool() {
+
+	if (c.maxENI > 0) && (c.maxENI <= c.dataStore.GetENIs()) {
```
rtripat: can c.maxENI be less than c.dataStore.GetENIs() ?

liwenwu-amazon: c.maxENI can NOT be less than c.dataStore.GetENIs(), unless it is 0.

rtripat: That's what I thought. Changing the condition to '==' or at least adding a comment would avoid this confusion

ipamd/ipamd.go
```diff
@@ -49,6 +50,10 @@ type IPAMContext struct {
 
 	currentMaxAddrsPerENI int
 	maxAddrsPerENI        int
+	// maxENI indicate the maximum number of ENIs can be attached to the instance
+	// It is initialized to 0 and it is set to current number of ENIs attached
+	// when ipamD receives AttachmentLimitExceeded error
+	maxENI int
```
SaranBalaji90: For now this should be ok, but in near future we should look into implementing something in a way that even if ipamD restarts we don't invoke AttachENI api to know if max number of ENIs are attached or not.

SaranBalaji90: We should create an issue or add TODO to visit this logic later.

SaranBalaji90: Or we should probably use the hardcoded map instead of this, till we have some sort of api to get these limits. Otherwise if all my 200 nodes in ASG gets restarted at the same time will cause us the same problem.

liwenwu-amazon: Here are Pros and Cons between **Using the hardcoded map** and **Dynamically discover maximum ENIs per instance from AttachmentLimitExceeded**:
## Dynamically discover maximum ENI per instance
### Pros:
* No code change nor new image whenever there is NEW instance type introduced or there any improvements on EC2 instance ENI limit.
### Cons:
* there is ONE extra call to discover the EC2 instance ENI limit during the life time of L-IPAMD.

## Using the hardcode map
### Pros:
* no extra ONE call to discovery the EC2 instance ENI limit
### Cons:
* Whenever there is a New instance type and there is any enhancement on EC2 instance ENI limits, we need to have a code change and new image
* This can add operation cost that you need to upgrade the aws-vpc-cni-k8s daemon set frequently to support new instance type and new EC2 ENI enhancement

SaranBalaji90: Agree. But we can also populate hardcoded map for known instance type and for unknown ones we can populate dynamically and store it.

SaranBalaji90: Also its not just one call, one call for CreateNetworkInterface, one for AttachNetworkInterface and other for DeleteNetworkInterface, which we can avoid using map. My only concern is if its larger cluster with 1000 nodes then all nodes will perform this operation right?

liwenwu-amazon: If static Map is used,  then it is operation hassle to upgrade aws-vpc-cni-k8s daemon for 1000 nodes every time there is a new instance type or any change in EC2 instance

liwenwu-amazon: Keep in mind, NOT all 1000 nodes will perform this operation unless its available IP address are below threshold.

SaranBalaji90: That's true.

PR #46, opened 2018-03-06
Check stderr() for error

liwenwu-amazon:


liwenwu-amazon:
Abandon this approach. Will use PR #48

Issue #39, opened 2018-03-01
amazon-k8s-cni pod fails on restart


SaranBalaji90:
When the amazon-k8s-cni pod is restarted either through docker restart CONTAINER_ID or when performing rolling update to DS/ when container is terminated, new pod ends up in crash loop. 
```
2018-03-01T06:28:10Z [INFO] Starting L-IPAMD 0.1.2  ...
2018-03-01T06:28:10Z [DEBUG] Discovered region: us-west-2
2018-03-01T06:28:10Z [DEBUG] Found avalability zone: us-west-2a
2018-03-01T06:28:10Z [DEBUG] Discovered the instance primary ip address: 10.0.107.27
2018-03-01T06:28:10Z [DEBUG] Found instance-id: i-022cd86d550d8d584
2018-03-01T06:28:10Z [DEBUG] Found primary interface's mac address: 06:d6:8a:ab:62:c6
2018-03-01T06:28:10Z [DEBUG] Discovered 3 interfaces.
2018-03-01T06:28:10Z [DEBUG] Found device-number: 2
2018-03-01T06:28:10Z [DEBUG] Found account ID: 908176817140
2018-03-01T06:28:10Z [DEBUG] Found eni: eni-242b8713
2018-03-01T06:28:10Z [DEBUG] Found device-number: 1
2018-03-01T06:28:10Z [DEBUG] Found eni: eni-d02488e7
2018-03-01T06:28:10Z [DEBUG] Found device-number: 0
2018-03-01T06:28:10Z [DEBUG] Found eni: eni-462c8071
2018-03-01T06:28:10Z [DEBUG] Found eni eni-462c8071 is a primary eni
2018-03-01T06:28:10Z [DEBUG] Found security-group id: sg-f4d6ad8b
2018-03-01T06:28:10Z [DEBUG] Found subnet-id: subnet-f98368b2
2018-03-01T06:28:10Z [DEBUG] Found vpc-ipv4-cidr-block: 10.0.0.0/16
2018-03-01T06:28:10Z [DEBUG] Total number of interfaces found: 3
2018-03-01T06:28:10Z [DEBUG] Found eni mac address : 06:31:b4:c1:38:80
2018-03-01T06:28:10Z [DEBUG] Found eni: eni-242b8713, mac 06:31:b4:c1:38:80, device 3
2018-03-01T06:28:10Z [DEBUG] Found cidr 10.0.96.0/19 for eni 06:31:b4:c1:38:80
[...]

PR #48, opened 2018-03-08
Use os.netLink.RuleAdd() APIs instead of using exec.cmd() 

liwenwu-amazon:
This PR  addresses issue #39 .
 * it removes calling exec.Cmd(""ip", "rule", "add", "not", "to", vpcCIDR.String()...)
 * instead, it uses os.netLInk.RuleAdd()/os.netLink.RuleDel() APIs 

**Tests Performed**
 * create a 1.9+ cluster using existing CNI 0.1.3
 * then perform rolling upgrade to CNI 0.1.4 
```
kubectl-osx apply -f  <cni.yaml which points to CNI 0.1.4>
``` 
* verify rolling upgrade finish OK 
```
kubectl-osx get ds -n kube-system
```
* deploy some pods and verify Pod-to-Pod ping OK


vendor/github.com/vishvananda/netlink/Makefile
```diff
@@ -3,7 +3,8 @@ DIRS := \
 	nl
 
 DEPS = \
-	github.com/vishvananda/netns
```
SaranBalaji90: Is it possible to separate these vendor into separate PR? Its bit harder to do review along with these changes.

liwenwu-amazon: @SaranBalaji90 It won't compile if I put them into separate PRs.  Right now, there are in 2 different commits, one for cni and one for vendor update

pkg/networkutils/network.go
```diff
@@ -72,32 +71,24 @@ func isDuplicateRuleAdd(err error) bool {
...
-	//TODO : temporary implement "ip rule not to vpc cider table main priority 1024
-	cmd := append([]string{"ip", "rule", "add", "not", "to", vpcCIDR.String(), "table", "main", "priority", "1024"})
+	hostRule := os.netLink.NewRule()
+	hostRule.Dst = vpcCIDR
+	hostRule.Table = mainRoutingTable
+	hostRule.Priority = hostRulePriority
+	hostRule.Invert = true
+
+	// if this is a restart, cleanup previous rule first
+	err := os.netLink.RuleDel(hostRule)
```
SaranBalaji90: Why not just invoke RuleAdd(...) and check if error code contains anything other than ENOENT?

liwenwu-amazon: hmmm, in some OS (e.g. ubuntu) it will create 2 Rule entries and not complaining anything.

PR #54, opened 2018-03-29
Ensure that logs are flushed when exiting on error

### LINKED ISSUE #57: Logs are not flushed on ipamd initialization failure
Logs are not flushed to disk (to [host]/var/log/aws-routed-eni/ipamd.log.xxx) when ipamd initialization fails, which makes debugging hard.

This is due to the defer clause in the main function: defer does not apply when using os.Exit() so logs are not flushed on initialization errors.

lbernail:
*Description of changes:*

Fixes #57

The defer clause does not apply when using os.Exit() so logs are not flushed on initialization errors, which makes debugging hard. 

This fix ensures that logs are flushed before exiting.

### DIFF
--- a/main.go
+++ b/main.go
@@ -28,6 +28,10 @@ const (
 )
 
 func main() {
+	os.Exit(_main())
+}
+
+func _main() int {
 	defer log.Flush()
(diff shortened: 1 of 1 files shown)


mattlandis:
Thanks for the PR.  Everything looks good.  Merging.

PR #62, opened 2018-04-25
Add 30 sec cooling for IP address reassignment

liwenwu-amazon:
**Issue #59**:
 Pods stuck in ContainerCreating due to CNI Failing to Assing IP to Container Until aws-node is deleted*

**Description of changes:**
When an IP is returned to IP WarmPool, keep it there for 30 sec before reassigning to a new Pod. This was planned in the [original proposal](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/proposals/cni-proposal.md#pod-ip-address-cooling-period).

**Tests Performed:**
performs back-to-back pods create and pods delete continuously.  Without fix, the issue can be reproduced easily uses "heavy pods"(e.g. from StatefulSet). With the fix, the issue is no-longer reproducible 

### DIFF
--- a/ipamd/datastore/data_store.go
+++ b/ipamd/datastore/data_store.go
@@ -24,8 +24,14 @@ import (
 )
 
 const (
-	minLifeTime          = 1 * time.Minute
-	addressCoolingPeriod = 1 * time.Minute
+	minLifeTime = 1 * time.Minute
+	// addressENICoolingPeriod is used to ensure ENI will NOT get freed back to EC2 control plane if one of
+	// its secondary IP addresses is used for a Pod within last addressENICoolingPeriod
+	addressENICoolingPeriod = 1 * time.Minute
+
+	// addressCoolingPeriod is used to ensure an IP not get assigned to a Pod if this IP is used by a different Pod
+	// in addressCoolingPeriod
+	addressCoolingPeriod = 30 * time.Second
 	// DuplicatedENIError is an error when caller tries to add an duplicate ENI to data store
 	DuplicatedENIError = "data store: duplicate ENI"
 
@@ -194,6 +200,7 @@ func (ds *DataStore) assignPodIPv4AddressUnsafe(k8sPod *k8sapi.K8SPodInfo) (stri
 		namespace: k8sPod.Namespace,
 		container: k8sPod.Container,
 	}
+	curTime := time.Now()
 	for _, eni := range ds.eniIPPools {
 		if (k8sPod.IP == "") && (len(eni.IPv4Addresses) == eni.AssignedIPv4Addresses) {
 			// skip this ENI, since it has no available IP address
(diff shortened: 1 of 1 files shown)


ipamd/datastore/data_store.go
```diff
@@ -213,7 +214,7 @@ func (ds *DataStore) assignPodIPv4AddressUnsafe(k8sPod *k8sapi.K8SPodInfo) (stri
 					addr.address, k8sPod.Name, k8sPod.Namespace)
 				return addr.address, eni.DeviceNumber, nil
 			}
-			if !addr.Assigned && k8sPod.IP == "" {
+			if !addr.Assigned && k8sPod.IP == "" && curTime.Sub(addr.unAssignedTime) > 30*time.Second {
```
mattlandis: How is this time related to the `addressCoolingPeriod` const?

dbenhur: +1 on Matt's question.  The values seem related, though the defined constant seems to be used to guard when it's ok to free an ENI.  If they are coupled, reuse the defined constant.  If not, make another named constant instead of the inline `30*time.Second`.

I note that `addr.unAssignedTime` may be nil.  In this case `Sub` appears to compute a duration from epoch, but this is accidental, [`not documented`](https://golang.org/pkg/time/#Time.Sub) behavior.  It would be safer for code to handle the nil case explicitly, to protect for this undocumented behavior changing.  I also note the time package has a [`Since`](https://golang.org/pkg/time/#Since) function that would simplify the expression to `Since(addr.unAssignedTime) > addressCoolingPeriod` and save having to set curTime above.

dbenhur: Upon deeper look, [time.Time](https://golang.org/pkg/time/#Time) is struct, only one element of which can be nil.  The Zero value of time is defined as the unix epoch, so the behavior is not undocumented.  I still advocate a named constant, commentary, and use of Since in preference to now.Sub(prior).

liwenwu-amazon:
The cooling period ([design-proposal](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/proposals/cni-proposal.md) ) is used to avoid an IP address being reassigned to a different Pod within cooling period.

ipamd/datastore/data_store.go
```diff
@@ -24,8 +24,14 @@ import (
...
-	minLifeTime          = 1 * time.Minute
-	addressCoolingPeriod = 1 * time.Minute
+	minLifeTime = 1 * time.Minute
+	// addressENICoolingPeriod is used to ensure ENI will NOT get freed back to EC2 control plane if one of
+	// its secondary IP addresses is used for a Pod within last addressENICoolingPeriod
+	addressENICoolingPeriod = 1 * time.Minute
+
+	// addressCoolingPeriod is used to ensure an IP not get assigned to a Pod if this IP is used by a different Pod
+	// in addressCoolingPeriod
+	addressCoolingPeriod = 30 * time.Second
```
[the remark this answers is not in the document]
liwenwu-amazon: addressCoolingPeriod is used to ensure an IP not get assigned to another Pod within addressCoolingPeriod. It impacts how soon the new Pods assigned to the node can re-use those IP addresses. For examples, a replica scale down which deletes all Pods on the node, and then immediately followed by a replica scale up that allocates all new Pods to the node. The addressCoolingPeriod can impact how soon these new Pods get their addresses.  addressENICoolingPeriod is used to ensure ENI will not get freed back to EC2 control plane within addressENICoolingPeriod. ENI allocation/free are expensive operations. So, the timer is longer and we planning to add this a configurable parameter in future. For example, in some cases, ENI are never freed.

PR #68, opened 2018-05-15
Add config option for number of ENIs get preallocated

liwenwu-amazon:
## Description of changes

**Current Behavior**
Today, ipamD dynamically allocate and free ENIs based on the number Pods running on the node.
It allocates one more ENI when the number of available IP addresses goes below the per ENI's  secondary IP limit.  It can free an ENI when the number of available IP addresses grows more than 2 * per ENI's secondary IP limit.  

 For example, for c3.xlarge, each ENI have 14 secondary IPs.   When there is one ENI attached to the instance(e.g. when the instance is first booted) and after 8 pods just get scheduled to run on this node. The available IPs becomes 6 (14-8) and this will trigger ipamD allocating a new ENI.

**Requirements**
There are 3 different deployment scenarios requires 3 different triggering threshold:

* scenario-1: default and same as today's behavior. Let ipamD dynamically allocate and free one ENI at time based on the available IPs
* scenario-2:  allocate all available ENIs and IPs right after the  instance is initialized.  Once all ENIs are allocated on all nodes, there will be no delay on Pod deployment (compare to scenario-1, that some pods may have to wait for nodes to get new ENI and  IPs)
* scenario-3: do NOT allocate any ENI.  And only use secondary IPs from instance's primary ENI. 

**New Behavior**
This PR add config option which allow user to define the number of ENIs get allocated.  If the config option is not defined, it is default to today's behavior 
 
### DIFF
--- a/ipamd/ipamd.go
+++ b/ipamd/ipamd.go
@@ -15,6 +15,8 @@ package ipamd
 
 import (
 	"net"
+	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -40,27 +42,34 @@ const (
 	ipPoolMonitorInterval = 5 * time.Second
 	maxRetryCheckENI      = 5
 	eniAttachTime         = 10 * time.Second
+	defaultWarmENITarget  = 1
 )
 
 var (
 	ipamdErr = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
 			Name: "ipamd_error_count",
-			Help: "the number of errors encountered in ipamd",
+			Help: "The number of errors encountered in ipamd",
 		},
 		[]string{"fn", "error"},
 	)
 	ipamdActionsInprogress = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Name: "ipamd_action_inprogress",
-			Help: "the number of ipamd actions inprogress",
+			Help: "The number of ipamd actions inprogress",
 		},
 		[]string{"fn"},
 	)
 	enisMax = prometheus.NewGauge(
 		prometheus.GaugeOpts{
 			Name: "eni_max",
-			Help: "The number of maximum ENIs can be attached to the instance",
+			Help: "The maximum number of ENIs that can be attached to the instance",
+		},
+	)
+	ipMax = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "ip_max",
+			Help: "The maximum number of IP addresses that can be allocated to the instance",
 		},
 	)
 	prometheusRegistered = false
@@ -86,6 +95,7 @@ func prometheusRegister() {
 		prometheus.MustRegister(ipamdErr)
 		prometheus.MustRegister(ipamdActionsInprogress)
 		prometheus.MustRegister(enisMax)
+		prometheus.MustRegister(ipMax)
 		prometheusRegistered = true
 	}
 }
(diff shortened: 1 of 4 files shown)


ipamd/ipamd.go
```diff
@@ -63,6 +66,12 @@ var (
 			Help: "The number of maximum ENIs can be attached to the instance",
 		},
 	)
+	ipMax = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "ip_max",
+			Help: "The number of maximum IPs can be allocated to the instance",
```
dbenhur: s/number of maximum IPs can/maximum number of IP addresses that can/

ipamd/ipamd.go
```diff
@@ -400,23 +418,44 @@ func (c *IPAMContext) waitENIAttached(eni string) (awsutils.ENIMetadata, error)
...
+			return defaultWarmENITarget
+		}
+		log.Debugf("Using WARM-ENI-TARGET %v", input)
+		return input
+	}
+	return defaultWarmENITarget
+}
+
+func logPoolStats(total, used, currentMaxAddrsPerENI, maxAddrsPerENI int) {
+	log.Debugf("IP pool stats: total=%d, used=%d, c.currentMaxAddrsPerENI =%d, c.maxAddrsPerENI = %d",
```
dbenhur: be consistent on your use of spaces around `=`

ipamd/ipamd.go
```diff
@@ -400,23 +418,44 @@ func (c *IPAMContext) waitENIAttached(eni string) (awsutils.ENIMetadata, error)
...
 //nodeIPPoolTooLow returns true if IP pool is below low threshhold
 func (c *IPAMContext) nodeIPPoolTooLow() bool {
+	warmENITarget := getWarmENITarget()
 	total, used := c.dataStore.GetStats()
-	log.Debugf("IP pool stats: total=%d, used=%d, c.currentMaxAddrsPerENI =%d, c.maxAddrsPerENI = %d",
-		total, used, c.currentMaxAddrsPerENI, c.maxAddrsPerENI)
+	logPoolStats(total, used, c.currentMaxAddrsPerENI, c.maxAddrsPerENI)
 
-	return ((total - used) <= c.currentMaxAddrsPerENI)
+	return ((total - used) <= c.currentMaxAddrsPerENI*warmENITarget)
```
dbenhur: this would be clearer if you said `available := total - used` then made the comparison against available.  add space around ' * '.

ipamd/ipamd.go
```diff
@@ -400,23 +418,44 @@ func (c *IPAMContext) waitENIAttached(eni string) (awsutils.ENIMetadata, error)
...
 func (c *IPAMContext) nodeIPPoolTooHigh() bool {
+	warmENITarget := getWarmENITarget()
 	total, used := c.dataStore.GetStats()
+	logPoolStats(total, used, c.currentMaxAddrsPerENI, c.maxAddrsPerENI)
 
-	log.Debugf("IP pool stats: total=%d, used=%d, c.currentMaxAddrsPerENI =%d, c.maxAddrsPerENI = %d",
-		total, used, c.currentMaxAddrsPerENI, c.maxAddrsPerENI)
-
-	return (total-used > 2*c.currentMaxAddrsPerENI)
+	return (total-used > (warmENITarget+1)*c.currentMaxAddrsPerENI)
```
dbenhur: ditto

ipamd/ipamd.go
```diff
@@ -117,8 +127,12 @@ func New() (*IPAMContext, error) {
 
 //TODO need to break this function down(comments from CR)
 func (c *IPAMContext) nodeInit() error {
+	ipamdActionsInprogress.WithLabelValues("nodeInit").Add(float64(1))
+	defer ipamdActionsInprogress.WithLabelValues("nodeInit").Sub(float64(1))
 	maxENIs, err := c.awsClient.GetENILimit()
 	enisMax.Set(float64(maxENIs))
+	maxIPs, err := c.awsClient.GetENIipLimit()
+	ipMax.Set(float64(maxIPs * int64(maxENIs)))
```
dbenhur: These can get an error which the code completely ignores.

ipamd/ipamd.go
liwenwu-amazon: go fmt removes spaces around '*'

ipamd/ipamd.go
dbenhur: please correct as advised.  this is user facing description, let's get the grammar right.

Issue #66, opened 2018-05-12
ENI and its secondary IP addresses are NOT installed into ipamd datastore


liwenwu-amazon:
### Problem Symptoms
During the scale test  of deploying 8000 pods over 150 nodes (c3.xlarge),  there are 12 pods stucked in ***ContainerCreating*** state. 

The node where these pods get assigned are complaining there are NO available IP addresses for these Pods.  There are total of 42 IP addresses and all of them have been assigned to Pods.
```
#e.g. in /var/log/aws-routed-eni/ipamd.log.2018-05-12-00
2018-05-12T00:59:59Z [INFO] DataStore has no available IP addresses
``` 
By current cni design, this node should have 4 enis and 56 IP addresses. And from EC2 console, you can see there are 4 enis and 60 ip addresses available for this node.

But ipamd datastore indicates that there are only 3 enis and 42 IP addresses.

### Workaround 
Reload aws-node container on this node and then 4 enis and 56 IP addresses get added to datastore. All remaining 12 pods transit into "Running" state

### Root Cause
If cni-ipamd daemonset pod is rebooted when it is in the middle of increasing IP Pool Size (create eni/ attach eni/add eni/ip to datastore),  the newly attached eni and its ip addresses can be lost and never get added to l-ipamd datastore.

liwenwu-amazon:
### Here are snips from *ipamd.log.yyyy-mm-dd-hh* logfile:

2018-05-12T00:00:27Z [DEBUG] Found ip addresses [192.168.64.205 192.168.110.74 192.168.99.64 192.168.68.162 192.168.100.131 192.168.107.38 192.168.65.199 192.168.89.231 192.168.79.231 192.168.108.89 192.168.111.91 192.168.109.222 192.168.92.144 192.168.67.145 192.168.76.87] on eni 06:87:9b:fe:65:28
2018-05-12T00:00:27Z [DEBUG] Not able to discover attached eni yet (attempt 5/5)
2018-05-12T00:00:37Z [DEBUG] Total number of interfaces found: 1
2018-05-12T00:00:37Z [DEBUG] Found eni mac address : 06:87:9b:fe:65:28
2018-05-12T00:00:37Z [DEBUG] Using device number 0 for primary eni: eni-029d728401ff5ea94
2018-05-12T00:00:37Z [DEBUG] Found eni: eni-029d728401ff5ea94, mac 06:87:9b:fe:65:28, device 0
[...]

PR #74, opened 2018-05-19
eni reconcile

liwenwu-amazon:
*Issue #66 , if available:*

### Summary

* query instance's metadata service and get up-to-date eni and IP addresses information
* then reconcile this up-to-date eni/IP information into local cache in L-IPAMD

### Testing Done

```
make unit-test
make lint
make vet
```
In addition, manually tested following:

* Manually remove worker node's IAM permission for EC2's CreateNetworkInterface/AttachNetworkInterface/DeleteNetworkInterface/DetachNetworkInterface/AssignPrivateIpAddresses
* Use AWS console and manually add and attach ENI and IP addresses to a worker node
* Verify cni-ipamD dynamically reconcile this ENI and IP address to its data store
* Use AWS console and manually detach and free ENI and IP addresses from a worker node
* Verify cni-ipamD dynamically reconcile and delete this ENI and its IP addresses from its data store

### DIFF
--- a/ipamd/datastore/data_store.go
+++ b/ipamd/datastore/data_store.go
@@ -38,6 +38,18 @@ const (
 
 	// DuplicateIPError is an error when caller tries to add an duplicate IP address to data store
 	DuplicateIPError = "datastore: duplicated IP"
+
+	// UnknownIPError is an error when caller tries to delele an IP which is unknown to data store
+	UnknownIPError = "datastore: unknown IP"
+
+	// IPInUseError is an error when caller tries to delete an IP where IP is still assigned to a Pod
+	IPInUseError = "datastore: IP is used and can not be deleted"
+
+	// ENIInUseError is an error when caller tries to delete an ENI where there are IP still assigned to a pod
+	ENIInUseError = "datastore: ENI is used and can not be deleted"
+
+	// UnknownENIError is an error when caller tries to access an ENI which is unknown to datastore
+	UnknownENIError = "datastore: unknown ENI"
 )
 
 // ErrUnknownPod is an error when there is no pod in data store matching pod name, namespace, container id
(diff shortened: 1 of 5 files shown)


ipamd/datastore/data_store.go
```diff
@@ -343,6 +387,29 @@ func (ds *DataStore) FreeENI() (string, error) {
 	return deletedENI, nil
 }
 
+// DeleteENI free a ENI.
+func (ds *DataStore) DeleteENI(eni string) error {
+	ds.lock.Lock()
+	defer ds.lock.Unlock()
+
+	_, ok := ds.eniIPPools[eni]
```
mogren: Save this in a variable here, `eniIPPool, ok = ds.eniIPPools[eni]` and replace the three lookups below.

ipamd/ipamd.go
```diff
@@ -232,16 +246,18 @@ func (c *IPAMContext) reconcileENIIP() {
...
 		}
 		ec2Addrs, _, err := c.getENIaddresses(eni.Id)
 		if err != nil {
-			ipamdErrInc("reconcileENIIPgetENIaddressesFailed", err)
+			ipamdErrInc("retryAllocENIIPgetENIaddressesFailed", err)
 			log.Warn("During eni repair: failed to get ENI ip addresses", err)
 			return
 		}
+		curTime := time.Now()
+		c.lastNodeIPPoolAction = curTime
```
mogren: Why not just `c.lastNodeIPPoolAction = time.Now()`?

ipamd/ipamd.go
```diff
@@ -257,6 +273,8 @@ func (c *IPAMContext) decreaseIPPool() {
 	}
 	log.Debugf("Start freeing eni %s", eni)
 	c.awsClient.FreeENI(eni)
+	curTime := time.Now()
+	c.lastNodeIPPoolAction = curTime
```
mogren: Same

ipamd/ipamd.go
```diff
@@ -318,6 +336,8 @@ func (c *IPAMContext) increaseIPPool() {
 		log.Errorf("Failed to increase pool size: %v", err)
 		return
 	}
+	curTime := time.Now()
+	c.lastNodeIPPoolAction = curTime
```
mogren: Same

ipamd/ipamd.go
```diff
@@ -207,21 +220,22 @@ func (c *IPAMContext) StartNodeIPPoolManager() {
 	for {
 		time.Sleep(ipPoolMonitorInterval)
 		c.updateIPPoolIfRequired()
+		c.nodeIPPoolReconcile(nodeIPPoolReconcileInterval)
```
mogren: Is there a reason this reconcile loop doesn't run in it's own go routine, and just uses the same approach with a sleep like the update loop? Then there would be no need to set and check against `lastNodeIPPoolAction`.

ipamd/datastore/data_store.go
liwenwu-amazon: done

ipamd/ipamd.go
liwenwu-amazon: done

ipamd/ipamd.go
liwenwu-amazon: done

ipamd/ipamd.go
liwenwu-amazon: done

ipamd/ipamd.go
liwenwu-amazon: The intend of ```lastNodeIPPoolAction``` is to make sure it does NOT do reconcile right after  increasing or decreasing the IP Pool, so it is needed even it is put inside ```c.updateIPPoolIfRequred()```

mogren: Ah, that makes sense. Forgot that `updateIPPoolIfRequired()` also updates that time.

PR #89, opened 2018-05-28
Added -max-pods flag to main binary

micahhausler:
* cleaned up main.go
* removed unused verify scripts 
* make now creates a static binary


Makefile
```diff
@@ -14,10 +14,10 @@
 
 # build binary
 static:
-	go build -o aws-k8s-agent main.go
-	go build -o aws-cni plugins/routed-eni/cni.go
-	go build verify-aws.go
-	go build verify-network.go
+	GOOS=linux CGO_ENABLED=0 go build -o aws-k8s-agent main.go
+	GOOS=linux CGO_ENABLED=0 go build -o aws-cni plugins/routed-eni/cni.go
+	go build scripts/verify-aws.go
```
mattlandis: Can we put them in scripts/verify-aws/verify-aws.go then go install will work correctly.

main.go
```diff
@@ -52,12 +70,10 @@ func _main() int {
 
 	if err != nil {
 		log.Error("initialization failure", err)
-		return 1
```
mattlandis: os.Exit causes the defer log.Flush() not to run, hence the exit status and subfunction.

mogren:
Cancelled

PR #92, opened 2018-05-29
Use UID for pod/container matching and only consider running containers

liwenwu-amazon:
*Description of changes*
When ipamD restart, it needs to discover currently running Pods on the node and reserves the IP addresses used by these running Pods in datastore.

there are 2 changes

* To find out the container ID of a Pod, the original code was using **name**.  The **name** could change if the pod got restarted. So, modify the logic to use UID. 
* Change logic to not consider non-running container/pod.

### DIFF
--- a/ipamd/ipamd.go
+++ b/ipamd/ipamd.go
@@ -210,6 +210,11 @@ func (c *IPAMContext) nodeInit() error {
 				ip.Name, ip.Namespace)
 			continue
 		}
+		if ip.IP == "" {
+			log.Infof("Skipping Pod %s, Namespace %s due to no IP",
+				ip.Name, ip.Namespace)
+			continue
+		}
 		log.Infof("Recovered AddNetwork for Pod %s, Namespace %s, Container %s",
 			ip.Name, ip.Namespace, ip.Container)
 		_, _, err = c.dataStore.AssignPodIPv4Address(ip)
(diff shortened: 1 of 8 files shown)


pkg/k8sapi/discovery.go
```diff
@@ -201,13 +201,12 @@ func (d *Controller) handlePodUpdate(key string) error {
 	}
 
 	if !exists {
-		podName := obj.(*v1.Pod).GetName()
+		//podName := obj.(*v1.Pod).GetName()
```
mattlandis: Remove instead of just commenting it out.

pkg/docker/docker.go
```diff
@@ -41,13 +43,48 @@ func (c *Client) GetRunningContainers() ([]*ContainerInfo, error) {
...
+		_, ok = containerInfos[uid]
+		if !ok {
+			containerInfos[uid] = &ContainerInfo{
 				ID:     container.ID,
 				Name:   container.Names[0],
-				K8SUID: container.Labels["io.kubernetes.pod.uid"]})
+				K8SUID: uid}
+			continue
+		}
+
```
mattlandis: Turn these into an if/else and a comment above about why you don't think that the code should reach here under normal circumstances.  Do we a theory on which of the containerInfos is the correct one in the case we have a collision on uuid but with different names?

ipamd/ipamd.go
```diff
@@ -243,21 +248,23 @@ func (c *IPAMContext) getLocalPodsWithRetry() ([]*k8sapi.K8SPodInfo, error) {
...
-		log.Errorf("Not able to get container info from docker: %v", err)
-		return nil, errors.Wrapf(err, "not able to get container info from docker")
+	for retry := 1; retry <= maxK8SRetries; retry++ {
+		containers, err = c.dockerClient.GetRunningContainers()
+		if err == nil {
+			break
+		}
+		log.Infof("Not able to get local containers yet (attempt %d/%d): %v", retry, maxK8SRetries, err)
+		time.Sleep(retryK8SInterval)
 	}
```
mattlandis: What happens if we exceed maxK8sRetries?  The loop below will iterate over the pods, but there are no containers so it will do nothing.  The function no longer returns an error if it cannot get container info from docker.

PR #93, opened 2018-05-30
Fix nil error in discovery of pods

mattlandis:
When an item was no longer in the index (delete) the interface returned
will be nil.

### DIFF
--- a/pkg/k8sapi/discovery.go
+++ b/pkg/k8sapi/discovery.go
@@ -4,7 +4,6 @@ package k8sapi
 import (
 	"fmt"
 	"os"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


mattlandis:
Tested by building the image and deploying to a cluster, then running conformance tests.

pkg/k8sapi/discovery.go
```diff
@@ -201,35 +200,35 @@ func (d *Controller) handlePodUpdate(key string) error {
...
+	if !ok {
+		log.Errorf("Updated object recieved was not a pod: %+v", obj)
+		return errors.New("recieved a non-pod object update")
+	}
+	// Note that you also have to check the uid if you have a local controlled resource, which
+	// is dependent on the actual instance, to detect that a Pod was recreated with the same name
+	podName := pod.GetName()
+
+	// check to see if this is one of worker pod on my nodes
+	if d.myNodeName == pod.Spec.NodeName && pod.Spec.HostNetwork {
```
liwenwu-amazon: should be !pod.Spec.HostNetwork

PR #98, opened 2018-06-05
Add m5 family to the IP limits

igalarzab:
*Description of changes:*

When the plugin is run in a `m5.*` node, the following error is shown in the logs:

```
[ERROR] Failed to get eni IP limit due to unknown instance type m5.large
```

This PR adds the ENI limits for all the m5 nodes.

PS: The whole array was reformated by gofmt so the diff is much easier to see if you ignore whitespaces (`&w=1` in github or `-w` in git diff)


mattlandis:
Thanks for the contribution.  Would you mind updating the `InstanceIPsAvailable` map below for these instance types as well.

igalarzab:
done!

dbenhur:
Thanks for the PR. @igalarzab .   I noticed a few other instance types omitted and reused a script we had that parses the authoritative table at https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html#AvailableIpPerENI to build the Go tables here.  So we're going to close this PR in favor of the more complete one: https://github.com/aws/amazon-vpc-cni-k8s/pull/101/

Issue #91, opened 2018-05-29
Need Continuous integration set up


liwenwu-amazon:
Whenever there is a PR,  **Continuous Integration** should get kick off automatically:

* compile and generate binary and docker image
* run **make unit-test** 
* run **end-to-end** function tests, such as
   * kubernetes e2e network related tests

igorvpcleao:
@liwenwu-amazon, do you still need help on this?
Did https://github.com/aws/amazon-vpc-cni-k8s/pull/116 solve this issue?

liwenwu-amazon:
@igorvpcleao  #116 only solves portion of the problem.  The next step is to figure out how to automatically trigger kubernetes e2e network related test or conformance test upon every PR or code commit.

jaypipes:
We've got the [test-automation project](https://github.com/aws/amazon-vpc-cni-k8s/projects/2) with a bunch of issues tracking testing and CI. Closing this one out.

PR #116, opened 2018-06-25
Initial travis CI integration

mogren:
*Resolves issue #91*

*Description of changes:*
- Adds basic Travis CI integration. 
- Updates Makefile to make clear it is a Linux only build.
- Updated README with Makefile changes.
- Added IntelliJ files to `.gitignore`.
- Removed two unused scripts.

Tested by enabling Travis CI for my fork: https://travis-ci.org/mogren/amazon-vpc-cni-k8s


.travis.yml
```diff
@@ -0,0 +1,30 @@
...
+git:
+  depth: 1
+
+go_import_path: github.com/aws/amazon-vpc-cni-k8s
+
+# Install needed tools
+install:
+  - go get -u github.com/golang/dep/cmd/dep
+  - go get -u golang.org/x/lint/golint
+  - dep ensure
```
mattlandis: You should not need to run dep ensure as part of the build.  All of the dependencies should be in the vendor directory already and properly versioned.  If they are not the build should fail.

mogren: You are right! It did fail at first on my mac, but that was because of OS X. Removing the dep stuff.

mogren: Passed without dep ensure: https://travis-ci.org/mogren/amazon-vpc-cni-k8s/builds/396655872

Issue #114, opened 2018-06-25
Ability to configure secondary IP preallocation 


cnidus:
EKS preallocates the max # of IPs that a given instance supports. This pre-warms and allows for fast pod spinup. However, it leads to over allocation of IPs and in some cases IP exhaustion of a given VPC / subnet.

In extreme examples, where a node always hosts a single pod, there could be upwards for 50 IPs wasted on each host.

Suggestion is for the IP pre-warming to be a configurable value.

liwenwu-amazon:
#119 propose to have a configuration option to use different subnet for Pods than worker node's subnet. This can also help to prevent IP exhaustion of node's subnet.

liwenwu-amazon:
This has been addressed with release 1.1.0.   Please refer https://aws.amazon.com/blogs/opensource/vpc-cni-plugin-v1-1-available/ for the details.

PR #125, opened 2018-07-07
Add WARM_IP_TARGET support

liwenwu-amazon:
*Issue #114:*

#### Description of changes

This PR allows user to use a OS environment variable `WARM_IP_TARGET` to configure the number of pre-warming IP addresses.

You can modify [amazon-vpc-cni.yaml](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/v1.0/aws-k8s-cni.yaml) to include `WARM_IP_TARGET` value

```

env:
          - name: AWS_VPC_K8S_CNI_LOGLEVEL
            value: DEBUG
          - name: MY_NODE_NAME
            valueFrom:
              fieldRef:
                fieldPath: spec.nodeName
          - name: "WARM_IP_TARGET"  <-------- define pre-warming IP
             value: "1"   <--------   This must be a integer, e.g. 1 IP address

```
If `WARM_IP_TARGET` is NOT defined, [ipamD](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/docs/cni-proposal.md) will fallback to current allocation behavior. 

Today, by-default, if number of available IP addresses is below ENI's IP address, ipamD will allocate a new ENI.  For example,  for a [t2.medium](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html#AvailableIpPerENI) worker node, if the number of **available IP address** is below 5, ipamD will allocate a new ENI, allocate IP addresses(5 addresses) on this new ENI. 

#### Caveat
If `WARM_IP_TARGET` is set too low and more Pods are expected to be scheduled on the node, ipamD will invoke more EC2 API calls to allocate IP addresses.  For example, for a m4.4xlarge worker node, each ENI can have up to 30 addresses.  If "WARM_IP_TARGET" is set to 1 and if there are more than 30 pods scheduled on the node,  there will be **30** EC2 AssignPrivateIpAddresses() API call for each ENI compare to **1** EC2 AssignPrivateIpAddresses() API call using default behavior (as today), and in worst case for a node,  **240** EC2 AssignPrivateIpAddresses() API calls compare to **8** EC2 AssignPrivateIpAddresses() API calls using default behavior(as today).  If the cluster is large and contains lots of worker nodes, this can cause ipamD running into **EC2 AssignPrivateIpAddresses() API call throttling** problems.

#### Tests 

* define WARM_IP_TARGET=1 and uses t2.medium worker node where each ENI can have up to 5 Pods IP addresses
    * allocate 2 Pods, and verify there is 1 more available IP addresses
    * allocate 5 pods, and verify 1 ENI allocated and there is only 1 more IP addressed pre-warmed

* regression test and make sure default behavior still works.  Do NOT define WARM_IP_TARGET, uses t2.medium worker
    * allocate 2 Pods, and verify ipamD will allocate 1 extra ENI and have 10 IP  addresses pre-warmed

### DIFF
--- a/ipamd/ipamd.go
+++ b/ipamd/ipamd.go
@@ -47,6 +47,7 @@ const (
 	nodeIPPoolReconcileInterval = 60 * time.Second
 	maxK8SRetries               = 12
 	retryK8SInterval            = 5 * time.Second
+	noWarmIPTarget              = 0
 )
 
 var (
@@ -195,14 +196,6 @@ func (c *IPAMContext) nodeInit() error {
 	for _, eni := range enis {
 		log.Debugf("Discovered ENI %s", eni.ENIID)
 
-		err = c.awsClient.AllocAllIPAddress(eni.ENIID)
-		if err != nil {
-			ipamdErrInc("nodeInitAllocAllIPAddressFailed", err)
-			//TODO need to increment ipamd err stats
-			log.Warn("During ipamd init:  error encountered on trying to allocate all available IP addresses", err)
-			// fall though to add those allocated OK addresses
-		}
-
 		err = c.setupENI(eni.ENIID, eni)
 		if err != nil {
 			log.Errorf("Failed to setup eni %s network: %v", eni.ENIID, err)
@@ -310,6 +303,12 @@ func (c *IPAMContext) updateIPPoolIfRequired() {
 func (c *IPAMContext) retryAllocENIIP() {
 	ipamdActionsInprogress.WithLabelValues("retryAllocENIIP").Add(float64(1))
 	defer ipamdActionsInprogress.WithLabelValues("retryAllocENIIP").Sub(float64(1))
+
+	curIPTarget, warmIPTargetDefined := c.getCurWarmIPTarget()
+	if warmIPTargetDefined && curIPTarget <= 0 {
+		log.Debugf("Skipping retry allocating ENI IP, warm IP target reached")
+		return
+	}
 	maxIPLimit, err := c.awsClient.GetENIipLimit()
 	if err != nil {
 		log.Infof("Failed to retrieve ENI IP limit: %v", err)
@@ -318,7 +317,12 @@ func (c *IPAMContext) retryAllocENIIP() {
 	eni := c.dataStore.GetENINeedsIP(maxIPLimit)
 	if eni != nil {
 		log.Debugf("Attempt again to allocate IP address for eni :%s", eni.ID)
-		err := c.awsClient.AllocAllIPAddress(eni.ID)
+		var err error
+		if warmIPTargetDefined {
+			err = c.awsClient.AllocIPAddresses(eni.ID, int64(curIPTarget))
+		} else {
+			err = c.awsClient.AllocIPAddresses(eni.ID, maxIPLimit)
+		}
 		if err != nil {
 			ipamdErrInc("retryAllocENIIPAllocAllIPAddressFailed", err)
 			log.Warn("During eni repair: error encountered on allocate IP address", err)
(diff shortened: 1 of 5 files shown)


ipamd/ipamd.go
```diff
@@ -303,7 +302,12 @@ func (c *IPAMContext) retryAllocENIIP() {
 	eni := c.dataStore.GetENINeedsIP(maxIPLimit)
 	if eni != nil {
 		log.Debugf("Attempt again to allocate IP address for eni :%s", eni.ID)
-		err := c.awsClient.AllocAllIPAddress(eni.ID)
+		var err error
+		if warmIPTargetDefined {
+			err = c.awsClient.AllocIPAddresses(eni.ID, curIPTarget)
```
nckturner: Do these need to be separate functions?

ipamd/ipamd.go
```diff
@@ -664,3 +700,35 @@ func (c *IPAMContext) eniIPPoolReconcile(ipPool map[string]*datastore.AddressInf
 	return nil
 
 }
+
+func getWarmIPTarget() int {
+	inputStr, found := os.LookupEnv("WARM_IP_TARGET")
```
nckturner: Instead of `WARM_IP_TARGET` which I don't think is very descriptive, I propose `NUM_WARM_IPS`, or `MAX_WARM_IPS` each of which signifies a number rather than a resource.

pkg/awsutils/awsutils.go
```diff
@@ -826,6 +829,36 @@ func (cache *EC2InstanceMetadataCache) GetENILimit() (int, error) {
 	return eniLimit, nil
 }
 
+// Allocate alloactes numIPs of IP address on a eni
+func (cache *EC2InstanceMetadataCache) AllocIPAddresses(eniID string, numIPs int) error {
```
nckturner: This function should be combined with `AllocAllIPAddress()`.

ipamd/ipamd.go
sdavids13: Out of curiosity how is this supposed to work if you assigned a value of 5? Right from the start there will be 5 secondary IPs available for pods to use, but let's say 3 pods are started on that particular machine... Does the pool of secondary IPs get expanded back out to 5, if so does it allocate three new addresses one at a time? Alternatively does the pool of available IPs become 2, then when those 2 IPs are eventually used then a new block of 5 IPs will be attached when needed. The naming for the former might be `NUM_WARM_IPS` while the later might be `IP_ALLOCATION_BLOCK_COUNT` or something along those lines.

liwenwu-amazon: if `WARM_IP_TARGET` is 5,
* at beginning, the available IPs is 5
* after 3 Pods assigned,  ipamD will try to allocate 3 more IPs  in the ipamD controller loops. ipamD will try to use 1 EC2 API if it is possible.  If it requires allocating 1 more ENIs, or EC2 control plane can not allocate 3 IP address at that time, ipamD controller loops will continuously retry to reach the `target` of 5 IPs

There are 2 reasons I choose WARM_IP_TARGET

* There is another configure knob `WARM_ENI_TARGET` which works similar but at ENI level.
  * the default is 1, e.g. if it is configured at 2, ipamD will try to make 2 ENI available. PR #68

* `...TARGET` means the `desired` value, and ipamD controller loops will try continuously to reach this `TARGET`.

sdavids13: Ok, I see `ipPoolMonitorInterval` is set to run every 5 seconds to check to see if it needs to increase the IP pool. Is that value perhaps set too low? Or if many pods are launched simultaneously will it all fall within the same time window to not cause excessive EC2 API calls.

devkid:
It would be nice to have any kind of documentation for that feature.

PR #130, opened 2018-07-10
Fix return path of NodePort traffic.

### LINKED ISSUE #75: NodePort not working properly for pods on secondary ENIs
When using nodeport services (with or without loadbalancer), if iptables redirects traffic to a local pod with an IP on a secondary interface, the traffic is dropped by the reverse path filter.

Everything seems to work OK because if the first SYN is dropped the client will retry (however queries load-balanced to the local pod take much longer) and will (probably) be sent to another host (or pod on primary interface).

This can be seen by logging martian packets. When traffic is sent to a local pod on a secondary interface, it will be dropped.

The reason is the following:
- traffic arrives on interface eth0 (for instance) with destination NodeIP:NodePort and is NATted in PREROUTING to PodIP:PortPort
- the reverse path for traffic from PodIP is through secondary ENI and not eth0
- rp_filter drops packet

To trigger the issue consistently:
- add externalTrafficPolicy: Local to the service definition so traffic arriving on NodePort will only use local pods
- make sure your pod is using an IP from a secondary ENI (but scheduling enough pods on the main interface)

fasaxc:
*Issue #, if available:*

Fixes #75

*Description of changes:*

Add iptables and routing rules that

- connmark traffic that arrives at the host over eth0
- restore the mark when the traffic leaves a pod veth
- force marked traffic to use the main routing table so that it
  exits via eth0.

* Testing performed

- Spun up a 2 node cluster and installed the plugin (and Calico with the `IptablesMangleAllowAction=RETURN`).
- Started a 2-pod nginx service, at least one ngingx pod was on a non-primary ENI (checked with `ip rule`)
- Exposed the nginx pods as a NodePort service
- Repeatedly checked connectivity from both nodes to the service, going via the other node's NodePort.  Worked as expected.  (Without fix, saw 50% connection failure rate.)
- Checked connectivity to the service from a busybox pod to its cluster IP and domain name.
- Killed/restarted the aws-node pod.
- Checked log for errors/ warnings.


fasaxc:
@liwenwu-amazon I think @lxpollitt is right, this PR only affects traffic that arrives at eth0 and then returns from a pod.  I think that:

- ALB/NLB direct to pod traffic will arrive over ethN and return via ethN.  (If it arrives at eth0 then it is going to a pod that is homed on eth0 so, although my change will be activated it'll be a no-op.)
- Any solution to the SG/subnet issue that uses ENIs will need to route NodePort traffic to the correct ENI, which means that it won't trigger my new rules so the behaviour will be correct.
- Doing direct ALB/NLB is desirable but it won't fix NodePorts, which are a basic feature of Kubernetes.  NodePorts can be used for other purposes in addition to ALB/NLB so I think we should fix them even if ALB/NLB is going to be improved.

liwenwu-amazon:
@fasaxc @lxpollitt , thanks for the PR.  Can you add more detail on how you verify the PR, such as
* what kind of tools you use to verify IP tables rules are invoked in the expected order by linux kernel? 
   * in case for node-port traffic for pods using non-eth0's secondary IPs
   * in other cases, non-node-port traffic     
* the analysis if there is any performance impact for non-node-port traffic 
* any extra guideline on how to troubleshoot

fasaxc:
@liwenwu-amazon the rules are static, and they amount to
```
iptables -t mangle -A PREROUTING -i eth0 -m addrtype --dst-type LOCAL --limit-iface-in -j CONNMARK --set-mark 0x80/0x80
iptables -t mangle -A PREROUTING -i eni+ -j CONNMARK --restore-mark --mask=0x80
```
and a routing rule that matches on the 0x80 mark bit to use the main routing table.

The mangle PREROUTING chain is executed just before the nat PREROUTING chain, before the routing decision.  

The first rule only takes effect if the incoming packet comes in via eth0 (`-i eth0`) and is going to an IP owned by eth0 (`--dst-type LOCAL --limit-iface-in`).  The latter check ensures that we're really dealing with a host IP and not a packet that is going to a pod that happens to be homed on eth0.  Traffic that comes in on a secondary ENI won't match so we won't set the connmark for those packets.

The second rule matches only packets that are leaving local pods and it restores the mark bit.  That means that the mark bit will only be set on packets that are part of a connection that started by coming in eth0 to a host IP.  Hence, only connections that arrived on eth0 and were NATted to a pod will have the mark bit set, which should only catch NodePort traffic.

This PR doesn't do anything for non-eth0 secondary IPs right now, those packets should be processed as before.  I wasn't sure what the desired behaviour of such IPs was.  The PR could be extended to use a mark value per ENI but that would require quite a bit of additional work.  If a user is already ENI-aware, presumably they'll only be accessing pods attached to that ENI so the routing should just work in that case?

We use similar rules in Calico and the impact of 1-2 rules of that kind should be negligible.  The non-NodePort traffic should fail the first match criteria on each rule, which amounts to a 4-byte string comparison.

For troubleshooting:

- I found it useful to use tcpdump on each ENI when diagnosing the issue originally.  We observed the correct packets (i.e. correct source and dest) being routed out of the wrong interface.
- The real rules have "AWS" in the comment, so you can see if they're being hit with `sudo iptables-save -c | grep AWS`, for example.
- Worth looking for any other uses of CONNMARK or the configured mark bit in the dataplane.  (I chose a bit for the default value that is not used by either kube-proxy or Calico.)

liwenwu-amazon:
@fasaxc  I am testing the change using https://kubernetes.io/docs/tasks/access-application-cluster/connecting-frontend-backend/ example on a cluster of 1 node.  I am not able to see
lb
 instance become healthy on ELB.

fasaxc:
@liwenwu-amazon That's odd, I'll try to reproduce here.  Please can you give me details of your set up?

- Was this using the AWS cloud provider to set up a LoadBalancer (or a manually configured ELB)?
- Was this an EKS system or set up using a different installer?
- Was the backing pod on the main ENI or another one?

fasaxc:
I was able to reproduce locally.  Previously I was testing on a kops cluster but after switching to EKS I hit issues with the fix.  I'll keep digging to see if I can get to the bottom of it.

fasaxc:
It looks like the difference between kops and EKS is that EKS has strict RPF filtering on eth0 whereas it was disabled on the kops cluster.  I find that the fix does seem to work if I set the RPF check on eth0 to "loose" mode with `sysctl -w net.ipv4.conf.eth0.rp_filter=2`.  Does that work for you?  Is it something that you'd consider accepting as part of the fix?

liwenwu-amazon:
@fasaxc thanks for the RPF finding.  Do you know why `unhealthy` instance target only happens to cluster which contains 1 node?  If a cluster have 2 nodes, both of them become `healthy`

liwenwu-amazon:
@fasaxc , can you  explain why change eth0 to "loose" mode make it working?  In another word, if it is not set (as it today in eks), which particular packet get dropped and make the instance `unhealthy`?  thanks

fasaxc:
@liwenwu-amazon With "strict" RPF, what happens is

- ELB packet arrives via **eth0**, gets DNATed to (local) pod IP on secondary ENI
- DNATted packet is RPF checked
  - kernel tries to find a route from <pod IP on secondary ENI> to <ELB IP>
  - it finds the route out of **eth1**
  - RPF check fails because strict mode requires that the reverse route is for the same interface.

With "loose" RPF, the check passes because the kernel accepts _any_ return route rather than requiring the return route to leave the same interface.

In my testing I used two nodes and 2 "frontend" pods, one on each host; in that scenario, I do understand why both nodes get marked as healthy.  This happens because the DNAT on each host chooses the frontend pod at random.  In addition, after an RPF failure, the kernel seems to drop the NAT state for the connection (so retried packets may go to a different DNAT rule).  That means that you get this scenario:

- remote host's SYN packet arrives via **eth0**, gets DNATed to (local) pod IP on secondary ENI
- RPF check fails, packet dropped
- remote host's resends same SYN packet (standard TCP behaviour)
- remote host's SYN packet arrives via **eth0**, gets DNATed to remote pod IP
- (since there are no special routing rules for remote pod IPs) RPF check passes and packet is forwarded to remote pod
- connection succeeds

In manual testing, I was seeing the SYN retries with tcpdump and the latency of a connection that went through this retry was 1s more than normal.  With an ELB it was showing the nodes as healthy but I was seeing random latency spikes through the ELB.

fasaxc:
I've updated the PR to set the "loose" setting on eth0. PTAL.

liwenwu-amazon:
@fasaxc  I am a bit confused on your notes why RPF check failed.
Here is my setup:
ELB (192.168.167.142) <----> Node (eth0, 192.168.132.34) <----> FrontEnd-Pod(192.168.157.136)
```
kubectl get pod -o wide
NAME                        READY     STATUS    RESTARTS   AGE       IP                NODE
frontend-766c875db4-wqr5p   1/1       Running   0          18h       192.168.157.136   ip-192-168-132-34.us-west-2.compute.internal
hello-7ff54bc875-4lfjq      1/1       Running   0          19h       192.168.154.231   ip-192-168-132-34.us-west-2.compute.internal
hello-7ff54bc875-4n429      1/1       Running   0          19h       192.168.131.110   ip-192-168-132-34.us-west-2.compute.internal
hello-7ff54bc875-9vqk7      1/1       Running   0          19h       192.168.173.131   ip-192-168-132-34.us-west-2.compute.internal
hello-7ff54bc875-gth7n      1/1       Running   0          19h       192.168.146.111   ip-192-168-132-34.us-west-2.compute.internal
hello-7ff54bc875-q4mbn      1/1       Running   0          19h       192.168.180.61    ip-192-168-132-34.us-west-2.compute.internal
hello-7ff54bc875-rwp57      1/1       Running   0          19h       192.168.161.31    ip-192-168-132-34.us-west-2.compute.internal
hello-7ff54bc875-tcdl5      1/1       Running   0          19h       192.168.159.100   ip-192-168-132-34.us-west-2.compute.internal
```
Here is route table for FrontEnd-Pod
```
ip route
default via 192.168.128.1 dev eth0 
169.254.169.254 dev eth0 
172.17.0.0/16 dev docker0 proto kernel scope link src 172.17.0.1 linkdown 
192.168.128.0/18 dev eth0 proto kernel scope link src 192.168.132.34 
192.168.131.110 dev eni6669baf3723 scope link 
192.168.146.111 dev eni4763e566de6 scope link 
192.168.154.231 dev enib36cd009927 scope link 
192.168.157.136 dev enibc52a89a1fc scope link <---FrontEnd-Pod
192.168.159.100 dev eni837ae39767c scope link 
192.168.161.31 dev enicea187a2214 scope link 
192.168.163.189 dev eni459f009cb4c scope link 
192.168.173.131 dev enia645ac38938 scope link 
192.168.180.61 dev eni6188ad2106a scope link 
```
ip rule show any packet destined to FrontEnd-Pod needs use main routing table
```
ip rule show
0:	from all lookup local 
512:	from all to 192.168.163.189 lookup main 
512:	from all to 192.168.173.131 lookup main 
512:	from all to 192.168.161.31 lookup main 
512:	from all to 192.168.131.110 lookup main 
512:	from all to 192.168.159.100 lookup main 
512:	from all to 192.168.180.61 lookup main 
512:	from all to 192.168.154.231 lookup main 
512:	from all to 192.168.146.111 lookup main 
512:	from all to 192.168.157.136 lookup main <----FrontEnd-Pod
1024:	not from all to 192.168.0.0/16 lookup main 
1024:	from all fwmark 0x80/0x80 lookup main 
1536:	from 192.168.173.131 lookup 2 
1536:	from 192.168.131.110 lookup 2 
1536:	from 192.168.159.100 lookup 2 
1536:	from 192.168.180.61 lookup 3 
1536:	from 192.168.154.231 lookup 3 
1536:	from 192.168.157.136 lookup 2 
32766:	from all lookup main 
32767:	from all lookup default 
```
Here is incoming ELB Health check flow
```
1. ELB --> Node eth0 (IP_SA=192.168.167.142, IP_DA=192.168.132.34, DA_PORT=32481)
2. Node NAT it  (IP_SA=192.168.132.34, IP_DA=192.168.157.136, DA_PORT=88)
3. route table should send it to enibc52a89a1fc
```
My understanding of RPF check is that it will drop the packet if packet is received from the interface where it should be the output interface for the IP-DA.   I don't quite follow why it fails PRF in this case.

spikecurtis:
@liwenwu-amazon strict RPF drops a packet if the routing lookup for the *reverse* path yields a different interface than it came in on. So, when it is NAT'ed to `(IP_SA=192.168.132.34, IP_DA=192.168.157.136)`, the kernel does a routing lookup for  `(IP_SA=192.168.157.136, IP_DA=192.168.132.34)`. The IP rules say it should use routing table 2

```
1536:	from 192.168.157.136 lookup 2 
```
which presumably would forward out a different interface than `eth0`, and thus it doesn't pass a strict RPF check.

A loose RPF check just verifies that it can route the reverse packet, but does not require that it goes out the same interface it came in, so would be successful in this case.

liwenwu-amazon:
@spikecurtis I thought routing table selection is done before DNAT, so it will use main routing table to perform RPF check.   Do you see any stats (e.g. netfilter stats) to prove your finding?  Also, do you have kernel code trace to prove kernel forwarding code will perform ip rule selection again during RPF checking? thanks

liwenwu-amazon:
@spikecurtis, @fasaxc  thanks for the reference. Interesting to know in this case it uses `main routing table` for destination lookup, and uses another `routing table` for RPF lookup.

For monitor/debug/audit purpose, do you know any statistics we can use for this change?  In another word,  shall we add any additional  to `/opt/cni/bin/aws-cni-support.sh`? thanks.

fasaxc:
The only diagnostic for the RPF check I'm aware of is the following sysctl:
```
net.ipv4.conf.all.log_martians=1 
```
It causes the kernel to log all packets that are dropped by the RPF check.  I haven't found any kernel stats for that.

For the new iptables rules that we added, I suggest including `iptables-save -c` output in the diagnostics dump.  That can be very useful; especially for spotting interactions between different apps that use iptables.

One slight correction to the health check flow; I think the SNAT happens after the routing decision so `IP_SA=192.168.132.34` should be `IP_SA=192.168.167.142` (the ELB's IP).

liwenwu-amazon:
@fasaxc  Thanks again for the notes.   The only concerns I have now is "disabling RPF check", where RPF is a security feature that can help to limit the malicious traffic and prevent IP address spoofing.

How about adding a configuration knob that

* for deployment that do *NOT* care about RPF security check and also that do *NOT* want to use NLB/ALB  for service/PodIP mapping, it can use this configuration knob to turn on node-port support
* for deployment that care about RPF security check and use NLB/ALB for service/PodIP mapping, they can disable this feature.

fasaxc:
OK, how about `AWS_VPC_IGNORE_STRICT_RP_FILTER=true`?

liwenwu-amazon:
@fasaxc how about use `AWS_VPC_CNI_NODE_PORT_SUPPORT`?   When `AWS_VPC_CNI_NODE_PORT_SUPPORT=true`, CNI will trigger the code in this PR, such as
* set RPF check false
* add 1 more IP table rules `1024:	from all fwmark 0x80/0x80 lookup main`
* add 2 more rules to IP mangle table

if `AWS_VPC_CNI_NODE_PORT_SUPPORT` is not set, it behaves same as today.

spikecurtis:
@liwenwu-amazon 

I think it's very important that nodePorts work correctly with the default settings, and would therefore advocate that the default value of this config flag is to enable the changes in this PR.

NodePorts are a part of standard Kubernetes networking, and if they don't work by default EKS and self-managed clusters that use the CNI plug in might not be considered conforming to the Kubernetes API spec.

While NLB/ALB with direct pod access is a good technology, the reality is that it will not be adopted overnight, nor will it be adopted by 100% of users. Having trouble getting NodePorts working could end up being a source of adoption friction for our users, and I think we should avoid it as much as possible.

I want to emphasize that this change does not *disable* RPF checks, it simply switches them from strict to loose.  Loose mode is the recommended mode "if using asymmetric routing or other complicated routing" according to the Linux man pages. https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt  It's also worth pointing out that unless the user has explicitly disable it, the AWS network will already check for the kind of IP spoofing that RPF is designed to prevent.

fasaxc:
@liwenwu-amazon I agree with @spikecurtis; I think it should default to making NodePorts work or users will get tripped up.  WDYT?

fasaxc:
I've updated the patch along those lines.

liwenwu-amazon:
@fasaxc can you add some unit-test too?  thanks

liwenwu-amazon:
@fasaxc @lxpollitt @spikecurtis I have following concerns with node-port as default:
* for each k8s service (e.g. [k8s frontend example](https://k8s.io/examples/service/access/frontend.yaml)),  Load Balancers needs to add `ALL` nodes in the cluster.  e.g. for a 5000 node cluster,  Load Balancer needs to have 5000 instance in its target group and constantly sending health-checks to all 5000 nodes.
* each k8s service will generate  a `ip_conntrack` state in `every node` in the cluster. This may increase `memory requirement` in every node in the cluster
*  health-check and traffic are sent multi-hops, not efficient and very difficult to debug:
      * for a endpoint (e.g. [nginx in frontend example](https://k8s.io/examples/service/access/frontend.yaml)) the incoming traffic can have any node IP address as source IP
      * all health-check and traffic can get sent to `any one of the node` in the cluster, SNATed with that node's IP and then forward to endpoint (nginx)
* for some deployment which requires multiple node groups and NO communication between nodes from different node group, these k8s service will trigger health-check traffic sent between nodes from different node group.

liwenwu-amazon:
Here is a comparison between node-port and using Pod VPC IP as target:
[git-node-port.pdf](https://github.com/aws/amazon-vpc-cni-k8s/files/2209880/git-node-port.pdf)

fasaxc:
@liwenwu-amazon I think we're all agreed that NLB/ALB/ELB -> Pod VPC IP is a good solution.  The reason for supporting NodePorts too is that they're a standard Kubernetes feature.  The user doesn't have to be using an NLB/ALB/ELB to use a NodePort.

fasaxc:
I've added some UT to the PR.

ahrkrak:
@liwenwu-amazon it sounds like your concerns are with the concept of service type node port vs load balancer. We are in agreement that the latter is preferable. However K8s includes both, so I think users will expect that AWS-CNI / EKS will support both.
Is there a specific concern you have with this PR, as opposed to the concept of node port in general?

liwenwu-amazon:
@ahrkrak @fasaxc  Thanks again for the PR.  I am still reviewing the code.  Also there is one outstanding debate on what's the default setting for AWS_VPC_CNI_NODE_PORT_SUPPORT.  
My argument for setting it to false is:
* using AWS NLB/ALB with Pods IP as target is much better solution
* using AWS NLB/ALB with Pods IP as target should able to solve the needs for all K8S application running on top of AWS that uses node-port (Please let me know if I missed anything here)

lxpollitt:
I'm not sure that people will regard EKS as a conformant cluster with this default setting.  I'm not 100% sure if it would fail the CNCF conformance test suite or not, but I can see no real downsides to make the default "node ports work".

pkg/networkutils/network.go
```diff
@@ -49,6 +51,19 @@ const (
...
 	// be installed and will be removed if they are already installed.
 	envExternalSNAT = "AWS_VPC_K8S_CNI_EXTERNALSNAT"
+
+	// envConnmark is the name of the environment variable that overrides the default connection mark, used to
+	// mark traffic coming from the primary ENI so that return traffic can be forced out of the same interface.
+	// Without using a mark, NodePort DNAT and our source-based routing do not work together if the target pod
+	// behind the node port is not on the main ENI.  In that case, the un-DNAT is done after the source-based
+	// routing, resulting in the packet being sent out of the pod's ENI, when the NodePort traffic should be
+	// sent over the main ENI.
+	envConnmark = "AWS_VPC_K8S_CNI_CONNMARK"
```
liwenwu-amazon: Is there any reason to make this configurable? In what usecases that we have to configure it different than default

pkg/networkutils/network.go
```diff
@@ -110,32 +109,51 @@ func (os *linuxNetwork) SetupHostNetwork(vpcCIDR *net.IPNet, primaryAddr *net.IP
...
 
+	// Configure the kernel's reverse path filter check on eth0 for "loose" filtering.  This is required because
+	// - NodePorts are exposed on eth0
+	// - The kernel's RPF check happens after incoming packets to NodePorts are DNATted to the pod IP.
+	// - For pods assigned to secondary ENIs, the routing table includes source-based routing.  When the kernel does
+	//   the RPF check, it looks up the route using the pod IP as the source.
+	// - Thus, it finds the source-based route that leaves via the secondary ENI.
+	// - In "strict" mode, the RPF check fails because the return path uses a different interface to the incoming
+	//   packet.  In "loose" mode, the check passes because some route was found.
+	f, err := os.OpenFile("/proc/sys/net/ipv4/conf/eth0/rp_filter", os.O_WRONLY, 0644)
```
liwenwu-amazon: Is there any API for this? (instead of directly manipulate /proc/....?  ) 
Can you also create a function for this?

pkg/networkutils/network.go
```diff
@@ -116,30 +122,33 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDR *net.IPNet, primaryAddr *net.IP)
...
-	f, err := os.OpenFile("/proc/sys/net/ipv4/conf/eth0/rp_filter", os.O_WRONLY, 0644)
-	if err != nil {
-		return errors.Wrapf(err, "failed to configure eth0 RPF check")
-	}
-	defer f.Close()
-	const rpFilterLoose = "2"
-	_, err = f.WriteString(rpFilterLoose)
-	if err != nil {
-		return errors.Wrapf(err, "failed to configure eth0 RPF check")
+	if nodePortSupportEnabled {
```
liwenwu-amazon: if NOT enabled, shall we revert this back to RPF?

pkg/networkutils/network_test.go
```diff
@@ -97,9 +107,164 @@ func TestSetupENINetwork(t *testing.T) {
...
+
+func TestSetupHostNetworkNodePortEnabled(t *testing.T) {
+	ctrl, mockNetLink, _, mockNS, mockIptables := setup(t)
+	defer ctrl.Finish()
+
+	var mockRPFilter mockFile
+	ln := &linuxNetwork{
+		useExternalSNAT:        true,
+		nodePortSupportEnabled: true,
+		mainENIMark:            0x80,
```
liwenwu-amazon: can u use defaultConnmark  instead of  0x80

pkg/networkutils/network.go
fasaxc: Most users shouldn't need to change this.   However, if the user is using another app that also uses mark bits (or their own iptables rules), this allows our mark bit to be changed to avoid conflicts.

pkg/networkutils/network.go
fasaxc: I was a little wary of doing that in case it was the user who set the value differently rather than the CNI plugin but it'd be easy to make that change if you think it's the right thing to do.

liwenwu-amazon: I am fine with current approach.  For debugging field problems, can you add following
* add a new command, e,g `curl http://localhost:61678/env-settings` which dumps out current setting for RPF, CNI_CONNMARKS 
* add this new command into /opt/cni/aws-cni-support.sh
* also modify /opt/cni/aws-cni-support.sh to save a copy of /proc/sys/net/ipv4/conf/eth0/rp_filter setting inside aws-cni-support.tar.gz

fasaxc:
@liwenwu-amazon I've updated the PR to include the diagnostics you requested.

liwenwu-amazon:
@fasaxc @spikecurtis , can you help to come up a e2e test which we can manually verify for now:
* when both `network policy` and  `node-port service` are configured,  both are working as expected.
* ideally, this should cover following cases:
    * sending Pod and receiving Pod are in 2 different worker nodes
    * sending Pod and receiving Pod are in `same` worker node

thanks

liwenwu-amazon:
@fasaxc ,  I am getting following (just use default setting) which does NOT seems right:
```
[root@ip-192-168-232-52 bin]# curl http://localhost:61678/v1/env-settings | python -m json.tool
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   100  100   100    0     0    100      0  0:00:01 --:--:--  0:00:01 20000
{
    "AWS_VPC_CNI_NODE_PORT_SUPPORT": "",
    "AWS_VPC_K8S_CNI_CONNMARK": "",
    "AWS_VPC_K8S_CNI_EXTERNALSNAT": ""
}
```

fasaxc:
I've updated the PR to print the active state of the environment variables instead of the raw text.

For a E2E tests, I think this should be sufficient:

- create 2-node cluster and external test node
- create server pod on each node, using non-default ENI
- create separate NodePort service for each server pod
- verify connectivity from each host to each NodePort on the primary ENI (local and remote)
- verify connectivity from external test node to NodePorts
- with network policy enabled (e.g. Calico policy-only mode installed and configured to expect `eni*` interfaces instead of `cali*` interfaces)
  - repeat test
  - add NetworkPolicy to pods to allow traffic from hosts via CIDR
  - repeat test
  - remove policy, install default-deny policy
  - verify traffic is dropped from other hosts (for Calico, the local host can always connect to itself)

liwenwu-amazon:
@fasaxc thanks for the E2E testcase!
Based on your suggestion, I have come up the following detail steps for testing this.  Please review it and see if it is accurate.
Also, can you share a network policy yaml which I can use for it?

Here are the detail steps I am going to use to test this:
* I am planning to use https://kubernetes.io/docs/tasks/access-application-cluster/connecting-frontend-backend/
* start out with 1 node (`t2.medium`) cluster first (I think this will force testing node-port feature)
    * modify [hello.yaml](https://k8s.io/examples/service/access/hello.yaml
) with `replicas : 12` 
    * `kubectl apply -f hello.yaml`
    * create `hello` service : `kubectl create -f https://k8s.io/examples/service/access/hello-service.yaml`
    * create frontend deployment and service: `kubectl create -f https://k8s.io/examples/service/access/frontend.yaml`
    * To test local to local test
          *`kubectl exec -ti hello-xxx sh` into one of the hello pod
          * verify `wget <elb's address>` returns index.html
    * To test remote to local test
          * verify curl <elb's address> from a external node
* scale up to 2 nodes and also scale `hello` pod up to 24, so some of `hello` pod get scheduled onto the new node
     * perform same tests as above
* deploy calico policy-engine add-on  `kubectl apply -f calico.yaml`
* `TODO:details-need-to-be added` add allow policy (?? yaml) and verify??
* `TODO:details-need-to-be-added`  add dis-allow policy (?? yaml) and verify??

fasaxc:
@liwenwu-amazon k8s is allow-by-default; this page explains how to install a default-deny policy (in the "Default Policies" section): https://kubernetes.io/docs/concepts/services-networking/network-policies/

Note: node ports and policy don't interact as you might hope.  kube-proxy SNATs the traffic to ensure that return traffic goes via the ingress node so kubernetes policy can actually see the ingress node as the source of the traffic.  Calico supports pre-NAT policy that allows for securing NodePorts but that is not exposed by the kubernetes API.  It has to be applied to the host instead of the workload since that is where the NAT happens.

liwenwu-amazon:
@fasaxc  thanks for the input.  Here is the yaml for policy part, please let me know if they are correct and if you have any comments/suggestions:

* After apply frontend and backend mentioned in   https://kubernetes.io/docs/tasks/access-application-cluster/connecting-frontend-backend/ to a `single node cluster`
* deploy calico policy engine add-on by `kubectl apply -f calico.yaml`
* create network policy such that:
    * frontend can communicate with backend
    * backend can NOT communicate with each other
     * backend can reach out external (e.g. www.yahoo.com)
```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: deny-hello2hello
  namespace: default
spec:
  podSelector:
    matchLabels:
      tier: backend
  policyTypes:
  - Ingress
  - Egress
  ingress:
  - from:
    - podSelector:
        matchLabels:
           tier: frontend
    - ipBlock:
        cidr: 0.0.0.0/0
        except:
        - 192.168.0.0/16 <-- VPC CIDR
  egress:
  - to:
    - podSelector:
        matchLabels:
           tier: frontend
    - ipBlock:
        cidr: 0.0.0.0/0
        except:
        - 192.168.0.0/16 <-- VPC CIDR
```
* here are the verification I have done:
  * find out dns name of frontend service
       * `kubectl describe svc frontend` and IP address
  * verify able to reach node-port service from outside cluster
       * curl http://<frontend's dns>, should return `{"message":"Hello"}` 
  * verify able to reach node-port service from a backend pod
       * kubectl exec -ti `hello-xxxx` sh
       * curl http://<frontend's IP> should return `{"message":"Hello"}`
  * verify backend can NOT communicate with another backend pod
       * kubectl exec -ti `hello-xxxx` sh
       * ping another Pod's IP, it should failed

liwenwu-amazon:
@spikecurtis  Thanks for suggestions.  Since  the `frontend` service only have 1 nginx Pod,  and  from a Pod (`kubectl exec  -ti <hello-pod> sh`) and do a `wget http://<node>: port` for every node in the cluster, I should have tested nodePort->local pod vs nodePort->remote pod,  right?

spikecurtis:
Yeah, if there is only one pod and you hit the node port on each node, you will exercise both nodePort->local pod vs nodePort->remote pod.  However, with only one pod, how do you ensure that the pod is on a secondary ENI?  The way I thought you were going to do it is to create more pods than can fit on a single ENI.

liwenwu-amazon:
@spikecurtis  Good catch!  Yes. I will need to deploy some  pods first to make sure IP of primary ENI are all used before deploy frontend Pod and sevice.

liwenwu-amazon:
@fasaxc @spikecurtis  Thank you very much for this PR!

PR #151, opened 2018-08-08
Choose correct VPC CIDR block even when more than 1 VPC CIDR block exists

andy-han:
- use "/vpc-ipv4-cidr-blocks/" instead of "/vpc-ipv4-cidr-block/" to get all VPC CIDR blocks, not just primary
- store CIDR blocks into string array
- loop through array and find correct VPC CIDR, based on localIP
- getter now always returns correct VPC CIDR, based on where localIP is
- tests to confirm correct CIDR is chosen and not just primary every time


andy-han:
sorry my commit history is a bit of a mess

PR #152, opened 2018-08-09
Choose correct VPC CIDR block even when more than 1 VPC CIDR block exists 

### LINKED ISSUE #35: Incorrect SNAT setup when multiple CIDRs on VPC
## Description 

I was using a VPC that had two IPv4 CIDRs and a subnet with only one of the CIDRs, this resulted in all traffic being dropped with any policy even when trying to match with a pod label.  Using tcpdump I observed on the host with the source pod that all traffic from the pod would be SNAT'ed before being sent to the destination host.

## Observed behavior

IP addresses being assigned to pods were in the subnet 172.32.0.0/16 while the SNAT iptables rule was setup for 172.31.0.0/16.  This resulted in all traffic between hosts being SNAT'ed which caused Calico network policy to block all traffic between hosts.

Here is the iptables SNAT rule:
```
-A POSTROUTING ! -d 172.31.0.0/16 -m comment --comment "AWS, SNAT" -m addrtype ! --dst-type LOCAL -j SNAT --to-source 172.32.0.27
```

## Expected behavior

SNAT will be setup correctly when there are more than one CIDR in a VPC.

andy-han:
*Issue #, if available:*

Fixes #35 

*Description of changes:*

Changes aim to make sure node level network configuration is set up in the correct VPC CIDR
- use "/vpc-ipv4-cidr-blocks/" instead of "/vpc-ipv4-cidr-block/" to get all VPC CIDR blocks, not just primary
- store CIDR blocks into string array
- loop through array and find correct VPC CIDR, based on localIP of EC2
- getter now always returns correct VPC CIDR, based on localIP 
- tests to confirm correct CIDR is chosen and not just primary every time

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -19,6 +19,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"net"
 
 	"github.com/pkg/errors"
 
@@ -42,7 +43,7 @@ const (
 	metadataMAC          = "mac"
 	metadataSGs          = "/security-group-ids/"
 	metadataSubnetID     = "/subnet-id/"
-	metadataVPCcidr      = "/vpc-ipv4-cidr-block/"
+	metadataVPCCidrs     = "/vpc-ipv4-cidr-blocks/"
 	metadataDeviceNum    = "/device-number/"
 	metadataInterface    = "/interface-id/"
 	metadataSubnetCIDR   = "/subnet-ipv4-cidr-block"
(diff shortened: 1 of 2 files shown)


liwenwu-amazon:
@andy-han can you add details how you verify your changes solves the issues?  thanks

liwenwu-amazon:
Here is a e2e test case which can be used to reproduce the issue and also be used to verify the fix:

* configure 2 CIDRs for a VPC, e.g. 10.0.0.0/16 and 10.1.0.0/16

* create a EKS k8s cluster using this VPC, following [EKS getting started guide](https://docs.aws.amazon.com/eks/latest/userguide/getting-started.html)

* create worker group1 using VPC CIDR1 (e.g. 10.0.0.0/16) using CloudFormation Template described in [EKS getting started guide](https://docs.aws.amazon.com/eks/latest/userguide/getting-started.html)

* create worker group2 using VPC CIDR2 (e.g. 10.1.0.0/16) using CloudFormation Template described in [EKS getting started guide](https://docs.aws.amazon.com/eks/latest/userguide/getting-started.html)

* use same AWS VPC route table (e.g. the route table used for CIDR1) for all subnets in CIDR1 and CIDR2 

* modify security groups to allow node group using CIDR1 to communicate with node group using CIDR2

* update aws-auth-cm.yaml with roles from both worker group1 and worker group2 cloudformation output, so that all nodes from CIDR1 and CIDR2 joins the cluster

* once all nodes joins, `kubectl apply -f hello.yaml`
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: hello
spec:
  selector:
    matchLabels:
      app: hello
      tier: backend
      track: stable
  replicas: 20 <-- pods are created among all nodes in CIDR1 and CIDR2
  template:
    metadata:
      labels:
        app: hello
        tier: backend
        track: stable
    spec:
      containers:
        - name: hello
          image: "gcr.io/google-samples/hello-go-gke:1.0"
          ports:
            - name: http
              containerPort: 80
```
* run `kubectl get pod -o wide` and find out the Pods running on CIDR1 node and CIDR2

* `kubectl exec -ti <Pod on a node of CIDR2> sh`  and ping a IP of a Pod running on node that uses CIDR1

* `ssh into the node using CIDR1` and do `tcpdump -i eth0 host <Pod's IP>`.
    * Without a fix, you should see IP source address are `NATed` with node's IP
    * With the right fix, you should see IP source address is the source Pod IP

liwenwu-amazon:
the issue has been fixed already #234

Issue #155, opened 2018-08-10
Minutely cronjob fails after swapping out workers with new ASG


hobbsh:
We use blue/green worker groups to update our AMIs for patching and other fixes. I've noticed that when pods are evicted during the draining, that the CNI starts having problems and the error rate of the cronjob skyrockets unless I delete and re-add it. I see these errors on multiple nodes and not consistently. The job succeeds many times between these errors. This is seemingly a dupe of #59 but I am running v1.1.0 here which supposedly fixed this issue? Our cronjob is not totally broken but our failed cronjob alert is flappy as hell now.

With `kubectl get events`, I see these errors:

> kubelet, ip-10-1-60-40.us-west-2.compute.internal    Failed create pod sandbox: rpc error: code = Unknown desc = NetworkPlugin cni failed to set up pod "REDACTED-1533940200-56dnc_lumo" network: add command: failed to setup network: setup NS network: failed to setup NS network: failed to Statfs "/proc/1577/ns/net": no such file or directory

> kubelet, ip-10-1-60-40.us-west-2.compute.internal    error killing pod: failed to "KillPodSandbox" for "7b36ece3-9ced-11e8-ad95-02af9b42ed2e" with KillPodSandboxError: "rpc error: code = Unknown desc = NetworkPlugin cni failed to teardown pod \"REDACTED-1533940440-z5rns_lumo\" network: rpc error: code = Unavailable desc = grpc: the connection is unavailable"

hobbsh:
@liwenwu-amazon I should be able to replicate this again and can send you the support tarball afterwards. Thanks

hobbsh:
I have tried to replicate this but have not been able to today. @tomweston [on slack](https://kubernetes.slack.com/archives/C09NXKJKA/p1534275684000100) seems to be having the same issue though. I will continue to try to replicate but hopefully his support logs can identify the issue.

lstoll:
We're seeing something similar - repro steps
* Scale node ASG to 0
* Create a pod directly in API, it'll sit in pending
* Scale node ASG up
* Pod will end up in Error state, with events showing items like:
```
  Normal   SandboxChanged          11m (x10 over 11m)  kubelet, ip-10-0-124-11.ec2.internal  Pod sandbox changed, it will be killed and re-created.
[...]

Issue #154, opened 2018-08-10
Inconsistent ENI count attached to instances running 1.1.0


hobbsh:
I have noticed an inconsistency in the number of ENI's attached to EKS nodes. I have not been able to correlate it to anything in particular - not hitting the ENI limit yet, doesn't seem related to an instance size but that was my initial thought. Seems like something was removed or added in 1.1.0 but the Changelog does not seem to call anything out related to this. 

In one VPC, all the m4.larges have 2 ENIs apiece so I assumed this is how the CNI is supposed to work. Then I spun up workers in a different VPC and some had 1 ENI, some had 2, and some had 3 (m5.xlarge instance size). So, I tried m4.xlarges and they all have 1 ENI - I would expect a larger instance to have more ENIs given it is going to be supporting more pods. 

Running `amazon-k8s-cni:1.1.0` on all the m4.xlarges and m5.xlarges but not on the m4.larges (running 1.0.0) which are the ones with an even distribution. If this is somehow normal/changed behavior or if I misconfigured something, it would be great to know. Additionally, if I can provide any information to help troubleshoot this issue, please let me know.

liwenwu-amazon:
@hobbsh in v1.1.0, the number of ENIs attached to the instance is triggered by number of concurrent running Pods running  on the instance.
* when a worker node first joins the cluster, there is only 1 ENI along with all of its addresses in the ENI
* when number of pods running on the node exceeds the number of addresses on a single ENI, the CNI backend start allocating a new ENI and start using following allocation scheme:

```
Without any configuration, ipamD always try to keep one extra ENI.  

For example, a m4.4xlarge node can have up to 8 ENIs, and each ENI can have up to 30 IP addresses. ( https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html ).   

* If the number of current running Pods is between 0 to 29, ipamD will allocate one more eni. And Warm-Pool size is 2 eni * (30 -1) = 58 
* If the number of current running Pods is between 30 and 58, ipamD will allocate 2 more eni. And Warm-Pool size is 3 eni * (30 -1) = 87
```

hobbsh:
@liwenwu-amazon thanks for the clarification. I see this is laid out in more complex language in the proposal - it would be great to add what you just said into the readme. Also, what do you mean by "without any configuration, ipamD will always try to keep one extra ENI" - does this mean that by deafult, all workers in an m4.xlarge should have at least 2 ENI?

[...]

PR #169, opened 2018-09-05
Maintain the right number of  ENIs and its IP addresses in WARM-IP pool

liwenwu-amazon:
*Issue #154, #155 *

*Description of changes:*
Today, curMaxAddrsPerENI is set to 1 initially and it is only set to the correct value after allocating the 1st ENI.  This can cause `nodeIPPoolLow()` and `nodeIPPoolTooHIgh()` returns incorrect value.

### DIFF
--- a/ipamd/ipamd.go
+++ b/ipamd/ipamd.go
@@ -108,8 +108,8 @@ type IPAMContext struct {
 	dockerClient  docker.APIs
 	networkClient networkutils.NetworkAPIs
 
-	currentMaxAddrsPerENI int
-	maxAddrsPerENI        int
+	currentMaxAddrsPerENI int64
+	maxAddrsPerENI        int64
 	// maxENI indicate the maximum number of ENIs can be attached to the instance
 	// It is initialized to 0 and it is set to current number of ENIs attached
 	// when ipamD receives AttachmentLimitExceeded error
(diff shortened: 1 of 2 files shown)


edmorley:
Hi! 

The description of this PR at first glance makes it seem like a small bug fix, however it in fact appears to be the cause of CNI plugin v1.2.0+ having double the IP address usage requirements of the prior v1.1.0 (under our workflow at least), which caused us to run out of free IP addresses until we adjusted our subnets accordingly (which in itself required an EKS cluster rebuild/migration). 

Specifically after this change, our worker nodes went from being allocated one ENI (which was only partly used, since we use smaller instances with fewer pods per instance) to always having two ENIs allocated (with the second being entirely unused), with both ENIs having the full 10 IP addresses allocated, doubling usage. (Whilst we plan on trying out `WARM_IP_TARGET` in the future, it sounds like it will need some testing/tuning, so we had stuck with the out of the box `WARM_ENI_TARGET` for now.)

As such, please could significant changes like these (that have the chance of being breaking) be called out more clearly in the PR description and changelog in the future? :-)

mogren:
@edmorley Thanks, good call-out. And like you said, `WARM_IP_TARGET` is another quite significant change, and it will only work well if the pod-churn is not too high. If pods are scheduled and terminated at a high and variable rate, the calls to EC2 will get throttled, and attaching new IPs and ENIs will take a long time.

Issue #168, opened 2018-09-05
CNI plugin went into CrashLoopBackoff while handing pod updates


rtripat:
I have a dev EKS cluster (1.10) on which I had the AWS CNI plugin installed and running for ~  16 days. I deleted and recreated the deployment of my application which resulted in the AWS CNI plugin to CrashLoopBackoff with the following stack trace. Unfortunately, I don't have access to the EC2 instance anymore to gather more diagnostic data. 

```
NAMESPACE     NAME                        READY     STATUS             RESTARTS   AGE
kube-system   aws-node-6729d              0/1       CrashLoopBackOff   15         16d
kube-system   aws-node-6kv9t              0/1       CrashLoopBackOff   16         16d
kube-system   aws-node-nc8fq              0/1       CrashLoopBackOff   16         16d
kube-system   kube-dns-75fddcb66f-jrbxm   3/3       Running            0          19d
kube-system   kube-proxy-dtfgg            1/1       Running            0          16d
kube-system   kube-proxy-jnvnz            1/1       Running            0          16d
kube-system   kube-proxy-kzc8s            1/1       Running            0          16d
```

```
=====Starting installing AWS-CNI =========
=====Starting amazon-k8s-agent ===========
ERROR: logging before flag.Parse: W0823 19:16:16.738062      12 client_config.go:533] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
ERROR: logging before flag.Parse: E0823 19:16:42.519233      12 runtime.go:66] Observed a panic: &runtime.TypeAssertionError{interfaceString:"interface {}", concreteString:"", assertedString:"*v1.Pod", missingMethod:""} (interface conversion: interface {} is nil, not *v1.Pod)
/home/ubuntu/workspace/src/github.com/aws/amazon-vpc-cni-k8s/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:72
/home/ubuntu/workspace/src/github.com/aws/amazon-vpc-cni-k8s/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:65
/home/ubuntu/workspace/src/github.com/aws/amazon-vpc-cni-k8s/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:51
/usr/local/go/src/runtime/asm_amd64.s:509
/usr/local/go/src/runtime/panic.go:491
/usr/local/go/src/runtime/iface.go:172
[...]

PR #170, opened 2018-09-06
Check to make sure it is a Pod object

liwenwu-amazon:
*Issue #168, :*

*Description of changes:*
Check to make sure the object is a Pod type

### DIFF
--- a/pkg/k8sapi/discovery.go
+++ b/pkg/k8sapi/discovery.go
@@ -208,7 +208,7 @@ func (d *Controller) handlePodUpdate(key string) error {
 	}
 
 	pod, ok := obj.(*v1.Pod)
-	if !ok {
+	if !ok || pod == nil {
 		log.Errorf("updated object received was not a pod: %+v", obj)
 		return errors.New("received a non-pod object update")
 	}


pkg/k8sapi/discovery.go
```diff
@@ -208,7 +208,7 @@ func (d *Controller) handlePodUpdate(key string) error {
 	}
 
 	pod, ok := obj.(*v1.Pod)
-	if !ok {
+	if !ok || pod == nil {
```
nckturner: What circumstances cause pod to be nil?

liwenwu-amazon: @nckturner it happens once in @rtripat setup once.  I am not able to reproduce it.

nckturner: Ok well it looks to me like it might help prevent #168 but I'm wondering why its receiving `nil` in the first place.

mattlandis: @ofiliz 

I think this is OK.
A nil interface is != nil but a nil interface converted to  struct is nil and pod is a struct.

https://play.golang.org/p/uAZJG2_mUuL

Unless I am missing something.

mattlandis: lgtm

I agree with nic that we should investigate further why we can get a nil pod here if exists is true.  Somewhere we are storing nil in the datastore.

Issue #122, opened 2018-07-02
Add better error message indicate why cni ipamD is not starting


liwenwu-amazon:
Today, in CNI daemonSet(aws-node) whenever ipamD restart,   it query kubernetes API server about Pods already running on the node.  If it can not reach kubernetes API server, ipamD will exit and you will see following logs in the /var/log/aws-routed-cni/ipamd.log.xxx

```
2018-07-02T15:00:33Z [INFO] Starting L-IPAMD 1.0.0 ...
2018-07-02T15:00:33Z [INFO] Testing communication with server

..
2018-07-02T15:00:33Z [INFO] Starting L-IPAMD 1.0.0 ...
2018-07-02T15:00:33Z [INFO] Testing communication with server

```

ipamD needs print out explicit error that it failed due to it can NOT communicate with API server.

To verify security groups are configured correctly between worker node and kubernetes API server, you can run following commands:
```
# find out kubernetes service IP
kubectl get svc
NAME         TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)   AGE
kubernetes   ClusterIP   10.100.0.1   <none>        443/TCP   24d

#verify  worker node can reach port 443 of master 
telnet 10.100.0.1 443
Trying 10.100.0.1...
Connected to 10.100.0.1.
Escape character is '^]'.
```

kuroneko25:
I think we are also running into this issue and maybe this one #104. But I don't believe it's because of security groups. I suspect it has something to do with service accounts or RBAC but right now it's extremely difficult to tell what the actual failure is since there is nothing in `ipamd.log` or the container log to help us debug. 

All I'm seeing is the crash loop output you mentioned above and this single line in the container log:

`ERROR: logging before flag.Parse: W0615 02:36:27.560592       9 client_config.go:533] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.` 

which is probably expected since you are invoking `clientcmd.BuildConfigFromFlags` but not explicitly setting `apiserver` or `kubeconfig`.

liwenwu-amazon:
@kuroneko25 , is this a EKS cluster?

kuroneko25:
Yes.

liwenwu-amazon:
on your worker node, are you able to get 443 port of master.  What's the output of  on your worker node
```
telnet 10.100.0.1 443
```

kuroneko25:
Should I try with the real master IP or in cluster VIP?

[...]

PR #174, opened 2018-09-10
Add more error messages during initialization

liwenwu-amazon:
*Issue #122 , if available:*

*Description of changes:*
One of the common scenario that CNI ipamD fails to start is that it can not communicate with K8S master. Here are 2  possible root causes:

* security groups  are not configured correctly on the across-account ENI attached to the master.
* or worker node HTTP proxy setting prevent ipamD communicate with master

This PR adds more error msgs when this happens.

### DIFF
--- a/pkg/k8sapi/discovery.go
+++ b/pkg/k8sapi/discovery.go
@@ -88,6 +88,9 @@ func CreateKubeClient(apiserver string, kubeconfig string) (clientset.Interface,
 	log.Infof("Testing communication with server")
 	v, err := kubeClient.Discovery().ServerVersion()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/k8sapi/discovery.go
```diff
@@ -88,6 +88,8 @@ func CreateKubeClient(apiserver string, kubeconfig string) (clientset.Interface,
 	log.Infof("Testing communication with server")
 	v, err := kubeClient.Discovery().ServerVersion()
 	if err != nil {
+		log.Infof("Failed to communicate with K8S Server. Please check instance security groups or http proxy setting")
+		fmt.Printf("Failed to communicate with K8S Server. Please check instance security groups or http proxy setting")
```
nckturner: Use `fmt.Sprintf()` to create the string and pass it to both.

Issue #188, opened 2018-10-01
CRD required even with `AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG` set to false


deiwin:
I upgraded to v1.2, but didn't create the `eniconfigs.crd.k8s.amazonaws.com` CRD, because I left `AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG` unset, which defaults to `false`, according to docs. `aws-node` pods now fail to start with the error message logged below. Why is the existence of the CRD a requirement when this feature is not used?

```
$ kubectl -n kube-system log aws-node-xt646
=====Starting installing AWS-CNI =========
=====Starting amazon-k8s-agent ===========
ERROR: logging before flag.Parse: W1001 07:34:01.315798      13 client_config.go:533] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
time="2018-10-01T07:34:01Z" level=error msg="failed to initialize service object for operator metrics: OPERATOR_NAME must be set"
time="2018-10-01T07:34:01Z" level=error msg="failed to get resource client for (apiVersion:crd.k8s.amazonaws.com/v1alpha1, kind:ENIConfig, ns:): failed to get resource type: failed to get the resource REST mapping for GroupVersionKind(crd.k8s.amazonaws.com/v1alpha1, Kind=ENIConfig): no matches for kind \"ENIConfig\" in version \"crd.k8s.amazonaws.com/v1alpha1\""
panic: failed to get resource type: failed to get the resource REST mapping for GroupVersionKind(crd.k8s.amazonaws.com/v1alpha1, Kind=ENIConfig): no matches for kind "ENIConfig" in version "crd.k8s.amazonaws.com/v1alpha1"

goroutine 15 [running]:
github.com/aws/amazon-vpc-cni-k8s/vendor/github.com/operator-framework/operator-sdk/pkg/sdk.Watch(0x16f7a6a, 0x1e, 0x16dbe38, 0x9, 0x0, 0x0, 0x12a05f200, 0x0, 0x0, 0x0)
        /usr/src/app/src/github.com/aws/amazon-vpc-cni-k8s/vendor/github.com/operator-framework/operator-sdk/pkg/sdk/api.go:49 +0x488
github.com/aws/amazon-vpc-cni-k8s/pkg/eniconfig.(*ENIConfigController).Start(0xc42043a4c0)
[...]

PR #192, opened 2018-10-03
Do not watch eniconfig CRD if cni is not configured to use pod config

liwenwu-amazon:
*Issue #188 , if available:*

*Description of changes:*


aleerizw:
Are you guys planning a new release? We can't really use the 1.2 version without this change.

dcarley:
> We can't really use the 1.2 version without this change.

If it helps, it did work for me when applying the complete config including the CRD: https://github.com/aws/amazon-vpc-cni-k8s/blob/v1.2.0/config/v1.2/aws-k8s-cni.yaml

PR #194, opened 2018-10-05
Ignore ENIs returning 404 on metadata request

deiwin:
*Issue #, if available:*

*Description of changes:*

This fixes an issue where sometimes new ENIs cannot be attached to an instance,
if an ENI was recently detached from the same instance. There's been cases
where this scenario causes new ENI attaching to be impossible for more than
5 minutes, causing new pods to fail to start on that node.

An example failure scenario:
* More than enough IPs available in the warm IP pool
* An ENI is deleted/detached
* Many CronJobs get scheduled on the node
* Not enough IPs available in the warm IP pool
* An ENI is added/attached
* IP allocation from the recently attached ENI fails for 5+ minutes because:
  * `network/interfaces/macs/` includes the recently detached ENI's mac, but
  * `network/interfaces/macs/<MAC>/device-number/` returns 404 Not Found
* New pods don't get IPs, because the IPs from the recently attached ENI are
  not available until the metadata cache clears

<details>
  <summary>Click to expand annotated log excerpt showing the issue in action</summary>

```
# Four ENIs
2018-10-04T22:54:22Z [DEBUG] Total number of interfaces found: 4
2018-10-04T22:54:22Z [DEBUG] Found eni: eni-02900675dacdab027, mac 06:4c:dd:d0:be:a8, device 0
2018-10-04T22:54:22Z [DEBUG] Found eni: eni-0b92f047b2e22725d, mac 06:8e:da:83:de:22, device 4
2018-10-04T22:54:22Z [DEBUG] Found eni: eni-0ec6a9cacf25ab7a3, mac 06:8f:44:b6:4d:7a, device 3
2018-10-04T22:54:22Z [DEBUG] Found eni: eni-086fb1ba2dd981648, mac 06:f9:4f:21:d9:94, device 2

# One can be removed
2018-10-04T22:54:27Z [DEBUG] FreeENI: found a deletable ENI eni-0b92f047b2e22725d
2018-10-04T22:54:27Z [INFO] FreeENI eni-0b92f047b2e22725d: IP address pool stats: free 14 addresses, total: 42, assigned: 27
2018-10-04T22:54:27Z [DEBUG] Start freeing eni eni-0b92f047b2e22725d

# Takes a bit to remove
2018-10-04T22:54:33Z [DEBUG] Not able to delete eni yet (attempt 2/20): InvalidParameterValue: Network interface 'eni-0b92f047b2e22725d' is currently in use.
[...]

PR #206, opened 2018-10-19
Add logs and metrics for cooling IP count

deiwin:
*Issue #, if available:*

*Description of changes:*

When a node is near full capacity in terms of IP addresses used, then logs can
become confusing for understanding why ipamD was unable to assign an IP to
a new pod. For example, the modified test case would previously report the
following in logs:

```
[Debug] UnAssignIPv4Address: IP address pool stats: total:1, assigned 1, pod(Name: pod-1, Namespace: ns-1, Container container-1)
[Info] UnAssignIPv4Address: pod (Name: pod-1, NameSpace ns-1 Container container-1)'s ipAddr 1.1.1.1, DeviceNumber1
[Debug] AssignIPv4Address: IP address pool stats: total:1, assigned 0
[Info] DataStore has no available IP addresses
```

The 3rd line indicates that there's an IP available, but the 4th one concludes
that no IPs are available. It is correct that no IPs are available, because all
unused IPs are in ["cooling mode"][1]. The problem is that that is not obvious
from the logs. With these changes, the logs for the test are changed to the
following:

```
[Debug] UnAssignIPv4Address: IP address pool stats: total: 1, assigned: 1, cooling: 0, pod(Name: pod-1, Namespace: ns-1, Container container-1)
[Info] UnAssignIPv4Address: pod (Name: pod-1, NameSpace ns-1 Container container-1)'s ipAddr 1.1.1.1, DeviceNumber1
[Debug] AssignIPv4Address: IP address pool stats: total: 1, assigned: 0, cooling: 1
[Info] DataStore has no available IP addresses
```

This makes it clear why IP allocation failed.

[1]: https://github.com/aws/amazon-vpc-cni-k8s/blob/3735b4a5c38524504cf11429bd731fa833256d2f/docs/cni-proposal.md#pod-ip-address-cooling-period

### DIFF
--- a/ipamd/datastore/data_store.go
+++ b/ipamd/datastore/data_store.go
@@ -77,6 +77,12 @@ var (
 			Help: "The number of IP addresses assigned to pods",
 		},
 	)
+	coolingIPs = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "cooling_ip_addresses",
+			Help: "The number of IP addresses in cooling mode",
+		},
+	)
 	prometheusRegistered = false
 )
 
@@ -122,6 +128,7 @@ type PodIPInfo struct {
 type DataStore struct {
 	total      int
 	assigned   int
+	cooling    int
 	eniIPPools map[string]*ENIIPPool
 	podsIP     map[PodKey]PodIPInfo
 	lock       sync.RWMutex
@@ -145,6 +152,7 @@ func prometheusRegister() {
 		prometheus.MustRegister(enis)
 		prometheus.MustRegister(totalIPs)
 		prometheus.MustRegister(assignedIPs)
+		prometheus.MustRegister(coolingIPs)
 		prometheusRegistered = true
 	}
 }
@@ -184,9 +192,7 @@ func (ds *DataStore) AddENIIPv4Address(eniID string, ipv4 string) error {
 	ds.lock.Lock()
 	defer ds.lock.Unlock()
 
-	log.Debugf("Adding ENI(%s)'s IPv4 address %s to datastore", eniID, ipv4)
-	log.Debugf("IP Address Pool stats: total: %d, assigned: %d",
-		ds.total, ds.assigned)
+	ds.withStats(log.Debugf, "Adding ENI(%s)'s IPv4 address %s to datastore", eniID, ipv4)
 
 	curENI, ok := ds.eniIPPools[eniID]
 	if !ok {
@@ -212,9 +218,7 @@ func (ds *DataStore) AddENIIPv4Address(eniID string, ipv4 string) error {
 func (ds *DataStore) DelENIIPv4Address(eniID string, ipv4 string) error {
 	ds.lock.Lock()
 	defer ds.lock.Unlock()
-	log.Debugf("Deleting ENI(%s)'s IPv4 address %s from datastore", eniID, ipv4)
-	log.Debugf("IP Address Pool stats: total: %d, assigned: %d",
-		ds.total, ds.assigned)
+	ds.withStats(log.Debugf, "Deleting ENI(%s)'s IPv4 address %s from datastore", eniID, ipv4)
 
 	curENI, ok := ds.eniIPPools[eniID]
 	if !ok {
@@ -246,8 +250,7 @@ func (ds *DataStore) AssignPodIPv4Address(k8sPod *k8sapi.K8SPodInfo) (string, in
 	ds.lock.Lock()
 	defer ds.lock.Unlock()
 
-	log.Debugf("AssignIPv4Address: IP address pool stats: total:%d, assigned %d",
-		ds.total, ds.assigned)
+	ds.withStats(log.Debugf, "AssignIPv4Address")
 	podKey := PodKey{
 		name:      k8sPod.Name,
 		namespace: k8sPod.Namespace,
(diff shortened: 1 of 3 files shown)


mogren:
Thanks! 

Could you please squash the commits into one?

deiwin:
Yes, of course. Done.

Just wondering, though, are you not able to [do so yourself, using GitHub merge button's dropdown selection](https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits)?

Also, I would appreciate feedback on #194, which is more critical for us.

deiwin:
Yes, of course. Done.

Just wondering, though, are you not able to [do so yourself, using GitHub merge button's dropdown selection](https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits)?

Also, I would appreciate feedback on #194, which is more critical for us.

ipamd/datastore/data_store.go
```diff
@@ -321,9 +323,31 @@ func (ds *DataStore) assignPodIPv4AddressUnsafe(k8sPod *k8sapi.K8SPodInfo) (stri
...
+				curTime.Sub(addr.unAssignedTime) <= addressCoolingPeriod {
+				count++
+			}
+		}
+	}
+	return count
+}
+
+func (ds *DataStore) withStats(logf func(string, ...interface{}), format string, a ...interface{}) {
+	total, assigned, cooling := ds.GetStats()
```
liwenwu-amazon: This will have `big performance` impact!  It can NOT loop through the all IP addresses  in ENI for every stats.
Also, the cooling IP only last 30 seconds.  Not sure if this is useful.

deiwin: > This will have `big performance` impact! It can NOT loop through the all IP addresses in ENI for every stats.

I understand your concern and thought about this as well, but in testing this on a cluster using m5.xlarge instances the impact on performance has been negligible. See the following graph. The update is clearly visible in the middle.

screen shot 2018-10-23 at 10 05 01

> Also, the cooling IP only last 30 seconds. Not sure if this is useful.

I'm making this change because it is useful for us. I spent half a day chasing ghosts, because the logs say (as explained in the PR description) that there's IPs available, but IP allocation fails. Also, you can see on the graph below, that in a cluster that uses CronJobs, some IPs are **always** in cooling mode (the thin purple area at the bottom). 

screen shot 2018-10-23 at 10 01 35
  
However, if you will not accept the change as is, then there's a few alternatives:

1. Increment a cooling stat every time an IP is unallocated and then decrement it 30 seconds later.
  a) This will work now, but is not very future proof. If the behavior for cooling IPs gets more complex, then this has to also be kept up to date, which might be difficult.
2. Re-calculate the stat every time an IP is unallocated and then 30 seconds later, by looping through all IPs.
  a) This will avoid looping through all IPs for every log message, but will still loop twice every  time an IP is unallocated. A compromise between the other two options.

I've gone ahead and implemented alternative nr 1, assuming that that will be your preference.

liwenwu-amazon:
@deiwin would it help debugging if we add a debug msg if it skip those IPs in `cooling` mode?
```

func (ds *DataStore) assignPodIPv4AddressUnsafe(k8sPod *k8sapi.K8SPodInfo) (string, int, error) {
        ...
      // Add a debug msg here if curTime.Sub(addr.unAssignedTime)  < addressCoolingPeriod

        if !addr.Assigned && k8sPod.IP == "" && curTime.Sub(addr.unAssignedTime) > addressCoolingPeriod { 
                                // This is triggered by a pod's Add Network command from CNI plugin
                                ds.assigned++
```

deiwin:
Yes, it would help. However, to confirm that all IPs are accounted for, I would then have to count all the debug messages, add that count to the assigned IP count and then compare it to the total IP count. In addition, that would leave out the metrics, which means I can't create a monitor for `available IP count == 0`.

Is there a problem with the PR as proposed? I've addressed your performance concern.

ipamd/datastore/data_store.go
liwenwu-amazon: @deiwin , 

 I am worrying about possible too many 30sec timers triggered by each IP-unallocation.  I am also worrying about the  "locking" when trying to decrement in those timeout routines and "locks" can introduce delay for allocating a IP for a pod

liwenwu-amazon:
@deiwin 

You can also add metrics along with the debug commands.   for example,
 * number of cooling IP skipped  for a successful allocation
* number of cooling IP skipped for an un-successful allocation

deiwin:
> I am worrying about possible too many 30sec timers triggered by each IP-unallocation.

The maximal number of IPs for the largest nodes is 737. Quickly creating and deleting all of the pods on such a node would create in the order of thousands of goroutines. This shouldn't be a problem at all for Go, AFAIK.  

> I am also worrying about the "locking" when trying to decrement in those timeout routines and "locks" can introduce delay for allocating a IP for a pod.

Oh, right. The decrement should be wrapped with [`atomic.AddInt64`](https://golang.org/pkg/sync/atomic/#AddInt64). Right now it doesn't lock at all and isn't safe. But again, I'm not sure if that has any impact at all on overall performance.

> You can also add metrics along with the debug commands.

If you prefer implementing it that way, then I would be OK with that.

deiwin:
Closing as there seems to be no interest in this

Issue #167, opened 2018-09-04
CNI does not support jumbo packets


oded-dd:
When using amazon CNI it seems that the eni created for the POD does not support jumbo packets

```
eni24e1650925c@if3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default
    link/ether 3a:76:a2:dc:c0:88 brd ff:ff:ff:ff:ff:ff link-netnsid 0
```

While the node itself does supports
```
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9001 qdisc mq state UP mode DEFAULT group default qlen 1000
    link/ether 06:db:1a:40:a5:1c brd ff:ff:ff:ff:ff:ff
```

oded-dd:
Any news about it? It's quite critical for us

hmizuma:
Primary interface is configured by DHCP for jumbo frames. Secondary ENIs are not handled by DHCP so changes needed to configure for jumbo frames after creation.

yuriipolishchuk-finsight:
@hmizuma moreover MTU is hardcoded here: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/plugins/routed-eni/driver/driver.go#L41

simplytunde:
is their a  reason why jumbo packet is not configured for the secondary interfaces?

PR #210, opened 2018-10-23
Configure MTU of ENI and veths to 9001

hmizuma:
*Issue #, if available:*
#167 

*Description of changes:*
* Add LinkSetMTU wrapper and  mocks
* Add LinkSetMTU call to ENI setup and network test
* Change hardcoded veth MTU to 9001

### DIFF
--- a/pkg/networkutils/network.go
+++ b/pkg/networkutils/network.go
@@ -70,6 +70,9 @@ const (
 	// - kube-proxy uses 0x0000c000
 	// - Calico uses 0xffff0000.
 	defaultConnmark = 0x80
+
+	// MTU of ENI - veth MTU defined in plugins/routed-eni/driver/driver.go
+	ethernetMTU = 9001
 )
 
 // NetworkAPIs defines the host level and the eni level network related operations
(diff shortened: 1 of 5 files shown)


pkg/networkutils/network.go
```diff
@@ -70,6 +70,8 @@ const (
 	// - kube-proxy uses 0x0000c000
 	// - Calico uses 0xffff0000.
 	defaultConnmark = 0x80
+
+	ethernetMTU = 9001
```
liwenwu-amazon: Can you add a comment here to indicate that 9001 also need to be defined in driver.go?

pkg/networkutils/network_test.go
```diff
@@ -44,6 +44,7 @@ const (
 	testeniIP        = "10.10.10.20"
 	testeniMAC       = "01:23:45:67:89:ab"
 	testeniSubnet    = "10.10.0.0/16"
+	testMTU          = 9001
```
liwenwu-amazon: same here, can u also add a comment that 9001 needs to be defined in network.go. thanks

pkg/networkutils/network.go
hmizuma: Updated to add comments where MTU is defined.

xrl:
@gtaylor yes, that's how I understand this issue. Only the primary, auto-created NIC gets jumbo frames by default. This code makes it so all NICs get jumbo frames.

PR #223, opened 2018-11-08
Add retry for plumbing route entry

liwenwu-amazon:
*Issue #, if available:*

*Description of changes:*


pkg/networkutils/network.go
```diff
@@ -454,12 +460,36 @@ func setupENINetwork(eniIP string, eniMAC string, eniTable int, eniSubnetCIDR st
...
+						return errors.Wrapf(err,
+							"eni network setup: failed unable to add route %s/0 via %s table %d",
+							r.Dst.IP.String(), gw.IP.String(), eniTable)
+					}
+					log.Debugf("Not able to add route route %s/0 via %s table %d (attempt %d/%d)",
+						r.Dst.IP.String(), gw.IP.String(), eniTable,
+						retry, maxRetryRouteAdd)
+					time.Sleep(retryRouteAddInterval)
+				} else if isRouteExistsError(err) {
+					if err := netlink.RouteReplace(&r); err != nil {
```
mogren: If this replace works, shouldn't we just `break` the loop and log a success?

liwenwu-amazon: good catch!

pkg/networkutils/network.go
```diff
@@ -454,12 +460,37 @@ func setupENINetwork(eniIP string, eniMAC string, eniTable int, eniSubnetCIDR st
...
+						return errors.Wrapf(err, "eni network setup: unable to replace route entry %s", r.Dst.IP.String())
+					}
+					break
```
mogren: Maybe a debug log as well? Something like
`log.Debugf("Successfully replaced route to be %s/0", r.Dst.IP.String())`

pkg/networkutils/network.go
```diff
@@ -454,12 +460,37 @@ func setupENINetwork(eniIP string, eniMAC string, eniTable int, eniSubnetCIDR st
...
+				} else {
+					return errors.Wrapf(err, "eni network setup: unable to add route %s/0 via %s table %d",
+						r.Dst.IP.String(), gw.IP.String(), eniTable)
+				}
+			} else {
+				log.Debugf("Successfully added route route %s/0 via %s table %d", r.Dst.IP.String(), gw.IP.String(), eniTable)
```
mogren: "route" is there twice.

Issue #166, opened 2018-09-03
Error message: failed to setup network: setup NS network: failed to add host route: file exists


max-rocket-internet:
In pod events:

```
Events:
  Type     Reason                  Age   From                                 Message
  ----     ------                  ----  ----                                 -------
  Normal   Scheduled               4m    default-scheduler                    Successfully assigned us01-stg-spinnaker-spinn-halyard-0 to ip-10-0-22-21.ec2.internal
  Normal   SuccessfulMountVolume   4m    kubelet, ip-10-0-22-21.ec2.internal  MountVolume.SetUp succeeded for volume "halyard-config"
  Normal   SuccessfulMountVolume   4m    kubelet, ip-10-0-22-21.ec2.internal  MountVolume.SetUp succeeded for volume "us01-stg-spinnaker-spinn-halyard-token-zcg7q"
  Normal   SuccessfulMountVolume   4m    kubelet, ip-10-0-22-21.ec2.internal  MountVolume.SetUp succeeded for volume "reg-secrets"
  Normal   SuccessfulAttachVolume  4m    attachdetach-controller              AttachVolume.Attach succeeded for volume "pvc-d17d527b-aad3-11e8-81e7-xxxxxxxxx"
  Normal   SuccessfulMountVolume   4m    kubelet, ip-10-0-22-21.ec2.internal  MountVolume.SetUp succeeded for volume "pvc-d17d527b-aad3-11e8-81e7-xxxxxxxxx"
  Warning  FailedCreatePodSandBox  4m    kubelet, ip-10-0-22-21.ec2.internal  Failed create pod sandbox: rpc error: code = Unknown desc = NetworkPlugin cni failed to set up pod "us01-stg-spinnaker-spinn-halyard-0_default" network: add command: failed to setup network: setup NS network: failed to add host route: file exists
  Normal   SandboxChanged          4m    kubelet, ip-10-0-22-21.ec2.internal  Pod sandbox changed, it will be killed and re-created.
  Normal   Pulled                  4m    kubelet, ip-10-0-22-21.ec2.internal  Container image "gcr.io/spinnaker-marketplace/halyard:stable" already present on machine
  Normal   Created                 4m    kubelet, ip-10-0-22-21.ec2.internal  Created container
```

Pod is created and started OK but these errors are still in the events and look bad.

max-rocket-internet:
I've emailed the `aws-cni-support.tar.gz` file to liwenwu@amazon.com.

[...]

PR #228, opened 2018-11-13
cleanup up the hostroute when perform CNI delete

liwenwu-amazon:
*Issue #166 , if available:*

*Description of changes:*


plugins/routed-eni/driver/driver.go
```diff
@@ -298,6 +298,17 @@ func tearDownNS(addr *net.IPNet, table int, netLink netlinkwrapper.NetLink) erro
 		log.Infof("Delete fromContainer rule for %s in table %d", addr.String(), table)
 	}
 
+	addrHostAddr := &net.IPNet{
+		IP:   addr.IP,
+		Mask: net.CIDRMask(32, 32)}
```
mogren: So this means all IPv4 subnets? (/32 or net mask 255.255.255.255)

plugins/routed-eni/driver/driver.go
```diff
@@ -298,6 +298,17 @@ func tearDownNS(addr *net.IPNet, table int, netLink netlinkwrapper.NetLink) erro
...
+
+	// cleanup host route:
+	if err = netLink.RouteDel(&netlink.Route{
+		Scope: netlink.SCOPE_LINK,
+		Dst:   addrHostAddr}); err != nil {
+		log.Errorf("delete NS network:failed to delete host route for %s, %v", addr.String(), err)
```
mogren: Nit: space missing after the `:`

plugins/routed-eni/driver/driver.go
liwenwu-amazon: no.  This is to cleanup route for Pod, for example, the support bundle collected from #166, 
```
ip route show table all
default via 10.0.22.1 dev eth0
10.0.22.0/23 dev eth0 proto kernel scope link src 10.0.23.172
10.0.22.56 dev enic0a80e334d5 scope link
10.0.22.94 dev enid336d91eb1a scope link
10.0.22.105 dev enie7834888824 scope link
10.0.22.137 dev enieda03e67426 scope link
10.0.22.168 dev eni995da829552 scope link
10.0.22.173 dev enicd72a6e3741 scope link
10.0.22.183 dev enic42d8b48626 scope link
10.0.23.8 dev enic382c044fa7 scope link
10.0.23.44 dev eni5fdca8f8634 scope link
10.0.23.93 dev eni7ce0ea50d68 scope link
10.0.23.133 dev enia7153239861 scope link <---- existing
10.0.23.140 dev enie5b1ee2eef9 scope link
10.0.23.145 dev eni9092c5a5932 scope link
10.0.23.174 dev enifd23e38f68c scope link
```
For Pod which is using `10.0.23.133`,  in most cases `10.0.23.133 dev enia7153239861 scope link` is deleted.   But we have seen in few cases reported in #166 (e.g. when the Pod is not started properly ), this route is left in the route table even when the Pod is deleted.

mogren: Ah, yes, `addr.IP` restricts that. Just wasn't sure what the `net.CIDRMask(32, 32)` was at first. Thanks for the details.

Issue #231, opened 2018-11-14
NodePort Connectivity Issue


nickdgriffin:
Hello,

We are experiencing an issue that is functionally identical to #75, using 1.2.1, where certain pods are not accessible from their NodePort on remote hosts and a tcpdump shows a SYN/SYN_ACK at the start followed by TCP retransmissions. As the mentioned ticket is about the rp_filter, here are the values collected by the support bundler:

```
/proc/sys/net/ipv4/conf/all/rp_filter = 1
/proc/sys/net/ipv4/conf/default/rp_filter = 1
/proc/sys/net/ipv4/conf/eth0/rp_filter = 2
```

We have this popping up across different clusters (although they are all identical in terms of setup) after pods are created (in this case Nginx, but it may be happening elsewhere) and the remedy is to delete the pods until the NodePort is functioning correctly.

I can send the support bundle and packet traces by email, and anything else that would help in identifying the cause of this and what we can do about it as it is quite problematic for us.

Thanks,
Nick

nickdgriffin:
Cracked it.

So, the problem stems from the fact that the secondary interfaces that are added still have the source/destination check enabled which must result in the ENI dropping the return packets from the pod. This can be proven by disabling the check on the ENI that the pod has an IP allocated on, and connections succeed.

I will be submitting a PR to disable the check when ENIs are allocated.

nickdgriffin:
Worth mentioning that this issue was brought out by setting WARM_ENI_TARGET to 20 as has been advised in numerous places to prevent issues scheduling containers on hosts that have run out of IPs, which meant that all nodes had the maximum number of ENIs attached which increased the chance that the CNI picked an IP that was associated with a secondary interface instead of the primary.

ikatson:
Have the same problem with 1.3.

The first packet hits eth0 (in my case 100.122.192.148) for the node port, gets DNAT'ed to the respective container on eth1 (in my case container ip is 100.122.199.244 which is on eth1).
Container reply gets routed back over eth1 because of routing rule like "from 100.122.199.244 lookup 2"

I was trying to figure out why does the reverse of DNAT is not applied before the routing decision, but could not figure this out, probably that's just how linux works. I.e. if the routing rule saw the un-DNATE'ed ip 100.122.192.148, it would get returned back over eth0 how it came from.

Will try that PR @nickdgriffin to alter the ENIs, thanks for it :)

ikatson:
[...]

PR #233, opened 2018-11-15
Setting SourceDestCheck to false in AllocENI

nickdgriffin:
*Issue #, if available:*
https://github.com/aws/amazon-vpc-cni-k8s/issues/231

*Description of changes:*
This disables the SourceDestCheck on the ENIs that are created by AllocENI so that packets are not dropped in the case where a NodePort service's backend pods have IPs allocated from a secondary interface. The NodePort traffic enters on eth0 and is then dropped by the secondary interface (e.g. eth1) when the source/destination check is true, as per the default.

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -555,6 +555,9 @@ func (cache *EC2InstanceMetadataCache) AllocENI(useCustomCfg bool, sg []*string,
 			DeleteOnTermination: aws.Bool(true),
 		},
 		NetworkInterfaceId: aws.String(eniID),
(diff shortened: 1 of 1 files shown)


mogren:
@nickdgriffin, is this still an issue?

infinitydon:
Currently we have a use-case for this (having the option to enable/disable SourceDestCheck):
-  As part of network connectivity requirement for some Telco cloud native functions.
- Aid debugging process but not of having to do this manually for every worker node ENIs, this will make it easier to automate enabling/disabling SourceDestCheck.

PR #245, opened 2018-11-27
Stop wrapping and returning nil

nak3:
*Description of changes:*

Currently `err` in arg of the second `errors.Wrapf()` in getENIaddresses() will never
receive error except for nil value.

- L#584 checks `err != nil` and return, hence L#595 will never get err except for nil.
https://github.com/aws/amazon-vpc-cni-k8s/blob/07363123942c6fa03da6a7737a2d15817a7c2f8e/ipamd/ipamd.go#L583-L595

When `errors.Wrapf()` is called with nil in first arg, the new message
is not wrapped. To make matters worse, `getENIaddresses()` returns
error as `nil` so the logic of code after the method could be collapsed.

Hence, this patch uses `errors.Errorf()` instead `errors.Wrapf()`.

Also, current unit tests for ipamd use two primary IPs for one
ENI. This patch fixes it as well.

### DIFF
--- a/ipamd/ipamd.go
+++ b/ipamd/ipamd.go
@@ -592,7 +592,7 @@ func (c *IPAMContext) getENIaddresses(eni string) ([]*ec2.NetworkInterfacePrivat
 		}
 	}
 
-	return nil, "", errors.Wrapf(err, "faind to find eni's primary address for eni %s", eni)
+	return nil, "", errors.Errorf("failed to find eni's primary address for eni %s", eni)
 }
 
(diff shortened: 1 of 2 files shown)


nak3:
Here is a small test code when `errors.Wrap` gets nil.

```
package main

import (
	"fmt"
	"github.com/pkg/errors"
)

func main() {
	var err error

	err = nil
	err = errors.Wrap(err, "I want to check this error")

	fmt.Printf("error w/ nil:  %v\n", err)
	if err == nil {
		fmt.Printf("error is still nil.\n")
	}

	/* OUTPUT
	$ go run a.go
	error w/ nil:  <nil>
	error is still nil.
	*/
}

```

PR #246, opened 2018-11-27
Randomise outgoing port for connections in the SNAT iptables rule.

taylorb-syd:
*Issue #, if available:* N/A

*Description of changes:* This change introduces the `--random` flag to outgoing SNAT connections, and enables it by default.

When making a large number of outgoing connections to the same endpoint there is a known race condition with SNAT rules that means that two connections can be assigned the same outgoing port from the ephemeral range. This results in connection traffic (for both the client and server) being unable to track both connections because they will have the same Source IP Address, Source Port, Destination IP Address, and Destination Port, resulting in one process being "orphaned".

To prevent this edge case from occurring you can enable the `--random` flag within the SNAT rule which will select a random (unused) port from the ephemeral range rather than assigning them sequentially.

The impact for enabling this should be minimal for most customers, so this option is enabled by default in this pull request.


taylorb-syd:
Realized that since kernel 3.14 there has been an option in the extensions called `--random-fully` that uses PRNG rather than hash based random allocation. Added `prng` as third option for the environment variable for people who need `--random-fully`

pkg/networkutils/network.go
```diff
@@ -55,6 +55,12 @@ const (
 	// be installed and will be removed if they are already installed.  Defaults to false.
 	envExternalSNAT = "AWS_VPC_K8S_CNI_EXTERNALSNAT"
 
+	// This environment is used to specify weather an the SNAT rule added to iptables should randomize port
```
miguelbernadi:
```suggestion
 	// This environment variable is used to specify wether the SNAT rule added to iptables should randomize port
```

pkg/networkutils/network.go
```diff
@@ -435,8 +445,26 @@ func useExternalSNAT() bool {
...
+			return defaultValue
+		}
+		// true is equal to hash based random
+		if parsedValue {
+			return randomHashSNAT
+		}
+		// false is equal to sequential
+		return sequentialSNAT
+	}
+	return defaultValue
```
miguelbernadi: Isn't this clearer? Also there's no need to use `strings.Compare` in this case, as we can just compare the strings themselves.
```go
strValue := os.Getenv(envRandomizeSNAT)
if strValue == "" {
	// empty string means default
	return defaultValue
}
if strValue == "prng" {
	// prng means to use --random-fully
	return randomPRNGSNAT
}
parsedValue, err := strconv.ParseBool(strValue)
if err != nil {
	log.Error("Failed to parse "+envRandomizeSNAT+"; using default: "+defaultString, err.Error())
	return defaultValue
}
// true is equal to hash based random
if parsedValue { return randomHashSNAT }
// false is equal to sequential
return sequentialSNAT
```

miguelbernadi:
As the PR adds the `AWS_VPC_K8S_CNI_RANDOMISESNAT` variable, I think it would be better to specify the different possible string values instead of relying on an obscure meaning of `true` and `false`. I'd suggest using the following terms:

- `none`: equal to `false`
- `hashrandom`: equal to `true`
- `prng`: equal to `prng`

I'm not aware of the reasoning behind the original naming scheme, but this should make clear what each option actually means and may also make the code a bit cleaner/less convoluted.

pkg/networkutils/network.go
taylorb-syd: Nice catch. Fixed

pkg/networkutils/network.go
taylorb-syd: strings.Compare was used in a trivial comparison elsewhere in the code in network.go. I assumed that it was being used for a reason.

Cleaned up the code to be easier to read.

taylorb-syd:
> As the PR adds the `AWS_VPC_K8S_CNI_RANDOMISESNAT` variable, I think it would be better to specify the different possible string values instead of relying on an obscure meaning of `true` and `false`. I'd suggest using the following terms:
> 
> * `none`: equal to `false`
> * `hashrandom`: equal to `true`
> * `prng`: equal to `prng`
> 
> I'm not aware of the reasoning behind the original naming scheme, but this should make clear what each option actually means and may also make the code a bit cleaner/less convoluted.

The first revision I wrote was tested on an old kernel that didn't support `--random-fully`, so I originally made it a `true`/`false` option. I then found that newer kernels, including Amazon Linux 2 (default for EKS) supports `--random-fully` and that there are benefits to using PRNG over the hash based random, especially with lots of calls to the same destination (say an API endpoint like S3). 

I modified the code to therefore support 'prng' as a "special case". Given that reasoning doesn't hold true (i.e. there will be no backwards compatibly reasons) I have modified as per your suggestion.

pkg/networkutils/network.go
```diff
@@ -448,22 +452,25 @@ func useExternalSNAT() bool {
...
-			}
-			log.Error("Failed to parse "+envRandomizeSNAT+"; using default: "+defaultString, err.Error())
-			return defaultValue
-		}
-		// true is equal to hash based random
-		if parsedValue {
-			return randomHashSNAT
-		}
-		// false is equal to sequential
+	strValue := os.Getenv(envRandomizeSNAT)
```
miguelbernadi: Nitpick: as all values are now strings we can also use a switch over the environment variable's value. 
```go
switch os.Getenv(envRandomizeSNAT) {
case "":
	// empty means default
	return defaultValue
case "prng":
	// prng means to use --random-fully
	return randomPRNGSNAT
case "none":
	// none means to disable randomisation (no flag)
	return sequentialSNAT
case "hashrandom":
	// hashrandom means to use --random
	return randomHashSNAT
default:
	// if we get to this point, the environment variable has an invalid value
	log.Error("Failed to parse " + envRandomizeSNAT + "; using default: " + defaultString + ". Provided string was " + strValue)
	return defaultValue
}
```
It's mostly a matter of taste to prefer the succession of `if` statements or the `switch`.

pkg/networkutils/network.go
```diff
@@ -115,10 +122,19 @@ type iptablesIface interface {
 	NewChain(table, chain string) error
 }
 
+type snatType uint32
+
+const (
+	sequentialSNAT snatType = 0
+	randomHashSNAT snatType = 1
+	randomPRNGSNAT snatType = 2
```
mogren: The idiomatic way to define this is to use `iota`. (See https://golang.org/ref/spec#Iota)

nak3:
I like this idea and believe that it is necessary. But I guess that updated SNAT rule is not available until the host was rebooted? I mean that if you change `none` to `prng`, iptables rule will be appended but previous `none`'s rule still exists like this:

```
iptables -t nat -A POSTROUTING -s x.x.x.x -j SNAT --to y.y.y.y
iptables -t nat -A POSTROUTING -s x.x.x.x -j SNAT --to y.y.y.y --random-fully
```

Even if the answer is yes, it might be better to leave it as a known limitation rather than fixing it, though.

pkg/networkutils/network.go
```diff
@@ -295,15 +311,21 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDR *net.IPNet, vpcCIDRs []*string,
...
 
 	curChain := fmt.Sprintf("AWS-SNAT-CHAIN-%d", len(vpcCIDRs))
+	snatRule := []string{"-m", "comment", "--comment", "AWS, SNAT",
+		"-m", "addrtype", "!", "--dst-type", "LOCAL",
+		"-j", "SNAT", "--to-source", primaryAddr.String()}
+	if n.randomizeSNAT == randomHashSNAT {
+		snatRule = append(snatRule, "--random")
+	}
+	if n.randomizeSNAT == randomPRNGSNAT {
+		snatRule = append(snatRule, "--random-fully")
```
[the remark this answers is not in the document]
taylorb-syd: Good catch. Have implemented.

pkg/networkutils/network.go
```diff
@@ -427,6 +449,31 @@ func useExternalSNAT() bool {
...
+	if strValue == "none" {
+		// none means to disable randomisation (no flag)
+		return sequentialSNAT
+	}
+	if strValue == "hashrandom" {
+		// hashrandom means to use --random
+		return randomHashSNAT
+	}
+	// if we get to this point, the environment variable has an invalid value
+	log.Error("Failed to parse " + envRandomizeSNAT + "; using default: " + defaultString + ". Provided string was " + strValue)
```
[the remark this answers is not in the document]
taylorb-syd: That makes sense. I will change the messaging.

pkg/networkutils/network.go
taylorb-syd: It does look neater.

README.md
```diff
@@ -99,6 +99,13 @@ Default: `false`
...
 Disable SNAT if you need to allow inbound communication to your pods from external VPNs, direct connections, and external VPCs, and your pods do not need to access the Internet directly via an Internet Gateway\. However, your nodes must be running in a private subnet and connected to the internet through an AWS NAT Gateway or another external NAT device\.
 
+`AWS_VPC_K8S_CNI_RANDOMIZESNAT`  
+Type: String 
+Default: `hashrandom` 
+Valid Values: `hashrandom`, `prng`, `none`
+Specifies weather the SNAT `iptables` rule should randomize the outgoing ports for connections. When enabled (`hashrandom`) the `--random` flag will be added to the SNAT `iptables` rule. To use pseudo random number generation rather than hash based (i.e. `--random-fully`) use `prng` for the environment variable.
+Disable (`none`) this functionality if you rely on sequential port allocation for outgoing connections. For old versions if iptables that do not support `--random-fully` this option will fall back to `--random`.
```
mogren: Nit pick, but I think "For old versions **if** iptables that do not support `--random-fully` this option will fall back to `--random`." sentence should be be before the "Disable (`none`).." one.  Also, it says "if iptables", should be "of iptables.

taylorb-syd: Fixed.

PR #251, opened 2018-11-30
add an environment variable to limit the number of ENIs

pdbogen:
*Issue #, if available:*

*Description of changes:*

This adds an environment variable to limit the number of ENIs assigned to a node. For the time being we're interested in only using one ENI per node, and this setting makes it possible to _ensure_ that this is the case.

(I also add a Make target to run tests in the same docker container used for building, but, you know, take it or leave it.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,12 @@ unit-test:
 	GOOS=linux CGO_ENABLED=1 go test -v -cover -race -timeout 10s ./pkg/networkutils/...
 	GOOS=linux CGO_ENABLED=1 go test -v -cover -race -timeout 10s ./ipamd/...
 
+docker-unit-test:
+	docker run -v $(shell pwd):/usr/src/app/src/github.com/aws/amazon-vpc-cni-k8s \
+		--workdir=/usr/src/app/src/github.com/aws/amazon-vpc-cni-k8s \
+		--env GOPATH=/usr/src/app \
+		golang:1.10 make unit-test
+
 # golint
 # To install: go get -u golang.org/x/lint/golint
 lint:
--- a/README.md
+++ b/README.md
@@ -113,10 +113,10 @@ Default: None
 Specifies the number of free IP addresses that the `ipamD` daemon should attempt to keep available for pod assignment on the node\. For example, if `WARM_IP_TARGET` is set to 10, then `ipamD` attempts to keep 10 free IP addresses available at all times\. If the elastic network interfaces on the node are unable to provide these free addresses, `ipamD` attempts to allocate more interfaces until `WARM_IP_TARGET` free IP addresses are available\.  
 This environment variable overrides `WARM_ENI_TARGET` behavior\.
 
-`MAX_ENI`
-Type: Integer
-Default: `-1`
-Specifies the maximum number of ENIs that will be attached to the node. When MAX_ENI is <= 1 (i.e., the default), the setting is not used, and the maximum number of ENIs is always equal to the maximum number for the instance type in question. Even when MAX_ENI is a positive number, it is limited by the maximum number for the instance type.
+`MAX_ENI`  
+Type: Integer  
+Default: None  
+Specifies the maximum number of ENIs that will be attached to the node. When MAX_ENI is unset or 0 (or lower), the setting is not used, and the maximum number of ENIs is always equal to the maximum number for the instance type in question. Even when MAX_ENI is a positive number, it is limited by the maximum number for the instance type.
 
 ### Notes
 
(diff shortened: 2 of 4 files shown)


ipamd/ipamd.go
```diff
@@ -194,10 +201,19 @@ func New(k8sapiClient k8sapi.K8SAPIs, eniConfig *eniconfig.ENIConfigController)
 func (c *IPAMContext) nodeInit() error {
 	ipamdActionsInprogress.WithLabelValues("nodeInit").Add(float64(1))
 	defer ipamdActionsInprogress.WithLabelValues("nodeInit").Sub(float64(1))
-	maxENIs, err := c.awsClient.GetENILimit()
-	if err == nil {
+
+	// If MAX_ENI is set, the maximum number of ENIs may be no more than that
+	// number (but might be less, depending on instance type.) If MAX_ENI is < 1,
+	// then we use the limit from the instance type.
+	maxENIs := getMaxENI()
```
mogren: If `getMaxENI()` took `instanceENILimit` as an argument, the limit checks could be in once place and the check would just be something like
```go
	instanceMaxENIs, err := c.awsClient.GetENILimit()
	if err == nil {
		maxENIs = getMaxENI(instanceMaxENIs)
	}
```

ipamd/ipamd_test.go
```diff
@@ -360,6 +360,23 @@ func TestGetWarmENITarget(t *testing.T) {
...
+	maxENI := getMaxENI()
+	assert.Equal(t, maxENI, 5)
+
+	os.Unsetenv("MAX_ENI")
+	maxENI = getMaxENI()
+	assert.Equal(t, maxENI, defaultMaxENI)
+
+	os.Setenv("MAX_ENI", "non-integer-string")
+	maxENI = getMaxENI()
+	assert.Equal(t, maxENI, defaultMaxENI)
```
mogren: Tests with `MAX_ENI = 0` and `MAX_ENI = 1` would be nice.

Makefile
```diff
@@ -52,6 +52,12 @@ unit-test:
 	GOOS=linux CGO_ENABLED=1 go test -v -cover -race -timeout 10s ./pkg/networkutils/...
 	GOOS=linux CGO_ENABLED=1 go test -v -cover -race -timeout 10s ./ipamd/...
 
+docker-unit-test:
+	docker run -v $(shell pwd):/usr/src/app/src/github.com/aws/amazon-vpc-cni-k8s \
+		--workdir=/usr/src/app/src/github.com/aws/amazon-vpc-cni-k8s \
+		--env GOPATH=/usr/src/app \
+		golang:1.10 make unit-test
```
mogren: Thanks, I like this.

pdbogen:
we're running EKS clusters in VPCs with public IP addresses but that need to communicate with peered VPCs and other things; basically networking only works properly when we run on the primary ENI, otherwise pods assigned to the secondary ENIs aren't able to communicate properly.

I.e., we're not running EKS in one of the two modes that the CNI seems to support, but, it works for primary ENIs with EXTERNALSNAT=true.

However, I'm wondering if https://github.com/aws/amazon-vpc-cni-k8s/pull/233 might have something to do with my troubles, so before I go farther down this path, I'm going to experiment with _that_.

liwenwu-amazon:
@pdbogen  Have you tried setting WARM_ENI_TARGET=0 ?

pdbogen:
I *think* I'd have to set both WARM_ENI_TARGET _and_ WARM_IP_TARGET to 0, otherwise when the CNI has maxed out the first ANI, WARM_IP_TARGET will result in configuring a second ENI. I'd like to avoid that situation, but still be able to keep warm IPs around.

README.md
```diff
@@ -113,6 +113,11 @@ Default: None
 Specifies the number of free IP addresses that the `ipamD` daemon should attempt to keep available for pod assignment on the node\. For example, if `WARM_IP_TARGET` is set to 10, then `ipamD` attempts to keep 10 free IP addresses available at all times\. If the elastic network interfaces on the node are unable to provide these free addresses, `ipamD` attempts to allocate more interfaces until `WARM_IP_TARGET` free IP addresses are available\.  
 This environment variable overrides `WARM_ENI_TARGET` behavior\.
 
+`MAX_ENI`
+Type: Integer
+Default: `-1`
```
mattlandis: With both 0 and -1 being treated as using the max for the instance I would prefer we use 0 as the default but am open to discussion.

mattlandis:
Good point.  I think this is good as is.  We can make using it to override when nothing is in the map a separate PR.

PR #252, opened 2018-12-03
Remove unused options from rpc.proto

nak3:
*Description of changes:*

Three options `java_multiple_files`, `java_package`, `java_outer_classname`
defined in `rpc.proto` are not used in current code.

This patch removes these options from rpc.proto


mogren:
Could you squash this to one commit? Thanks for cleaning this up!

Issue #259, opened 2018-12-06
support eniConfig set based on node label


etopeter:
Feature request to support setting ENI Config via node label.
Kubelet has a way to set node labels when bootstrapping and registering node to Kubernetes cluster by using extra args. We can use this function to set correct configuration.

etopeter:
Feature merged [#280](https://github.com/aws/amazon-vpc-cni-k8s/pull/280)

PR #260, opened 2018-12-06
#259 Support for node labels to set eniConfig

etopeter:
#259

Added support for node labels and unit tests.

### DIFF
--- a/pkg/eniconfig/eniconfig.go
+++ b/pkg/eniconfig/eniconfig.go
@@ -36,6 +36,7 @@ import (
 
 const (
 	eniConfigAnnotationDef = "k8s.amazonaws.com/eniConfig"
+	eniConfigLabelDef      = "k8s.amazonaws.com/eniConfig"
 	eniConfigDefault       = "default"
 )
 
@@ -104,7 +105,7 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 
 	case *corev1.Node:
 
-		log.Infof("Handle corev1.Node: %s, %v", o.GetName(), o.GetAnnotations())
+		log.Infof("Handle corev1.Node: %s, %v, %v", o.GetName(), o.GetAnnotations(), o.GetLabels())
 		if h.controller.myNodeName == o.GetName() {
 			annotation := o.GetAnnotations()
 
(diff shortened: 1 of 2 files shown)


md2k:
Tested this PR with my EKS and it is awesome. probably for a while we going to use custom build for aws-node to ensure that we able to do automated cni configuration.

pkg/eniconfig/eniconfig.go
```diff
@@ -115,10 +116,22 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
-				h.controller.eniLock.Lock()
-				defer h.controller.eniLock.Unlock()
-				h.controller.myENI = eniConfigDefault
-				log.Infof(" Setting myENI to: %s", eniConfigDefault)
+
+				label := o.GetLabels()
+
+				val, ok := label[eniConfigLabelDef]
+				if ok {
+					h.controller.eniLock.Lock()
```
nak3: There are three duplicated `Lock`, `Unlock` lines around this block. Would you like to refactor them?

For example, you can write under L106 as:

```
        case *corev1.Node:

                log.Infof("Handle corev1.Node: %s, %v, %v", o.GetName(), o.GetAnnotations(), o.GetLabels())
                if h.controller.myNodeName == o.GetName() {
                        val, ok := o.GetAnnotations()[eniConfigAnnotationDef]
                        if !ok {
                                val, ok = o.GetLabels()[eniConfigLabelDef]
                                if !ok {
                                        val = eniConfigDefault
                                }
                        }

                        h.controller.eniLock.Lock()
                        defer h.controller.eniLock.Unlock()
                        h.controller.myENI = val
                        log.Infof(" Setting myENI to: %s", val)
                }
        }
        return nil
```

pkg/eniconfig/eniconfig_test.go
```diff
@@ -68,6 +68,33 @@ func updateNodeAnnotation(hdlr sdk.Handler, nodeName string, configName string,
 	hdlr.Handle(nil, event)
 }
 
+func updateNodeLabel(hdlr sdk.Handler, nodeName string, configName string, toDelete bool) {
```
nak3: This function is almost same with `updateNodeAnnotation()`. It is better to combine them?

pkg/eniconfig/eniconfig_test.go
```diff
@@ -144,3 +171,44 @@ func TestNodeENIConfig(t *testing.T) {
 	assert.Equal(t, defaultCfg, *outputCfg)
 
 }
+
+func TestNodeENIConfigLabel(t *testing.T) {
```
nak3: Same with above comment. Is it not possible to merge this test into `TestNodeENIConfig()`?

nak3:
I am not an official reviewer so please disregard my comments if you disagree.

pkg/eniconfig/eniconfig_test.go
mattlandis: Can we extract the strictly duplicate code.  I think two methods is fine, if we remove the code that is the same toa  `createEvent` function.

pkg/eniconfig/eniconfig_test.go
mattlandis: I'm planning a bigger refactor of the test code so I am OK with this.

md2k:
when this can be merged?

etopeter:
There is another PR with same functionality, more features and cleaner logic here [280](https://github.com/aws/amazon-vpc-cni-k8s/pull/280) So hope that one gets merged and you will also be able to set custom annotations and labels.

mogren:
Can we cancel this PR since #280 has been merged?

PR #278, opened 2018-12-23
Detect if mockgen and goimports are in the path

nak3:
*Description of changes:*

This patch adds validation to detect if `mockgen` and `goimports` are
in the executable file path before running commands.

Currently, if either `mockgen` or `goimports` does not exists, the
program stops, but mock codes are gone.

### DIFF
--- a/scripts/mockgen.go
+++ b/scripts/mockgen.go
@@ -44,13 +44,23 @@ const (
   `
 )
 
-// TODO detect if mockgen or goimports are in the path
+var (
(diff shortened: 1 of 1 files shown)


scripts/mockgen.go
```diff
@@ -44,13 +44,23 @@ const (
   `
 )
 
-// TODO detect if mockgen or goimports are in the path
+var (
+	requiredCommands = [2]string{"mockgen", "goimports"}
```
mogren: Thanks!

PR #280, opened 2019-01-03
Feature: ENIConfig set by custom annotation or label names

etopeter:
This feature enables user to set custom annotation or label key to define ENIConfig name. One use case if you have nodes in different failure domains. By default AWS deployments set node label failure-domain.beta.kubernetes.io/zone and its value is availability zone name. We can use availability zone name as ENIConfig to map to proper subnets.
Added 2 new environmental variables:
ENI_CONFIG_ANNOTATION_DEF
ENI_CONFIG_LABEL_DEF

### DIFF
--- a/pkg/eniconfig/eniconfig.go
+++ b/pkg/eniconfig/eniconfig.go
@@ -35,9 +35,19 @@ import (
 )
 
 const (
-	eniConfigAnnotationDef = "k8s.amazonaws.com/eniConfig"
-	eniConfigLabelDef      = "k8s.amazonaws.com/eniConfig"
-	eniConfigDefault       = "default"
+	defaultEniConfigAnnotationDef = "k8s.amazonaws.com/eniConfig"
+	defaultEniConfigLabelDef      = "k8s.amazonaws.com/eniConfig"
+	eniConfigDefault              = "default"
+
+	// when "ENI_CONFIG_LABEL_DEF is defined, ENIConfigController will use that label key to
+	// search if is setting value for eniConfigLabelDef
+	// Example:
+	//   Node has set label k8s.amazonaws.com/eniConfigCustom=customConfig
+	//   We can get that value in controller by setting environmental variable ENI_CONFIG_LABEL_DEF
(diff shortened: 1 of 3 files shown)


pkg/eniconfig/eniconfig.go
```diff
@@ -35,8 +35,19 @@ import (
 )
 
 const (
-	eniConfigAnnotationDef = "k8s.amazonaws.com/eniConfig"
-	eniConfigDefault       = "default"
+	defaultEniConfigAnnotationDef = "k8s.amazonaws.com/eniConfig"
```
mattlandis: Can we store the value being used in the `ENIConfigController`?  The `New*` methods can read the ENVs or use the defaults.  Then the ENV lookup doesn't happen on every request.  I don't think they should change during the execution of the program so that shouldn't be an issue.

etopeter: made a change to get the label and annotation definition only once.

pkg/eniconfig/eniconfig.go
```diff
@@ -104,21 +119,34 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
+					defer h.controller.eniLock.Unlock()
+					h.controller.myENI = val
+					log.Infof(" Setting myENI to: %s", val)
+				} else {
+
+					h.controller.eniLock.Lock()
+					defer h.controller.eniLock.Unlock()
+					h.controller.myENI = eniConfigDefault
+					log.Infof(" Setting myENI to: %s", eniConfigDefault)
+				}
```
mogren: I would much prefer this to look like @nak3 suggested in [his comment on #260](https://github.com/aws/amazon-vpc-cni-k8s/pull/260/files#r243752421)

etopeter: Added changes for cleaner logic. I agree it looks much better the way @nak3 suggested.

mattlandis:
Thanks for the contribution.

/lgtm

adammw:
@mattlandis @mogren any chance we can get a release that includes this patch? Finding it difficult to start a node that comes up with the annotation set, compared to a label which is supported via kubelet's `--node-labels` flag.

Issue #288, opened 2019-01-16
Log message why failed to connect k8s server is missing 


nak3:
Please refer to https://github.com/aws/amazon-vpc-cni-k8s/issues/283

The message must come from here: (and returning error)

https://github.com/aws/amazon-vpc-cni-k8s/blob/92980a5578b6acb855d9ad886b02fcb1f85e7526/pkg/k8sapi/discovery.go#L90-L94

However, we cannot see any message from here L#48:

https://github.com/aws/amazon-vpc-cni-k8s/blob/92980a5578b6acb855d9ad886b02fcb1f85e7526/main.go#L46-L50

We should get below message: 

`Failed to create client: error communicating with apiserver: <SOME ERROR MESSAGE WHY FAILED TO CONNECT TO SERVER>`

nak3:
This issue be fixed by https://github.com/aws/amazon-vpc-cni-k8s/pull/289

PR #289, opened 2019-01-16
Return with exit codes to call log.Flush()

nak3:
Before `os.Exit()` is fired, `github.cihub/seelog`'s log.Flush() needs to be
called. Otherwise, the log message was not output and we lost some
important logs.

This patch changes surely `returns 1` to call `defer log.Flush()` in
main().

Fixes https://github.com/aws/amazon-vpc-cni-k8s/issues/288

### DIFF
--- a/main.go
+++ b/main.go
@@ -46,7 +46,7 @@ func _main() int {
 	kubeClient, err := k8sapi.CreateKubeClient("", "")
 	if err != nil {
 		log.Errorf("Failed to create client: %v", err)
-		os.Exit(1)
+		return 1
 	}
 
 	discoverController := k8sapi.NewController(kubeClient)


ewbankkit:
Related: https://github.com/aws/amazon-vpc-cni-k8s/issues/57, https://github.com/aws/amazon-vpc-cni-k8s/pull/54.

main.go
```diff
@@ -46,7 +46,7 @@ func _main() int {
 	kubeClient, err := k8sapi.CreateKubeClient("", "")
 	if err != nil {
 		log.Errorf("Failed to create client: %v", err)
-		os.Exit(1)
+		return 1
```
mogren: Thanks, good catch.

PR #291, opened 2019-01-17
Fix warning message from makefile

nak3:
Currently `make arg` produces following error:

```
Makefile:24: extraneous text after 'else' directive
```

Please also refer to previous travis build log - https://travis-ci.org/aws/amazon-vpc-cni-k8s/jobs/480537040#L474
This patch fixes this warning message.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,9 @@ ARCH ?= $(shell uname -m)
 
 ifeq ($(ARCH),aarch64)
   ARCH = arm64
-else ($(ARCH),x86_64)
+else
+endif
+ifeq ($(ARCH),x86_64)
   ARCH = amd64
 endif
 


Makefile
```diff
@@ -21,7 +21,9 @@ ARCH ?= $(shell uname -m)
 
 ifeq ($(ARCH),aarch64)
   ARCH = arm64
-else ($(ARCH),x86_64)
+else
```
nak3: Following `else ifeq` style is much clear, but according to [stackflow](https://stackoverflow.com/questions/34761685/extraneous-text-after-else-directive), this style does not work on Windows. Hence I changed safer way.

```
ifeq ($(ARCH),aarch64)
  ARCH = arm64
else ifeq ($(ARCH),x86_64)
  ARCH = amd64
endif
```

mogren: Thanks

ruhlem: Thanks! I have missed that "else ifeq" , sorry for that

PR #292, opened 2019-01-18
Allow cross compile on different platform

nak3:
This patch adds `GOARCH` option to `go build`, so any platform can
perform the build for arm64, amd64 and so on from different platform.

- Example usage:
```
$ ARCH=amd64 make build-linux
GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o aws-k8s-agent -ldflags "-X main.version=v1.3.0-49-g729fcd41"
GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o aws-cni -ldflags "-X main.version=v1.3.0-49-g729fcd41" ./plugins/routed-eni/
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@
 # language governing permissions and limitations under the License.
 #
 
-.PHONY: build-linux clean docker docker-build lint unit-test vet
+.PHONY: build-linux clean docker docker-build lint unit-test vet download-portmap
 
 VERSION ?= $(shell git describe --tags --always --dirty)
 LDFLAGS ?= -X main.version=$(VERSION)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -38,13 +38,14 @@ download-portmap:
...
 # Default to build the Linux binary
 build-linux:
-	GOOS=linux CGO_ENABLED=0 go build -o aws-k8s-agent -ldflags "$(LDFLAGS)"
-	GOOS=linux CGO_ENABLED=0 go build -o aws-cni -ldflags "$(LDFLAGS)" ./plugins/routed-eni/
+	GOOS=linux GOARCH=$(ARCH) CGO_ENABLED=0 go build -o aws-k8s-agent -ldflags "$(LDFLAGS)"
+	GOOS=linux GOARCH=$(ARCH) CGO_ENABLED=0 go build -o aws-cni -ldflags "$(LDFLAGS)" ./plugins/routed-eni/
 
 docker-build:
-	docker run -v $(shell pwd):/usr/src/app/src/github.com/aws/amazon-vpc-cni-k8s \
+	docker run -v $(shell pwd):/usr/src/app/src/github.com/aws/amazon-vpc-cni-k8s:z \
```
mogren: Why do we need the `:z` suffix for shared bind mounts? Is it just for building on SELinux?

nak3: Exactly, it is just for building on SELinux. I put it as my machine is enabled SELinux (and sorry I forgot to remove it when I sent this PR). The option was removed now.

mogren: Seems like it doesn't have any negative effects on non SE systems, so maybe we could have it around. I had no idea it was needed on SELinux, thanks for letting me know 🙃

PR #296, opened 2019-01-22
Restrict the number of IPs per ENI to 30 even on larger instances

mattlandis:
While some instance types allow for more than 30 secondary IPs on ENIs,
there is an issue where the 32 IP forward will not be able to access
instance metadata, VPC DNS and Time Sync services.

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html


mogren:
Some unit tests check `SecondaryPrivateIpAddressCount: 49`

mattlandis:
Yea.  Saw that.  Forgot we didn't have unit tests running as part of docker-build.  I am also testing 30 vs 31 for these instance types.  I think we can do 31 since the CNI does the match and removes the primary and the docs indicate that  it is the 32nd that breaks.

mogren:
Yes, 31 should be fine.

pkg/awsutils/vpc_ip_resource_limit.go
```diff
@@ -278,26 +278,26 @@ var InstanceIPsAvailable = map[string]int64{
...
-	"m5d.24xlarge":  50,
+	"m5d.24xlarge":  31,
 	"p2.xlarge":     15,
 	"p2.8xlarge":    30,
 	"p2.16xlarge":   30,
 	"p3.2xlarge":    15,
 	"p3.8xlarge":    30,
 	"p3.16xlarge":   30,
-	"p3dn.24xlarge": 50,
+	"p3dn.24xlarge": 31,
```
mogren: Lets only reduce it for `p3dn.24xlarge`

mattlandis: I split this into two commits.  One for the p3dn.24xlarge and one for all others.  This way we can cherry-pick the p3dn fix into 1.3.1 (since it is not currently in 1.3.0 it won't break anything).

PR #196, opened 2018-10-07
Add logic to dynamically discover primary interface name

liwenwu-amazon:
*Issue #171 , if available:*

*Description of changes:*
* use metadata service to find out the mac address on primary interface
* go through all the interfaces and find the interface which has same mac address as primary interface's mac.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #305, opened 2019-01-30
Use primary interface to add iptables for connmark entry

nak3:
*Description of changes:*

Currently iptables for CONNMARK entry always uses `eth0`, even though
AWS EC2 instance does not always use eth0.

This fix uses primary interface name which was discovered dynamically.

Please also refer to https://github.com/aws/amazon-vpc-cni-k8s/pull/196

### DIFF
--- a/pkg/networkutils/network.go
+++ b/pkg/networkutils/network.go
@@ -207,9 +207,10 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDR *net.IPNet, vpcCIDRs []*string, 
 		return errors.Wrapf(err, "host network setup: failed to delete old host rule")
 	}
 
+	primaryIntf := "eth0"
 	if n.nodePortSupportEnabled {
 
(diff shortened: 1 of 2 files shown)


nak3:
Sorry, I will fix the unit test.

pkg/networkutils/network.go
```diff
@@ -207,9 +207,10 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDR *net.IPNet, vpcCIDRs []*string,
 		return errors.Wrapf(err, "host network setup: failed to delete old host rule")
 	}
 
+	primaryIntf := "eth0"
```
mogren: It's not optimal to define the variable here, then use it 144 lines later, but I also don't want to call `findPrimaryInterfaceName()` again, so I guess this is the best option.

Btw, would you mind updating the error message on line 233, it should use the value of `primaryIntf` and not be hardcoded to "eth0", that might be misleading.

nak3: Sure, that makes sense. Thank you!

PR #340, opened 2019-03-07
Set mainENIRule mask

tustvold:
- Re-add the mask for the mainENIRule. Without this calico breaks NodePort traffic on secondary ENIs, as whilst the traffic is correctly marked by iptables, the ip rule that forces the traffic out of the primary ENI is impacted by other bits sets by calico in the mask


mogren:
I'm afraid I don't understand exactly what this change does or how this impacts Calico. @2ffs2nns or @caseydavenport, could you please help me out?

tustvold:
If you don't set the mask the routing rule looks for packets with the exact mark instead of just the bit that is set by conntrack. As calico uses the marks in order to implement its own filtering, this breaks the routing that forces NodePort traffic out of the primary ENI.

mogren:
Ah, thanks @tustvold, that makes sense. Also found the iptables documentation saying:

**--mark value[/mask]**
> Matches packets in connections with the given mark value (if a mask is specified, this is logically ANDed with the mark before the comparison).

Issue #135, opened 2018-07-13
Allow logging output to stdout/stderr


stuartnelson3:
Logging is currently hardcoded to write to a file. It would be nice to support logging to stdout/stderr, as many logging systems are based on this.

PR #342, opened 2019-03-08
add support for stdout logging

rudoi:
*Issue #, if available:*

https://github.com/aws/amazon-vpc-cni-k8s/issues/135

*Description of changes:*

- Add env var (`AWS_VPC_K8S_CNI_LOG_OUTPUT`) for log output (log file is default, stdout is only other option)

Note: I took the least-friction approach I could think of for adding this. If anyone has a better idea/design for this support, I'd be happy to give it a shot.

### DIFF
--- a/pkg/utils/logger/logger.go
+++ b/pkg/utils/logger/logger.go
@@ -23,7 +23,6 @@ import (
 
 const (
 	envLogLevel    = "AWS_VPC_K8S_CNI_LOGLEVEL"
-	envLogOutput   = "AWS_VPC_K8S_CNI_LOG_OUTPUT"
 	envLogFilePath = "AWS_VPC_K8S_CNI_LOG_FILE"
 	// logConfigFormat defines the seelog format, with a rolling file
 	// writer. We cannot do this in code and have to resort to using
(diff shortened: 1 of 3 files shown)


pkg/utils/logger/logger.go
```diff
@@ -68,3 +69,12 @@ func getLogLevel() string {
 
 	return seelogLevel.String()
 }
+
+func getLogOutput(logFilePath string) string {
+	switch strings.ToLower(os.Getenv(envLogOutput)) {
```
mogren: What about just reusing `AWS_VPC_K8S_CNI_LOG_FILE` and adding the environment variable to the documentation? This would just become:
```
    switch logFilePath {
	case "stdout":
		return `<console />`
	default:
return fmt.Sprintf(`<rollingfile filename="%s" type="date" datepattern="2006-01-02-15" archivetype="none" maxrolls="24" />`, logFilePath)
	}
```

And something like

`AWS_VPC_K8S_CNI_LOG_FILE`
Type: String  
Default: Unset
Valid Values: `stdout` or a file path
Specifies where to write the logging output. Either to `stdout` or to override the default file.

in the README.md?

rudoi: I completely agree with you. Updated!

PR #354, opened 2019-03-22
Comment and logging change only

mogren:
*Description of changes:*
* Use ENI and IP consistently instead of mixing cases.
* Some minor spelling fixes as well.
* Removed some excessive newlines

### DIFF
--- a/pkg/networkutils/network.go
+++ b/pkg/networkutils/network.go
@@ -163,11 +163,9 @@ type stringWriteCloser interface {
 
 // find out the primary interface name
 func findPrimaryInterfaceName(primaryMAC string) (string, error) {
-
 	log.Debugf("Trying to find primary interface that has mac : %s", primaryMAC)
 
 	interfaces, err := net.Interfaces()
-
 	if err != nil {
 		log.Errorf("Failed to read all interfaces: %v", err)
 		return "", errors.Wrapf(err, "findPrimaryInterfaceName: failed to find interfaces")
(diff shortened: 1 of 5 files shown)


pkg/networkutils/network.go
```diff
@@ -622,7 +618,7 @@ func setupENINetwork(eniIP string, eniMAC string, eniTable int, eniSubnetCIDR st
 						log.Errorf("Failed to add route %s/0 via %s table %d",
 							r.Dst.IP.String(), gw.String(), eniTable)
 						return errors.Wrapf(err,
-							"eni network setup: failed unable to add route %s/0 via %s table %d",
+							"ENI network setup: failed unable to add route %s/0 via %s table %d",
```
tiffanyfay: This or a different PR (also to match the log.Errorf)
s/failed unable to add route/failed to add route/

pkg/utils/utils.go
```diff
@@ -152,7 +152,7 @@ func RetryNWithBackoff(backoff Backoff, n int, fn func() error) error {
 // If the number of tries is exhausted, the last error will be returned.
 func RetryNWithBackoffCtx(ctx context.Context, backoff Backoff, n int, fn func() error) error {
 	var err error
-	RetryWithBackoffCtx(ctx, backoff, func() error {
+	_ = RetryWithBackoffCtx(ctx, backoff, func() error {
```
tiffanyfay: What about erroring if it fails to retry?

pkg/networkutils/network.go
mogren: Thanks, good catch.

pkg/utils/utils.go
mogren: I'll revert this. I think one issue is that we don't even use this code except in the tests. I'd much rather import the code from the official AWS go SDK once we actually use a retryer:

https://github.com/aws/aws-sdk-go/blob/master/aws/client/default_retryer.go

PR #366, opened 2019-03-29
Add new instance types

mogren:
*Description of changes:*

* Added m5ad and r5ad families
* Updated some stale limits for 16xlarge types


mogren:
Values from https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html#AvailableIpPerENI

pkg/awsutils/vpc_ip_resource_limit.go
```diff
@@ -233,17 +245,17 @@ var InstanceIPsAvailable = map[string]int64{
 	"d2.8xlarge":    30,
 	"f1.2xlarge":    15,
 	"f1.4xlarge":    30,
-	"f1.16xlarge":   24,
+	"f1.16xlarge":   31,
```
tiffanyfay: Do we not match it to https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html ?

tiffanyfay: Basically, where is 31 coming from? Doc says 8 ENI 50 IPs

mogren: There is a note just below the table:

> If f1.16xlarge, g3.16xlarge, h1.16xlarge, i3.16xlarge, and r4.16xlarge instances use more than 31 IPv4 or IPv6 addresses per interface, they cannot access the instance metadata, VPC DNS, and Time Sync services from the 32nd IP address onwards. If access to these services is needed from all IP addresses on the interface, we recommend using a maximum of 31 IP addresses per interface

Also, for all the ones that have 50, like `c5.18xlarge` we still have at 31 in this file. I would like to check which ones still work after 31 pods, but we don't have a test for that yet and it's a pain to do manually.

tiffanyfay: Ah ok. Thanks.

Issue #365, opened 2019-03-28
Does this CNI support containerd?


rudoi:
Is there any part of this CNI that would not function well with containerd (aka not docker)? If so, why?

mogren:
Currently [ipamd](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/ipamd/ipamd.go#L323) uses the go docker client to call `GetRunningContainers()` to find the local pods on the node, but that is the only place. I have not tested another runtime myself, so I'm not sure what would be needed to support that.

rudoi:
Thanks @mogren. I'm curious... what is the reason for retrieving the container ID from Docker, as opposed to checking a container status via Kube API? I have a PR ready to go that completely skips checking Docker, but I'm wondering if I'm missing something critical about the CNI's functionality that requires the container ID from Docker specifically.

If this is required, I will have to implement a client for interfacing with containerd instead, but that is a much larger piece of work. 😄

mogren:
I really don't know why Liwen wrote it that way. The last related change was in #86, way before my time.

rudoi:
I opened #371 proposing getting the container ID from the Kubernetes API instead. Happy to discuss alternative paths, but I can't think of one that doesn't involve writing a bespoke containerd package.

mogren:
Merged in v1.6, release candidate published.

mogren:
It does now! CRI-O as well. That said, we will require containerd to use the CRI plugin in order to get the Sandbox ID.

PR #371, opened 2019-04-01
fix: get container ID from kube rather than docker

rudoi:
*Issue #, if available:*

#365 

*Description of changes:*

Gets container ID from Kubernetes rather than Docker. This makes the CNI runtime-agnostic. 

When using containerd instead of Docker, the CNI essentially skips allocating addresses for all pods because it can't find containers to associate them with. The current retry logic doesn't fail hard when the CNI is unable to connect to a Docker sock. New pods appear to successfully obtain IP addresses, but when the CNI pod restarts, these IPs will all be freed because the CNI will be unable to connect to Docker and therefore will skip allocating IPs for all pods on the host.


pkg/k8sapi/discovery.go
```diff
@@ -215,7 +215,14 @@ func (d *Controller) handlePodUpdate(key string) error {
 		d.workerPodsLock.Lock()
 		defer d.workerPodsLock.Unlock()
 
+		var containerID string
+
+		if len(pod.Status.ContainerStatuses) > 0 {
+			containerID = pod.Status.ContainerStatuses[0].ContainerID
```
mogren: Could you add the same debug logging as when we fetched it from Docker?
```
log.Debugf("Found pod(%v)'s container ID: %v ", podName, containerID)
```

rudoi: updated!

mogren: Ouch, that won't work, you haven't assigned `containerID` the value yet.

rudoi: note to self: no PRs before coffee

liwenwu-amazon:
This change needs to be carefully tested out to make sure **no  same IP** can be allocated to different Pods.
ipamD does NOT preserve IP allocations pool state through its **restart**
During ipamD restart,  it first builds IP warm Pool info from instance metadata service, then it talks to Docker to find out the all running Pod's IPs and take running Pod IPs out of IP warm pool.   **After this**, it then responds CNI-ADD request.

To get same info from kube, you need to make sure ipamD have learned all running Pod IPs before start accepting CNI-ADD.

mogren:
Thanks @liwenwu-amazon, that is good to know. We won't merge this without some more testing.

rudoi:
@mogren / @liwenwu-amazon 

I dumped the list of ENIs on a host running 5 pods, 3 of which were hostNetwork=true, and saw that 2 were allocated as expected. Then, I `kill`'d the aws-k8s-agent process, causing the pod to restart. After the CNI plugin came back online, I captured the ENI list again and there was no diff between the two ENI dumps.

FWIW, I've been running a cluster off of this branch with several nodes and several dozen pods for about 10 days. A few of the CNI daemonset pods have restarts for various reasons, but I haven't seen any issues so far with duplicate IP allocation.

If there's anything more specific I should run some tests for, please let me know. I'm curious to understand how getting the container ID from kube API vs Docker has an impact on how IPs are allocated. It doesn't seem like the container ID is ever read again after the initial fetch.

rudoi:
Understood, @liwenwu-amazon. Thanks for clarifying. I'll try my hardest to cause that race. 😄

mogren:
@rudoi Hi again! 

I like this change, and I think we have some great integration tests to test this. I still bumped this change out to the v1.6 Milestone since I'm trying to get v1.5 out soon. Would you care to rebase this change against the latest master?

rudoi:
> @rudoi Hi again!
> 
> I like this change, and I think we have some great integration tests to test this. I still bumped this change out to the v1.6 Milestone since I'm trying to get v1.5 out soon. Would you care to rebase this change against the latest master?

Hi @mogren! Rebased, made sure tests are passing, etc. Looking forward to seeing the integration tests! I did some stress testing last week but I wasn't totally satisfied with my scenarios.

dadux:
@rudoi - thanks for this PR. 

We are about to deploy some of our clusters with this fix - so hopefully we'll get some real world use case before it gets merged into 1.6.

rudoi:
@dadux - awesome, looking forward to seeing the results!

We've been running this for about 3 weeks in real clusters and have not run into anything yet. -fingers crossed-

dadux:
Hi @mogren - what is preventing this to get merged ?

Happy to provide some help if needed !

rudoi:
Happy to help out as well!

mogren:
Hey @dadux and @rudoi! I just wanted to run a few more tests to try and catch any potential edge cases. Everything looks good on my end, so thanks a lot. Sorry it took so long to verify the change.

PR #387, opened 2019-04-09
Clean up function names, no change in functionality.

mogren:
*Description of changes:*

* Make function names match what the code does
* Removed some duplicated code
* Cleaned up logging

### DIFF
--- a/ipamd/datastore/data_store.go
+++ b/ipamd/datastore/data_store.go
@@ -186,8 +186,7 @@ func (ds *DataStore) AddENIIPv4Address(eniID string, ipv4 string) error {
 	defer ds.lock.Unlock()
 
 	log.Debugf("Adding ENI(%s)'s IPv4 address %s to datastore", eniID, ipv4)
-	log.Debugf("IP Address Pool stats: total: %d, assigned: %d",
-		ds.total, ds.assigned)
+	log.Debugf("IP Address Pool stats: total: %d, assigned: %d", ds.total, ds.assigned)
 
 	curENI, ok := ds.eniIPPools[eniID]
 	if !ok {
@@ -200,6 +199,7 @@ func (ds *DataStore) AddENIIPv4Address(eniID string, ipv4 string) error {
 	}
 
 	ds.total++
+	// Prometheus gauge
 	totalIPs.Set(float64(ds.total))
 
 	curENI.IPv4Addresses[ipv4] = &AddressInfo{address: ipv4, Assigned: false}
@@ -229,12 +229,12 @@ func (ds *DataStore) DelENIIPv4Address(eniID string, ipv4 string) error {
 	}
 
 	ds.total--
+	// Prometheus gauge
 	totalIPs.Set(float64(ds.total))
 
 	delete(curENI.IPv4Addresses, ipv4)
 
 	log.Infof("Deleted ENI(%s)'s IP %s from datastore", eniID, ipv4)
-
 	return nil
 }
 
@@ -244,7 +244,7 @@ func (ds *DataStore) AssignPodIPv4Address(k8sPod *k8sapi.K8SPodInfo) (string, in
 	ds.lock.Lock()
 	defer ds.lock.Unlock()
 
-	log.Debugf("AssignIPv4Address: IP address pool stats: total:%d, assigned %d", ds.total, ds.assigned)
+	log.Debugf("AssignIPv4Address: IP address pool stats: total: %d, assigned %d", ds.total, ds.assigned)
 	podKey := PodKey{
 		name:      k8sPod.Name,
 		namespace: k8sPod.Namespace,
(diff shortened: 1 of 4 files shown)


ipamd/datastore/data_store.go
```diff
@@ -285,37 +285,36 @@ func (ds *DataStore) assignPodIPv4AddressUnsafe(k8sPod *k8sapi.K8SPodInfo) (stri
 			if k8sPod.IP == addr.address {
 				// After L-IPAM restart and built IP warm-pool, it needs to take the existing running pod IP out of the pool.
 				if !addr.Assigned {
-					ds.assigned++
-					eni.AssignedIPv4Addresses++
-					assignedIPs.Set(float64(ds.assigned))
-					addr.Assigned = true
+					increaseAssignedCount(ds, eni, addr)
```
tiffanyfay: increase or increment?

ipamd/datastore/data_store.go
```diff
@@ -285,37 +285,36 @@ func (ds *DataStore) assignPodIPv4AddressUnsafe(k8sPod *k8sapi.K8SPodInfo) (stri
...
 				}
-				ds.podsIP[podKey] = PodIPInfo{IP: addr.address, DeviceNumber: eni.DeviceNumber}
 				log.Infof("AssignPodIPv4Address Reassign IP %v to pod (name %s, namespace %s)",
 					addr.address, k8sPod.Name, k8sPod.Namespace)
+				ds.podsIP[podKey] = PodIPInfo{IP: addr.address, DeviceNumber: eni.DeviceNumber}
```
tiffanyfay: Did you swap the order since the above says reassign vs reassigned? Also may be good to add the colon after AssignPodIPv4Address

ipamd/datastore/data_store.go
```diff
@@ -372,17 +372,17 @@ func (ds *DataStore) FreeENI() string {
 		return ""
 	}
 
-	ds.total -= len(ds.eniIPPools[deletableENI.ID].IPv4Addresses)
-	ds.assigned -= deletableENI.AssignedIPv4Addresses
-	log.Infof("FreeENI %s: IP address pool stats: free %d addresses, total: %d, assigned: %d",
-		deletableENI.ID, len(ds.eniIPPools[deletableENI.ID].IPv4Addresses), ds.total, ds.assigned)
-	deletedENI := deletableENI.ID
-	delete(ds.eniIPPools, deletableENI.ID)
+	removedENI := deletableENI.ID
```
tiffanyfay: I think this is slightly misleading because it's not removed yet here.

ipamd/datastore/data_store.go
```diff
@@ -372,17 +372,17 @@ func (ds *DataStore) FreeENI() string {
...
+	perENIIPAddressCount := len(ds.eniIPPools[removedENI].IPv4Addresses)
```
tiffanyfay: Why per? It would be more like removed if you go that way or just deletableENIPAddressCount.

mogren: How about just `eniIPCount`?

ipamd/introspect.go
```diff
@@ -132,48 +128,48 @@ func podV1RequestHandler(ipam *IPAMContext) func(http.ResponseWriter, *http.Requ
...
+			log.Error("Failed to marshal ipamd env var data: %v", err)
 			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
 			return
 		}
-		w.Write(responseJSON)
+		logErr(w.Write(responseJSON))
 	}
 }
 
-func metricsHandler(ipam *IPAMContext) func(http.ResponseWriter, *http.Request) {
```
nckturner: Was this unused?

ipamd/ipamd.go
```diff
@@ -746,26 +744,26 @@ func (c *IPAMContext) nodeIPPoolReconcile(interval time.Duration) {
...
 
-		// add new ENI
+		// Add new ENI
 		log.Debugf("Reconcile and add a new ENI %s", attachedENI)
 		err = c.setupENI(attachedENI.ENIID, attachedENI)
 		if err != nil {
 			log.Errorf("IP pool reconcile: Failed to setup ENI %s network: %v", attachedENI.ENIID, err)
 			ipamdErrInc("eniReconcileAdd", err)
-			//continue if having trouble with ONLY 1 ENI, instead of bailout here?
+			// Continue if having trouble with ONLY 1 ENI, instead of bailout here?
```
nckturner: This shouldn't be a question either

ipamd/datastore/data_store.go
```diff
@@ -338,31 +337,32 @@ func (ds *DataStore) getDeletableENI() *ENIIPPool {
...
 	return nil
 }
 
-// FreeENI frees a deletable ENI.
-// It returns the name of the ENI which is deleted out the data store
-func (ds *DataStore) FreeENI() string {
+// RemoveUnusedENIFromDataStore removes a deletable ENI from the data store.
+// It returns the name of the ENI which has been removed from the data store and needs to be deleted,
+// or empty string if no ENI could be removed.
+func (ds *DataStore) RemoveUnusedENIFromDataStore() string {
```
nckturner: I was going to say this should just be simply`RemoveENIFromDataStore` but then I realized there is also a DeleteENI and they neither calls the other they just do really similar things.  So I guess this should be refactored in a future PR.

ipamd/introspect.go
mogren: Yes, not used at all.

ipamd/datastore/data_store.go
mogren: Definitely. This rename was just so we don't mix up the two FreeENI methods.

Issue #314, opened 2019-02-05
L-IPAMD IP allocation and usage configuration possibility


daugustus:
I have been experiencing some challenges with IP starvation where an EKS cluster uses all, or almost all available IP addresses within each subnet. I understand the reasoning about pre-assigning ENIs and IPs due to latency that can occur. However, we have some use cases where this behavior is unwanted. Is there a configuration setting that can control/limit the number of IPs assigned? For example, some workloads require higher resources so that larger instances are created for worker nodes. In these cases, alot of the IPs are not going to be used as there will be few pods assigned. Thoughts? Ideas?

nak3:
Do you want to control less than the num of max IPs per ENI? For example, `m4.4xlarge` consumes 30 IP addresses per 1 ENI. You want to limit the num of IPs less than 30.

If No (you don't need to control less than 30 for `m4.4xlarge`), you can set `WARM_ENI_TARGET` or `WARM_IP_TARGET`. Or you can use `MAX_ENI` https://github.com/aws/amazon-vpc-cni-k8s/pull/251 in the future release. I believe `MAX_ENI` is not implemented in current image, though.

If Yes (you want to control less than 30 for `m4.4xlarge`), I am pretty sure that it is not possible for now. ipamD always allocates IPs at least `maxAddrsPerENI` (=`30` IPs for `m4.4xlarge`) as:

https://github.com/aws/amazon-vpc-cni-k8s/blob/4a51ed41abf61b7445f8226e5a086a8ec9b0581b/ipamd/ipamd.go#L723

daugustus:
Yes- I would want to limit the max number of IPs per instance.

mogren:
@daugustus If your pod churn is low, and you only have a few pods per node, setting `WARM_IP_TARGET` should solve this issue for you. Setting `MAX_ENI` would also cap your upper limit.

PR #412, opened 2019-04-25
Unassign IPs over warm IP target

nckturner:
This is a change to the behavior of WARM_IP_TARGET.  When the plugin's IP address pool goes beyond WARM_IP_TARGET, it will unassign excess IP addresses in order to reduce the total number of allocated IP addresses.

The way it does this is when the number of available IPs is short of WARM_IP_TARGET, we allocate the remaining IPs on the ENI.  Every 30 seconds, if IPs remain ununsed, the `decreaseIPPool` function will deallocate IPs left over on ENIs down to WARM_IP_TARGET.  

*Issue #, if available:*
Helps with concerns like #314 where total IP usage in a VPC is a problem

*Description of changes:*
* Moved IP allocation and retry logic into `increaseIPPool()`, 
* Removed `retryAllocENIIP()`.  
* Modified c.decreaseIPPool() to be `c.decreaseIPPool(decreaseIPPoolInterval)`, so that it does not run as frequently (now every 30 seconds) to allow for IPs to be used by kubelet before being unassigned and removed from the pool.  
* Additionally renamed a number of functions and variables to better reflect their function.


pkg/awsutils/awsutils.go
```diff
@@ -883,6 +886,34 @@ func (cache *EC2InstanceMetadataCache) AllocIPAddresses(eniID string, numIPs int
 	return nil
 }
 
+// DeallocIPAddresses allocates numIPs of IP address on an ENI
+func (cache *EC2InstanceMetadataCache) DeallocIPAddresses(eniID string, ips []string) error {
+	ctx := context.Background()
```
nckturner: Currently just using a dummy context, but in a future change we should pipe context from higher up so that we can take advantage of cancellation.

ipamd/ipamd.go
```diff
@@ -348,177 +352,254 @@ func (c *IPAMContext) StartNodeIPPoolManager() {
 }
 
 func (c *IPAMContext) updateIPPoolIfRequired() {
-	c.retryAllocENIIP()
```
nckturner: This logic was moved into increaseIPPool, since that's what its doing, and if WARM_IP_TARGET is set then its not always a retry.

ipamd/ipamd.go
```diff
@@ -348,177 +352,254 @@ func (c *IPAMContext) StartNodeIPPoolManager() {
...
+		skipPrimary := UseCustomNetworkCfg()
+		if skipPrimary && eni.IsPrimary {
+			log.Debugf("Skip the primary ENI for need IP check")
+			continue
+		}
+		if int64(len(eni.IPv4Addresses)) < maxIPPerENI {
+			log.Debugf("Found ENI %s that has less than the maximum number of IP addresses allocated: cur=%d, max=%d",
+				eni.ID, len(eni.IPv4Addresses), maxIPPerENI)
+
+			err = c.awsClient.AllocIPAddresses(eni.ID, maxIPPerENI-int64(len(eni.IPv4Addresses)))
```
nckturner: Here is the IP allocation logic for each ENI.  We attempt to allocate the remaining IPs on the ENI, keep them around until the next `decreaseIPPool` (which is called if we are over WARM_IP_TARGET), and then they will be unassigned if they are still not used.  We only get here if we have less than `WARM_IP_TARGET`, so this doesn't continuously assign and unassign IPs.

ipamd/ipamd.go
```diff
@@ -348,177 +352,254 @@ func (c *IPAMContext) StartNodeIPPoolManager() {
...
-	maxIPLimit, err := c.awsClient.GetENIipLimit()
-	if err != nil {
-		log.Infof("Failed to retrieve ENI IP limit: %v", err)
-		return
-	}
-	eni := c.dataStore.GetENINeedsIP(maxIPLimit, UseCustomNetworkCfg())
-	if eni != nil {
-		log.Debugf("Attempt again to allocate IP address for ENI :%s", eni.ID)
-		var err error
-		if warmIPTargetDefined {
```
nckturner: See line 587 to see where the IP allocation logic moved for cases where ENIs should have more assigned IPs than they currently have.

ipamd/ipamd.go
```diff
@@ -850,17 +929,25 @@ func getWarmIPTarget() int {
 	return noWarmIPTarget
 }
 
-func (c *IPAMContext) getCurWarmIPTarget() (int64, bool) {
+func (c *IPAMContext) ipTargetState() (short int, over int, enabled bool) {
```
nckturner: This now returns short (the number of IPs that we are short of the target), over (the number of IPs that we are over the target) and enabled.  I thought it made more sense than using a negative number to represent short, but maybe its more confusing.  Appreciate feedback.

ipamd/ipamd.go
```diff
@@ -348,177 +352,254 @@ func (c *IPAMContext) StartNodeIPPoolManager() {
...
+	c.lastDecreaseIPPool = now
+	c.lastNodeIPPoolAction = now
 	total, used := c.dataStore.GetStats()
 	log.Debugf("Successfully decreased IP pool")
 	logPoolStats(int64(total), int64(used), c.currentMaxAddrsPerENI, c.maxAddrsPerENI)
 }
 
+// findFreeableIPs finds and returns IPs that are not assigned to Pods but are attached
+// to ENIs on the node.
+func (c *IPAMContext) findFreeableIPs(eni string) ([]string, error) {
```
mogren: Would be nice with a unit test for this one.

ipamd/ipamd.go
```diff
@@ -348,177 +352,254 @@ func (c *IPAMContext) StartNodeIPPoolManager() {
...
+				return
+			}
+
+			if len(ips) == 0 {
+				continue
+			}
+
+			// Deallocate IPs from the instance if they aren't used by pods.
+			if err := c.awsClient.DeallocIPAddresses(eniID, ips); err != nil {
+				log.Debugf(fmt.Sprintf("Failed to decrease IP pool by removing IPs %v from ENI %s: %s", ips, eniID, err))
```
mogren: What happens if we try to free 5 IPs and the call fails. Could 3 have been freed? Might be handled by the reconcile loop.

go.mod
```diff
@@ -3,58 +3,63 @@ module github.com/aws/amazon-vpc-cni-k8s
 go 1.12
 
 require (
```
mogren: What is the actual dependency change? How easy is this PR to cherry pick over to the `release-1.4` branch?

go.mod
```diff
@@ -3,58 +3,63 @@ module github.com/aws/amazon-vpc-cni-k8s
...
 	github.com/aws/aws-sdk-go v1.15.22
-	github.com/beorn7/perks v0.0.0-20180321164747-3a771d992973
+	github.com/beorn7/perks v0.0.0-20180321164747-3a771d992973 // indirect
 	github.com/cihub/seelog v0.0.0-20151216151435-d2c6e5aa9fbf
 	github.com/containernetworking/cni v0.5.2
 	github.com/coreos/go-iptables v0.4.0
-	github.com/davecgh/go-spew v1.1.1
-	github.com/docker/distribution v2.6.2+incompatible
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/deckarep/golang-set v1.7.1
```
nckturner: `golang-set` is the only dependency I know of that actually changed.

jqmichael:
Not sure what happened. But I couldn't find any references to `c.decreaseIPPool(decreaseIPPoolInterval)` in the latest commit.

mogren:
Superseded by #461

PR #406, opened 2019-04-22
Adding additional CRD for Calico that was missing.

wmorgan6796:
Description of changes: The Calico deployment was missing a CRD causing it to have errors. I have added the missing CRD to remove the errors.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #421, opened 2019-04-26
Remove globalfelixconfigs and globalbgpconfigs from calico config

tiffanyfay:
*Issue #, if available:*
N/A

*Description of changes:*
Removed globalfelixconfigs and globalbgpconfigs from calico config following conversation https://github.com/aws/amazon-vpc-cni-k8s/pull/406#issuecomment-487150036

### DIFF
--- a/config/v1.4/calico.yaml
+++ b/config/v1.4/calico.yaml
@@ -362,10 +362,8 @@ rules:
       - list
   - apiGroups: ["crd.projectcalico.org"]
     resources:
-      - globalfelixconfigs
       - felixconfigurations
       - bgppeers
-      - globalbgpconfigs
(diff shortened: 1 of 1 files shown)


caseydavenport:
@tiffanyfay @mogren I think I led you guys in the wrong direction on this one, it turns out the new code _does_ in fact check for this as part of the migration code.

It doesn't need the CRD to exist, but it needs the permission otherwise it will fail.

We should probably fix upstream in Calico so that it doens't need this any more, but for now I think we'll need to revert this change... Sorry for the runaround.

mogren:
Thanks @caseydavenport, I agree this seem to cause an issue. We will revert the change for now.

PR #422, opened 2019-04-26
Support for unassigning IPs

nckturner:
*Description of changes:*
* When WARM_IP_TARGET is set, unassign IPs if the number of
available IPs is greater than the WARM_IP_TARGET.
* Unassigns from all ENIs at once.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ install:
 
 # Tests to run
 script:
-  - make build-linux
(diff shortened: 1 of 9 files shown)


nckturner:
@jqmichael I think you missed the changes in ipamd.go, which is what the description is referring to (it doesn't load by default in the diff, you have to click on it).

PR #427, opened 2019-04-30
add support for t3a.* instance types

jhoblitt:
Note that the limits are copies of the t3 values and may need to be adjusted correct.

This may (?) require a newer version of aws/aws-sdk-go.

blocks https://github.com/awslabs/amazon-eks-ami/pull/255
related to #428


jhoblitt:
https://github.com/aws/aws-sdk-go/pull/2571

mogren:
Included in #459

pkg/awsutils/vpc_ip_resource_limit.go
```diff
@@ -180,6 +180,13 @@ var InstanceENIsAvailable = map[string]int{
 	"t3.large":      3,
 	"t3.xlarge":     4,
 	"t3.2xlarge":    4,
+	"t3a.nano":      2,
```
mogren: We need IPs per ENI as well. I added this in #459

Issue #426, opened 2019-04-30
Use -w in support script


mogren:
In the support script, we should wait for the xlock to be released.

Zyqsempai:
Sorry, i am new here, and I am looking for first issue;)
Can you provide little bit more info, do you mean this [script](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/aws-cni-support.sh) ?

mogren:
@Zyqsempai Yes, exactly. The lines with `iptables` should have the `-w` option enabled.

PR #439, opened 2019-05-02
Added -W (wait for xlock's) flag to iptables commands

Zyqsempai:
*Issue #426*

*Description of changes: Added -W (wait for xlock's) flag to iptables commands*

### DIFF
--- a/scripts/aws-cni-support.sh
+++ b/scripts/aws-cni-support.sh
@@ -52,14 +52,14 @@ ip rule show > ${LOG_DIR}/iprule.out
 # iptables-save
 iptables-save > ${LOG_DIR}/iptables-save.out
 
-# iptables -nvL
-iptables -nvL > ${LOG_DIR}/iptables.out
+# iptables -w1 -nvL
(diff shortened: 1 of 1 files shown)


scripts/aws-cni-support.sh
```diff
@@ -52,14 +52,14 @@ ip rule show > ${LOG_DIR}/iprule.out
 # iptables-save
 iptables-save > ${LOG_DIR}/iptables-save.out
 
-# iptables -nvL
-iptables -nvL > ${LOG_DIR}/iptables.out
+# iptables -wnvL
+iptables -wnvL > ${LOG_DIR}/iptables.out
```
mogren: Have you tested that it works? I think you need to give `-w` an argument of how many seconds to wait. I'd guess 1s would be enough, so how about `iptables -w1 -nvL`?

Zyqsempai: "-w" has a default time to wait, but honestly, I am not sure which one, so I agree with your idea.

Zyqsempai:
@mogren Done

PR #443, opened 2019-05-03
Fix issue with ipamd ENI allocation

nckturner:
*Description of changes:*
* ipamd would free and reallocate an ENI sometimes when
  it shouldn't have.

### DIFF
--- a/ipamd/ipamd.go
+++ b/ipamd/ipamd.go
@@ -850,7 +850,7 @@ func (c *IPAMContext) shouldRemoveExtraENIs() bool {
 	warmENITarget := getWarmENITarget()
 	total, used := c.dataStore.GetStats()
 	available := total - used
-  // We need the +1 to account for the Primary ENI.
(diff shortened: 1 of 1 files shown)


ipamd/ipamd.go
```diff
@@ -844,7 +844,7 @@ func (c *IPAMContext) shouldRemoveExtraENIs() bool {
 	warmENITarget := getWarmENITarget()
 	total, used := c.dataStore.GetStats()
 	available := total - used
-	return int64(available) > int64(warmENITarget) * c.maxAddrsPerENI
+	return int64(available) - int64(c.maxAddrsPerENI) > int64(warmENITarget) * c.maxAddrsPerENI
```
mogren: IIRC, this used to say 
```
return int64(available) >= int64(warmENITarget + 1) * c.maxAddrsPerENI
```
I see why now. We should have a unit test for this.

PR #451, opened 2019-05-07
Switch from skel.PluginMain to skel.PluginMainWithError, as the log h…

venkatesh-eb:
…as to be flushed before os.Exit is called.

https://github.com/aws/amazon-vpc-cni-k8s/blob/957f835f437fd76953f37e6b409ba6426bb4ce37/vendor/github.com/containernetworking/cni/pkg/skel/skel.go#L226

*Description of changes:*
- Flush the log before exiting when cni plugin has an error.

### DIFF
--- a/plugins/routed-eni/cni.go
+++ b/plugins/routed-eni/cni.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net"
+	"os"
 	"runtime"
 
 	"golang.org/x/net/context"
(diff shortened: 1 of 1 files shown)


plugins/routed-eni/cni.go
```diff
@@ -142,7 +143,7 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 			K8S_POD_NAME:               string(k8sArgs.K8S_POD_NAME),
 			K8S_POD_NAMESPACE:          string(k8sArgs.K8S_POD_NAMESPACE),
 			K8S_POD_INFRA_CONTAINER_ID: string(k8sArgs.K8S_POD_INFRA_CONTAINER_ID),
-			IfName: args.IfName})
+			IfName:                     args.IfName})
```
mogren: Thanks.

PR #452, opened 2019-05-07
No pods is not an error

mogren:
*Description of changes:*
Having a node with no pods is fine.


ipamd/ipamd.go
```diff
@@ -319,12 +319,16 @@ func (c *IPAMContext) getLocalPodsWithRetry() ([]*k8sapi.K8SPodInfo, error) {
...
 	}
 
+	if err != nil {
+		return nil, err
+	}
+
 	if pods == nil {
-		return nil, errors.New("unable to get local pods, giving up")
+		log.Info("No pods found on this node")
+		return pods, nil
```
nckturner: nit: could return nil here just to make it a little more explicit

PR #456, opened 2019-05-09
Avoid unnecessary call to EC2

mogren:
*Description of changes:*

No need to add 0 IPs.

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -866,6 +866,11 @@ func (cache *EC2InstanceMetadataCache) AllocIPAddresses(eniID string, numIPs int
 		needIPs = ipLimit
 	}
 
+	// If we don't need any more IPs, exit
+	if needIPs < 1 {
+		return nil
+	}
+
 	log.Infof("Trying to allocate %d IP address on ENI %s", needIPs, eniID)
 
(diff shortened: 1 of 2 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -866,6 +866,11 @@ func (cache *EC2InstanceMetadataCache) AllocIPAddresses(eniID string, numIPs int
 		needIPs = ipLimit
 	}
 
+	// If we don't need any more IPs, exit
```
tiffanyfay: 864 has no check for err != nil.

pkg/awsutils/awsutils.go
```diff
@@ -866,6 +866,11 @@ func (cache *EC2InstanceMetadataCache) AllocIPAddresses(eniID string, numIPs int
 		needIPs = ipLimit
```
tiffanyfay: Should we log to let it be known that the ipLimit is less than the needed #?

pkg/awsutils/awsutils.go
mogren: We should handle unknown instances better, but that's a different change.

pkg/awsutils/awsutils.go
mogren: Right now we log that it the `tryAssignIPs()`

PR #457, opened 2019-05-09
If GetAttachedENIs has problems, still exit the loop

mogren:
*Description of changes:*

The loop could get stuck.


ipamd/ipamd.go
```diff
@@ -707,9 +707,14 @@ func (c *IPAMContext) waitENIAttached(eni string) (awsutils.ENIMetadata, error)
 	// Wait until the ENI shows up in the instance metadata service
 	retry := 0
 	for {
+		retry++
```
tiffanyfay: This code appears to retry one time more than the max retries.

mogren: Moved the increase to the end of the loop.

Issue #359, opened 2019-03-23
aws-node stuck in crash recycle attempting to use an ENI it deleted


peterbroadhurst:
One of our nodes encountered an issue where the `aws-node` CNI pod is constantly crash recycling, and the ipamd logs show:
```
2019-03-22T16:12:20Z [ERROR] Failed to setup eni eni-0c18e47af27178549 network: failed to retrieve eni eni-0c18e47af27178549 ip addresses: fail to find eni addresses for eni eni-0c18e47af27178549: failed to describe network interface: InvalidNetworkInterfaceID.NotFound: The networkInterface ID 'eni-0c18e47af27178549' does not exist
```
Node is tainted to isolate while we investigate.

> We have a ticket open with AWS support, as the conclusion is the metadata API is returning stale data, and we have an ENI stuck in attaching status on this instance

I am raising this and will report root cause back once determined, in case others see the same symptoms.

Reviewing the earlier ipamd logs I can see that the ENI was released and deleted, but just a minute later the `Reconciling ENI/IP pool info` logic re-added the deleted ENI back into the pool

```
2019-03-22T14:26:53Z [DEBUG] FreeENI: found a deletable ENI eni-0c18e47af27178549
2019-03-22T14:26:53Z [INFO] FreeENI eni-0c18e47af27178549: IP address pool stats: free 14 addresses, total: 28, assigned: 14
2019-03-22T14:26:53Z [DEBUG] Start freeing eni eni-0c18e47af27178549
2019-03-22T14:26:53Z [INFO] Trying to free eni: eni-0c18e47af27178549
2019-03-22T14:26:53Z [INFO]  Pods deleted on my node: a0ptn8lpi5/a0ptn8lpi5-a0ox4r7pd8-quorum-node-0
2019-03-22T14:26:54Z [DEBUG] Found eni eni-0c18e47af27178549 attachment id: eni-attach-0263261d435bf0466
2019-03-22T14:26:54Z [DEBUG] Trying to delete eni: eni-0c18e47af27178549
2019-03-22T14:26:54Z [DEBUG] Not able to delete eni yet (attempt 1/20): InvalidParameterValue: Network interface 'eni-0c18e47af27178549' is currently in use.
        status code: 400, request id: 17b480ad-6a54-4959-90da-ff6e9d7d443b
[...]

PR #458, opened 2019-05-09
Avoid using force detach of ENIs

mogren:
*Issue #, if available:* #359

*Description of changes:*

* Not using force
* Added retries

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -731,22 +731,31 @@ func (cache *EC2InstanceMetadataCache) FreeENI(eniName string) error {
 		log.Errorf("Failed to retrieve ENI %s attachment id: %v", eniName, err)
 		return errors.Wrap(err, "FreeENI: failed to retrieve ENI's attachment id")
 	}
-
 	log.Debugf("Found ENI %s attachment id: %s ", eniName, aws.StringValue(attachID))
 
-	// Detach it first
 	detachInput := &ec2.DetachNetworkInterfaceInput{
 		AttachmentId: attachID,
-		Force:        aws.Bool(true),
 	}
 
-	start := time.Now()
-	_, err = cache.ec2SVC.DetachNetworkInterface(detachInput)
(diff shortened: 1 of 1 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -731,22 +731,33 @@ func (cache *EC2InstanceMetadataCache) FreeENI(eniName string) error {
...
-	start := time.Now()
-	_, err = cache.ec2SVC.DetachNetworkInterface(detachInput)
-	awsAPILatency.WithLabelValues("DetachNetworkInterface", fmt.Sprint(err != nil)).Observe(msSince(start))
-	if err != nil {
-		awsAPIErrInc("DetachNetworkInterface", err)
-		log.Errorf("Failed to detach ENI %s %v", eniName, err)
-		return errors.Wrap(err, "FreeENI: failed to detach ENI from instance")
+	// Retry detaching the ENI from the instance
+	var retry int
+	for retry := 0; retry < maxENIDeleteRetries; retry++ {
```
tiffanyfay: Remove the :

pkg/awsutils/awsutils.go
```diff
@@ -731,22 +731,33 @@ func (cache *EC2InstanceMetadataCache) FreeENI(eniName string) error {
...
+		if err == nil {
+			log.Infof("Successfully detached ENI: %s", eniName)
+			break
+		}
+
+		log.Debugf("Not able to detach ENI yet (attempt %d/%d): %v ", retry, maxENIDeleteRetries, err)
+		time.Sleep(retryDeleteENIInternal)
+	}
+
+	if retry >= maxENIDeleteRetries {
```
tiffanyfay: Retry can never be more than max.
Also, this will give a failure even if the for loop passes on the last retry. 
Some options are a bool, or turning this into a function.
Another is to move 759 (and make it ==) to 749.

mogren: Ah, made it `==`, but then the loop needs to start at 1. PR updated

Issue #391, opened 2019-04-10
CNI can assign IP address that are no longer attached to the instance


nckturner:
I had an m4.large (2 ENIs at 10 IP addresses each) which was saturated with pods, and I detached one ENI, then scaled down the pods to 0, then deleted the ENI.  Then I scaled the pods back up so that the node was saturated, and every pod entered the running state and and was allocated an IP address by the CNI.  But, when I look at the ENIs allocated to the node, there is only 1 ENI.  That is, the CNI never requested a new ENI, and instead handed out stale IP addresses to the Pods.  This results in connection timeouts when packets are routed to the pods that are behind non-existent IP addresses.

Issue #377, opened 2019-04-04
IPs are allocated to ENIs in an inconsistent way


nckturner:
When IPAMD attempts to meet pod IP demands by allocating ENIs and IP addresses, it allocates new IP addresses and attaches them to the ENI in an inconsistent way.  This behavior is easy to reproduce.  
```
# Using https://github.com/aws/aws-k8s-tester:
aws-k8s-tester eks create config --path ./cni-test.yaml
```
Check some of the config values: 
```
# cni-test.yaml
kubernetes-version: "1.11"
...
worker-node-asg-desired-capacity: 1
worker-node-asg-max: 1
worker-node-asg-min: 1
worker-node-instance-type: m3.xlarge
```

```
aws-k8s-tester eks create cluster --path ./cni-test.yaml
```
After cluster is ready and KUBECONFIG is set to generated kubeconfig file:
```
k run nginx --image nginx -r 50
```
Using the aws cli or console, detach an unused IP address, (determine it is unused by any pod by choosing one and cross referencing with `kubectl get pod -o wide --all-namespaces`.

Then, scale `nginx` to 60 replicas.  The deployment should never get passed 55 pods in RUNNING state, when it should be 56.

mogren:
We should verify that the reconcile loop in ipamd detects this and removes the detached IP from the data-store.

mogren:
This has been added in #1177 where we check with EC2 in the reconcile loop in the case where the datastore and IMDS does not match.

Issue #294, opened 2019-01-22
IP addresses still allocated to completed pods


hibbert:
Hi,

We have some CronJobs that run hourly and then the pods return to Completed. It looks like the IP address is still in use. The ENI itself is still in use with other secondary IPs. But wandering why this stays as "In Use" despite the cronjob being complete and no pods are running? Also what is the cooling period?

Thanks

mogren:
Hi @hibbert! 

The reasons we have the IPs around is an optimization. Also, they are not allocated to the pods, but kept as a "warm pool" per ENI for the CNI plugin to pick from when attaching IPs to new pods. Some more details in my reply on [issue 123](https://github.com/aws/amazon-vpc-cni-k8s/issues/123#issuecomment-455403330).

mogren:
I've added some improved logging, but not been able to track down the issue. Punting this issue to 1.5.

mogren:
Resolving since v1.5.0 is released. If you want the IPs to be released from attached ENIs, please set the `WARM_IP_TARGET` environment variable.

Issue #265, opened 2018-12-13
Information Required : POD IP address limit while using amazon-vpc-routed-eni


sam0104:
I am testing amazon-vpc-routed-eni inside AWS with KOPS deployed using --topology=private
**My scenario** 
I am deploying multiple kubernetes cluster ( assume 100 ) and each kubernetes cluster use separate VPC with unique VPC CIDR. 

No VPC is going to use same CIDR for avoiding conflict 
Deployed Kubernetes cluster using KOPS with amazon-vpc-routed-eni  networking , which allow eacg POD to have IP address from CIDR range 

Q : Is it fine to use amazon-vpc-routed-eni for large Kubernetes setup , where each VPC has unique CIDR ?

Q: What is the limit for amazon-vpc-routed-eni for providing POD IP from a /16 CIDR. Example : 172.16.0.0/16 , 10.0.0.0/16 

Q: How many maximum POD IP can be assigned using amazon-vpc-routed-eni if I am using the large ec2 VM m2.4xlarge  which supports 8 network interface and 30 IPV4 address per interface. Which provides me 30x8 = 240 IPV4 IP addresses from **one Kubernetes worker node** . If I am using 6 Kubernetes worker node and each worker node is m2.large then I will get 6 x 240 = **1440 IPV4** IP addresses per Kubernetes cluster. Am I in the right direction for calculating POD IP addresses per cluster ?

Q: Is there a maximum limit for amazon-vpc-routed-eni for deploying worker nodes. Can I deploy 5000  worker nodes with c4.8xlarge using amazon-vpc-routed-eni in a VPC with 172.16.0.0/16 CIDR to get maximum POD  IPV4 addresses ?
I am referring  to Kubernetes : https://kubernetes.io/docs/setup/cluster-large/

labria:
1. Yes, it's ok to use amazon-vpc-routed-eni in this case.
2. /16 is ~32k addresses in total (if you assign the whole /16 to the subnet you're gonna use with Kube).
3. Yes, you're right. Here's a complete table for calculations: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html#AvailableIpPerENI. Although, notice that https://kubernetes.io/docs/setup/cluster-large/ says **No more than 100 pods per node**, so you might want smaller instances (not sure what's the reasoning behind this limit, or how hard a limit this is).
[...]

PR #461, opened 2019-05-10
Improve WARM_IP_TARGET handling

mogren:
*Issue #, if available:* #377, #294, #314, #391, #265 and others.

*Description of changes:*
Merge of all fixes that has gone into the 1.4 branch back into master. Add support for actually freeing unused IPs from an attached ENI. Requires the `ec2:UnassignPrivateIPs` permission.


ipamd/ipamd.go
```diff
@@ -208,7 +253,11 @@ func (c *IPAMContext) nodeInit() error {
 
 	log.Debugf("Start node init")
 
-	instanceMaxENIs, _ := c.awsClient.GetENILimit()
+	instanceMaxENIs, err := c.awsClient.GetENILimit()
+	if err != nil {
+		log.Errorf("Failed to get ENI limit: %s")
```
vsiddharth: Should we not error out in case the limits cannot be determined

ipamd/ipamd.go
```diff
@@ -256,7 +319,7 @@ func (c *IPAMContext) nodeInit() error {
 		ipamdErrInc("nodeInitK8SGetLocalPodIPsFailed", err)
 		// This can happens when L-IPAMD starts before kubelet.
 		// TODO  need to add node health stats here
-		return nil
+		return errors.Wrap(err, "Failed to get running pods!")
```
vsiddharth: nit: s/Failed/failed

ipamd/ipamd.go
```diff
@@ -316,12 +379,11 @@ func (c *IPAMContext) getLocalPodsWithRetry() ([]*k8sapi.K8SPodInfo, error) {
 	}
 
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "no pods because apiserver not running")
```
vsiddharth: nit: add `.` in the end

ipamd/ipamd.go
mogren: I agree we should handle unknown instance types in a better way, but this change was just to keep it consistent with line 576 were we do the same thing.

PR #472, opened 2019-05-16
Don't modify maxENI

nckturner:
*Description of changes:*
* Don't "discover" maxENI and modify the value dynamically, which
  can lead to strange situations

### DIFF
--- a/ipamd/ipamd.go
+++ b/ipamd/ipamd.go
@@ -158,10 +158,6 @@ type IPAMContext struct {
 
 	currentMaxAddrsPerENI int
 	maxAddrsPerENI        int
-	// maxENI indicate the maximum number of ENIs can be attached to the instance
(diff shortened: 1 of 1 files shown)


ipamd/ipamd.go
```diff
@@ -590,13 +590,8 @@ func (c *IPAMContext) increaseIPPool() {
 	}
 
 	if c.dataStore.GetENIs() < maxENIs {
-		// c.maxENI represent the discovered maximum number of ENIs
-		if (c.maxENI > 0) && (c.maxENI == c.dataStore.GetENIs()) {
```
mogren: Do we want to remove `maxENI` from `IPAMContext` as well, since it was only ever used here?

nckturner: Yes, I thought I had.

PR #480, opened 2019-05-18
On start up, wait for pods with no IP

mogren:
*Description of changes:*
If ipamd gets restarted while a pod is registering it's newly assigned IP with the apiserver, we might miss the assigned IP because of the eventual consistency of Kubernetes. This change will wait for all running pods for up to 1 minute to get an IP set.

### DIFF
--- a/ipamd/ipamd.go
+++ b/ipamd/ipamd.go
@@ -368,7 +368,18 @@ func (c *IPAMContext) getLocalPodsWithRetry() ([]*k8sapi.K8SPodInfo, error) {
 	for retry := 1; retry <= maxK8SRetries; retry++ {
 		pods, err = c.k8sClient.K8SGetLocalPodIPs()
 		if err == nil {
-			break
+			// Check for pods with no IP since the API server might not have the latest state of the node.
+			allPodsHaveAnIP := true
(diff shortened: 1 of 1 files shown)


ipamd/ipamd.go
```diff
@@ -368,7 +368,19 @@ func (c *IPAMContext) getLocalPodsWithRetry() ([]*k8sapi.K8SPodInfo, error) {
...
+			for _, pod := range pods {
+				if pod.IP == "" {
+					log.Infof("Pod %s, Namespace %s, has no IP", pod.Name, pod.Namespace)
+					allPodsHaveAnIP = false
+				}
+			}
+			if allPodsHaveAnIP {
+				break
+			} else {
+				log.Warnf("Not all pods has an IP, trying again in %s more seconds.", retryK8SInterval.Seconds())
```
tiffanyfay: *have

I might drop "more"

tiffanyfay: Why have this else when you have line 385? Can just add the retry interval to 385?

mogren: Mostly to log that we did find pods, but that we keep retrying since they might not be in sync.

Issue #460, opened 2019-05-09
Errors not handled


tiffanyfay:
- [ ] `AllocIPAddresses` `ipLimit, err := cache.GetENIipLimit()` https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/awsutils/awsutils.go

Zyqsempai:
I will take care of this one.

PR #484, opened 2019-05-24
Added error handling for GetENIipLimit in AllocIPAddresses func

Zyqsempai:
Issue #460 

*Description of changes: Added error handling for GetENIipLimit in AllocIPAddresses func*

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -870,7 +870,12 @@ func (cache *EC2InstanceMetadataCache) AllocIPAddresses(eniID string, numIPs int
 	var needIPs = numIPs
 
 	ipLimit, err := cache.GetENIipLimit()
-	if err == nil && ipLimit < needIPs {
+	if err != nil {
+		awsUtilsErrInc("UnknownInstanceType", err)
+		return err
+	}
+
+	if ipLimit < needIPs {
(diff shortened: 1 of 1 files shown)


Zyqsempai:
@tiffanyfay PTAL

pkg/awsutils/awsutils.go
```diff
@@ -870,7 +870,13 @@ func (cache *EC2InstanceMetadataCache) AllocIPAddresses(eniID string, numIPs int
 	var needIPs = numIPs
 
 	ipLimit, err := cache.GetENIipLimit()
-	if err == nil && ipLimit < needIPs {
+	if err != nil {
+		awsUtilsErrInc("UnknownInstanceType", err)
+		// for unknown instance type, will allocate one ip address at a time
+		ipLimit = 1
```
mogren: On error, we should probably just return it, same as for `getMaxEni()`. I added #485 to remove the unused code where the 1 came from. We shouldn't default to 1 IP, that won't work.

mogren: Btw to clarify. We should call `awsUtilsErrInc("UnknownInstanceType", err)`, but after that just `return err`.

Zyqsempai: @mogren Done

PR #487, opened 2019-05-28
Add IPs to existing ENIs first

mogren:
*Description of changes:*
* When `WARM_IP_TARGET` is set, try to allocate all unallocated IPs for the current ENI. If that fails, try to just add  `1` IP to an existing ENI before allocating a new ENI.
* Remove `currentMaxAddrsPerENI` since it will not change.
* Only read `customNetworkCfg` once

### DIFF
--- a/ipamd/ipamd.go
+++ b/ipamd/ipamd.go
@@ -148,15 +148,14 @@ var (
 
 // IPAMContext contains node level control information
 type IPAMContext struct {
-	awsClient        awsutils.APIs
-	dataStore        *datastore.DataStore
-	k8sClient        k8sapi.K8SAPIs
-	customNetworkCfg bool
-	eniConfig        eniconfig.ENIConfig
-	dockerClient     docker.APIs
-	networkClient    networkutils.NetworkAPIs
-	maxAddrsPerENI   int
-	// maxENI is the maximum number of ENIs that can be attached to the instance
+	awsClient            awsutils.APIs
+	dataStore            *datastore.DataStore
+	k8sClient            k8sapi.K8SAPIs
+	useCustomNetworking  bool
+	eniConfig            eniconfig.ENIConfig
+	dockerClient         docker.APIs
+	networkClient        networkutils.NetworkAPIs
+	maxIPsPerENI         int
 	maxENI               int
 	warmENITarget        int
 	warmIPTarget         int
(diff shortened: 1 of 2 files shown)


ipamd/ipamd.go
```diff
@@ -582,44 +579,36 @@ func (c *IPAMContext) increaseIPPool() {
...
-		c.tryAllocateENI()
-		c.updateLastNodeIPPoolAction()
-	} else {
-		log.Debugf("Skipping ENI allocation due to max ENI already attached to the instance: %d", c.maxENI)
-	}
-
+	// Try to add more IPs to existing ENIs first.
 	increasedPool, err := c.tryAssignIPs()
 	if err != nil {
 		log.Errorf(err.Error())
```
tiffanyfay: What about `ipamdErrInc`?

mogren: I only increase it after both trying all and then 1 ip:

https://github.com/aws/amazon-vpc-cni-k8s/pull/487/files#diff-3fbbb0bd392fd47bd85a56d6221de538R682

ipamd/ipamd.go
```diff
@@ -582,44 +579,36 @@ func (c *IPAMContext) increaseIPPool() {
...
 	}
 	if increasedPool {
 		c.updateLastNodeIPPoolAction()
+	} else {
+		// If we did not add an IP, try to add an ENI instead.
+		if c.dataStore.GetENIs() < c.maxENI {
+			c.tryAllocateENI()
+			c.updateLastNodeIPPoolAction()
+		} else {
+			log.Debugf("Skipping ENI allocation due to max ENI already attached to the instance: %d", c.maxENI)
```
tiffanyfay: Trying to read this and it keeps feeling weird. Thoughts on this?

"Skipping ENI allocation as the max ENI limit (%d) was reached"
or 
"Skipping ENI allocation as the instance's max ENI limit (%d) was reached"

ipamd/ipamd.go
```diff
@@ -149,15 +148,14 @@ var (
...
-	currentMaxAddrsPerENI int
-	maxAddrsPerENI        int
+	awsClient        awsutils.APIs
+	dataStore        *datastore.DataStore
+	k8sClient        k8sapi.K8SAPIs
+	customNetworkCfg bool
+	eniConfig        eniconfig.ENIConfig
+	dockerClient     docker.APIs
+	networkClient    networkutils.NetworkAPIs
+	maxAddrsPerENI   int
```
tiffanyfay: I would rename this to maxIPsPerENI to be consistent with the functions and the previous var name.

ipamd/ipamd.go
```diff
@@ -713,8 +697,8 @@ func (c *IPAMContext) tryAssignIPs() (increasedPool bool, err error) {
 
 // setupENI does following:
 // 1) add ENI to datastore
-// 2) add all ENI's secondary IP addresses to datastore
-// 3) setup linux ENI related networking stack.
+// 2) setup linux ENI related networking stack.
```
tiffanyfay: set up

PR #501, opened 2019-06-05
Add env var to override introspection bind address

jacksontj:
2b08772fff0281f3dca70f5f1c2f0aae7fb87397 (inadvertently?) changed the
bind address from `:61678` to `localhost:61678` which is backwards
incompatible and for those actually scraping the metrics via prometheus
makes the metrics endpoint entirely unusable (since prometheus isn't
local to each node in the cluster).

This path simply adds an env var to override the bind address.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,6 +185,11 @@ Default: Unset
 Valid Values: `stdout` or a file path
 Specifies where to write the logging output. Either to stdout or to override the default file.
 
+`INTROSPECTION_BIND_ADDRESS`
+Type: String
+Default: `127.0.0.1:61679`
+Specifies the bind address for the introspection endpoint.
+
 `DISABLE_INTROSPECTION`
 Type: Boolean
 Default: `false`
--- a/ipamd/introspect.go
+++ b/ipamd/introspect.go
@@ -27,8 +27,11 @@ import (
 )
 
 const (
-	// introspectionAddress is listening on localhost 61679 for ipamd introspection
-	introspectionAddress = "127.0.0.1:61679"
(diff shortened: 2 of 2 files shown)


jacksontj:
I have added another commit which changes the default back because the introspection endpoint is largely useless when bound to localhost.

mogren:
@jacksontj Hi, the prometheus metrics port has not changed, it was just moved to another file: [ipamd/metrics.go](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/ipamd/metrics.go#L30). You can still query `:61678/metrics` on the aws-node to get the metrics.

PR #433 moved the introspection endpoint that returns data in json format and that is used for debugging the CNI's internal state to another port and bound it to localhost: [ipamd/introspect.go](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/ipamd/introspect.go#L31). The [debug script](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/aws-cni-support.sh#L25) was updated to handle this. 

Are there some additional metrics that you are missing that can only be found through the introspection endpoints?

jacksontj:
@mogren OIC, it changed the `/` response code for the metrics endpoint as well (used to be 200, now 404 -- this is what we were using for our liveliness check). The bind interface for this introspection endpoint was still changed without mention in the release notes. I'll remove the patch to change the default back, but I still think it would be good to add the env variable override -- but it isn't a huge blocker for my specific use-case

jacksontj:
I also just added the new env var to README (apparently I missed that in the previous commit).

PR #502, opened 2019-06-05
Error when newly required env var is missing

jacksontj:
https://github.com/aws/amazon-vpc-cni-k8s/commit/463e880d06c850246374838a384355d87506de8c swapped discovery of local running pods from the docker daemon to the k8s API. In doing so it now requires the `MY_NODE_NAME` environment variable -- which was both (1) not documented and (2) no error if it is not set. In the case that the variable wasn't set the local daemon would then attempt to add *all* pods that aren't assigned to any node (as the node name would be "").

This patch (1) adds an entry in readme about this option and (2) returns
an error if it is not set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,11 @@ Type: Boolean
 Default: `false`
 Specifies whether prometeus metrics endpoints are enabled on a worker node.
 
+`MY_NODE_NAME`
+Type: String
+Default: Unset
+Specifies the local node name to use when filtering pods from the kubernetes API.
+
 ### Notes
 
 `L-IPAMD`(aws-node daemonSet) running on every worker node requires access to kubernetes API server. If it can **not** reach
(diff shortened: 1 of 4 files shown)


mogren:
Hey @jacksontj, thanks for the PR, but there are multiple things to consider here. First of all, this environment variable should already have be [set in the configuration](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/v1.5/aws-k8s-cni.yaml#L93-L96) and that should probably be documented.

Also, it is currently fetched from the environment in [two](https://github.com/aws/amazon-vpc-cni-k8s/blob/ccf1c29e451cf2bebc3e490b83db05245a14cc27/pkg/k8sapi/discovery.go#L77) [places](https://github.com/aws/amazon-vpc-cni-k8s/blob/ef7e0785172950bb4a38b33cebb1a9d18348702b/pkg/eniconfig/eniconfig.go#L81) which is weird, and there is no error handling if it's not set. I think this might need some more refactoring to resolve these issues, I'll try to get some time to go through it some more.

PR #507, opened 2019-06-12
Add 1.10 compatible config files

mogren:
*Description of changes:*
- Add milliseconds to log format
- Add 1.10 compatible config files, using `version: v1` for CRDs instead of `versions:`

### DIFF
--- a/config/v1.5/aws-k8s-cni-1.10.yaml
+++ b/config/v1.5/aws-k8s-cni-1.10.yaml
@@ -1,0 +1,130 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: aws-node
+rules:
+  - apiGroups:
+      - crd.k8s.amazonaws.com
+    resources:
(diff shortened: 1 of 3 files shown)


tiffanyfay:
It looks ok, but we might want to make sure it works.

config/v1.5/aws-k8s-cni-1.10.yaml
```diff
@@ -0,0 +1,130 @@
...
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: eniconfigs.crd.k8s.amazonaws.com
+spec:
+  scope: Cluster
+  group: crd.k8s.amazonaws.com
+  # 1.10 uses "version", all newer use "versions"
+  version: v1
```
nithu0115: spec.version field is immutable. Probably, we need to change it back to v1alpha1 ?

nithu0115:
I tested it on my environment. LGTM :)

Issue #469, opened 2019-05-14
Feature request: SNAT exclusions for vpc peers


casret:
Would it be possible to write the IPtables rules so that it would not SNAT for other specified CIDR blocks (along with the ones it grabbed from the metadata servers)?  That would allow pods to talk to other VPC through peering right?

mogren:
Change by @totahuanocotl, @rewiko and @yorg1st to add this feature merged in #520.

PR #520, opened 2019-07-03
Allow additional CIDRs to be excluded from SNAT

totahuanocotl:
Connects to https://github.com/aws/amazon-vpc-cni-k8s/issues/469

Introduces a mechanism to provide an exclusion list of CIDRs that should
not be SNATed.

Adds an environment variable `AWS_VPC_K8S_CNI_EXCLUDE_SNAT_CIDRS` to
provide a comma separated list of ipv4 CIDRs to exclude from SNAT.
The value of this environment variable will only be use when
`AWS_VPC_K8S_CNI_EXTERNALSNAT` is false.

The SNAT exclusion CIDRs will be used to:
- Generate `iptable` rules to prevent SNATing for packets directed to
the excluded CIDRs
- Generate `ip rule` entries to route packets directed to the excluded
CIDRs through the pod's `eni`. This configuration is done both at the
`cni` plugin level via the `rpc_handler`, and the runtime `network` api.

Given that the list of excluded CIDRs may vary by configuration it was
necessary to include a mechanism to clean up stale SNAT rules. If a
CIDR is removed from the exclusion list, the corresponding `iptable`
rule will be removed as well, and the chain will be adjusted.

Co-authored-by: @rewiko
Co-authored-by: @yorg1st

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ Disable (`none`) this functionality if you rely on sequential port allocation fo
 `AWS_VPC_K8S_CNI_EXCLUDE_SNAT_CIDRS`
 Type: String
 Default: empty
-Specify a comma separated list of ipv4 CIDRs to exclude from SNAT. For every item in the list an `iptables` rule and off\-VPC
+Specify a comma separated list of IPv4 CIDRs to exclude from SNAT. For every item in the list an `iptables` rule and off\-VPC
 IP rule will be applied. If an item is not a valid ipv4 range it will be skipped. This should be used when `AWS_VPC_K8S_CNI_EXTERNALSNAT=false`.
 
 `WARM_ENI_TARGET`
--- a/README.md
+++ b/README.md
@@ -151,6 +151,12 @@ Disable (`none`) this functionality if you rely on sequential port allocation fo
 
 *Note*: Any options other than `none` will cause outbound connections to be assigned a source port that's not necessarily part of the ephemeral port range set at the OS level (/proc/sys/net/ipv4/ip_local_port_range). This is relevant for any customers that might have NACLs restricting traffic based on the port range found in ip_local_port_range
 
+`AWS_VPC_K8S_CNI_EXCLUDE_SNAT_CIDRS`
+Type: String
+Default: empty
+Specify a comma separated list of ipv4 CIDRs to exclude from SNAT. For every item in the list an `iptables` rule and off\-VPC
+IP rule will be applied. If an item is not a valid ipv4 range it will be skipped. This should be used when `AWS_VPC_K8S_CNI_EXTERNALSNAT=false`.
+
 `WARM_ENI_TARGET`
 Type: Integer
 Default: `1`
--- a/pkg/networkutils/network.go
+++ b/pkg/networkutils/network.go
@@ -15,11 +15,13 @@ package networkutils
 
 import (
 	"encoding/binary"
+	"encoding/csv"
 	"fmt"
 	"io"
 	"math"
 	"net"
 	"os"
+	"reflect"
 	"strconv"
 	"strings"
 	"syscall"
@@ -55,6 +57,11 @@ const (
 	// secondary ENI IP addresses.  If set to "true", the SNAT iptables rule and off-VPC ip rule will not
 	// be installed and will be removed if they are already installed.  Defaults to false.
 	envExternalSNAT = "AWS_VPC_K8S_CNI_EXTERNALSNAT"
+
+	// This environment is used to specify a comma separated list of ipv4 CIDRs to exclude from SNAT. An additional rule 
+	// will be written to the iptables for each item. If an item is not an ipv4 range it will be skipped.
+	// Defaults to empty.
+	envExcludeSNATCIDRs = "AWS_VPC_K8S_CNI_EXCLUDE_SNAT_CIDRS"
 
 	// This environment is used to specify weather the SNAT rule added to iptables should randomize port
 	// allocation for outgoing connections. If set to "hashrandom" the SNAT iptables rule will have the "--random" flag
@@ -101,6 +108,7 @@ type NetworkAPIs interface {
 	// SetupENINetwork performs eni level network configuration
 	SetupENINetwork(eniIP string, mac string, table int, subnetCIDR string) error
 	UseExternalSNAT() bool
+	GetExcludeSNATCIDRs() []string
 	GetRuleList() ([]netlink.Rule, error)
 	GetRuleListBySrc(ruleList []netlink.Rule, src net.IPNet) ([]netlink.Rule, error)
 	UpdateRuleListBySrc(ruleList []netlink.Rule, src net.IPNet, toCIDRs []string, toFlag bool) error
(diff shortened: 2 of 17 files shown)


totahuanocotl:
Hi @mogren , do you think this PR could be looked at?
We are quite interested in this festure.

pkg/networkutils/network_test.go
```diff
@@ -188,20 +189,22 @@ func TestUpdateRuleListBySrc(t *testing.T) {
...
-		oldRule  netlink.Rule
-		toFlag   bool
-		toCIDRs  []string
-		ruleList []netlink.Rule
-		newRules []netlink.Rule
-		expDst   []*net.IPNet
-		expTable []int
+		name               string
+		oldRule            netlink.Rule
+		requiresSNAT       bool
```
mogren: Thanks for fixing the name.

pkg/networkutils/network_test.go
```diff
@@ -305,6 +322,183 @@ func TestSetupHostNetworkNodePortEnabled(t *testing.T) {
...
+			"mangle": {
+				"PREROUTING": [][]string{
+					{"-m", "comment", "--comment", "AWS, primary ENI", "-i", "lo", "-m", "addrtype", "--dst-type", "LOCAL", "--limit-iface-in", "-j", "CONNMARK", "--set-mark", "0x80/0x80"},
+					{"-m", "comment", "--comment", "AWS, primary ENI", "-i", "eni+", "-j", "CONNMARK", "--restore-mark", "--mask", "0x80"},
+				},
+			},
+		}, mockIptables.dataplaneState)
+}
+
+func TestSetupHostNetworkCleansUpStaleSNATRules(t *testing.T) {
```
mogren: Nice tests!

README.md
```diff
@@ -151,6 +151,12 @@ Disable (`none`) this functionality if you rely on sequential port allocation fo
 
 *Note*: Any options other than `none` will cause outbound connections to be assigned a source port that's not necessarily part of the ephemeral port range set at the OS level (/proc/sys/net/ipv4/ip_local_port_range). This is relevant for any customers that might have NACLs restricting traffic based on the port range found in ip_local_port_range
 
+`AWS_VPC_K8S_CNI_EXCLUDE_SNAT_CIDRS`
+Type: String
+Default: empty
+Specify a comma separated list of ipv4 CIDRs to exclude from SNAT. For every item in the list an `iptables` rule and off\-VPC
+IP rule will be applied. If an item is not a valid ipv4 range it will be skipped. This should be used when `AWS_VPC_K8S_CNI_EXTERNALSNAT=false`.
```
mogren: Nit: `IPv4` instead of `ipv4`, just to be consistent.

pkg/networkutils/network.go
```diff
@@ -329,6 +359,24 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDR *net.IPNet, vpcCIDRs []*string,
 		rule:        snatRule,
 	})
 
+	var snatStaleRulesToClear []iptablesRule
+	log.Debugf("Setup Host Network: synchronising SNAT stale rules")
+	for _, staleRule := range snatStaleRulesToCheck {
+		keepRule := false
+		for _, newRule := range iptableRules {
+			if staleRule.chain == newRule.chain && reflect.DeepEqual(newRule.rule, staleRule.rule) {
```
mogren: `reflect.DeepEqual()` is a very slow way to compare two string slices, but I agree that it's fine to use here since we only set up the host networking once.

pkg/networkutils/network.go
```diff
@@ -393,6 +441,42 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDR *net.IPNet, vpcCIDRs []*string,
...
+			r := csv.NewReader(strings.NewReader(rule))
+			r.Comma = ' '
+			ruleSpec, err := r.Read()
+			if err != nil {
+				return nil, errors.Wrap(err, fmt.Sprintf("host network setup: failed to parse iptables nat chain %s rule %s", chain, rule))
+			}
+			log.Debugf("host network setup: found potentially stale SNAT rule for chain %s: %v", chain, ruleSpec)
+			toClear = append(toClear, iptablesRule{
+				name:        fmt.Sprintf("[%d] %s", i, chain),
+				shouldExist: false,
```
mogren: It's not completely obvious that the `false` here will trigger the `ipt.Delete` on line 434 unless the newly created rules include it (`iptableRules` on line 366). Maybe just adding a comment would help, something like
```
				shouldExist: false, // To trigger ipt.Delete for stale rules
```

pkg/networkutils/network.go
```diff
@@ -286,24 +313,26 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDR *net.IPNet, vpcCIDRs []*string,
 			"-m", "comment", "--comment", "AWS SNAT CHAN", "-j", "AWS-SNAT-CHAIN-0",
 		}})
 
-	for i, cidr := range vpcCIDRs {
+	for i, cidr:= range allCIDRs {
```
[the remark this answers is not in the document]
totahuanocotl: @mogren I could add two targets to the Makefile: 
- `format` to apply formatting to all go files
- `check-format` to check all go files and list those which require formatting applied.

I could then add the `check-format` target to the travis build, to guarantee that any given commit only goes green if it is properly formatted, as it would fail the build if formatting changes are required. 

If you like this idea, let me know if I should add it as part of this PR, or maybe a different one.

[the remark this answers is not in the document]
totahuanocotl: (the above would not be applied to the vendor package)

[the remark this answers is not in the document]
totahuanocotl: If applied now, it would format ~ 11 files, most of them tests, and most of the changes are aliasing `mock` imports to the actual package name. e.g
 `mockec2metadatawrapper "github.com/aws/amazon-vpc-cni-k8s/pkg/ec2metadatawrapper/mocks"`

pkg/networkutils/network.go
totahuanocotl: I am glad you agree. I wasn't too sure about keeping it like this, but writing a custom comparison just seem a bit over the top for this use case.

pkg/networkutils/network.go
```diff
@@ -286,24 +313,26 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDR *net.IPNet, vpcCIDRs []*string,
...
 		curChain := chains[i]
-		nextChain := chains[i+1]
 		curName := fmt.Sprintf("[%d] AWS-SNAT-CHAIN", i)
-
-		log.Debugf("Setup Host Network: iptables -A %s ! -d %s -t nat -j %s", curChain, *cidr, nextChain)
+		nextChain := chains[i+1]
+		comment := "AWS SNAT CHAN"
```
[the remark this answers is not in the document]
totahuanocotl: I renamed all the instances, except for one in one test just to check that it would not cause problems for current rules with the typo.

totahuanocotl:
Thanks @mogren !

Issue #330, opened 2019-02-22
Kubelet will start scheduling pods before amazon-vpc-cni-k8s daemon set is fully functional causing workloads to error with "failed to assign an IP address to container"


nithu0115:
# Issue description
Pods are failing to create with sanbox error. We are using latest CNI (1.3.0) and there sufficient IPs available for pods

# Analysis
I went through the Kubelet and aws-node logs. This happens as soon as the Worker Nodes joins the cluster and if there are any pending work loads which needs to be scheduled as soon as the Worker Nodes goes to Ready state. I see Kubelet failed to start pod because of no available IP address which tally with the ipamd daemon not having IP address in its warm pool. 
This is what I found so far about issue: As soon as the aws-node CNI was initialized at 2019-01-29T01:57:01Z, Kubelet gave the aws-node cni todo pods networking to assign a private ip addresses. However, aws-node CNI took over 7seconds to add ip addresses to warm pool which happened around  2019-01-29T01:57:08Z. This caused the pods to fail in those 7 seconds as they could not get a IP address with error - "failed to assign an IP address to container". 

Below are the last occurrences of the error where it is complaining about "failed to assign an IP address to container". 

## Kubelet logs
```
cmd: failed to assign an IP address to container
Jan 29 01:57:08 ip-10-77-10-10.ec2.internal kubelet[5122]: E0129 01:57:08.022505    5122 cni.go:260] Error adding network: add cmd: failed to assign an IP address to container
Jan 29 01:57:08 ip-10-77-10-10.ec2.internal kubelet[5122]: E0129 01:57:08.022519    5122 cni.go:228] Error while adding to cni network: add cmd: failed to assign an IP address to container
Jan 29 01:57:08 ip-10-77-10-10.ec2.internal kubelet[5122]: E0129 01:57:08.377159    5122 remote_runtime.go:92] RunPodSandbox from runtime service failed: rpc error: code = Unknown desc = failed to set up sandbox container "b849b621ba3c224c1b42912252ded857d2a798d031d8f81065d2b1736d07771a" network for pod "workbench-rules-cc7764ccb-stjc8": NetworkPlugin cni failed to set up pod "workbench-rules-cc7764ccb-stjc8_placeable-regression" network: add cmd: failed to assign an IP address to container
[...]

PR #523, opened 2019-07-09
Added liveness and readiness probes

stijndehaes:
*Issue #, if available:*

https://github.com/aws/amazon-vpc-cni-k8s/issues/330

*Description of changes:*

Added a liveness and readiness probe to see if the application is up and running. This can help with the above issue. However this is still not a full solution, since the application can be up but it could be it is still not ready to distribute ip adresses.


stijndehaes:
I also only added this to version 1.4 and 1.5 yaml. As I was not sure if before those versions this would work.

PR #535, opened 2019-07-21
set iptables FORWARD rules  ACCEPT for inside VPC traffic

Komey:
*Description of changes:*
For some AMI doesn't ACCEPT FORWARD traffic by default, this will enable those traffic FORWARD in VPC CIDR

### DIFF
--- a/pkg/networkutils/network.go
+++ b/pkg/networkutils/network.go
@@ -331,6 +331,20 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDR *net.IPNet, vpcCIDRs []*string, 
 			rule: []string{
 				"!", "-d", cidr.cidr, "-m", "comment", "--comment", comment, "-j", nextChain,
 			}})
+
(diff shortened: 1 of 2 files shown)


mogren:
Hi @Komey,

Could you explain the use case for this? It seems like a configuration that should belong in the AMI and not the CNI?

Komey:
> Hi @Komey,
> 
> Could you explain the use case for this? It seems like a configuration that should belong in the AMI and not the CNI?

Hi @mogren

I was testing a self-hosted k8s cluster with AMI --> ubuntu/images/hvm-ssd/ubuntu-bionic-18.04-amd64-server-20190212.1 (ami-0c55b159cbfafe1f0) , which is not accept forward traffic in any cidr.
In my opinion, CNI should take care of the traffic which has been configured within the VPC cird

mogren:
The reason we do it outside the CNI is that this applies outside of Kubernetes as well, for example if you have Docker containers running on an EC2 instance.

Issue #282, opened 2019-01-08
Race condition between CNI plugin install and aws-k8s-agent startup


kc8apf:
My employer runs GitLab CI using the Kubernetes executor which runs each CI job as a separate pod.  Since the workload is bursty, cluster-autoscaler is also enabled.  When certain CI stages begin, 100+ pods are created rapidly and a large number of them are unscheduleable until the autoscaler has spun up enough new nodes.  GitLab CI will wait up to 10m for the pod to start before assuming the job failed.

Occasionally, one of these pods will hit the 10m timeout with the following events according to `kubectl describe po`:
```
Events:
  Type     Reason                  Age                From                                                 Message
  ----     ------                  ----               ----                                                 -------
  Normal   Scheduled               14m                default-scheduler                                    Successfully assigned gitlab-ci-runner/runner-cad974a7-project-1-concurrent-0cvwc7 to ip-172-30-13-60.us-west-2.compute.internal

  Warning  FailedCreatePodSandBox  14m                kubelet, ip-172-30-13-60.us-west-2.compute.internal  Failed create pod sandbox: rpc error: code = Unknown desc = [failed to set up sandbox container "c96fe97cd035ce410f537254a20102485c83eb275f3fc317f0e2b5ea5c4b7720" network for pod "runner-cad974a7-project-1-concurrent-0cvwc7": NetworkPlugin cni failed to set up pod "runner-cad974a7-project-1-concurrent-0cvwc7_gitlab-ci-runner" network: rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = "transport: Error while dialing dial tcp 127.0.0.1:50051: connect: connection refused", failed to clean up sandbox container "c96fe97cd035ce410f537254a20102485c83eb275f3fc317f0e2b5ea5c4b7720" network for pod "runner-cad974a7-project-1-concurrent-0cvwc7": NetworkPlugin cni failed to teardown pod "runner-cad974a7-project-1-concurrent-0cvwc7_gitlab-ci-runner" network: rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = "transport: Error while dialing dial tcp 127.0.0.1:50051: connect: connection refused"]
  Normal   SandboxChanged          4m (x46 over 14m)  kubelet, ip-172-30-13-60.us-west-2.compute.internal  Pod sandbox changed, it will be killed and re-created.
```

[...]

Issue #537, opened 2019-07-22
Unhandled rate limit which causes other errors


uthark:
We found the following behavior:

1. Because of #431, we decided to warming maximum number of ENIs that can be attached to an instance.
2. We have medium size clusters (hundreds of nodes)
3. During rolling out, we observed spike of AWS API Errors in the plugin.
4. During investigation, we found the error in the plugin.

Steps to reproduce
1. Setup `WARM_ENI_TARGET` to max available to your instance type.
2. Rollout change in a big cluster.
3. CNI Plugins on each nodes will start to warming ENIs.

ipamd logs: 
```
2019-07-19T17:40:34Z [INFO]     Exceeded instance ENI attachment limit: 7
2019-07-19T17:40:34Z [ERROR]    Failed to attach ENI eni-029644ab1bc40fc71: AttachmentLimitExceeded: Interface count 9 exceeds the limit for m4.4xlarge
        status code: 400, request id: b14ba702-ca17-4772-a0a8-7c577257d2ab
2019-07-19T17:40:34Z [DEBUG]    Trying to delete ENI: eni-029644ab1bc40fc71
```

Upon further investigation, we found the issue in `AllocENI` function:
https://github.com/aws/amazon-vpc-cni-k8s/blob/73ab7f5807137705f6122d0b056dd63abd3fa462/pkg/awsutils/awsutils.go#L547-L584
1. If `ModifyNetworkInterfaceAttribute` call was unsuccessful (it was in our case - cloudtrail logs indicates that this call was rate-limited), then plugin will try to delete ENI. 
2. ENI can't be deleted because it is attached to an instance.
```
62019-07-19T17:25:55Z [DEBUG]	Trying to delete ENI: eni-0a0b3f58e0fe6d7d0
2019-07-19T17:29:30Z [DEBUG]	Not able to delete ENI yet (attempt 1/20): InvalidParameterValue: Network interface 'eni-0a0b3f58e0fe6d7d0' is currently in use.
	status code: 400, request id: 9c360487-187e-4573-a072-4a234b8d1946 
```

This error causes another issue: we will have an ENI instance that is attached to an EC2 instance, but CNI Plugin doesn't know about it, and it will try to warm another CNI plugin and it will fail wit the error message `AttachmentLimitExceeded`.

[...]

Issue #525, opened 2019-07-10
ENI warming is delayed for at least for 1 minute, probably caused by #480


uthark:
We also found that #480 changes causes the following interesting behavior:

1. New node joins the cluster.
2. API Server assigns new pods to it (particularly from system daemonsets, like aws-cni itself, fluentd, kube-node-monitor and some other that have tolerations)
3. aws-cni plugins starts and immediately discovers that there are pods without IP address. It then wait for 12*5=1 minute before trying to assign IP addresses to these pods.
4. And it looks like warming ENIs starts *only* after this timeout. At least, according to the logs we got:

Relevant log entries from IPAMD:
```
2019-07-10T20:30:30Z [INFO]     Starting L-IPAMD v1.5.0  ...
...
2019-07-10T20:30:30Z [DEBUG]    Using WARM_ENI_TARGET 4
...
2019-07-10T20:31:32Z [INFO]     Serving metrics on port 61678
2019-07-10T20:31:32Z [INFO]     Serving introspection endpoints on 127.0.0.1:61679
...
2019-07-10T20:31:34Z [INFO]     Received DelNetwork for IP <nil>, Pod kube-node-monitor-ckbsh, Namespace default, Container 9a8410ed9f541e1961f47b5c797fe10cd3d006c09301ba88ccb3c389c51a6017
2019-07-10T20:31:34Z [INFO]     Received AddNetwork for NS /proc/38956/ns/net, Pod kube-node-monitor-ckbsh, NameSpace default, Container 35265389458c297371c825f2ae3f5277583c9c5ab2acebc599ba02a3bb3b6920, ifname eth0
2019-07-10T20:31:34Z [DEBUG]    AssignIPv4Address: IP address pool stats: total: 0, assigned 0
2019-07-10T20:31:34Z [DEBUG]    AssignPodIPv4Address: Skip ENI eni-036df2e82333ebf90 that does not have available addresses
2019-07-10T20:31:34Z [ERROR]    DataStore has no available IP addresses
...
2019-07-10T20:31:35Z [DEBUG]    IP pool stats: total = 0, used = 0, c.maxIPsPerENI = 29
2019-07-10T20:31:35Z [DEBUG]    IP pool is too low: available (0) < ENI target (4) * addrsPerENI (29)
2019-07-10T20:31:35Z [DEBUG]    Starting to increase IP pool size
2019-07-10T20:31:35Z [DEBUG]    Skip the primary ENI for need IP check
...
[...]

PR #538, opened 2019-07-22
Detach ENI before deleting

uthark:
Fix for issue #537.

Try to detach ENI before deleting it.


uthark:
@mogren  Is there any plans to release a minor 1.5.1 just with the bugfixes or do you plan to release 1.6.0 only?

mogren:
@uthark I would really like to get #371 merged first, just need a little more testing. So far everything looks good.

The top priority is to resolve #282, #330 and #525, which are all basically the same issue. So once we get those fixed I'll make a v1.6-rc build and start some more deep testing.

There is some work to get a new version out, but I'd prefer if we could be more frequent with new releases instead of trying to put too many things in before cutting the next one.

PR #539, opened 2019-07-22
Fix message - data is returned from API Server, not from kubelet

uthark:
Fix message - data is returned from API Server, not from kubelet


uthark:
@mogren Please take a look.

PR #545, opened 2019-07-26
Reduce noise in logs

uthark:
Change log level from info to debug for *Handle...* messages. They are useless for production usage and just generate a lot of noise in logs.

### DIFF
--- a/pkg/eniconfig/eniconfig.go
+++ b/pkg/eniconfig/eniconfig.go
@@ -102,7 +102,7 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 	case *v1alpha1.ENIConfig:
 		eniConfigName := o.GetName()
 		if event.Deleted {
-			log.Infof("Deleting ENIConfig: %s", eniConfigName)
+			log.Debugf("Deleting ENIConfig: %s", eniConfigName)
 			h.controller.eniLock.Lock()
 			defer h.controller.eniLock.Unlock()
 			delete(h.controller.eni, eniConfigName)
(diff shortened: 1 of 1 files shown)


uthark:
@mogren Please take a look.

pkg/eniconfig/eniconfig.go
```diff
@@ -131,8 +131,10 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 
 			h.controller.eniLock.Lock()
 			defer h.controller.eniLock.Unlock()
-			h.controller.myENI = val
-			log.Infof("Setting myENI to: %s", val)
+			if h.controller.myENI != val {
```
mogren: If you're checking this variable, you might as well move the lock inside the if-statement as well.

pkg/eniconfig/eniconfig.go
```diff
@@ -111,14 +111,14 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 
 		curENIConfig := o.DeepCopy()
 
-		log.Infof("Handle ENIConfig Add/Update: %s, %v, %s", eniConfigName, curENIConfig.Spec.SecurityGroups, curENIConfig.Spec.Subnet)
+		log.Debugf("Handle ENIConfig Add/Update: %s, %v, %s", eniConfigName, curENIConfig.Spec.SecurityGroups, curENIConfig.Spec.Subnet)
```
mogren: I agree that Info is high for this.

pkg/eniconfig/eniconfig.go
uthark: Ah, nice catch. Fixed.

PR #553, opened 2019-08-01
Update start script to wait for ipamd health check

mogren:
*Issue #, if available:* #282 

*Description of changes:*
* Update start script to wait for ipamd health check to return SERVING before copying in the CNI binary and config file, making the node Ready.
* Tested by scheduling pods with 0 worker nodes in the ASG, then scaling up. The nodes do not become ready until ipamd is running.

This is a quick fix for the release-1.5 branch. It will also add additional log lines to the aws-node stdout logs. The long term goal will probably be to move the set up into go instead and have ipamd copy the files once it's ready.

### DIFF
--- a/scripts/install-aws.sh
+++ b/scripts/install-aws.sh
@@ -1,14 +1,42 @@
 #!/usr/bin/env bash
-echo "===== Starting installing AWS-CNI ========="
-sed -i s/__VETHPREFIX__/${AWS_VPC_K8S_CNI_VETHPREFIX:-"eni"}/g /app/10-aws.conflist
-cp /app/aws-cni /host/opt/cni/bin/
-cp /app/portmap /host/opt/cni/bin/
-cp /app/aws-cni-support.sh /host/opt/cni/bin/
-cp /app/10-aws.conflist /host/etc/cni/net.d/
 
-if [[ -f /host/etc/cni/net.d/aws.conf ]]; then
+grpcHealthCheck () {
+  /app/grpc_health_probe -addr 127.0.0.1:50051
+}
+
+waitIPamDServing () {
+   until grpcHealthCheck; do
+    echo "Waiting for ipamd health check";
+    sleep 1;
+  done
+}
+
+main () {
+  # Remove old config files that might have been baked into older AMIs
+  if [[ -f /host/etc/cni/net.d/aws.conf ]]; then
     rm /host/etc/cni/net.d/aws.conf
-fi
+  fi
 
-echo "===== Starting amazon-k8s-agent ==========="
-/app/aws-k8s-agent
+  # Copy standard files
+  cp /app/aws-cni-support.sh /host/opt/cni/bin/
+  cp /app/portmap /host/opt/cni/bin/
+
+  echo "===== Starting amazon-k8s-agent ==========="
+  /app/aws-k8s-agent &
+
+  # Check ipamd health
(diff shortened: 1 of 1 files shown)


scripts/install-aws.sh
```diff
@@ -1,14 +1,42 @@
...
+
+waitIPamDServing () {
+   until grpcHealthCheck; do
+    echo "Waiting for ipamd health check";
+    sleep 1;
+  done
+}
+
+main () {
+  # Remove old config files that might have been baked in to older AMIs
```
tiffanyfay: *into

scripts/install-aws.sh
```diff
@@ -1,14 +1,42 @@
...
-sed -i s/__VETHPREFIX__/${AWS_VPC_K8S_CNI_VETHPREFIX:-"eni"}/g /app/10-aws.conflist
-cp /app/aws-cni /host/opt/cni/bin/
-cp /app/portmap /host/opt/cni/bin/
-cp /app/aws-cni-support.sh /host/opt/cni/bin/
-cp /app/10-aws.conflist /host/etc/cni/net.d/
-
-if [[ -f /host/etc/cni/net.d/aws.conf ]]; then
+
+grpcHealthCheck () {
+  /app/grpc_health_probe -addr 127.0.0.1:50051
```
tiffanyfay: Where is the `grpc_health_probe` coming from?

Wouldn't Dockerfile.release need something like this?
```
RUN GRPC_HEALTH_PROBE_VERSION=v0.3.0 && \
    wget -qO/app/grpc_health_probe https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/${GRPC_HEALTH_PROBE_VERSION}/grpc_health_probe-linux-amd64 && \
    chmod +x /app/grpc_health_probe
```

scripts/install-aws.sh
```diff
@@ -1,14 +1,42 @@
...
+
+  # Check that ipamd is healthy, exit if the check fails
+  while grpcHealthCheck; do
+      sleep 10;
+  done
+}
 
-echo "===== Starting amazon-k8s-agent ==========="
-/app/aws-k8s-agent
+main
```
nithu0115: Even though there are no positional  parameters, I personally would add `main "$@" `

scripts/install-aws.sh
```diff
@@ -1,14 +1,42 @@
 #!/usr/bin/env bash
-echo "===== Starting installing AWS-CNI ========="
-sed -i s/__VETHPREFIX__/${AWS_VPC_K8S_CNI_VETHPREFIX:-"eni"}/g /app/10-aws.conflist
-cp /app/aws-cni /host/opt/cni/bin/
-cp /app/portmap /host/opt/cni/bin/
-cp /app/aws-cni-support.sh /host/opt/cni/bin/
-cp /app/10-aws.conflist /host/etc/cni/net.d/
-
-if [[ -f /host/etc/cni/net.d/aws.conf ]]; then
+
```
nithu0115: Probably fail fast, by adding `set -eo pipefail`

mogren: Good point, will add.

xrl:
I've been eagerly anticipating something like this. When should we expect it to show up in an official image I can start running?

nithu0115:
Issue #555 is a follow up to this change which has new release candidate image with have these changes.

PR #556, opened 2019-08-01
Correct docs for AWS_VPC_K8S_CNI_RANDOMIZESNAT

### LINKED ISSUE #380: Documentation correction for AWS_VPC_K8S_CNI_RANDOMIZESNAT
Documentation seems to be wrong for AWS_VPC_K8S_CNI_RANDOMIZESNAT which states use it when AWS_VPC_K8S_CNI_EXTERNALSNAT=true is set. But setting the value to true means `The SNAT iptables rule and off-VPC IP rule are not applied, and these rules are removed if they have already been applied.` AWS_VPC_K8S_CNI_RANDOMIZESNAT should be used only when the AWS_VPC_K8S_CNI_EXTERNALSNAT=false. Here is the code link

https://github.com/taylorb-syd/amazon-vpc-cni-k8s/blob/e59edf200179e82d6cd141d8e394f4d4b7978bb4/pkg/networkutils/network.go#L338

orirawlings:
*Issue #, if available:*
Fixes #380

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,10 +143,10 @@ private subnet and connected to the internet through an AWS NAT Gateway or anoth
 Type: String
 Default: `hashrandom`
 Valid Values: `hashrandom`, `prng`, `none`
-Specifies weather the SNAT `iptables` rule should randomize the outgoing ports for connections\. When enabled (`hashrandom`)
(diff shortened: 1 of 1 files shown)


orirawlings:
Looks like the travis build failed because of a tcp timeout:

PR #528, opened 2019-07-12
Adding metal instance support

pauljrob:
*Description of changes:*
Added Metal support

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #572, opened 2019-08-06
update list of instances to be more metal \m/

pdbogen:
*Issue #, if available:* n/a

*Description of changes:* add missing `metal` instances, data from https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html#AvailableIpPerENI


pdbogen:
n.b., the CI failures look spurious / network-related, don't seem to be a result of this change.

mogren:
@pdbogen Yes, TravisCI seem to have timeout issues the last few days. 

Also, #528 is basically exactly the same PR as this one, except that you fixed the `i3.metal` IP count as well.

pdbogen:
ah! I didn't think to check for an existing PR. In response to the query on _that_ one, we haven't tested on a `.metal` instance yet, but I think we plan to in the next few days.

pdbogen:
Ya. And to close the loop, we've tested on some `metal` instances and it's working well.

PR #574, opened 2019-08-06
Release v1.5.2

mogren:
*Description of changes:*

* Cherry pick of v1.5.2 release commit back to master.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Changelog
 
+## v1.5.2
+* Bug fix - [Fix formatting flag](https://github.com/aws/amazon-vpc-cni-k8s/pull/521) (#521, @uthark)
+* Bug fix - [Fix formatting issue](https://github.com/aws/amazon-vpc-cni-k8s/pull/524) (#524, @uthark)
+* Bug fix - [Detach ENI before deleting](https://github.com/aws/amazon-vpc-cni-k8s/pull/538) (#538, @uthark)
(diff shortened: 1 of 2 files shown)


edmorley:
Hi! There appears to other differences between the example configs on master and the 1.5.2 tag?
Eg: The `WATCH_NAMESPACE` env var was removed on master, but not on:
https://github.com/aws/amazon-vpc-cni-k8s/blob/e5f2b85c5a0ab7301e095dd8a25ddae9efaf1af9/config/v1.5/aws-k8s-cni.yaml#L105-L108

Also, are the commented out readinessProbe/livenessProbe a leftover that can be removed, or something we should be uncommenting ourselves if experiencing #282?
https://github.com/aws/amazon-vpc-cni-k8s/blob/ffaf737145ab3262b7afd0ddbdf613a2174f30dd/config/v1.5/aws-k8s-cni.yaml#L90-L97

Many thanks :-)

mogren:
@edmorley Hey, thanks for checking this.

The `WATCH_NAMESPACE` was only ever used by the CNI when having a custom network config. Since the v1.5.1 release change, it isn't needed any more, thanks for catching that.

The readinessProbe/livenessProbe is there and commented out to show that the option is available. If some user points to an image tag of v1.5.1 or earlier, with the liveness probe enabled, the aws-nodes will never go into running state. That said, something like #575 in v1.5.2 could avoided by enabling the probes. (#576 should fix that issue though)

Issue #575, opened 2019-08-07
Waiting for ipamd health check deadlocks node bootstrapping outside EKS


drakedevel:
Tested 1.5.2 on Kubernetes 1.15.2

The change in #553 introduced a node bootstrapping problem on our `kubeadm` test cluster. With this change, nodes get tainted with `node.kubernetes.io/not-ready` until the ipamD is healthy. ipamD can't become healthy until it's able to reach the API server. On at least `kubeadm` and `kops` clusters, the API server is a ClusterIP, which requires `kube-proxy` to be up and running. This circular dependency means that nodes booting up simply sit there forever.

drakedevel:
Sorry, this explanation is bogus -- I misread my `kubectl` output and `kube-proxy` is running fine. I am still seeing this in `ipamd.log` (with the result that the node is NotReady):
```
2019-08-07T02:24:04.835Z [INFO] Failed to communicate with K8S Server. Please check instance security groups or http proxy setting
2019-08-07T02:24:04.835Z [ERROR]        Failed to create client: error communicating with apiserver: Get https://100.127.0.1:443/version?timeout=32s: dial tcp 100.127.0.1:443: i/o timeout
```

The cluster worked fine with 1.5.0 w/ Kubernetes 1.15.0, but I need to investigate more tomorrow. Will update here when I figure out what happened -- sorry for the noise!

mogren:
@drakedevel Ok, thanks for the follow up. 

We did test v1.5.2 quite a lot, both on new clusters and upgrading from older versions. I'll close this issue since since kube-proxy does start, but feel free to open another issue if you can't figure out why ipamd can't talk to the api-server.

drakedevel:
@mogren It looks like the actual issue is that the `aws-node` pods started up faster than the `kube-proxy` pods, so the API server was unreachable initially. The `aws-k8s-agent` process apparently crashes when this happens (the process was gone when I looked), but `install-aws.sh` doesn't notice and continues to poll forever.

Everything works fine if the pod is manually deleted, but the node is broken until then as nothing will automatically get the pod out of this state. In 1.5.1, a `aws-k8s-agent` crash always results in the pod exiting and getting restarted, so if it "wins" the race against `kube-proxy` it's at least self-healing that way.

Ideas:
- `install-aws.sh` could detect the child process crashing before becoming healthy and bail
- The readiness probe (commented out in the example config) would work as a fallback to prevent a pod from getting stuck forever in this state

[...]

PR #576, opened 2019-08-07
Copy the binary and config after ipamd is ready

mogren:
*Issue #, if available:* #282, #575

*Description of changes:*
* Copy the plugin and config from within ipamd instead to avoid having to poll in bash.

### DIFF
--- a/main.go
+++ b/main.go
@@ -14,6 +14,7 @@
 package main
 
 import (
+	"io"
 	"os"
 
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -57,27 +58,78 @@ func _main() int {
...
+	// Copy the CNI plugin and config. This will mark the node as Ready.
+	log.Info("Copying /app/aws-cni to /host/opt/cni/bin/aws-cni")
+	err = copyFileContents("/app/aws-cni", "/host/opt/cni/bin/aws-cni")
 	if err != nil {
-		log.Error("Failed to set up gRPC handler ", err)
+		log.Errorf("Failed to copy aws-cni: %v", err)
 		return 1
 	}
 
+	log.Info("Copying /app/10-aws.conflist to /host/etc/cni/net.d/10-aws.conflist")
```
micahhausler: I think this is fine for now, but do we want to parameterize this filename? I'm take it or leave it

main.go
```diff
@@ -57,27 +58,78 @@ func _main() int {
...
+	err = ipamContext.RunRPCHandler()
+	if err != nil {
+		log.Errorf("Failed to set up gRPC handler: %v", err)
+		return 1
+	}
 	return 0
 }
+
+// copyFileContents copies a file
+func copyFileContents(src, dst string) error {
```
micahhausler: Should we have a test for this?

mogren: Adding in a separate PR.

PR #582, opened 2019-08-09
Exit early if MAC address doesn't match

mogren:
*Description of changes:*
* Change `retryLinkByMacInterval` from `5` to `3` seconds.
* Don't add an ENI to the datastore if we fail to describe it.
* If we fail on `find the link which uses MAC address`, don't retry for that ENI.
* Removed some unneeded casts.

### DIFF
--- a/ipamd/ipamd.go
+++ b/ipamd/ipamd.go
@@ -18,6 +18,7 @@ import (
 	"net"
 	"os"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -305,6 +306,12 @@ func (c *IPAMContext) nodeInit() error {
 			}
 
 			if err != nil {
+				log.Warnf("Error trying to set up ENI %s: %v", eni.ENIID, err)
+				if strings.Contains(err.Error(), "setupENINetwork: failed to find the link which uses MAC address") {
+					// If we can't find the matching link for this MAC address, there is no point in retrying for this ENI.
+					log.Errorf("Unable to match link for this ENI, going to the next one.")
+					break
+				}
 				log.Debugf("Unable to discover IPs for this ENI yet (attempt %d/%d)", retry, maxRetryCheckENI)
 				time.Sleep(eniAttachTime)
 				continue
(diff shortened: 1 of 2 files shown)


ipamd/ipamd.go
```diff
@@ -700,21 +706,22 @@ func (c *IPAMContext) tryAssignIPs() (increasedPool bool, err error) {
...
-	if err != nil && err.Error() != datastore.DuplicatedENIError {
-		return errors.Wrapf(err, "failed to add ENI %s to data store", eni)
-	}
-
 	ec2Addrs, eniPrimaryIP, err := c.getENIaddresses(eni)
 	if err != nil {
 		return errors.Wrapf(err, "failed to retrieve ENI %s IP addresses", eni)
 	}
 
+	// Add the ENI to the datastore
```
tiffanyfay: @mogren I think I am missing something. Why did you switch the order?

mogren: @tiffanyfay Thanks, good question! I tested this a few times, and if `c.getENIaddresses()` fails, it means that `c.awsClient.DescribeENI()` that does an actual call to EC2 failed. That in turn means we don't get the list of IP for that ENI, so there is no point in adding the ENI to the datastore just to have it be [removed in the next reconcile loop](https://github.com/aws/amazon-vpc-cni-k8s/blob/c6f082ed6fed7cc9b59e5365d97dcc36b9f0716b/ipamd/ipamd.go#L947).

mogren:
Tested in a cluster with a "stuck" ENI:
```
2019-08-09T16:55:05.894Z [DEBUG]	no interface found which uses mac address 11:7b:d7:5b:b1:64 (attempt 5/5)
2019-08-09T16:55:05.894Z [ERROR]	Failed to set up networking for ENI eni-0ee111c4949700f74
2019-08-09T16:55:05.894Z [WARN]	Error trying to set up ENI eni-0ee111c4949700f74: failed to set up ENI eni-0ee111c4949700f74 network: setupENINetwork: failed to find the link which uses MAC address 11:7b:d7:5b:b1:64: no interface found which uses mac address 11:7b:d7:5b:b1:64 (attempt 5/5)
2019-08-09T16:55:05.894Z [ERROR]	Unable to match link for this ENI, going to the next one.
```

This change will make the aws-node start up and not get stuck because of 1 bad ENI. If the ENI gets correctly detached, the reconcile loop will clean it up. (Or attach a new ENI if needed.)

ipamd/ipamd.go
tiffanyfay: Ahh ok, I get it--I was wondering if ec2Addrs would be outdated, but it's just putting the info in the datastore.

PR #612, opened 2019-09-10
Introduce a minimum target for ENI IPs

asheldon:
Adds a MINIMUM_IP_TARGET environment variable to inform the AWS CNI that a particular number of total IPs is anticipated for use with a particular node. This is useful to ensure a sufficient supply of IPs on a node up-front without the 2x IP usage overhead of setting WARM_IP_TARGET to the same value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,7 +234,26 @@ Specifies the number of free IP addresses that the `ipamD` daemon should attempt
 For example, if `WARM_IP_TARGET` is set to 10, then `ipamD` attempts to keep 10 free IP addresses available at all times. If the
 elastic network interfaces on the node are unable to provide these free addresses, `ipamD` attempts to allocate more interfaces
 until `WARM_IP_TARGET` free IP addresses are available.
+If both `WARM_IP_TARGET` and `MINIMUM_WARM_TARGET` are set, `ipamD` will attempt to meet both constraints.
 This environment variable overrides `WARM_ENI_TARGET` behavior.
+
+`MINIMUM_IP_TARGET`
+Type: Integer
+Default: None
+
+Specifies the number of total IP addresses that the `ipamD` daemon should attempt to allocate for pod assignment on the node.
+`MINIMUM_IP_TARGET` behaves identically to `WARM_IP_TARGET` except that instead of setting a target number of free IP
+addresses to keep available at all times, it sets a target number for a floor on how many total IP addresses are allocated.
+
+`MINIMUM_IP_TARGET` is for pre-scaling, `WARM_IP_TARGET` is for dynamic scaling. For example, suppose a cluster has an
+expected pod density of approximately 30 pods per node. If `WARM_IP_TARGET` is set to 30 to ensure there are enough IPs
+allocated up front by the CNI, then 30 pods are deployed to the node, the CNI will allocate an additional 30 IPs, for
+a total of 60, accelerating IP exhaustion in the relevant subnets. If instead `MINIMUM_IP_TARGET` is set to 30 and
+`WARM_IP_TARGET` to 2, after the 30 pods are deployed the CNI would allocate an additional 2 IPs. This still provides
+elasticity, but uses roughly half as many IPs as using WARM_IP_TARGET alone (32 IPs vs 60 IPs).
+
+This also improves reliability of the EKS cluster by reducing the number of calls necessary to allocate or deallocate
+private IPs, which may be throttled, especially at scaling-related times.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -234,7 +234,7 @@ Specifies the number of free IP addresses that the `ipamD` daemon should attempt
 For example, if `WARM_IP_TARGET` is set to 10, then `ipamD` attempts to keep 10 free IP addresses available at all times. If the
 elastic network interfaces on the node are unable to provide these free addresses, `ipamD` attempts to allocate more interfaces
 until `WARM_IP_TARGET` free IP addresses are available.
-If both `WARM_IP_TARGET` and `MINIMUM_WARM_TARGET` are set, `ipamD` will attempt to meet both constraints.
+If both `WARM_IP_TARGET` and `MINIMUM_IP_TARGET` are set, `ipamD` will attempt to meet both constraints.
 This environment variable overrides `WARM_ENI_TARGET` behavior.
 
 `MINIMUM_IP_TARGET`
--- a/ipamd/datastore/data_store.go
+++ b/ipamd/datastore/data_store.go
@@ -324,7 +324,19 @@ func (ds *DataStore) isRequiredForWarmIPTarget(warmIPTarget int, eni *ENIIPPool)
 	return otherWarmIPs < warmIPTarget
 }
 
-func (ds *DataStore) getDeletableENI(warmIPTarget int) *ENIIPPool {
+// IsRequiredForMinimumIPTarget determines if this ENI is necessary to fulfill whatever MINIMUM_IP_TARGET is
+// set to.
+func (ds *DataStore) isRequiredForMinimumIPTarget(minimumIPTarget int, eni *ENIIPPool) bool {
+	otherIPs := 0
+	for _, other := range ds.eniIPPools {
+		if other.ID != eni.ID {
+			otherIPs += len(other.IPv4Addresses)
+		}
+	}
+	return otherIPs < minimumIPTarget
+}
+
+func (ds *DataStore) getDeletableENI(warmIPTarget int, minimumIPTarget int) *ENIIPPool {
 	for _, eni := range ds.eniIPPools {
 		if eni.IsPrimary {
 			log.Debugf("ENI %s cannot be deleted because it is primary", eni.ID)
@@ -348,6 +360,11 @@ func (ds *DataStore) getDeletableENI(warmIPTarget int) *ENIIPPool {
 
 		if warmIPTarget != 0 && ds.isRequiredForWarmIPTarget(warmIPTarget, eni) {
 			log.Debugf("ENI %s cannot be deleted because it is required for WARM_IP_TARGET: %d", eni.ID, warmIPTarget)
+			continue
+		}
+
+		if minimumIPTarget != 0 && ds.isRequiredForMinimumIPTarget(minimumIPTarget, eni) {
+			log.Debugf("ENI %s cannot be deleted because it is required for MINIMUM_IP_TARGET: %d", eni.ID, minimumIPTarget)
 			continue
 		}
 
@@ -391,11 +408,11 @@ func (ds *DataStore) GetENINeedsIP(maxIPperENI int, skipPrimary bool) *ENIIPPool
 // RemoveUnusedENIFromStore removes a deletable ENI from the data store.
 // It returns the name of the ENI which has been removed from the data store and needs to be deleted,
 // or empty string if no ENI could be removed.
-func (ds *DataStore) RemoveUnusedENIFromStore(warmIPTarget int) string {
+func (ds *DataStore) RemoveUnusedENIFromStore(warmIPTarget int, minimumIPTarget int) string {
 	ds.lock.Lock()
 	defer ds.lock.Unlock()
 
-	deletableENI := ds.getDeletableENI(warmIPTarget)
+	deletableENI := ds.getDeletableENI(warmIPTarget, minimumIPTarget)
 	if deletableENI == nil {
 		return ""
 	}
--- a/ipamd/datastore/data_store_test.go
+++ b/ipamd/datastore/data_store_test.go
@@ -288,16 +288,73 @@ func TestPodIPv4Address(t *testing.T) {
 	assert.Equal(t, ds.eniIPPools["eni-2"].AssignedIPv4Addresses, 0)
 
 	noWarmIPTarget := 0
+	noMinimumIPTarget := 0
 
 	// Should not be able to free this ENI
-	eni := ds.RemoveUnusedENIFromStore(noWarmIPTarget)
+	eni := ds.RemoveUnusedENIFromStore(noWarmIPTarget, noMinimumIPTarget)
 	assert.True(t, eni == "")
 
 	ds.eniIPPools["eni-2"].createTime = time.Time{}
 	ds.eniIPPools["eni-2"].lastUnassignedTime = time.Time{}
-	eni = ds.RemoveUnusedENIFromStore(noWarmIPTarget)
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -178,6 +178,17 @@ Specifies the number of free IP addresses that the `ipamD` daemon should attempt
 For example, if `WARM_IP_TARGET` is set to 10, then `ipamD` attempts to keep 10 free IP addresses available at all times. If the
 elastic network interfaces on the node are unable to provide these free addresses, `ipamD` attempts to allocate more interfaces
 until `WARM_IP_TARGET` free IP addresses are available.
+If both `WARM_IP_TARGET` and `MINIMUM_WARM_TARGET` are set, `ipamD` will attempt to meet both constraints.
```
edmorley:
```suggestion
If both `WARM_IP_TARGET` and `MINIMUM_IP_TARGET` are set, `ipamD` will attempt to meet both constraints.
```

ipamd/ipamd.go
```diff
@@ -67,6 +67,17 @@ const (
 	envWarmIPTarget = "WARM_IP_TARGET"
 	noWarmIPTarget  = 0
 
+	// This environment variable is used to specify the desired minimum number of total IPs.
+	// When it is not set, ipamd defaults to 0.
+	// For example, for a m4.4xlarge node,
+	//     If WARM-IP-TARGET is set to 1 and MINIMUM_IP_TARGET is set to 12, and there are 9 pods running on the node,
+	//     ipamd will make the "warm pool" have 12 IP addresses with 9 being assigned to pods and 3 free IPs.
+	//
+	//     If "MINIMUM-WARM-IP-TARGET is not set, it will default to 0, which causes WARM-IP-TARGET settings to be the
```
edmorley:
```suggestion
	//     If "MINIMUM_IP_TARGET is not set, it will default to 0, which causes WARM_IP_TARGET settings to be the
```

ipamd/ipamd.go
```diff
@@ -67,6 +67,17 @@ const (
 	envWarmIPTarget = "WARM_IP_TARGET"
 	noWarmIPTarget  = 0
 
+	// This environment variable is used to specify the desired minimum number of total IPs.
+	// When it is not set, ipamd defaults to 0.
+	// For example, for a m4.4xlarge node,
+	//     If WARM-IP-TARGET is set to 1 and MINIMUM_IP_TARGET is set to 12, and there are 9 pods running on the node,
```
edmorley:
```suggestion
	//     If WARM_IP_TARGET is set to 1 and MINIMUM_IP_TARGET is set to 12, and there are 9 pods running on the node,
```

ipamd/datastore/data_store.go
```diff
@@ -358,6 +370,11 @@ func (ds *DataStore) getDeletableENI(warmIPTarget int) *ENIIPPool {
 			continue
 		}
 
+		if minimumWarmIPTarget != 0 && ds.isRequiredForMinimumWarmIPTarget(minimumWarmIPTarget, eni) {
+			log.Debugf("ENI %s cannot be deleted because it is required for MINIMUM_WARM_IP_TARGET: %d", eni.ID, minimumWarmIPTarget)
```
edmorley:
```suggestion
			log.Debugf("ENI %s cannot be deleted because it is required for MINIMUM_IP_TARGET: %d", eni.ID, minimumWarmIPTarget)
```

ipamd/datastore/data_store.go
```diff
@@ -331,7 +331,19 @@ func (ds *DataStore) isRequiredForWarmIPTarget(warmIPTarget int, eni *ENIIPPool)
 	return otherWarmIPs < warmIPTarget
 }
 
-func (ds *DataStore) getDeletableENI(warmIPTarget int) *ENIIPPool {
+// IsRequiredForMinimumWarmIPTarget determines if this ENI is necessary to fulfill whatever MINIMUM_WARM_IP_TARGET is
```
edmorley:
```suggestion
// IsRequiredForMinimumWarmIPTarget determines if this ENI is necessary to fulfill whatever MINIMUM_IP_TARGET is
```

asheldon:
Thanks, @edmorley!

asheldon:
Hi @jaypipes. MINIMUM_IP_TARGET is for pre-scaling, WARM_IP_TARGET is for dynamic scaling.

Suppose I have a cluster and know I will deploy approximately 30 pods to each node. What I've observed in practice is that people set WARM_IP_TARGET to 30 in this case to ensure there are enough IPs allocated up-front by the CNI. Then, when their 30 pods are running, the CNI allocates an additional 30 IPs to the node that are never used. This accelerates IP exhaustion in the relevant subnets.

If instead they set the proposed MINIMUM_IP_TARGET to 30 and WARM_IP_TARGET to 2, they would deploy their 30 expected pods and the CNI would allocate an additional 2 IPs. This still provides elasticity, but uses roughly half as many IPs as the prior behavior (32 vs 60 IPs).

This also improves reliability of the EKS cluster by reducing the number of calls necessary to allocate or deallocate private IPs, which may be throttled, especially at scaling-related times.

jaypipes:
> Hi @jaypipes. MINIMUM_IP_TARGET is for pre-scaling, WARM_IP_TARGET is for dynamic scaling.
> 
> Suppose I have a cluster and know I will deploy approximately 30 pods to each node. What I've observed in practice is that people set WARM_IP_TARGET to 30 in this case to ensure there are enough IPs allocated up-front by the CNI. Then, when their 30 pods are running, the CNI allocates an additional 30 IPs to the node that are never used. This accelerates IP exhaustion in the relevant subnets.
> 
> If instead they set the proposed MINIMUM_IP_TARGET to 30 and WARM_IP_TARGET to 2, they would deploy their 30 expected pods and the CNI would allocate an additional 2 IPs. This still provides elasticity, but uses roughly half as many IPs as the prior behavior (32 vs 60 IPs).
> 
> This also improves reliability of the EKS cluster by reducing the number of calls necessary to allocate or deallocate private IPs, which may be throttled, especially at scaling-related times.

Ahh... I see now, thank you for the excellent explanation. OK, I will re-review with that understanding.

Cheers,
-jay

jaypipes:
> Hi @jaypipes. MINIMUM_IP_TARGET is for pre-scaling, WARM_IP_TARGET is for dynamic scaling.
> 
> Suppose I have a cluster and know I will deploy approximately 30 pods to each node. What I've observed in practice is that people set WARM_IP_TARGET to 30 in this case to ensure there are enough IPs allocated up-front by the CNI. Then, when their 30 pods are running, the CNI allocates an additional 30 IPs to the node that are never used. This accelerates IP exhaustion in the relevant subnets.
> 
> If instead they set the proposed MINIMUM_IP_TARGET to 30 and WARM_IP_TARGET to 2, they would deploy their 30 expected pods and the CNI would allocate an additional 2 IPs. This still provides elasticity, but uses roughly half as many IPs as the prior behavior (32 vs 60 IPs).
> 
> This also improves reliability of the EKS cluster by reducing the number of calls necessary to allocate or deallocate private IPs, which may be throttled, especially at scaling-related times.

Ahh... I see now, thank you for the excellent explanation. OK, I will re-review with that understanding.

Cheers,
-jay

ipamd/datastore/data_store_test.go
```diff
@@ -288,14 +288,15 @@ func TestPodIPv4Address(t *testing.T) {
...
 	// Should not be able to free this ENI
-	eni := ds.RemoveUnusedENIFromStore(noWarmIPTarget)
+	eni := ds.RemoveUnusedENIFromStore(noWarmIPTarget, noMinimumIPTarget)
 	assert.True(t, eni == "")
 
 	ds.eniIPPools["eni-2"].createTime = time.Time{}
 	ds.eniIPPools["eni-2"].lastUnassignedTime = time.Time{}
-	eni = ds.RemoveUnusedENIFromStore(noWarmIPTarget)
+	eni = ds.RemoveUnusedENIFromStore(noWarmIPTarget, noMinimumIPTarget)
 	assert.Equal(t, eni, "eni-2")
```
jaypipes: Can you please add a couple new tests that validates the new behaviour when MINIMUM_IP_TARGET > 0 and when WARM_IP_TARGET is set and not set?

asheldon: Sure thing.

ipamd/datastore/data_store.go
```diff
@@ -331,7 +331,19 @@ func (ds *DataStore) isRequiredForWarmIPTarget(warmIPTarget int, eni *ENIIPPool)
 	return otherWarmIPs < warmIPTarget
 }
 
-func (ds *DataStore) getDeletableENI(warmIPTarget int) *ENIIPPool {
+// IsRequiredForMinimumWarmIPTarget determines if this ENI is necessary to fulfill whatever MINIMUM_IP_TARGET is
```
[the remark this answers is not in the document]
asheldon: I think this might be a misleading change to make since this method considers only the minimum IP target and the prior method, 'isRequiredForWarmIPTarget', is responsible for the warm IP target.

README.md
```diff
@@ -234,6 +234,17 @@ Specifies the number of free IP addresses that the `ipamD` daemon should attempt
...
+This environment variable overrides `WARM_ENI_TARGET` behavior.
+
+`MINIMUM_IP_TARGET`
+Type: Integer
+Default: None
+Specifies the minimum number of IP addresses that the `ipamD` daemon should attempt to allocate for pod assignment on the node.
+For example, if `MINIMUM_IP_TARGET` is set to 10, then `ipamD` attempts to prevent the total number of IP addresses from falling
+below 10 at all times. If the elastic network interfaces on the node are unable to provide this many addresses, `ipamD` attempts
+to allocate more interfaces until `MINIMUM_IP_TARGET` IP addresses exist in total.
+If both `WARM_IP_TARGET` and `MINIMUM_IP_TARGET` are set, `ipamD` will attempt to meet both constraints.
```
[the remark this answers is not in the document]
asheldon: Please let me know if the new text looks good to you.

ipamd/datastore/data_store.go
```diff
@@ -324,7 +324,19 @@ func (ds *DataStore) isRequiredForWarmIPTarget(warmIPTarget int, eni *ENIIPPool)
...
+// set to.
+func (ds *DataStore) isRequiredForMinimumWarmIPTarget(minimumWarmIPTarget int, eni *ENIIPPool) bool {
```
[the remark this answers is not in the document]
asheldon: Yes, I'll fix this.

asheldon:
Thanks for the feedback, @nckturner! I believe all your concerns have been addressed.

ipamd/datastore/data_store.go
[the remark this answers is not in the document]
asheldon: I had this broken nomenclature in a few other places as well and I think I've swept them all out now.

ipamd/ipamd.go
```diff
@@ -1046,10 +1059,27 @@ func getWarmIPTarget() int {
...
-// ipTargetState determines the number of IPs `short` or `over` our WARM_IP_TARGET
+func getMinimumIPTarget() int {
+	inputStr, found := os.LookupEnv(envMinimumIPTarget)
+
+	if !found {
+		return noMinimumIPTarget
+	}
+
+	if input, err := strconv.Atoi(inputStr); err == nil {
+		if input >= 0 {
```
mogren: We do know the maximum possible value here as well, `c.maxENI  * c.maxIPsPerENI`. It's probably best to check that as well, since on line 1099 you do `over = min(over, total-c.minimumIPTarget)`, and that would return a negative value if someone set `MINIMUM_IP_TARGET=1000`.

asheldon: We don't know `c.maxENI` / `c.maxIPsPerENI` yet because that is calculated as part of `nodeInit()`, which is called after this method is called during `New` and `nodeInit()` depends on `c.ipTargetState()` having these values already. I don't see anything obvious preventing additional refactoring to make these specific values available earlier, but could you please walk me through a bit of the motivation? It seems like this behavior is now the same as it was with a very large `WARM_IP_TARGET`.

I've updated line 1099 to prevent returning a negative number. I had assumed that calling code would treat negative numbers the same as 0, but it looks like that's not the case. Great catch!

mogren: @asheldon Ah, you're right. We need the instance type first... Thanks for adding the check.

README.md
```diff
@@ -234,8 +234,27 @@ Specifies the number of free IP addresses that the `ipamD` daemon should attempt
...
+Specifies the number of total IP addresses that the `ipamD` daemon should attempt to allocate for pod assignment on the node.
+`MINIMUM_IP_TARGET` behaves identically to `WARM_IP_TARGET` except that instead of setting a target number of free IP
+addresses to keep available at all times, it sets a target number for a floor on how many total IP addresses are allocated.
+
+`MINIMUM_IP_TARGET` is for pre-scaling, `WARM_IP_TARGET` is for dynamic scaling. For example, suppose a cluster has an
+expected pod density of approximately 30 pods per node. If `WARM_IP_TARGET` is set to 30 to ensure there are enough IPs
+allocated up front by the CNI, then 30 pods are deployed to the node, the CNI will allocate an additional 30 IPs, for
+a total of 60, accelerating IP exhaustion in the relevant subnets. If instead `MINIMUM_IP_TARGET` is set to 30 and
+`WARM_IP_TARGET` to 2, after the 30 pods are deployed the CNI would allocate an additional 2 IPs. This still provides
+elasticity, but uses roughly half as many IPs as using WARM_IP_TARGET alone (32 IPs vs 60 IPs).
```
jaypipes: bueno. :+1:

ipamd/datastore/data_store_test.go
```diff
@@ -288,16 +288,73 @@ func TestPodIPv4Address(t *testing.T) {
...
+	assert.Contains(t, []string{"eni-2", "eni-3"}, removedEni)
+
+	// Should not be able to free an ENI because minimumWarmIPTarget requires at least two ENIs and no warm IP target
+	eni = ds.RemoveUnusedENIFromStore(noWarmIPTarget, 3)
+	assert.Equal(t, "", eni)
+	// Should be able to free an ENI because one ENI can provide a minimum count of 2 IPs
+	secondRemovedEni := ds.RemoveUnusedENIFromStore(noWarmIPTarget, 2)
+	assert.Contains(t, []string{"eni-2", "eni-3"}, secondRemovedEni)
+
+	assert.NotEqual(t, removedEni, secondRemovedEni, "The two removed ENIs should not be the same ENI.")
```
jaypipes: Good tests. :heavy_check_mark:

PR #622, opened 2019-09-23
Add environment variable to configure MTU

### LINKED ISSUE #356: CNI doesn't support configuring MTU
As far as I understand the maximum transmission unit for each POD is hard coded to 9001 after the completion of #167. This currently causes problems since paths to some hosts our applications need to access has a much lower MTU.

What happens is typically that TCP connections hang waiting for response packages that are to big to be delivered since the agreed on maximum segment size (MSS) of the connection is too high.

This would not happen if the network paths to these hosts where letting through ICMP “Destination Unreachable, fragmentation required, and DF flag set“ packets, but alas, they don't.

So I request a way to configure the MTU to something lower than 9001.

mogren:
*Issue #, if available:* Resolves #356

*Description of changes:*
* Added `AWS_VPC_ENI_MTU` environment variable to set MTU
* Added unit tests to verify limits.
* Corrected some comments

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,16 @@ To select an `ENIConfig` based upon availability zone set this to `failure-domai
 
 ---
 
+`AWS_VPC_ENI_MTU` 
+
+Type: Integer
+
+Default: 9001 
+
+Used to configure the MTU size for attached ENIs. The valid range is from `576` to `9001`.
+
+---
+
 `AWS_VPC_K8S_CNI_EXTERNALSNAT`
 
 Type: Boolean
--- a/pkg/networkutils/network.go
+++ b/pkg/networkutils/network.go
@@ -77,18 +77,22 @@ const (
 	// envConnmark is the name of the environment variable that overrides the default connection mark, used to
 	// mark traffic coming from the primary ENI so that return traffic can be forced out of the same interface.
 	// Without using a mark, NodePort DNAT and our source-based routing do not work together if the target pod
-	// behind the node port is not on the main ENI.  In that case, the un-DNAT is done after the source-based
+	// behind the node port is not on the main ENI. In that case, the un-DNAT is done after the source-based
 	// routing, resulting in the packet being sent out of the pod's ENI, when the NodePort traffic should be
 	// sent over the main ENI.
 	envConnmark = "AWS_VPC_K8S_CNI_CONNMARK"
 
-	// defaultConnmark is the default value for the connmark described above.  Note: the mark space is a little crowded,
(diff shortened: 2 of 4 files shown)


pkg/networkutils/network.go
```diff
@@ -929,3 +935,27 @@ func (n *linuxNetwork) UpdateRuleListBySrc(ruleList []netlink.Rule, src net.IPNe
...
+		mtu, err := strconv.Atoi(envMTUValue)
+		if err != nil {
+			log.Error("Failed to parse "+envMTU+"; will use ", maximumMTU, err.Error())
+			return maximumMTU
+		}
+		// Restrict range between jumbo frame and the maximum required size to assemble.
+		// Details in https://tools.ietf.org/html/rfc879 and
+		// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/network_mtu.html
+		if mtu < minimumMTU {
+			log.Error(""+envMTU+" too low; will use ", minimumMTU)
```
shyamjvs: Why are you appending empty string in front of envMTU?

mogren: That was just sloppy, copied from the previous error log line. I've updated to use `log.Errorf()`

shyamjvs:
/lgtm

Nice!

README.md
```diff
@@ -152,6 +152,16 @@ To select an `ENIConfig` based upon availability zone set this to `failure-domai
 
 ---
 
+`AWS_VPC_ENI_MTU`
```
vsiddharth: Is this option cluster scoped against node scope?

pkg/networkutils/network.go
```diff
@@ -77,18 +77,22 @@ const (
...
-	// defaultConnmark is the default value for the connmark described above.  Note: the mark space is a little crowded,
+	// defaultConnmark is the default value for the connmark described above. Note: the mark space is a little crowded,
 	// - kube-proxy uses 0x0000c000
 	// - Calico uses 0xffff0000.
 	defaultConnmark = 0x80
 
-	// MTU of ENI - veth MTU defined in plugins/routed-eni/driver/driver.go
-	ethernetMTU = 9001
+	// envMTU gives a way to configure the MTU size for new ENIs attached. Range is from 576 to 9001.
+	envMTU = "AWS_VPC_ENI_MTU"
```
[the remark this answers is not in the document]
mogren: No, they will all get the same setting to keep things simple.

README.md
mogren: All nodes in the cluster will get the same setting.

PR #623, opened 2019-09-24
Lower delete delay

mogren:
*Issue #, if available:*

*Description of changes:*
* Update retry code 
* Use exponential backoff when deleting ENIs
* Always return success for deletes of unknown pods


pkg/awsutils/awsutils.go
```diff
@@ -16,6 +16,7 @@ package awsutils
 import (
 	"context"
 	"fmt"
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/retry"
```
vsiddharth: nit: move this down

Issue #257, opened 2018-12-05
Debug level logging enabled in yaml


gflarity:
Just checking. It seems like enabling debug logging by default in the yaml might lead to excessive logging? 

https://github.com/aws/amazon-vpc-cni-k8s/blob/07363123942c6fa03da6a7737a2d15817a7c2f8e/config/v1.3/aws-k8s-cni.yaml#L79

mogren:
Tracked as part of #511, discussing a switch to a structured logger. Configurability of the logger should be handled in a Helm chart.

PR #631, opened 2019-09-27
Reduce log level to Trace for the most common Debug lines

mogren:
*Description of changes:*
* Reduce logging level to Trace
* Added space to make the tab align the output.


ipamd/ipamd.go
```diff
@@ -824,9 +824,9 @@ func (c *IPAMContext) nodeIPPoolTooLow() bool {
 	available := total - used
 	poolTooLow := available < c.maxIPsPerENI*c.warmENITarget || (c.warmENITarget == 0 && available == 0)
 	if poolTooLow {
-		log.Debugf("IP pool is too low: available (%d) < ENI target (%d) * addrsPerENI (%d)", available, c.warmENITarget, c.maxIPsPerENI)
+		log.Tracef("IP pool is too low: available (%d) < ENI target (%d) * addrsPerENI (%d)", available, c.warmENITarget, c.maxIPsPerENI)
```
shyamjvs: These seem to be informative logs, what's your motivation to move those to the lowest log level? Are you planning to not log your binary at this level?

mogren: Yes, we log too much. Another option would be to change the default to be Info instead of Debug (see #257), but when testing stuff Trace would be helpful. I've updated the PR.

ipamd/ipamd.go
```diff
@@ -570,8 +570,10 @@ func (c *IPAMContext) increaseIPPool() {
 	} else {
 		// If we did not add an IP, try to add an ENI instead.
 		if c.dataStore.GetENIs() < c.maxENI {
-			c.tryAllocateENI()
-			c.updateLastNodeIPPoolAction()
+			err = c.tryAllocateENI()
+			if err == nil {
```
shyamjvs: nit: You can move the above statement and this if into a single line.

ipamd/ipamd.go
```diff
@@ -628,15 +631,17 @@ func (c *IPAMContext) tryAllocateENI() {
...
-		return
+		return err
 	}
 
 	err = c.setupENI(eni, eniMetadata)
 	if err != nil {
 		ipamdErrInc("increaseIPPoolsetupENIFailed")
 		log.Errorf("Failed to increase pool size: %v", err)
-		return
+		return err
```
shyamjvs: Maybe you can simply return the error in the above line instead of logging it here and let the caller take care of logging that error?

ipamd/ipamd.go
shyamjvs: So are we ok to log at `Info` level in production? Wouldn't it make it harder for us to debug issues after they've happened if we skip these logs?

mogren: Yes, I'd like to keep the default at Debug level for now, but some of them are very verbose, that's why I'm lowering them to Trace for some of these.

Issue #592, opened 2019-08-21
cni-metrics-helper continues accessing a node after it's gone


tdmalone:
Hi there,

After terminating a node, it appears that `cni-metrics-helper` is still trying to access the no-longer-existing `aws-node` pod:

```
E0821 01:02:31.304392       1 cni_metrics.go:144] grabMetricsFromTarget: Failed to grab CNI endpoint: the server could not find the requested resource (get pods aws-node-ngnhh:61678)
E0821 01:01:31.303667       1 metrics.go:406] Failed to getMetricsFromTarget: the server could not find the requested resource (get pods aws-node-ngnhh:61678)
```

The above messages are logged every 30 seconds after node termination, until the `cni-metrics-helper` pod is restarted.

Zyqsempai:
I will take this one!

shyamjvs:
This is expected behaviour and I'm not sure if there is anything to fix here. You can see the following code:

https://github.com/aws/amazon-vpc-cni-k8s/blob/ccf1c29e451cf2bebc3e490b83db05245a14cc27/cni-metrics-helper/metrics/metrics.go#L404-L409

If an underlying cloud instance is deleted, it takes a while for the node to become unready (because kubelet stops posting heartbeats) and node-controller eventually deletes the node when it checks that the underlying instance has been deleted in the cloud-provider. Soon after, the pod is also deleted by k8s pod_gc_controller which sees that the pod is orphaned (i.e running on a non-existing node). All of this can take around 60-90 sec from the point where the instance was deleted. So any attempts to fetch metrics from the pod in the meanwhile *will* fail, which is why we simply log the error and continue with other pods inside the above code.

shyamjvs:
Let me know if there's still some issue I'm missing. Or we can close it.

tdmalone:
I get this log message turning up for much longer than 60-90 seconds after termination - possibly over half an hour or more, but I'll need to test to verify that. My assumption was that, currently, it would keep happening forever (until I restart the `cni-metrics-helper` pod).

I only noticed it because I have a log alert tracking words like 'failed' to pick up on issues I might not've noticed otherwise. I can easily exclude this particular event from that alert and then the problem essentially 'goes away' - but it seemed to me like something was wrong if this keeps happening for a long time.

jaypipes:
@tdmalone ping on this. have you continued to see the behaviour over long(er than 60-90s) periods?

jaypipes:
[...]

PR #639, opened 2019-10-02
Added exists check for cni pods

Zyqsempai:
issue #592 

Added check if cniPod still exists and not terminated

### DIFF
--- a/pkg/k8sapi/discovery.go
+++ b/pkg/k8sapi/discovery.go
@@ -110,7 +110,9 @@ func (d *Controller) GetCNIPods() []string {
 	defer d.cniPodsLock.Unlock()
 
 	for k := range d.cniPods {
-		cniPods = append(cniPods, k)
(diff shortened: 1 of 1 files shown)


Zyqsempai:
@mogren PTAL, I know that it's not so efficient but I didn't found more elegant way to do that.

shyamjvs:
I don't think #592 is a issue - replied there. This change doesn't seem to offer much given that we're already updating `d.cniPods` inside `handlePodUpdate`.

@Zyqsempai Let me know if I'm missing sth.

mogren:
Like @shyamjvs said, `delete(d.cniPods, key)` is already called on any pod event in `handlePodUpdate()` so I don't think this change would help.

jaypipes:
@Zyqsempai see comments from @shyamjvs and @mogren. Closing this pull request out.

Issue #347, opened 2019-03-14
Do exponential backoffs for retries


mogren:
Currently calls with retries have a 5 second pause between each try, they should be exponential instead.

Zyqsempai:
I will take this one!

mogren:
Resolved.

PR #640, opened 2019-10-02
Made timeouts exponential

Zyqsempai:
Issue #347

Made timeouts exponential

### DIFF
--- a/pkg/networkutils/network.go
+++ b/pkg/networkutils/network.go
@@ -761,15 +761,19 @@ func setupENINetwork(eniIP string, eniMAC string, eniTable int, eniSubnetCIDR st
 			return errors.Wrap(err, "setupENINetwork: failed to clean up old routes")
 		}
 
-		err = retry.RetryNWithBackoff(retry.NewSimpleBackoff(500*time.Microsecond, retryRouteAddInterval, 0.15, 2.0), maxRetryRouteAdd, func() error {
+		err = retry.RetryNWithBackoff(retry.NewSimpleBackoff(500*time.Millisecond, retryRouteAddInterval, 0.15, 2.0), maxRetryRouteAdd, func() error {
 			if err := netLink.RouteReplace(&r); err != nil {
-				log.Debugf("Not able to add/replace route route %s/0 via %s table %d ",
-					r.Dst.IP.String(), gw.String(), eniTable, maxRetryRouteAdd)
+				log.Debugf("Not able to set route %s/0 via %s table %d",
+					r.Dst.IP.String(), gw.String(), eniTable)
 				return errors.Wrapf(err, "setupENINetwork: unable to replace route entry %s", r.Dst.IP.String())
 			}
+
 			log.Debugf("Successfully added/replaced route to be %s/0", r.Dst.IP.String())
 			return nil
 		})
+		if err != nil {
+			return err
+		}
 	}
 
 	// Remove the route that default out to ENI-x out of main route table
--- a/pkg/networkutils/network.go
+++ b/pkg/networkutils/network.go
@@ -17,6 +17,7 @@ import (
 	"encoding/binary"
 	"encoding/csv"
 	"fmt"
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/retry"
 	"io"
 	"math"
 	"net"
@@ -645,7 +646,7 @@ func LinkByMac(mac string, netLink netlinkwrapper.NetLink, retryInterval time.Du
 		if attempt > maxAttemptsLinkByMac {
 			return nil, lastErr
 		} else if attempt > 1 {
-			time.Sleep(retryInterval * time.Duration(attempt))
+			time.Sleep(retryInterval)
 		}
 
 		links, err := netLink.LinkList()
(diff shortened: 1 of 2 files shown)


pkg/networkutils/network.go
```diff
@@ -745,36 +746,28 @@ func setupENINetwork(eniIP string, eniMAC string, eniTable int, eniSubnetCIDR st
...
-							r.Dst.IP.String(), gw.String(), eniTable)
-					}
-					log.Debugf("Not able to add route route %s/0 via %s table %d (attempt %d/%d)",
-						r.Dst.IP.String(), gw.String(), eniTable, retry, maxRetryRouteAdd)
+					log.Debugf("Not able to add route route %s/0 via %s table %d ",
+						r.Dst.IP.String(), gw.String(), eniTable, maxRetryRouteAdd)
 					time.Sleep(retryRouteAddInterval)
 				} else if netlinkwrapper.IsRouteExistsError(err) {
 					if err := netLink.RouteReplace(&r); err != nil {
 						return errors.Wrapf(err, "setupENINetwork: unable to replace route entry %s", r.Dst.IP.String())
```
mogren: Hmm, honestly I'm not sure why we try to Add, then Replace. We should be able to use `netLink.RouteReplace()` right away, since it will also add it if it doesn't already exist. That would make this code even simpler.

pkg/networkutils/network.go
```diff
@@ -745,36 +746,28 @@ func setupENINetwork(eniIP string, eniMAC string, eniTable int, eniSubnetCIDR st
 			return errors.Wrap(err, "setupENINetwork: failed to clean up old routes")
 		}
 
-		// In case of route dependency, retry few times
-		retry := 0
-		for {
+		err = retry.RetryNWithBackoff(retry.NewSimpleBackoff(time.Millisecond*500, retryRouteAddInterval*time.Duration(maxRetryRouteAdd), 0.15, 2.0), maxRetryRouteAdd, func() error {
```
mogren: Nice! I do think we could reduce the max time though to just the `retryRouteAddInterval`, since that is 5s:
```
retry.NewSimpleBackoff(0.5 * time.Second, retryRouteAddInterval, 0.15, 2.0)
```
This would give the delays (not counting the jitter): 0.5, 1, 2, 4, 5 seconds.

mogren:
Thanks, looks a lot cleaner, but `TestSetupENINetwork` fails because we use `RouteReplace` now instead of `RouteAdd`. You can run the tests locally with `make docker-unit-tests` to test any changes.

Zyqsempai:
@mogren Done, thanks for your time.

pkg/networkutils/network.go
```diff
@@ -745,36 +747,15 @@ func setupENINetwork(eniIP string, eniMAC string, eniTable int, eniSubnetCIDR st
...
-					log.Debugf("Successfully replaced route to be %s/0", r.Dst.IP.String())
-					break
-				} else {
-					return errors.Wrapf(err, "setupENINetwork: unable to add route %s/0 via %s table %d",
-						r.Dst.IP.String(), gw.String(), eniTable)
-				}
-			} else {
-				log.Debugf("Successfully added route route %s/0 via %s table %d", r.Dst.IP.String(), gw.String(), eniTable)
-				break
+		err = retry.RetryNWithBackoff(retry.NewSimpleBackoff(500*time.Microsecond, retryRouteAddInterval, 0.15, 2.0), maxRetryRouteAdd, func() error {
```
mogren: Thanks a lot for simplifying this code! 

That said, `500*time.Microsecond` should be `500 * time.Millisecond` (or `0.5 * time.Second`), or we will call way to often.

pkg/networkutils/network.go
```diff
@@ -745,36 +747,15 @@ func setupENINetwork(eniIP string, eniMAC string, eniTable int, eniSubnetCIDR st
...
+			if err := netLink.RouteReplace(&r); err != nil {
+				log.Debugf("Not able to add/replace route route %s/0 via %s table %d ",
+					r.Dst.IP.String(), gw.String(), eniTable, maxRetryRouteAdd)
```
mogren: We don't need `maxRetryRouteAdd`, it's not printed:
```
log.Debugf("Not able to set route %s/0 via %s table %d"
	r.Dst.IP.String(), gw.String(), eniTable)
```

pkg/networkutils/network.go
```diff
@@ -745,36 +747,15 @@ func setupENINetwork(eniIP string, eniMAC string, eniTable int, eniSubnetCIDR st
...
+				return errors.Wrapf(err, "setupENINetwork: unable to replace route entry %s", r.Dst.IP.String())
 			}
-		}
+			log.Debugf("Successfully added/replaced route to be %s/0", r.Dst.IP.String())
+			return nil
+		})
 	}
```
mogren: We should check that `err != nil` here, since if all the retries fail, `retry.RetryNWithBackoff()` will return the last error. Add
```
if err != nil {
	return err
}
```

Zyqsempai:
@mogren Sorry for delay, I was in vacation, now I a back! PTAL

PR #648, opened 2019-10-08
Add IPs to the first ENI on startup

mogren:
*Description of changes:*
* Make sure we have IPs available before making the node ready
* This change will add a 400 - 600ms delay before marking a node as *Ready*
* For new nodes, this saves at least 2.5 seconds before IPs are available

### DIFF
--- a/ipamd/ipamd.go
+++ b/ipamd/ipamd.go
@@ -368,7 +368,12 @@ func (c *IPAMContext) nodeInit() error {
 			log.Errorf("UpdateRuleListBySrc in nodeInit() failed for IP %s: %v", ip.IP, err)
 		}
 	}
-	return nil
+	// For a new node, attach IPs
+	increasedPool, err := c.tryAssignIPs()
+	if err == nil && increasedPool {
+		c.updateLastNodeIPPoolAction()
+	}
+	return err
 }
 
 func (c *IPAMContext) getLocalPodsWithRetry() ([]*k8sapi.K8SPodInfo, error) {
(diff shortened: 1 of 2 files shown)


jaypipes:
Is there an issue this is resolving/addresssing?

ipamd/ipamd.go
```diff
@@ -368,7 +368,12 @@ func (c *IPAMContext) nodeInit() error {
 			log.Errorf("UpdateRuleListBySrc in nodeInit() failed for IP %s: %v", ip.IP, err)
 		}
 	}
-	return nil
+	// For a new node, attach IPs
+	increasedPool, err := c.tryAssignIPs()
+	if increasedPool {
```
jaypipes: Should we check whether `err != nil` before checking if increasedPool is true?

ipamd/ipamd.go
```diff
@@ -651,9 +656,7 @@ func (c *IPAMContext) tryAssignIPs() (increasedPool bool, err error) {
 	eni := c.dataStore.GetENINeedsIP(c.maxIPsPerENI, c.useCustomNetworking)
 	if eni != nil && len(eni.IPv4Addresses) < c.maxIPsPerENI {
 		currentNumberOfAllocatedIPs := len(eni.IPv4Addresses)
-		log.Debugf("Found ENI %s that has less than the maximum number of IP addresses allocated: cur=%d, max=%d", eni.ID, currentNumberOfAllocatedIPs, c.maxIPsPerENI)
```
jaypipes: any reason to remove this debug line?

ipamd/ipamd.go
```diff
@@ -651,9 +656,7 @@ func (c *IPAMContext) tryAssignIPs() (increasedPool bool, err error) {
...
 		// Try to allocate all available IPs for this ENI
-		// TODO: Retry with back-off, trying with half the number of IPs each time
```
jaypipes: Is this no longer a TODO? Are we not going to try a backoff (other than "if >1 allocation failed, try a single IP allocation" strategy)?

ipamd/ipamd.go
mogren: Yes, we log the exact same line in `GetENINeedsIP()`, so it's always just duplicated.

ipamd/ipamd.go
mogren: Yes, this is a left over comment from another approach of allocating ENIs and it is not valid any more.

Issue #653, opened 2019-10-14
Add support for m5n/m5dn/r5n/r5dn instance types


Jeffwan:
r5dn instances are not supported yet. Let's enable the support and unlock some customers.

https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/awsutils/vpc_ip_resource_limit.go

Jeffwan:
just notice https://aws.amazon.com/about-aws/whats-new/2019/10/introducing-amazon-ec2-m5n-m5dn-r5n-and-r5dn-instances-featuring-100-gbps-of-network-bandwidth/

There're also m5n, m5dn, r5n there. I will file a follow up PR.

PR #657, opened 2019-10-14
Add support for m5n/m5dn/r5n instances

Jeffwan:
*Issue #, if available:*
#653 

This is a follow up PR of #656. I just notice there're few other instance releases together with r5dn. We should add support for them together. 

*Description of changes:*
Add eni and maximum ips for m5n/m5dn/r5n instances


mogren:
> I wish there was a way to automate/generate this information :)

Yeah, like an EC2 API that would give us this info....

pkg/awsutils/vpc_ip_resource_limit.go
```diff
@@ -148,6 +148,22 @@ var InstanceENIsAvailable = map[string]int{
 	"m5d.16xlarge":  15,
 	"m5d.24xlarge":  15,
 	"m5d.metal":     15,
+	"m5dn.large":    3,
```
mogren: We need the number of IPs per ENI for all types as well in this map:
```
var InstanceIPsAvailable = map[string]int{
```

Jeffwan: Ah, I forget to add this part. Just add to the list

Jeffwan:
Maybe we can run some utility to parse page and auto generate a go file?
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html

Instead of adding instance types, we should generate full snapshot every time.

mogren:
@Jeffwan we do have a script like that already, but it needs some updates 🙂

pkg/awsutils/vpc_ip_resource_limit.go
mogren: Thanks!

PR #666, opened 2019-10-21
Log plugin at debug level

mogren:
*Description of changes:*
* Fix comments
* Remove unneeded casts
* Default to Debug level instead of info

### DIFF
--- a/pkg/networkutils/network.go
+++ b/pkg/networkutils/network.go
@@ -40,12 +40,12 @@ import (
 )
 
 const (
-	// 0- 511 can be used other higher priorities
+	// 0 - 511 can be used other higher priorities
 	toPodRulePriority = 512
 
 	// 513 - 1023, can be used priority lower than toPodRulePriority but higher than default nonVPC CIDR rule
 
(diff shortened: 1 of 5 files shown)


pkg/networkutils/network.go
```diff
@@ -809,14 +809,14 @@ func incrementIPv4Addr(ip net.IP) (net.IP, error) {
...
-	intIP := binary.BigEndian.Uint32([]byte(ip4))
+	intIP := binary.BigEndian.Uint32(ip4)
 	if intIP == (1<<32 - 1) {
 		return nil, fmt.Errorf("%q will be overflowed", ip)
 	}
 	intIP++
 	bytes := make([]byte, 4)
 	binary.BigEndian.PutUint32(bytes, intIP)
-	return net.IP(bytes), nil
+	return bytes, nil
```
nckturner: Is this ok?  You're no longer casting to net.IP but the function still returns net.IP.

mogren: Yes, `type IP []byte`, so they are equivalent.

mogren: Renamed the variable to make it clearer.

Issue #641, opened 2019-10-02
Upgrade CNI version broke pod-to-pod communication within the same worker node


rimaulana:
After upgrading the CNI version from v1.5.1-rc1 to v1.5.4, we are seeing issue where pod was unable to communicate with other pod on the same worker node. We have the following schema

CoreDNS pod on eth0
Kibana pod on eth0
App1 on eth1
App2 on eth2

What we are seeing is that DNS query from App1 and App2 failed with no server found when we tried it using dig command

dig @CoreDNS-ip amazonaws.com

Meanwhile, executing the same command from Kibana pod, the worker node and pod on a different worker node works as expected.

When collecting the logs using https://github.com/nithu0115/eks-logs-collector, we found out that CoreDNS IP was not found anywhere on the output of the ip rule show command. I would expect for each IP address of a pod running on the worker node it should have at least this associated rule on the ip rule

512:    from all to POD_IP lookup main

However, we do not see one for the CoreDNS pod IP. Therefore, we believe that this is an issue with the CNI plugin unable to rebuild the rule after upgrade. There is an internal issue open for this if you want to get the collected logs

MartiUK:
Downgrading to v1.5.3 resolved this issue on (EKS) K8s v1.14 with CoreDNS v1.3.1. 
Required node reboots first.

mogren:
Glad you found a work-around (rebooting the nodes), but I'll keep trying to reproduce this.

igor-pinchuk:
Facing the same issue

ueokande:
We encountered the issue with Kubernetes 1.13 (eks.4) and amazon-vpc-cni-k8s v1.5.4.  Its not only on CoreDNS, but also inter-pods communication.

It occurs immediately after cluster created.  We just repaired by restarting pods (release and reassign an IP address on the pod):

```console
$ kubectl delete pod --all
$ kubectl delete pod -nkube-system --all
```

dmarkey:
I've been tearing my hair out all day after upgrading a cluster. Please change https://docs.aws.amazon.com/eks/latest/userguide/update-cluster.html to suggest v1.5.3 and not v1.5.4 as to not break more clusters until it's verified that this bug is fixed.

mogren:
[...]

PR #667, opened 2019-10-21
Avoid trying tearing down NS

mogren:
*Issue #641*

*Description of changes:*
When deleting a deleted pod, we should not call TeardownNS
* Added Trace logging of pod events
* Check for IP before trying to tear down network namespace


nithu0115:
Thanks for change @mogren. I am running custom built CNI that include this commit over night. So far, I do not see issues with pod to pod connectivity or pod to coreDNS connectivity or pod to external service connectivity.

PR #674, opened 2019-10-23
Calling the veth interfaces eni is confusing

mogren:
*Description of changes:*
* Rename the veth interfaces to `vethxxxxx` instead of `enixxxxx`


dadux:
@mogren - This change breaks calico / kube2iam - basically anything that relies on the interface name to setup iptables rules.

Example : https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/v1.5/calico.yaml#L47

dadux:
What worries me slightly here is that it got merged to master, without a question - no integration tests ?

mogren:
@dadux Good call out. We should definitely improve the integration tests around this to catch this earlier.

PR #680, opened 2019-10-25
Revert "Calling the veth interfaces eni is confusing"

mogren:
Reverts aws/amazon-vpc-cni-k8s#674

Pains me to revert this since having it called `eni` doesn't feel right at all, but changing it now will probably cause more trouble than it's worth.


jaypipes:
Yeah, we'll do this as a 2.0 CNI plugin thing.

Issue #697, opened 2019-10-30
Script to trigger integration/e2e tests


nckturner:
We need a script to trigger e2e/integration tests that optionally provisions a test cluster, tears it down, is configurable, etc.  It should be executable from a CI job.

PR #698, opened 2019-10-30
Add run-integration-tests.sh script

nckturner:
*Issue #, if available:*
#697 

*Description of changes:*
- Runs integration tests
- Uses aws-k8s-tester
- Adapted from CSI test script, there may still be some stuff in here
that won't apply


scripts/up-test-cluster.sh
```diff
@@ -0,0 +1,32 @@
...
+        AWS_K8S_TESTER_EKS_ENABLE_WORKER_NODE_PRIVILEGED_PORT_ACCESS=true \
+        AWS_K8S_TESTER_EKS_WORKER_NODE_ASG_MIN=3 \
+        AWS_K8S_TESTER_EKS_WORKER_NODE_ASG_MAX=3 \
+        AWS_K8S_TESTER_EKS_WORKER_NODE_ASG_DESIRED_CAPACITY=3 \
+        AWS_K8S_TESTER_EKS_WORKER_NODE_PRIVATE_KEY_PATH=$SSH_KEY_PATH \
+        AWS_K8S_TESTER_EKS_WORKER_NODE_INSTANCE_TYPE=m3.xlarge \
+        $TESTER_PATH eks create cluster --path /tmp/aws-k8s-tester-eks.yaml
+
+    # Wait for cluster creation
+    while [[ 1 ]]; do
```
nckturner: I think I can remove this, aws-k8s-tester was waiting built in.

scripts/run-integration-tests.sh
```diff
@@ -0,0 +1,73 @@
...
+    # Push test image
+    eval $(aws ecr get-login --region $REGION --no-include-email)
+    AWS_ACCOUNT_ID=$(aws sts get-caller-identity --query Account --output text)
+    IMAGE_NAME="$AWS_ACCOUNT_ID.dkr.ecr.$REGION.amazonaws.com/amazon/amazon-k8s-cni"
+    IMAGE_VERSION=$(git describe --tags --always --dirty)
+
+    make docker IMAGE=$IMAGE_NAME VERSION=$IMAGE_VERSION
+    docker push $IMAGE_NAME:$IMAGE_VERSION
+
+    sed -i'.bak' "s,602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni,$IMAGE_NAME," ./config/v1.5/aws-k8s-cni.yaml
```
nckturner: Not sure if this works on linux

scripts/run-integration-tests.sh
```diff
@@ -0,0 +1,74 @@
...
+    sed -i'.bak' "s,v1.5.3,$IMAGE_VERSION," ./config/v1.5/aws-k8s-cni.yaml
+fi
+
+echo "Deploying CNI"
+export KUBECONFIG=/tmp/aws-k8s-tester/kubeconfig
+kubectl apply -f ./config/v1.5/aws-k8s-cni.yaml
```
jaypipes: This needs to have "v1.5" parameterized with an envvar.

scripts/run-integration-tests.sh
```diff
@@ -0,0 +1,74 @@
...
+    # Push test image
+    eval $(aws ecr get-login --region $REGION --no-include-email)
+    AWS_ACCOUNT_ID=$(aws sts get-caller-identity --query Account --output text)
+    IMAGE_NAME="$AWS_ACCOUNT_ID.dkr.ecr.$REGION.amazonaws.com/amazon/amazon-k8s-cni"
+    IMAGE_VERSION=$(git describe --tags --always --dirty)
+
+    make docker IMAGE=$IMAGE_NAME VERSION=$IMAGE_VERSION
+    docker push $IMAGE_NAME:$IMAGE_VERSION
+
+    sed -i'.bak' "s,602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni,$IMAGE_NAME," ./config/v1.5/aws-k8s-cni.yaml
```
jaypipes: This needs to have "v1.5" parameterized with an envvar.

scripts/run-integration-tests.sh
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+DIR=$(cd "$(dirname "$0")"; pwd)
+source $DIR/up-test-cluster.sh
+source $DIR/down-test-cluster.sh
```
jaypipes: femto-nit: you can just have a single `scripts/lib/cluster.sh` script that contains these functions.

scripts/run-integration-tests.sh
```diff
@@ -0,0 +1,74 @@
...
+mkdir -p $REPORT_DIR
+
+# Download aws-k8s-tester if not yet
+if [[ ! -e $TESTER_PATH ]]; then
+  echo "Downloading aws-k8s-tester from $TESTER_DOWNLOAD_URL to $TESTER_PATH"
+  curl -L -X GET $TESTER_DOWNLOAD_URL -o $TESTER_PATH
+  chmod +x $TESTER_PATH
+fi
+
+if [[ "$PROVISION" = true ]]; then
```
jaypipes: Instead of envvars, might be nice to accept normal CLI args. just a suggestion, though :)

scripts/run-integration-tests.sh
```diff
@@ -0,0 +1,74 @@
...
+fi
+
+echo "Deploying CNI"
+export KUBECONFIG=/tmp/aws-k8s-tester/kubeconfig
+kubectl apply -f ./config/v1.5/aws-k8s-cni.yaml
+
+# Run the test
+pushd ./test/integration
+go test -v -timeout 0 ./... --kubeconfig=$KUBECONFIG --ginkgo.focus="\[cni-integration\]" --ginkgo.skip="\[Disruptive\]" \
+    --assets=${DIR}/../test/integration/assets
```
jaypipes: Why specify `../test/integration` if you `pushd ./test/integration` above here?

scripts/run-integration-tests.sh
```diff
@@ -0,0 +1,74 @@
...
+DIR=$(cd "$(dirname "$0")"; pwd)
+source $DIR/up-test-cluster.sh
+source $DIR/down-test-cluster.sh
+
+OS=$(go env GOOS)
+ARCH=amd64
+TEST_ID=$RANDOM
+CLUSTER_NAME=test-cluster-$TEST_ID
+BASE_DIR=$(dirname $0)
+TEST_DIR=/tmp/cni-test
```
jaypipes: Consider using `mktemp -d` here instead.

scripts/run-integration-tests.sh
```diff
@@ -0,0 +1,74 @@
...
+go test -v -timeout 0 ./... --kubeconfig=$KUBECONFIG --ginkgo.focus="\[cni-integration\]" --ginkgo.skip="\[Disruptive\]" \
+    --assets=${DIR}/../test/integration/assets
+TEST_PASS=$?
+popd
+
+if [[ "$DEPROVISION" = true ]]; then
+    down-test-cluster
+fi
+
+rm -rf $TEST_DIR
```
jaypipes: If you use `mktemp -d` to create the test dir, you could just as well leave off this `rm -rf` and let the normal tmpdir cleanup happen.

scripts/up-test-cluster.sh
jaypipes: Does it?

scripts/run-integration-tests.sh
```diff
@@ -0,0 +1,74 @@
...
+REGION=${AWS_REGION:-us-west-2}
+K8S_VERSION=${K8S_VERSION:-1.14.1}
+PROVISION=${PROVISION:-true}
+DEPROVISION=${DEPROVISION:-true}
+BUILD=${BUILD:-true}
+
+echo "Testing in region: $REGION"
+
+TESTER_DOWNLOAD_URL=https://github.com/aws/aws-k8s-tester/releases/download/v0.4.3/aws-k8s-tester-v0.4.3-$OS-$ARCH
+TESTER_PATH=$TEST_DIR/aws-k8s-tester
```
jaypipes: Recommend *not* putting the `aws-k8s-tester` binary in `$TEST_DIR` because you `rm -rf $TEST_DIR` at the end of each test run...

scripts/up-test-cluster.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+function up-test-cluster() {
+    echo "Creating cluster $CLUSTER_NAME"
+    CLUSTER_YAML_PATH=$TEST_DIR/$CLUSTER_NAME.yaml
+    SSH_KEY_PATH=$TEST_DIR/id_rsa
+    ssh-keygen -P cni-test -f $SSH_KEY_PATH
+
+
+    $TESTER_PATH eks create config --path /tmp/aws-k8s-tester-eks.yaml
```
jaypipes: Recommend putting `/tmp/aws-k8s-tester-eks.yaml` inside `$TEST_DIR` so you can test different clusters simultaneously.

scripts/run-integration-tests.sh
nckturner: Done

scripts/run-integration-tests.sh
nckturner: Done

scripts/run-integration-tests.sh
nckturner: Femto-Done

scripts/run-integration-tests.sh
nckturner: I don't disagree, but I also don't love doing args in bash...

scripts/run-integration-tests.sh
nckturner: We may never know...

scripts/run-integration-tests.sh
nckturner: I was just thinking, we probably don't want to clean this directory up, rather just put it in tmp so it can be viewed after the run is over.

scripts/run-integration-tests.sh
nckturner: Removed `rm -rf $TEST_DIR` in favor of leaving behind `/tmp/$TEST_DIR` for examination

scripts/up-test-cluster.sh
nckturner: I added a separate test directory per cluster, so that both "1 cluster per test" and "1 cluster multiple tests" are supported.

scripts/up-test-cluster.sh
nckturner: Yup

nckturner:
Updated the PR and addressed your comments @jaypipes

nckturner:
Depends on: https://github.com/aws/aws-k8s-tester/pull/62

test/integration/integration_test.go
```diff
@@ -0,0 +1,112 @@
...
+var _ = ginkgo.BeforeSuite(func() {
+	fmt.Printf("Using KUBECONFIG=\"%s\"\n", framework.TestContext.KubeConfig)
+})
+
+var _ = ginkgo.Describe("[cni-integration]", func() {
+	var f *framework.Framework
+	f = framework.NewDefaultFramework("cni-integration")
+
+	ginkgo.It("should enable pod-pod communication", func() {
+		serverPod := newTestPod()
```
jaypipes: I think you need to clean up this resource, no? Otherwise, pods will continue to build up on the test cluster until the cluster is destroyed.

nckturner: Added a clean up.

nckturner:
The aws-k8s-tester PR is merged, can you take another look @jaypipes?

jaypipes:
> The aws-k8s-tester PR is merged, can you take another look @jaypipes?

Build was failing, though it looks to be transient failure in getting go lint tool. I've restarted the build jobs and will approve once I see a green.

test/integration/integration_test.go
wongma7: Prefer to set DeleteNamespaceOnFailure https://github.com/kubernetes/kubernetes/blob/cf06a2efd6ec950dc6707b57cd55cb4f50d29f33/test/e2e/framework/test_context.go#L278 true.

wongma7:
did not review bash because I hate it, also hate that we are calling these "integration" when we're spinning up a kubernetes cluster and testing cni "end to end" :)

test/integration/integration_test.go
jaypipes: > Prefer to set DeleteNamespaceOnFailure https://github.com/kubernetes/kubernetes/blob/cf06a2efd6ec950dc6707b57cd55cb4f50d29f33/test/e2e/framework/test_context.go#L278 true.

The problem with that is that you cannot diagnose something if a failure occurs.

PR #724, opened 2019-11-15
Don't use primary ENI IPs if custom networking is enabled

### LINKED ISSUE #723: IPAMd not allocating additional IPs on primary ENI
I noticed in my clusters that the primary ENI for my nodes aren't getting additional IPs assigned to them. After digging into this I found that the primary interface is skipped if `skipPrimary` is set [here](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/ipamd/datastore/data_store.go#L393) which is passed as the value of `useCustomNetworking` [here](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/ipamd/ipamd.go#L682)-- which is controlled through the `AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG` environment variable [as shown here](https://github.com/aws/amazon-vpc-cni-k8s/blob/5aa67e9a659d2d098b13c8709ddbd000a6637326/ipamd/ipamd.go#L1035)

From the README:

```
Specifies that your pods may use subnets and security groups that are independent of your worker node's VPC configuration. By default, pods share the same subnet and security groups as the worker node's primary interface. Setting this variable to true causes ipamD to use the security groups and VPC subnet in a worker node's ENIConfig for elastic network interface allocation.
```

From reading that I would expect the exact opposite behavior. As it stands now if you `AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG` set to false (the default) you cannot use IPs on the primary interface -- even though this config specifically requests that all pod subnets/securitygroups match. Wheras if you have it set to true (meaning you don't want them to match) then you *can* allocate IPs on the primary interface -- which would mean you'd get inconsistent subnet/securitygroups.

This seems like a simple case of a missing `!`?

jacksontj:
Fixes #723

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/ipamd/ipamd.go
+++ b/ipamd/ipamd.go
@@ -679,7 +679,7 @@ func (c *IPAMContext) tryAssignIPs() (increasedPool bool, err error) {
 	}
 
 	// Find an ENI where we can add more IPs
-	eni := c.dataStore.GetENINeedsIP(c.maxIPsPerENI, c.useCustomNetworking)
+	eni := c.dataStore.GetENINeedsIP(c.maxIPsPerENI, !c.useCustomNetworking)
 	if eni != nil && len(eni.IPv4Addresses) < c.maxIPsPerENI {
 		currentNumberOfAllocatedIPs := len(eni.IPv4Addresses)
 		// Try to allocate all available IPs for this ENI


mogren:
Issue closed

Issue #725, opened 2019-11-16
ipamd uses and deletes user-managed ENIs


euank:
We have a use-case that I can't figure out how to get working safely while using the aws-vpc-cni-eks-k8s-eni plugin.

### Issue

We have a service that we wish to run in host networking and have listen on a public IP with a specific security group. Previously (before using this project), we managed this by associating an ENI with a public IP and the right security group, and that worked well. This service can't be modeled adequately with k8s's networking primitives, so it's necessary we keep doing this same thing, even in the k8s world.

If we try to do the same thing while using the aws-vpc-cni-eks-k8s-eni plugin, we find that this ENI has its secondary IPs used for other pods (not great), and if we drain pods it sometimes is detached from the node and deleted, even though we were the ones to create it (even worse).

### Proposed solution

The solution that makes sense to me is to allow a user of this plugin to denote certain ENIs as being "unmanaged" by this plugin. It then becomes the user's responsibility to configure that ENI and use it as they will.

Is there already a way to accomplish this somehow? Is this a use-case that is in-scope of this project to support?

jaypipes:
I like the concept of being able to reserve ENIs (or elastic IPs, frankly) for a workload that is either not managed by Kubernetes (but running on a k8s worker node) or using host networking and/or specialized offloads or networking capabilities.

I'll take a gander at your PR. Looks promising from an initial skim.

mogren:
@euank You are correct in that this CNI assumes that it owns all ENIs on the node, and currently there is no way for the user to configure this. Tagging the ENIs that you want to be left alone sounds good. I'll take a look at the PR, thanks a lot for working on this!

My only concern so far is that just like when using custom networking, the `--max-pods` value in the default EKS AMI will no longer be correct, since it assumes all ENIs are available and all secondary IPs on them can be used for pods. Also, does this work correctly if the primary ENI is tagged? The CNI will use the private IP of the first ENI for doing SNAT of traffic going outside the cluster.

euank:
> My only concern so far is that just like when using custom networking, the --max-pods value in the default EKS AMI will no longer be correct

Indeed. I think it's up to the user to configure that if they plan to reserve ENIs unless we can think of a better option.

> Also, does this work correctly if the primary ENI is tagged?

I don't think it would, but I think it's reasonable to document the primary ENI for the instance as being special (it is used for the k8s nodeIP after all).

[...]

PR #726, opened 2019-11-16
 Add support for a 'no manage' tag

euank:
One possible solution for #725 

This tag results in an ENI being unmanaged by the ipamd plugin. This
allows someone to create an ENI, associate it with an instance, and then
use eks per normal while also using that ENI for whatever.

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,17 +375,17 @@ This plugin interacts with the following tags on ENIs:
 * `node.k8s.amazonaws.com/instance_id`
 * `node.k8s.amazonaws.com/no_manage`.
 
-### Cluster Name tag
+#### Cluster Name tag
 
 The tag `cluster.k8s.amazonaws.com/name` will be set to the cluster name of the
 aws-node daemonset which created the ENI.
 
-### Instance ID tag
+#### Instance ID tag
 
 The tag `node.k8s.amazonaws.com/instance_id` will be set to the instance ID of
 the aws-node instance that allocated this ENI.
 
-### No Manage tag
+#### No Manage tag
 
 The tag `node.k8s.amazonaws.com/no_manage` is read by the aws-node daemonset to
 determine whether an ENI attached to the machine should not be configured or
--- a/README.md
+++ b/README.md
@@ -373,7 +373,7 @@ This plugin interacts with the following tags on ENIs:
 
 * `cluster.k8s.amazonaws.com/name`
 * `node.k8s.amazonaws.com/instance_id`
-* `node.k8s.amazonaws.com/no_manage`.
+* `node.k8s.amazonaws.com/no_manage`
 
 #### Cluster Name tag
 
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ The default manifest expects `--cni-conf-dir=/etc/cni/net.d` and `--cni-bin-dir=
 
 L-IPAM requires following [IAM policy](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies.html):
 
-```      
+```
  {
      "Effect": "Allow",
      "Action": [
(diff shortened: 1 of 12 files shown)


ipamd/ipamd.go
```diff
@@ -305,17 +313,9 @@ func (c *IPAMContext) nodeInit() error {
...
-			err = c.setupENI(eni.ENIID, eni)
-			if retry > maxRetryCheckENI {
-				log.Errorf("Unable to discover attached IPs for ENI from metadata service")
-				ipamdErrInc("waitENIAttachedMaxRetryExceeded")
-				break
-			}
-
-			if err != nil {
+		setup := false
+		for retry := 0; retry <= maxRetryCheckENI; retry++ {
```
jaypipes: again, try to keep style-only changes like this in a separate PR. this change isn't critical to the new functionality being added and while I prefer your code over the original style, it's tougher for reviewers to review the functional changes added in this PR from the style-only changes you've made.

ipamd/ipamd.go
```diff
@@ -129,6 +133,12 @@ var (
 			Help: "The maximum number of ENIs that can be attached to the instance",
 		},
 	)
+	reservedENIs = prometheus.NewGauge(
```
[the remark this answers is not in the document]
euank: I agree that the naming is a bit inconsistent, and also that it's of dubious value in the first place.
I've removed it, but kept the change that `ipMax` refers to the maximum ips available taking into account unmanaged ENIs.
I'm also not sure if ipMax should be changed when an unmanaged ENI is noticed or not.
Do you know how the ipMax metric is typically used right now? I think that might inform what we do here.

euank:
Thanks for the review. I've removed the two small bits of refactoring that didn't relate to this change. I skimmed through the rest of the diff and didn't see any other unrelated changes, but if there are any, let me know and I can rebase them out too.

ipamd/ipamd.go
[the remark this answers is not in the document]
jaypipes: I actually don't know whether/how the ip_max value is currently used, no. :man_shrugging: Perfectly fine with me to keep using it and adapt for the unmanaged ENIs, though.

pkg/networkutils/mocks/network_mocks.go
```diff
@@ -133,15 +145,3 @@ func (m *MockNetworkAPIs) UseExternalSNAT() bool {
 func (mr *MockNetworkAPIsMockRecorder) UseExternalSNAT() *gomock.Call {
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "UseExternalSNAT", reflect.TypeOf((*MockNetworkAPIs)(nil).UseExternalSNAT))
 }
-
```
jaypipes: AFAICT, these mocks weren't changed in content, right? Just moved for some reason?

pkg/netlinkwrapper/mocks/netlinkwrapper_mocks.go
```diff
@@ -231,18 +231,6 @@ func (mr *MockNetLinkMockRecorder) RouteAdd(arg0 interface{}) *gomock.Call {
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "RouteAdd", reflect.TypeOf((*MockNetLink)(nil).RouteAdd), arg0)
 }
 
-// RouteReplace mocks base method
```
jaypipes: AFAICT, these mocks were not changed, just moved within this file... correct?

pkg/networkutils/mocks/network_mocks.go
euank: Yeah. I checked for a makefile target or such that would install the right version of `mockgen`, but since I didn't see one I just went with the latest, and the latest version seems to generate slightly different code.

euank:
Thanks for the review. I've added a readme section which documents the existing eni tags and this new one.

README.md
```diff
@@ -343,6 +343,49 @@ Default: `eni`
...
+
+### ENI tags related to Allocation
+
+This plugin interacts with the following tags on ENIs:
+
+* `cluster.k8s.amazonaws.com/name`
+* `node.k8s.amazonaws.com/instance_id`
+* `node.k8s.amazonaws.com/no_manage`.
+
+### Cluster Name tag
```
jaypipes: This should be a level 4 heading (add another pound sign) :)  Same for the "Instance ID tag" and "No Manage tag" headings below.

euank: Indeed, thanks! Updated to an h4

mogren:
Hi @euank, thanks, this looks like a great feature to add. Also, thanks a lot for the documentation updates. 

I have one question though, it looks like `datastore.GetENINeedsIP()` that is called in `tryAssignIPs()` could return the ENI that is marked with a `no_manage` tag. That would result in secondary IPs being attached to that ENI and the total count updated. I assume that's not what we want is it?

README.md
```diff
@@ -343,6 +343,49 @@ Default: `eni`
...
+
+Specifies the cluster name to tag allocated ENIs with. See the "Cluster Name tag" section below.
+
+### ENI tags related to Allocation
+
+This plugin interacts with the following tags on ENIs:
+
+* `cluster.k8s.amazonaws.com/name`
+* `node.k8s.amazonaws.com/instance_id`
+* `node.k8s.amazonaws.com/no_manage`.
```
mogren: Nit: Remove the `.`

euank:
@mogren That only returns ENIs in the datastore's pool. The two places ENIs are added to that pool is `nodeInit` and `nodeIPPoolReconcile`, and in both of those places I've added code to filter out `no_manage` tagged ENIs.

The idea is such ENIs never end up in the datastore's pool in the first place.

It is possible that someone could tag an ENI after already attaching it, at which point one of those two functions might have added it to the pool, but the intent is that anyone using this will tag the ENI before attaching it to the instance to ensure that doesn't happen.

Are there other paths by which an ENI can end up in the datastore? Is there some other case you're thinking of?

mogren:
@euank Ah, you're right. It should never be in the data store at all. 

One final concern though. If we don't support tagging existing ENIs, maybe we should store the `numUnmanaged` count in the context so that we can use it later? 

For example, if we try to run more pods than can fit on the ENIs we have the function `nodeIPPoolTooLow()` will always count against the instance max ENI limit, and `increaseIPPool()` will be called on every reconcile loop even if we can't attach any more ENIs. 

Another option would be to update the [ENI count check on line 607](https://github.com/aws/amazon-vpc-cni-k8s/pull/726/files#diff-3fbbb0bd392fd47bd85a56d6221de538R606-R608). Just trying to avoid calling `tryAllocateENI()` when we are already at the maximum ENI count for the instance.

jaypipes:
> One final concern though. If we don't support tagging existing ENIs, maybe we should store the `numUnmanaged` count in the context so that we can use it later?

++

Originally @euank had created a num_reserved metric gauge and I'd said that I didn't see much value in that as a gauge. But I can see it being useful as a field in the context.

euank:
Yeah, that makes sense to me. I'll give it another look through and try to figure out what makes sense.

One thing that has me wary here is that I think the user has to do manual configuration anyway, so if setting `--max-pods` and `MAX_ENI=$numnode-$numUnmanaged` fixes it without any code changes, that might be a reasonable option. If we do try and do something a bit smarter, we'll have to document how it interacts with that option clearly.

euank:
I updated one check to make it avoid trying and failing to allocate an ENI in the common case.

It will still think it's short IPs and complain though I believe.

After looking through the code agian, I think the best choice is to rely on the user setting `MAX_ENI`, and the readme already does suggest that.

If you can think of a more elegant way to handle this, I'd be happy to hear it.

PR #727, opened 2019-11-17
use init container for CNI setup

jaypipes:
Instead of using a shell script as the entrypoint for the CNI image and
relying on that shell script to copy the CNI plugin binary to a host
volume, perform those steps using an initContainer and set the CNI
aws-k8s-agent binary to be the CNI image's CMD.

This has the benefit of not embedding setup steps in either a shell
script that needs to be copied into the target image /app directory nor
hard-coded into the aws-k8s-agent's main.go file (which was performing
the copy of the CNI plugin binary to a host volume).

See discussion in
https://github.com/aws/amazon-vpc-cni-k8s/pull/706#issuecomment-552379546
for more details.


mogren:
I like the idea of having an init container, but I would also like to have the config file copied only once ipamd is up and running, so that we know there are at least a few IPs available in the subnet where the node is launched.

jaypipes:
Will need to reevaluate this one. Closing and going with a different approach.

jaypipes:
#735 contains my new approach...

PR #735, opened 2019-11-26
copy CNI plugin and config in entrypoint not agent

jaypipes:
Replaces the hard-coded copying of the CNI plugin binary and
configuration file from the main.go of the aws-k8s-agent (the IPAMd
daemon) into a new entrypoint.sh script that backgrounds the IPAM
daemon, waits for it to listen on port 50051 and then proceeds to copy
the CNI plugin binaries and configuration files into the well-known
directory for Kubelet to see them as ready.

Issue #706 


scripts/dockerfiles/Dockerfile.release
```diff
@@ -21,6 +21,7 @@ FROM amazonlinux:2
 RUN yum update -y && \
     yum install -y iptables && \
     yum install -y iproute && \
+    yum install -y nmap && \
```
nithu0115: Instead of installing nmap package, probably, we check if gRPC is listening on the port by writing to it like below: 
`(echo >/dev/tcp/localhost/50051) &>/dev/null`

This is a lightweight alternative to using lsof/nc/netcat. wget/curl doesn't work here because it is a TCP connection, not an HTTP.

scripts/entrypoint.sh
```diff
@@ -0,0 +1,58 @@
...
+# turn on bash's job control
+set -m
+
+echo -n "starting IPAM daemon in background ... "
+./app/aws-k8s-agent &
+echo "ok."
+
+echo -n "checking for IPAM connectivity ..."
+
+while ! nc -z localhost 50051 </dev/null;
```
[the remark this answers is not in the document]
jaypipes: Unless I'm mistaken, the `nc -z localhost 50051` call will not return 0 until the aws-k8s-agent has fully started up and is successfully able to contact the k8s API and successfully able to contact IMDS to get the local host networking information.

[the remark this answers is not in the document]
jaypipes: I can add a timeout/short-circuit, though, no problem

scripts/dockerfiles/Dockerfile.release
jaypipes: Yes, I thought about that. The issue with that is we'd need to mount yet another host-path volume into this container.

jaypipes: @mogren what's your thought on this one? prefer another host-path volume mount versus installing nmap?

jaypipes: We could use the `app/grpc_health_probe` binary instead!

config/v1.5/aws-k8s-cni.yaml
```diff
@@ -98,6 +98,10 @@ spec:
           env:
             - name: AWS_VPC_K8S_CNI_LOGLEVEL
               value: DEBUG
+            - name: AWS_VPC_K8S_CNI_VETHPREFIX
```
nithu0115: IMO, I would like to move these variables into a configMap so that they persisted upon upgrades. May be a separate PR and add //To do here  ?

scripts/entrypoint.sh
```diff
@@ -0,0 +1,73 @@
...
+        if /app/grpc_health_probe -addr 127.0.0.1:50051 </dev/null; then
+            return 0
+        fi
+    done
+    return 1
+}
+
+echo -n "starting IPAM daemon in background ... "
+./app/aws-k8s-agent &
+echo "ok."
```
nithu0115: Probably, to something more meaning full message sed s/ok/"Started IPamD and back grounded the process"/g

jaypipes: I use the `echo -n` variant on line 41, then execute something, then `echo "ok"` after successful execution. This results in the following on successful execution:

```
starting IPAM daemon in background ... ok.
```

config/v1.5/aws-k8s-cni.yaml
jaypipes: ack. good idea.

scripts/entrypoint.sh
nithu0115: Thanks for clarifying @jaypipes

nithu0115:
LGTM! Thanks @jaypipes 

Between, what are your thoughts on adding `sleep 10` for IPamD daemon to pre-warm/reconcile before copying the net plugin binaries to avoid kubelet errors:`failed to assign an IP address to container` ?

jaypipes:
> LGTM! Thanks @jaypipes
> 
> Between, what are your thoughts on adding `sleep 10` for IPamD daemon to pre-warm/reconcile before copying the net plugin binaries to avoid kubelet errors:`failed to assign an IP address to container` ?

For my thoughts on that, see https://github.com/aws/amazon-vpc-cni-k8s/pull/738/files :) Basically, I believe we should get rid of the async loading and just make the `ipamd.NewController` function return a properly-constructed object with a populated SharedInformer cache and don't use a separate goroutine for the cache-loading operation.

mogren:
@jaypipes Yes, starting up synchronously sounds a lot cleaner to me.

scripts/entrypoint.sh
```diff
@@ -0,0 +1,73 @@
...
+
+echo " ok."
+
+if [[ -f /host/etc/cni/net.d/aws.conf ]]; then
+    rm /host/etc/cni/net.d/aws.conf
+fi
+
+# bring the aws-k8s-agent process back into the foreground
+echo -n "foregrounding IPAM daemon ... "
+fg %1
```
mogren: Nit: Normally I would store the job ID when planning to move a background to to foreground again, but not really needed in this script.

scripts/entrypoint.sh
```diff
@@ -0,0 +1,73 @@
...
+# service. This is why this entrypoint script exists; we start the IPAM daemon
+# and wait until we know it is up and running successfully before copying the
+# CNI plugin binary and its configuration file to the well-known directory that
+# Kubelet looks in.
+
+# turn on bash's job control
+set -m
+
+# Checks for IPAM connectivity on localhost port 50051, retrying connectivity
+# check with a timeout of 36 seconds
```
mogren: Linearly increasing backoff! 😄

scripts/entrypoint.sh
```diff
@@ -0,0 +1,73 @@
...
+echo "ok."
+
+echo -n "copying CNI plugin binaries and config files ... "
+
+sed -i s/__VETHPREFIX__/"${AWS_VPC_K8S_CNI_VETHPREFIX:-"eni"}"/g /app/10-aws.conflist
+sed -i s/__MTU__/"${AWS_VPC_ENI_MTU:-"9001"}"/g /app/10-aws.conflist
+mv /app/10-aws.conflist /host/etc/cni/net.d/
+mv /app/portmap /host/opt/cni/bin/
+mv /app/aws-cni /host/opt/cni/bin/
+mv /app/aws-cni-support.sh /host/opt/cni/bin/
```
mogren: Not sure if there is any actual risk of a race condition here, but I think we should copy the config file after the cni binary has been moved.

Also, will using `cp` or `mv` make any difference? Isn't the `/app` folder read only in the image, so making a copy would not require an overlay update?

jaypipes: Good point on the overlay update. Will modify.

Issue #711, opened 2019-11-08
Duplicate IP Addresses for pods w/o host network


uruddarraju:
We are running a 1.12.6 cluster provisioned with Kops.

Networking setup: aws-k8s-cni with calico for network policies
Using: `602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.5.2`

We saw a few networking issues for pods created in our integration testing cluster. This cluster experiences a lot of churn where we consistently spin up 100s of pods and delete them periodically. We observed a few pods experiencing some network issues and upon digging a little more, we found the following:

```
admin@ip-10-20-89-225:~$ kubectl get pods --all-namespaces -o wide | grep Running | awk '{print $7}' |  sort | uniq -c | grep "   2"
      2 10.20.106.106
      2 10.20.126.85
      2 10.20.33.133
      2 10.20.35.126
....<10s of more entries>
```

Trying to find if these pods are using host network,
```
admin@ip-10-20-89-225:~$ kubectl get pods --all-namespaces -o wide | grep 10.20.72.196
uday                       test-data-ui-6644c8f79-k8g5v                                               0/1     Running             1          9d      10.20.72.196    ip-10-20-75-71.us-west-2.compute.internal     <none>
marco                       mock-server                                                              1/1     Running             0          25m     10.20.72.196    ip-10-20-75-71.us-west-2.compute.internal     <none>
```

And you can see both those pods using the ip `10.20.72.196`

We cleaned up the pods experiencing these duplicate IP issue a few times and that does not seem to solve the problem(which makes sense as we still dont know the root cause)

One common scenario we observer in all cases above, atleast one pod has a restart count > 0 or they are in CrashLoopBackoff/Error/Completed states.

Ideally, the cni should only be reclaiming IPs of pods not running that have a restartPolicy of Never. I am not sure if this is the behavior today.

mogren:
Hi @uruddarraju, the CNI will not reuse IPs directly, but after a [1 minute cool down](https://github.com/aws/amazon-vpc-cni-k8s/blob/1726afd4395c6c02f5caab09daf56b447719bb31/ipamd/ipamd.go#L55-L57) they will be available to be assigned to pods again.

Do you use the default configuration? What kind of nodes are you running on, and how many pods per node? If you have a lot of churn, it helps to pre-allocate the IPs.

Also, how come you use v1.5.2? I'd recommend upgrading to v1.5.3.

[...]

PR #738, opened 2019-12-02
Wait until the discover controller cache is synced

uruddarraju:
https://github.com/aws/amazon-vpc-cni-k8s/issues/711

This change makes sure we don't have a race condition when the ipamD agent initially starts. In the current state, without these changes, there is a possibility that ipamD would start assigning IPs to pods before its internal state is synced

### DIFF
--- a/main.go
+++ b/main.go
@@ -50,7 +50,12 @@ func _main() int {
 		return 1
 	}
 
-	discoverController := k8sapi.NewController(kubeClient)
+	discoverController, err := k8sapi.NewController(kubeClient)
+	if err != nil {
+		log.Errorf("Failed to create new Discover Controller: %v", err)
+		return 1
+	}
+
 	go discoverController.DiscoverK8SPods()
 
 	eniConfigController := eniconfig.NewENIConfigController()
@@ -59,7 +64,6 @@ func _main() int {
 	}
 
 	ipamContext, err := ipamd.New(discoverController, eniConfigController)
-
 	if err != nil {
 		log.Errorf("Initialization failure: %v", err)
 		return 1
--- a/pkg/k8sapi/discovery.go
+++ b/pkg/k8sapi/discovery.go
@@ -68,15 +68,31 @@ type Controller struct {
 	controller *controller
 	kubeClient kubernetes.Interface
 	myNodeName string
-	synced     bool
 }
 
 // NewController creates a new DiscoveryController
-func NewController(clientset kubernetes.Interface) *Controller {
-	return &Controller{kubeClient: clientset,
+func NewController(clientset kubernetes.Interface) (*Controller, error) {
+	stopCh := wait.NeverStop
+
+	d := &Controller{kubeClient: clientset,
 		myNodeName: os.Getenv("MY_NODE_NAME"),
 		cniPods:    make(map[string]string),
 		workerPods: make(map[string]*K8SPodInfo)}
+
+	log.Info("Starting Pod informer")
+
+	go d.controller.informer.Run(stopCh)
+
+	log.Info("Waiting for controller cache sync")
+	// Wait for all involved caches to be synced, before processing items from the queue is started
+	if !cache.WaitForCacheSync(stopCh, d.controller.informer.HasSynced) {
+		log.Error("Timed out waiting for caches to sync!")
+		return nil, fmt.Errorf("timed out waiting for caches to sync")
+	}
+
+	log.Info("Synced successfully with APIServer")
+
+	return d, nil
 }
 
 // CreateKubeClient creates a k8s client
@@ -166,11 +182,6 @@ func (d *Controller) DiscoverK8SPods() {
 // K8SGetLocalPodIPs return the list of pods running on the local nodes
 func (d *Controller) K8SGetLocalPodIPs() ([]*K8SPodInfo, error) {
 	var localPods []*K8SPodInfo
-
-	if !d.synced {
-		log.Info("GetLocalPods: informer not synced yet")
-		return nil, ErrInformerNotSynced
-	}
 
 	log.Debug("GetLocalPods start ...")
 	d.workerPodsLock.Lock()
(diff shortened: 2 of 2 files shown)


pkg/k8sapi/discovery.go
```diff
@@ -185,6 +185,11 @@ func (d *Controller) K8SGetLocalPodIPs() ([]*K8SPodInfo, error) {
 	return localPods, nil
 }
 
+func (d *Controller) WaitUntilReady() {
+	stopCh := make(chan struct{})
+	cache.WaitForCacheSync(stopCh, d.controller.informer.HasSynced)
```
mogren: This sounds like something we should do, but the go docs say `WaitForNamedCacheSync` should be preferred. Could you change to used that function instead? Also, I'll try to test this some before it gets merged.

main.go
```diff
@@ -53,6 +53,10 @@ func _main() int {
 	discoverController := k8sapi.NewController(kubeClient)
 	go discoverController.DiscoverK8SPods()
```
jaypipes: If you want to wait for pods to be discovered before moving on, then why not just move the `DiscoverK8SPods()` call into the `NewController` function and perform this initialization in the controller's construction instead of in a separate goroutine here? Wouldn't that be simpler?

pkg/k8sapi/discovery.go
jaypipes: What about the `Controller.synced` field? In `Controller.run`, the `synced` field is set to true:

```go
	go d.controller.informer.Run(stopCh)
	log.Info("Waiting for controller cache sync")
	// Wait for all involved caches to be synced, before processing items from the queue is started
	if !cache.WaitForCacheSync(stopCh, d.controller.informer.HasSynced) {
		log.Error("Timed out waiting for caches to sync!")
		runtime.HandleError(fmt.Errorf("timed out waiting for caches to sync"))
		return
	}

	log.Info("Synced successfully with APIServer")
	d.synced = true
```

It seems to me that the simplest solution to this problem is to constrain all this setup/cache-prep logic to the construction of the `Controller` object itself by creating the `SharedInformer` structs in the `NewController` function and then returning the ready-to-run `Controller` object. You could remove the `Controller.synced` field entirely if you did this, and the code would become a whole lot easier to read.

main.go
uruddarraju: I thought briefly about this before I made this change. This is used for DiscoveryController today but might be a too strong assumption to say this would be the case going forward. I see this logic to be equivalent to an initialization logic, why start the RPC server or any other process when the required initialization is not done.

uruddarraju:
@mogren @jaypipes Thanks for the feedback. That makes sense. I made some changes to address the feedback. Let me know if this looks better now.

pkg/k8sapi/discovery.go
```diff
@@ -68,15 +68,31 @@ type Controller struct {
...
 		cniPods:    make(map[string]string),
 		workerPods: make(map[string]*K8SPodInfo)}
+
+	log.Info("Starting Pod informer")
+
+	go d.controller.informer.Run(stopCh)
+
+	log.Info("Waiting for controller cache sync")
+	// Wait for all involved caches to be synced, before processing items from the queue is started
+	if !cache.WaitForCacheSync(stopCh, d.controller.informer.HasSynced) {
```
jaypipes: On a future PR (not this one), we should use the `WaitForNamedCacheSync` method that @mogren alluded to in an earlier comment:

https://github.com/kubernetes/client-go/blob/bc9b51d240b2f5000adc2fa96231ab219ab0d702/tools/cache/shared_informer.go#L196

Which would basically remove the need for the log lines here.

pkg/k8sapi/discovery.go
```diff
@@ -68,15 +68,31 @@ type Controller struct {
 	controller *controller
 	kubeClient kubernetes.Interface
 	myNodeName string
-	synced     bool
 }
 
 // NewController creates a new DiscoveryController
```
jaypipes: Please update the docstring comment here to indicate that the function constructs the caches (and connects to the k8s API server to do so) before returning a fully-constructed controller object.

main.go
```diff
@@ -50,7 +50,12 @@ func _main() int {
...
 	}
 
-	discoverController := k8sapi.NewController(kubeClient)
+	discoverController, err := k8sapi.NewController(kubeClient)
+	if err != nil {
+		log.Errorf("Failed to create new Discover Controller: %v", err)
+		return 1
+	}
+
 	go discoverController.DiscoverK8SPods()
```
jaypipes: This needs to go inside the `NewController()` function and be synchronously executed (see below)

pkg/k8sapi/discovery.go
```diff
@@ -68,15 +68,31 @@ type Controller struct {
...
+	stopCh := wait.NeverStop
+
+	d := &Controller{kubeClient: clientset,
 		myNodeName: os.Getenv("MY_NODE_NAME"),
 		cniPods:    make(map[string]string),
 		workerPods: make(map[string]*K8SPodInfo)}
+
+	log.Info("Starting Pod informer")
+
+	go d.controller.informer.Run(stopCh)
```
jaypipes: `d.controller` will be `nil` here, since `d.controller` is created in the `DiscoverK8sPods()` method:

https://github.com/aws/amazon-vpc-cni-k8s/blob/7c6dec755c61703b966799cc5af26012044a9410/pkg/k8sapi/discovery.go#L155

What you need to do is basically get rid of the `DiscoverK8SPods()` method entirely, and move the code that is in there into the `NewController()` function.

The only reason that the `DiscoverK8SPods()` method even exists as a separate struct method apparently is to "start the controller" by doing:

```go
select {}
```

but that's really kinda hacky IMHO. A cleaner solution would be to simply move all this setup code into `NewController()`:

```go
	// create the pod watcher
	podListWatcher := cache.NewListWatchFromClient(d.kubeClient.CoreV1().RESTClient(), "pods", metav1.NamespaceAll, fields.OneTermEqualSelector("spec.nodeName", d.myNodeName))

	// create the workqueue
	queue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())

	// Bind the workqueue to a cache with the help of an informer. This way we make sure that
	// whenever the cache is updated, the pod key is added to the workqueue.
	// Note that when we finally process the item from the workqueue, we might see a newer version
	// of the Pod than the version which was responsible for triggering the update.
	indexer, informer := cache.NewIndexerInformer(podListWatcher, &v1.Pod{}, 0, cache.ResourceEventHandlerFuncs{
		AddFunc: func(obj interface{}) {
			key, err := cache.MetaNamespaceKeyFunc(obj)
			if err == nil {
				queue.Add(key)
			}
		},
		UpdateFunc: func(old interface{}, new interface{}) {
			key, err := cache.MetaNamespaceKeyFunc(new)
			if err == nil {
				queue.Add(key)
			}
		},
		DeleteFunc: func(obj interface{}) {
			// IndexerInformer uses a delta queue, therefore for deletes we have to use this
			// key function.
			key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
			if err == nil {
				queue.Add(key)
			}
		},
	}, cache.Indexers{})

	d.controller = newController(queue, indexer, informer)
```

and move this code:

```go
	// Now let's start the controller
	stop := make(chan struct{})
	defer close(stop)
	go d.run(1, stop)

	// Wait forever
	select {}
```

to a separate `Controller.Listen()` method that you can call after returning successfully from `NewController()`

adammw:
is this still required after 1.6 rewrites?

mogren:
Replaced by changes in #972

Issue #748, opened 2019-12-10
scripts/entrypoint.sh executable check not working


jaypipes:
There is a typo in the executable check in `scripts/entrypoint.sh`:

https://github.com/aws/amazon-vpc-cni-k8s/blob/88b52cc90d900eeca757ceee5d66860ae7b31f5c/scripts/entrypoint.sh#L32

Luckily, the `[ ! -x $(command -v grp_health_probe) ]` check for an executable isn't working properly (always returns 0 I guess) so this is not preventing the entrypoint.sh script from actually working.

PR #749, opened 2019-12-10
Use simple filename check in entrypoint script

jaypipes:
The `[ ! -x $(command -v grp_health_probe) ]` check contained a typo
(grp instead of grpc) but the `command -v` check was always returning 0
for some reason. This meant the executable check wasn't working but the
entrypoint script was executing properly anyway.

This patch fixes the typo and replaces the command -v check with a
simpler bash `if [ -f $FILE ]` check instead.

Also adds a DOCKER_ARGS envvar to facilitate local testing in
environments that don't handle Docker non-host networking well.

Closes Issue #748

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,6 +40,7 @@ require (
 	golang.org/x/net v0.0.0-20191004110552-13f9640d40b9
 	golang.org/x/sys v0.0.0-20190826190057-c7b8b68b1456
 	golang.org/x/time v0.0.0-20180412165947-fbb02b2291d2 // indirect
+	google.golang.org/appengine v1.4.0 // indirect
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -40,6 +40,7 @@ require (
 	golang.org/x/net v0.0.0-20191004110552-13f9640d40b9
 	golang.org/x/sys v0.0.0-20190826190057-c7b8b68b1456
 	golang.org/x/time v0.0.0-20180412165947-fbb02b2291d2 // indirect
+	google.golang.org/appengine v1.4.0 // indirect
```
mogren: Hmm.. `go mod tidy`...

PR #751, opened 2019-12-11
Use dockershim.sock for CRI

mogren:
*Issue #714 related*

*Description of changes:*
In the original commit, we only checked `/var/run/cri.sock` to get the Sandbox ID. This works fine for CRI-O or containerd with the CRI plugin, as long as the socket is mounted correctly, but when using docker we need the `/var/run/dockershim.sock` created by Kubelet, since it talks gRPC with the CRI API, while `/var/run/docker.sock` is http only.

* Use `dockershim.sock` provided by kubelet to get a gRPC CRI to call. (Follow up in #752)
* Fix entrypoint script
* ~~Fix delete check of pod that failed to start~~ (Merged #750 first)
* Log empty sandbox id using quotes.

### DIFF
--- a/config/v1.5/aws-k8s-cni.yaml
+++ b/config/v1.5/aws-k8s-cni.yaml
@@ -101,7 +101,7 @@ spec:
             - name: AWS_VPC_K8S_CNI_VETHPREFIX
               value: eni
             - name: AWS_VPC_K8S_CNI_MTU
-              value: 9001
+              value: "9001"
             - name: MY_NODE_NAME
               valueFrom:
                 fieldRef:
@@ -120,6 +120,8 @@ spec:
               name: log-dir
             - mountPath: /var/run/docker.sock
               name: dockersock
+            - mountPath: /var/run/dockershim.sock
+              name: dockershim
       volumes:
         - name: cni-bin-dir
           hostPath:
@@ -133,6 +135,9 @@ spec:
         - name: dockersock
           hostPath:
             path: /var/run/docker.sock
+        - name: dockershim
+          hostPath:
+            path: /var/run/dockershim.sock
 
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
--- a/go.mod
+++ b/go.mod
@@ -23,8 +23,8 @@ require (
 	github.com/json-iterator/go v1.1.5 // indirect
 	github.com/matttproud/golang_protobuf_extensions v1.0.1 // indirect
 	github.com/modern-go/reflect2 v0.0.0-20180701023420-4b7aa43c6742 // indirect
-	github.com/onsi/ginkgo v1.8.0
-	github.com/onsi/gomega v1.5.0
+	github.com/onsi/ginkgo v1.8.0 // indirect
+	github.com/onsi/gomega v1.5.0 // indirect
 	github.com/operator-framework/operator-sdk v0.0.7
 	github.com/peterbourgon/diskv v2.0.1+incompatible // indirect
 	github.com/pkg/errors v0.8.1
@@ -40,7 +40,6 @@ require (
 	golang.org/x/net v0.0.0-20191004110552-13f9640d40b9
 	golang.org/x/sys v0.0.0-20190826190057-c7b8b68b1456
 	golang.org/x/time v0.0.0-20180412165947-fbb02b2291d2 // indirect
-	google.golang.org/appengine v1.4.0 // indirect
 	google.golang.org/grpc v1.23.1
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	k8s.io/api v0.0.0-20180712090710-2d6f90ab1293
--- a/ipamd/ipamd.go
+++ b/ipamd/ipamd.go
@@ -334,7 +334,6 @@ func (c *IPAMContext) nodeInit() error {
 		}
 	}
 	localPods, err := c.getLocalPodsWithRetry()
-	log.Debugf("getLocalPodsWithRetry() found %d local pods", len(localPods))
 	if err != nil {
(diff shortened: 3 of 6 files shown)


config/v1.5/aws-k8s-cni.yaml
```diff
@@ -101,7 +101,7 @@ spec:
             - name: AWS_VPC_K8S_CNI_VETHPREFIX
               value: eni
             - name: AWS_VPC_K8S_CNI_MTU
-              value: 9001
+              value: "9001"
```
jaypipes: interesting. this was a problem?

go.mod
```diff
@@ -40,7 +40,6 @@ require (
 	golang.org/x/net v0.0.0-20191004110552-13f9640d40b9
 	golang.org/x/sys v0.0.0-20190826190057-c7b8b68b1456
 	golang.org/x/time v0.0.0-20180412165947-fbb02b2291d2 // indirect
-	google.golang.org/appengine v1.4.0 // indirect
```
jaypipes: hmm. this is likely due to different Golang versions in our local development environments trampling on each other... My local go version is 1.12.7. I suspect you're on 1.13 and beyond?

ipamd/ipamd.go
```diff
@@ -334,14 +334,14 @@ func (c *IPAMContext) nodeInit() error {
...
 	localPods, err := c.getLocalPodsWithRetry()
-	log.Debugf("getLocalPodsWithRetry() found %d local pods", len(localPods))
 	if err != nil {
 		log.Warnf("During ipamd init, failed to get Pod information from Kubernetes API Server %v", err)
 		ipamdErrInc("nodeInitK8SGetLocalPodIPsFailed")
 		// This can happens when L-IPAMD starts before kubelet.
 		// TODO  need to add node health stats here
 		return errors.Wrap(err, "failed to get running pods!")
 	}
+	log.Debugf("getLocalPodsWithRetry() found %d local pods", len(localPods))
```
jaypipes: good move...

pkg/cri/cri.go
```diff
@@ -33,7 +34,12 @@ func New() *Client {
 }
 
 func (c *Client) GetRunningPodSandboxes() (map[string]*SandboxInfo, error) {
-	conn, err := grpc.Dial(criSocketPath, grpc.WithInsecure())
+	socketPath := dockerSocketPath
+	if info, err := os.Stat("/var/run/cri.sock"); err == nil && !info.IsDir() {
+		socketPath = criSocketPath
+	}
+	log.Debugf("Getting running pod sandboxes from %q", socketPath)
+	conn, err := grpc.Dial(socketPath, grpc.WithInsecure())
```
jaypipes: I suppose we can turn this into a blocking call (WithBlock() with a backoff timeout) in a future commit...

scripts/entrypoint.sh
```diff
@@ -36,7 +36,7 @@ fi
 
 AGENT_LOG_PATH=${AGENT_LOG_PATH:-aws-k8s-agent.log}
 HOST_CNI_BIN_PATH=${HOST_CNI_BIN_PATH:-/host/opt/cni/bin}
-HOST_CNI_CONFDIR_PATH=${HOST_CNI_CONFDIR_PATH:-/host/opt/cni/net.d}
+HOST_CNI_CONFDIR_PATH=${HOST_CNI_CONFDIR_PATH:-/host/etc/cni/net.d}
```
jaypipes: gahhhhhh!

scripts/entrypoint.sh
```diff
@@ -84,8 +84,8 @@ fi
 
 # bring the aws-k8s-agent process back into the foreground
 echo "foregrounding IPAM daemon ... "
-fg %1 >/dev/null 2>&1 || $(echo "failed (process terminated)" && cat $AGENT_LOG_PATH && exit 1)
+fg %1 >/dev/null 2>&1 || $(echo "failed (process terminated)" && cat "$AGENT_LOG_PATH" && exit 1)
 
 # Best practice states we should send the container's CMD output to stdout, so
 # let's tee back up the log file into stdout
-cat $AGENT_LOG_PATH
+cat "$AGENT_LOG_PATH"
```
jaypipes: interesting that quote-encapping the variable in the entrypoint.sh script fixed anything here and above. I know it's safer to quote-encap vars but I'm curious whether this was an actual fix for a problem in the script.

config/v1.5/aws-k8s-cni.yaml
mogren: Yeah, for some reason I got an error applying the config without the quotes.

scripts/entrypoint.sh
mogren: It didn't, but we were mixing quoted and unquoted, so I just made it consistent.

drakedevel:
@mogren any reason not to just mount `dockershim.sock` as `cri.sock`? That's was the original idea, leaving it up to the DaemonSet to mount the appropriate path from the host (per https://github.com/aws/amazon-vpc-cni-k8s/pull/714#issuecomment-553764081). That was how all of my tests on Docker in EKS were done, as well.

I don't think there's anything broken by doing this instead, but it's the only place in the codebase where there's a special case for Docker, and it appears unnecessary :stuck_out_tongue:

mogren:
Heh, thanks @drakedevel I think you are right. I did try to use the `containerd` socket at first, but since the cri-plugin was not loaded, that didn't work. I then tried just using the default docker.sock, but that of course doesn't have the right API, so then I switched to the shim. I should have iterated some more and done that, mount it as `cri.sock`... Time for another follow up PR I guess.  😄

Issue #686, opened 2019-10-28
Automate conformance testing


jaypipes:
The EKS team currently runs the Sonobouy conformance test suite manually before releasing the CNI plugin. We want to both automate this process and place all test definitions into the public repo.

- [x] Create CircleCI plumbing (use eks-charts repo as example)
- [ ] Flesh out the following manual steps into a Bash script that is executed from the CircleCI YAML file
```
# modify security group and open lower ports
# egress on controlplane security group (1-65535)
# ingress on worker security group (1-65535)
# use c4.xlarge node as worker node

# get sonobuoy if you don't have it
go get -u -v github.com/heptio/sonobuoy

# Deploy a Sonobuoy pod to your cluster with:
sonobuoy run

#View actively running pods:
sonobuoy status 

#To inspect the logs:
sonobuoy logs

# Once sonobuoy status shows the run as completed, copy the output directory from the main Sonobuoy pod to a local directory:
sonobuoy retrieve .

# This copies a single .tar.gz snapshot from the Sonobuoy pod into your local . directory. Extract the contents into ./results with:
mkdir ./results; tar xzf *.tar.gz -C ./results
# NOTE: The two files required for submission are located in the tarball under plugins/e2e/results/{e2e.log,junit.xml}.

# To clean up Kubernetes objects created by Sonobuoy, run:
sonobuoy delete
```
- [ ] ensure successful Sonobuoy test run gates merges to master

Alternately, to avoid waiting for Sonobuoy logs, we could do this instead (suggested from @wongma7):

```
wget https://dl.k8s.io/v1.14.8/kubernetes-test.tar.gz
tar -xzf kubernetes-test.tar.gz
./kubernetes/platforms/darwin/amd64/e2e.test --kubeconfig=$KUBECONFIG --ginkgo.focus="Conformance"
````

**Comment by wongma7** [comment]
I don't mind sonobuoy, but if we can achieve the same thing using the e2e.test binary directly I would prefer that mostly because it's one less potential source of bugs to worry about.

For example, something like this should be possible and we'll get the test output in real-time straight to stdout rather than having to wait for `sonobuoy logs`
```
#wget https://dl.k8s.io/v1.14.8/kubernetes-test.tar.gz
#tar -xzf kubernetes-test.tar.gz
#./kubernetes/platforms/darwin/amd64/e2e.test --kubeconfig=$KUBECONFIG --ginkgo.focus="Conformance"
```

[...]

PR #766, opened 2019-12-19
integration tests running with updated aws-k8s-tester

jaypipes:
Checks that there is an ECR repo before we use aws-k8s-tester to actually
create the test cluster. In addition, trap errors in the main
run-integration-tests.sh script and deprovision cluster if an error
occurred after creating the cluster.

Adds some guardrails to the scripts/run-integration_tests.sh script by
checking (before creating an EKS cluster) that AWS credentials are
present and that the required ECR repository is readable.

The names of environment variables accepted by aws-k8s-tester changed
when the managed node group functionality was introduced. This commit
updates the integration testing scripts to call aws-k8s-tester (v.0.5.4
which is the release needed with the fix for aws/aws-k8s-tester#70) with
these updated environment variables.

We decrease the number of parallel builds of the echo job from 100 to
3 and the number of completions for that job from 1000 to 30. This
decreases the setup time of the cluster by about 10 minutes.

Finally, I added in a short-circuit to prevent double-deprovisioning of
the cluster if, say, a stacktrace occurred when running the
aws-k8s-tester tool.

Issue #686
Issue #784
Issue #786


scripts/run-integration-tests.sh
```diff
@@ -50,25 +83,16 @@ mkdir -p $TEST_DIR
...
-if [[ ! -e $TESTER_PATH ]]; then
-  mkdir -p $TESTER_DIR
-  echo "Downloading aws-k8s-tester from $TESTER_DOWNLOAD_URL to $TESTER_PATH"
-  curl -s -L -X GET $TESTER_DOWNLOAD_URL -o $TESTER_PATH
-  chmod +x $TESTER_PATH
-fi
-
 if [[ "$PROVISION" = true ]]; then
     up-test-cluster
+    __cluster_created=1
```
mogren: Nice!

scripts/run-integration-tests.sh
```diff
@@ -23,55 +42,62 @@ REPORT_DIR=${TEST_DIR}/report
...
+AWS_ECR_REPO_NAME=${AWS_ECR_REPO_NAME:-"amazon-k8s-cni"}
+IMAGE_NAME=${IMAGE_NAME:-"$AWS_ECR_REGISTRY/$AWS_ECR_REPO_NAME"}
+IMAGE_VERSION=${IMAGE_VERSION:-$(git describe --tags --always --dirty)}
+
+if [[ "$BUILD" = true ]]; then
+    # `aws ec2 get-login` returns a docker login string, which we eval here to
+    # login to the ECR registry
+    eval $(aws ecr get-login --region $AWS_REGION --no-include-email) >/dev/null 2>&1
+    ensure_ecr_repo "$AWS_ACCOUNT_ID" "$AWS_ECR_REPO_NAME"
+    make docker IMAGE=$IMAGE_NAME VERSION=$IMAGE_VERSION
```
SaranBalaji90: This doesn't pull source code from given tag(version) right?

jaypipes: No, it doesn't. It runs `docker build -t aws-vpc-cni:$IMAGE_VERSION`. If IMAGE_VERSION envvar has been overridden, we should probably set the BUILD envvar to false automatically.

jaypipes: Either that, or check that IMAGE_VERSION is overridden and if so, do a `git checkout $IMAGE_VERSION` here to switch the working tree to that git tag.

SaranBalaji90: using git-checkout makes sense. Can you help me understand how disabling BUILD will help? Do you mean to say, if CNI image is already available in my repo then disabling BUILD will help?

jaypipes: yes, exactly. if you've already pushed, say, a v1.5.5 CNI image built from that git tag/branch, then setting BUILD=false will prevent the run-integration-tests.sh script from `docker push`ing an image and the eks-k8s-tester binary will simply pull the image that was in the ECR repo for v1.5.5

SaranBalaji90: Ah ok makes sense. Thanks for clarifying.

jaypipes: In short, I need to make some changes to this script to facilitate the not-building-from-local-checkout scenarios...

scripts/run-integration-tests.sh
```diff
@@ -1,18 +1,37 @@
...
 K8S_VERSION=${K8S_VERSION:-1.14.6}
 PROVISION=${PROVISION:-true}
 DEPROVISION=${DEPROVISION:-true}
 BUILD=${BUILD:-true}
 
+__cluster_created=0
+__cluster_deprovisioned=0
+
+on_error() {
+    # Make sure we destroy any cluster that was created if we hit run into an
```
nckturner: Nit: if we run into

nckturner: Also, don't we want to leave behind test clusters on error for investigations in some cases?  For example if you are running the tests by hand?

jaypipes: I can add back in the `if [[ "$DEPROVISION" = true ]];` condition here, sure. This was added because running into  errors in running the integration tests resulted in me having to manually clean up the created test cluster (many times, and I got tired of the manual cleanup :) )

nckturner: Makes sense, I think there is still value in giving the runner the ability to choose.

jaypipes: I went ahead and pushed up a change that adds back the DEPROVISION = true check.

Issue #508, opened 2019-06-18
aws-cni on vanilla K8s 


tvalasek:
We build K8s clusters in AWS using CF/kubeadm from upstream vanilla K8s. For networking we use aws-cni plugin. Our out-of-the-box setup is 3x masters (etcd running inside) and 3x worker nodes.

AWS-cni plugin runs as deamonset, thus on all 6 members of a cluster (masters + workers).

Now, the behaviour I'm seeing is that aws-cni plugin does not differenciate masters from workers.

The result of that is (looking at the cni-metrics-helper stats) aws-cni on demand creates new ENIs and secondary IP address pool (warm pool) also on master nodes which by default have pod scheduling disabled (for obvious reasons). That leaves us with huge amount of warm pool unused IPs on master nodes that can't ever be allocated.

I believe aws-cni was primarily build for EKS (where control plane / master and etcd) are hidden from the EKS admin, but I wonder if aws-cni has feature to distinguish masters from workers (and thus different warm pool scheduling) for those of us who decided not to use EKS. (because of availability in regions)

Thanks

mogren:
Hi @tvalasek,

You are right that we have not yet optimized the plugin much for use outside of EKS, so there is more work to be done here. I think it sounds like a good idea to make the CNI more configurable in order work better on the masters. Do you have any more concrete suggestions for what configurations you would need?

tvalasek:
How I see it is something to what already has been done: https://github.com/aws/amazon-vpc-cni-k8s/pull/68

Ain't sure if is it part of aforementioned PR but for our use case I would like to have something like `max number of ENIs for given node that can be created` config option. (that way we could control how many warm pool IPs can be created for that specific node)

Secondly these config options would work globally on all members of a cluster (like it's now) or only work on nodes with either specific labels on K8s level (e.g. `node-role.kubernetes.io`) or specific tags on AWS EC2 level. I reckon the later sounds more like generic AWS approach, kinda like it.

Does it make sense?

mogren:
We do have the `MAX_ENI` setting already, but that would get applied to both master and worker nodes. 

In this case, I guess it would be better to have a way to tag master nodes, or that the CNI is aware of common taints like `node-role.kubernetes.io/master` and behave differently in that case.

danbeaulieu:
@mogren is creating separate daemonsets an option? One for control plane nodes with the right tolerations and selectors and one for non control plane nodes?

jaypipes:
[...]

PR #776, opened 2019-12-30
Update aws-k8s-cni.yaml

cheng0214:
*Issue #, if available:*
when i try to create kops cluster with vpc-cni network plugin,  the master node has these taints, so 
pod is not scheduled to master node.

*Description of changes:*
Add toleration to daemonset can fix this.

### DIFF
--- a/config/v1.5/aws-k8s-cni.yaml
+++ b/config/v1.5/aws-k8s-cni.yaml
@@ -79,7 +79,10 @@ spec:
       serviceAccountName: aws-node
       hostNetwork: true
       tolerations:
-        - operator: Exists
+        - key: node-role.kubernetes.io/master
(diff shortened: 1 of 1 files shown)


mogren:
@cheng0214 By default, the CNI will tolerate any taint, see the [official documentation](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/#concepts):

> An empty key with operator `Exists` matches all keys, values and effects which means this will tolerate everything.

That means the `aws-node` daemonset should schedule a pod on all nodes by default. For a related issue, see https://github.com/aws/amazon-vpc-cni-k8s/issues/508

cheng0214:
ok, I will close the PR. thanks you

Issue #777, opened 2019-12-30
vpc peering secondary eni pod choud not communicate crossing vpc.


phantooom:
```
                10.0.53.0/22                  172.16.12.0/22
pod1-eni0----node1---vpc1-------peer----------vpc2----node2--eni0 pod2
                                                           \-eni1 pod3
```

pod1 ping pod2 ok
pod1 ping pod3 failed

becasue of 
1536:   from 172.16.13.174 to 172.16.12.0/22 lookup 2
1536:   from 172.16.13.248 to 172.16.12.0/22 lookup 2
1536:   from 172.16.13.197 to 172.16.12.0/22 lookup 2
1536:   from 172.16.13.145 to 172.16.12.0/22 lookup 2
1536:   from 172.16.13.250 to 172.16.12.0/22 lookup 2

policy route define dst, i think dst shoud not set.

1536:   from 172.16.13.174  lookup 2
1536:   from 172.16.13.248  lookup 2
1536:   from 172.16.13.197  lookup 2
1536:   from 172.16.13.145  lookup 2
1536:   from 172.16.13.250  lookup 2

jaypipes:
This is not a bug, AFAICT. You need to use external SNAT option:

```
AWS_VPC_K8S_CNI_EXTERNALSNAT

Type: Boolean

Default: false

Specifies whether an external NAT gateway should be used to provide SNAT of secondary ENI IP addresses. If set to true, the SNAT iptables rule and off-VPC IP rule are not applied, and these rules are removed if they have already been applied. Disable SNAT if you need to allow inbound communication to your pods from external VPNs, direct connections, and external VPCs, and your pods do not need to access the Internet directly via an Internet Gateway. However, your nodes must be running in a private subnet and connected to the internet through an AWS NAT Gateway or another external NAT device.
```

jaypipes:
@phantooom it may be possible you were running into #737. Please check out that issue and perform one of the two workarounds suggested by @0xlen.

PR #778, opened 2019-12-30
fix vpc peering secondary eni pod communicate

phantooom:
*Issue #, if available:*
https://github.com/aws/amazon-vpc-cni-k8s/issues/777
*Description of changes:*
```
change policy route 

1536: from podIP to vpcCIDR lookup 2

to 

1536: from podIP lookup 2

reslove vpc peering secondary eni pod can not communicate.

```

### DIFF
--- a/plugins/routed-eni/driver/driver.go
+++ b/plugins/routed-eni/driver/driver.go
@@ -233,30 +233,22 @@ func setupNS(hostVethName string, contVethName string, netnsPath string, addr *n
 			}
 			log.Infof("Added rule priority %d from %s table %d", fromContainerRulePriority, addr.String(), table)
 		} else {
-			// add rule: 1536: list of from <podIP> to <vpcCIDR> use table <table>
-			for _, cidr := range vpcCIDRs {
-				podRule := netLink.NewRule()
-				_, podRule.Dst, _ = net.ParseCIDR(cidr)
-				podRule.Src = addr
-				podRule.Table = table
-				podRule.Priority = fromContainerRulePriority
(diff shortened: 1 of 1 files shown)


plugins/routed-eni/driver/driver.go
```diff
@@ -233,30 +233,22 @@ func setupNS(hostVethName string, contVethName string, netnsPath string, addr *n
...
+			podRule.Table = table
+			podRule.Priority = fromContainerRulePriority
+
+			err = netLink.RuleAdd(podRule)
+			if isRuleExistsError(err) {
+				log.Warn("Rule already exists [%v]", podRule)
+			} else {
+				if err != nil {
+					log.Errorf("Failed to add pod IP rule [%v]: %v", podRule, err)
+					return errors.Wrapf(err, "setupNS: failed to add pod rule [%v]", podRule)
```
jaypipes: what you are doing here is basically what the block above this does (when `useExternalSNAT` is `true`).

plugins/routed-eni/driver/driver.go
```diff
@@ -233,30 +233,22 @@ func setupNS(hostVethName string, contVethName string, netnsPath string, addr *n
 			}
 			log.Infof("Added rule priority %d from %s table %d", fromContainerRulePriority, addr.String(), table)
 		} else {
-			// add rule: 1536: list of from <podIP> to <vpcCIDR> use table <table>
-			for _, cidr := range vpcCIDRs {
```
mogren: The rule needs to be set for all CIDR ranges in the VPC I can't see how this is supposed to work.

PR #781, opened 2019-12-31
Speed up tests using exponential backoff

mogren:
*Description of changes:*
* Override the max-timeout in the tagging tests as well to make them run faster (Saves 22s of sleep)
* Lowercase error messages in tests
* Removed test for failing to create publisher because of no context (Saves over 1 minute)

All unit tests now run in around 8 seconds, even if we have added a lot of them. 8 months ago they took around 10 minutes to run.

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -578,7 +578,7 @@ func (cache *EC2InstanceMetadataCache) AllocENI(useCustomCfg bool, sg []*string,
 	}
 
 	// Once the ENI is attached, tag it.
-	cache.tagENI(eniID)
+	cache.tagENI(eniID, maxENIBackoffDelay)
 
 	// Also change the ENI's attribute so that the ENI will be deleted when the instance is deleted.
 	attributeInput := &ec2.ModifyNetworkInterfaceAttributeInput{
@@ -666,7 +666,7 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
 	return aws.StringValue(result.NetworkInterface.NetworkInterfaceId), nil
 }
 
-func (cache *EC2InstanceMetadataCache) tagENI(eniID string) {
+func (cache *EC2InstanceMetadataCache) tagENI(eniID string, maxBackoffDelay time.Duration) {
 	// Tag the ENI with "node.k8s.amazonaws.com/instance_id=<instance_id>"
 	tags := []*ec2.Tag{
 		{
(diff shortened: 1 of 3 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -823,7 +823,7 @@ func (cache *EC2InstanceMetadataCache) freeENI(eniName string, maxBackoffDelay t
 	}
 
 	// It does take awhile for EC2 to detach ENI from instance, so we wait 2s before trying the delete.
-	time.Sleep(2 * time.Second)
+	time.Sleep(2 * maxBackoffDelay)
```
nithu0115: Isn't 1 minute too long of a wait to detach ENI from instance? I can foresee ENI dangling issues when instances are terminated by ASGs.

nithu0115: IMO, we should describe the network interfaces to check if it is available instead of random sleep.

jaypipes:
I can verify this does indeed shave a bunch of time off tests:

```
jaypipes@thelio:~/go/src/github.com/aws/amazon-vpc-cni-k8s$ git describe --tags --always --dirty
v1.6.0-rc1-72-g3b0f8761
jaypipes@thelio:~/go/src/github.com/aws/amazon-vpc-cni-k8s$ time make docker-unit-test >/dev/null 2>&1

real	0m52.440s
user	0m0.447s
sys	0m0.139s
jaypipes@thelio:~/go/src/github.com/aws/amazon-vpc-cni-k8s$ git fetch upstream pull/781/head:issue781 && git checkout issue781
remote: Enumerating objects: 8, done.
remote: Counting objects: 100% (8/8), done.
remote: Compressing objects: 100% (4/4), done.
remote: Total 8 (delta 4), reused 6 (delta 4), pack-reused 0
Unpacking objects: 100% (8/8), done.
From github.com:aws/amazon-vpc-cni-k8s
 * [new ref]           refs/pull/781/head -> issue781
Switched to branch 'issue781'
jaypipes@thelio:~/go/src/github.com/aws/amazon-vpc-cni-k8s$ time make docker-unit-test >/dev/null 2>&1

real	0m15.757s
user	0m0.409s
sys	0m0.163s
```

pkg/awsutils/awsutils.go
jaypipes: > Isn't 1 minute too long of a wait to detach ENI from instance? I can foresee ENI dangling issues when instances are terminated by ASGs.

Yes, @nithu0115 is correct. This should stay as `time.Second`. Nithish, if you want to tackle a less naive approach that uses a call to DescribeNetworkInterfaces, feel free to do so in a different PR. Keep in mind, though, that this cache is all about reducing the number of calls to EC2 to prevent API rate limiting :)

mogren: @nithu0115 Thanks, good catch. I made the sleep configurable in the tests.

I'm not sure we should call EC2 even more just to check, all requests counts against the TPS limit, so we might as well try to delete. When testing, most waits were between 1.8 to 2.5 seconds delay from detach until the ENI could be deleted. I've never seen it faster, but sometimes up to 10 seconds.

nithu0115: > I'm not sure we should call EC2 even more just to check, all requests counts against the TPS limit, so we might as well try to delete.

That's true, but logs would have lot of noise about `ENI failed to delete`

Issue #755, opened 2019-12-12
Anti-affinity for Fargate nodes missing from this repo?


lukehoban:
When I stand up a cluster with `eksctl create --fargate`, I see the CNI daemonset as shown below.

This is different from what's in this repo in that it has a node anti-affinity set for `fargate` nodes - which is necessary to work correctly with fully-Fargate EKS clusters.

Is it intentional that this is missing from the copy of this config in https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/v1.5/aws-k8s-cni.yaml?  If so - why?

```yaml
$ kubectl get daemonset aws-node -n kube-system -o yaml
apiVersion: extensions/v1beta1
kind: DaemonSet
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
[...]

PR #792, opened 2020-01-14
adding node affinity for Fargate

nithu0115:
*Issue #, if available:* N/A 

*Description of changes:*

Upon creating new EKS version ( 1.14 k8s version), it comes with default 1.5.3 CNI version which comes with nodeAffinity for Fargate. However, the same parameters are missing on Github.

### DIFF
--- a/config/v1.5/aws-k8s-cni.yaml
+++ b/config/v1.5/aws-k8s-cni.yaml
@@ -76,6 +76,10 @@ spec:
                     operator: In
                     values:
                       - amd64
+                  - key: eks.amazonaws.com/compute-type
+                    operator: NotIn
+                    values:
+                      - fargate
       serviceAccountName: aws-node
       hostNetwork: true
       tolerations:


config/v1.5/aws-k8s-cni.yaml
```diff
@@ -76,6 +76,10 @@ spec:
                     operator: In
                     values:
                       - amd64
+                  - key: eks.amazonaws.com/compute-type
+                    operator: NotIn
+                    values:
+                      - fargate
```
jaypipes: @nithu0115 is this addressing a specific problem you've run into with customers? I closed a similar issue (#755) stating that we didn't want to place EKS-specific configuration values into the CNI plugin's YAML manifests because CNI can and should run everywhere, not just EKS. Do you disagree with that?

nithu0115: @jaypipes I agree up to a point, but the source of truth for customer to upgrade to latest CNI version (to 1.5.5) from default 1.5.3 version is still GitHub as per EKS public documentation.  

IMO, I still think we should have this node affinity params to provide clean upgrade experience, and should have a strategy to remove (like introducing customer to helms charts) going forward.

jaypipes: Chatted with @mogren about this some more this morning. He has convinced me that this won't do any harm, and although my objective is to keep EKS-specific items out of the VPC CNI configuration, this won't harm anyone running the VPC CNI plugin on non-EKS k8s on AWS

Issue #808, opened 2020-01-28
nodeInit() unnecessarily calls DescribeNetworkInterfaces() IMDS more than once


jaypipes:
nodeInit() - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/ipamd/ipamd.go#L271) function grabs all the ENIs for the worker node, then grabs all pods on the worker node and loops through all the pods setting up iptables rules.

This piece of code: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/ipamd/ipamd.go#L373-L378

can and should be pulled out of that loop since it's not dependent on the contents of the loop iteration. This would eliminate N-1 calls to IMDS to get the subnets associated with the node, where N == #pods

jaypipes:
Actually, it's even less efficient than that... We already call GetVPCIPv4CIDRs() on line 305 and throw away the result :(

https://github.com/aws/amazon-vpc-cni-k8s/blob/master/ipamd/ipamd.go#L305

jaypipes:
nodeInit() is making all these calls to DescribeNetworkInterfaces(), but it's not actually the GetVPCIPv4CIDRs() call that is doing it. It's all the callers of IPAMContext.getENIAddresses(): https://github.com/aws/amazon-vpc-cni-k8s/blob/4bf10a04402b1249f755fff36a79f608136d79ff/ipamd/ipamd.go#L792

IPAMContext.getENIAddresses() calls the awsClient.DescribeENI() method, which bypasses the "cache" and calls the IMDS's DescribeNetworkInterfaces() API call.
[January 28, 2020, 8:48 AM] Pipes, Jay: getENIAddresses() is called from two places: IPAMContext.tryAssignIPs() and IPAMContext.setupENI(). Both of those locations are called in nodeInit(), with the setupENI() called inside a loop across each ENI for the instance type and tryAssignIPs() doing yet another loop across ENIs for the instance type.

In short, the solution here is to call DescribeNetworkInterfaces() *once* at the start of nodeInit() and pass the results of that call along to subroutines like setupENI() and tryAssignIPs().

PR #809, opened 2020-01-28
[ipamd] de-dup calls to DescribeNetworkInterfaces

jaypipes:
In `IPAMContext.nodeInit()` and `IPAMContext.nodeIPPoolReconcile()`, the
`IPAMContext.setupENI()` method was being called, passing in an
`ENIMetadata` struct that contained the "local" IPv4 addresses that had
been queried from IMDS. `IPAMContext.setupENI()` called the
`IPAMContext.getENIAddresses()` method, however, which in turn calls the
AWS EC2 DescribeNetworkInterfaces API. The DescribeNetworkInterfaces API
call is expensive and throttled, and so we need to be careful how often
we call it.

In the case of `IPAMContext.nodeInit()` and
`IPAMContext.nodeIPPoolReconcile()`, they were both calling
`IPAMContext.setupENI()` and passing along an `ENIMetadata` struct that both
methods had previously constructed using the
`EC2InstanceMetadataCache.GetAttachedENIs()` method, which in turn called
`EC2InstanceMetadataCache.getENIMetadata()`.
`EC2InstanceMetadataCache.getENIMetadata()` itself ended up calling the
DescribeNetworkInterfaces EC2 API for each ENI on the host machine. So,
the slice of `ENIMetadata` structs returned from
`EC2InstanceMetadataCache.GetAttachedENIs()` -- which was being called by
`IPAMContext.nodeInit()` and `IPAMContext.nodeIPPoolReconcile()` *already
had the most up-to-date information from the DescribeNetworkInterfaces
EC2 API call*.

Therefore, what this patch does is simply modify the `ENIMetadata` struct
to store the slice of pointer to `ec2.NetworkInterfacePrivateIpAddress`
structs that is returned from the DescribeNetworkInterfaces EC2 API
call. That way, for methods like `IPAMContext.setupENI()`, which passes in
an `ENIMetadata` struct, we can remove the call to
`IPAMContext.getENIAddresses()` which cuts down on the number of duplicate
calls to DescribeNetworkInterfaces EC2 API.

Issue #808

### DIFF
--- a/ipamd/ipamd.go
+++ b/ipamd/ipamd.go
@@ -295,6 +295,13 @@ func (c *IPAMContext) nodeInit() error {
 	}
 	c.updateIPStats(numUnmanaged)
 
+	var pbVPCcidrs []string
+	vpcCIDRs := c.awsClient.GetVPCIPv4CIDRs()
+
+	for _, cidr := range vpcCIDRs {
+		pbVPCcidrs = append(pbVPCcidrs, *cidr)
+	}
+
 	_, vpcCIDR, err := net.ParseCIDR(c.awsClient.GetVPCIPv4CIDR())
 	if err != nil {
 		log.Error("Failed to parse VPC IPv4 CIDR", err.Error())
@@ -302,7 +309,7 @@ func (c *IPAMContext) nodeInit() error {
 	}
 
 	primaryIP := net.ParseIP(c.awsClient.GetLocalIPv4())
-	err = c.networkClient.SetupHostNetwork(vpcCIDR, c.awsClient.GetVPCIPv4CIDRs(), c.awsClient.GetPrimaryENImac(), &primaryIP)
+	err = c.networkClient.SetupHostNetwork(vpcCIDR, vpcCIDRs, c.awsClient.GetPrimaryENImac(), &primaryIP)
 	if err != nil {
 		log.Error("Failed to set up host network", err)
 		return errors.Wrap(err, "ipamd init: failed to set up host network")
@@ -370,12 +377,6 @@ func (c *IPAMContext) nodeInit() error {
 
 		// Update ip rules in case there is a change in VPC CIDRs, AWS_VPC_K8S_CNI_EXTERNALSNAT setting
 		srcIPNet := net.IPNet{IP: net.ParseIP(ip.IP), Mask: net.IPv4Mask(255, 255, 255, 255)}
-		vpcCIDRs := c.awsClient.GetVPCIPv4CIDRs()
-
-		var pbVPCcidrs []string
-		for _, cidr := range vpcCIDRs {
-			pbVPCcidrs = append(pbVPCcidrs, *cidr)
-		}
 
 		err = c.networkClient.UpdateRuleListBySrc(rules, srcIPNet, pbVPCcidrs, !c.networkClient.UseExternalSNAT())
 		if err != nil {
--- a/ipamd/ipamd.go
+++ b/ipamd/ipamd.go
@@ -746,19 +746,16 @@ func (c *IPAMContext) tryAssignIPs() (increasedPool bool, err error) {
 // 2) set up linux ENI related networking stack.
 // 3) add all ENI's secondary IP addresses to datastore
 func (c *IPAMContext) setupENI(eni string, eniMetadata awsutils.ENIMetadata) error {
-	ec2Addrs, eniPrimaryIP, err := c.getENIaddresses(eni)
-	if err != nil {
-		return errors.Wrapf(err, "failed to retrieve ENI %s IP addresses", eni)
-	}
-
 	// Add the ENI to the datastore
-	err = c.dataStore.AddENI(eni, eniMetadata.DeviceNumber, eni == c.awsClient.GetPrimaryENI())
+	err := c.dataStore.AddENI(eni, eniMetadata.DeviceNumber, eni == c.awsClient.GetPrimaryENI())
 	if err != nil && err.Error() != datastore.DuplicatedENIError {
 		return errors.Wrapf(err, "failed to add ENI %s to data store", eni)
 	}
 
 	// For secondary ENIs, set up the network
 	if eni != c.awsClient.GetPrimaryENI() {
+
+		eniPrimaryIP := eniMetadata.PrimaryIPv4Address()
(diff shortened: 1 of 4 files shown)


ipamd/ipamd.go
```diff
@@ -370,12 +376,6 @@ func (c *IPAMContext) nodeInit() error {
 
 		// Update ip rules in case there is a change in VPC CIDRs, AWS_VPC_K8S_CNI_EXTERNALSNAT setting
 		srcIPNet := net.IPNet{IP: net.ParseIP(ip.IP), Mask: net.IPv4Mask(255, 255, 255, 255)}
-		vpcCIDRs := c.awsClient.GetVPCIPv4CIDRs()
-
-		var pbVPCcidrs []string
-		for _, cidr := range vpcCIDRs {
-			pbVPCcidrs = append(pbVPCcidrs, *cidr)
-		}
```
mogren: Yep, no need to redo this every time, even if it's cached.

ipamd/ipamd.go
```diff
@@ -1000,17 +997,18 @@ func (c *IPAMContext) nodeIPPoolReconcile(interval time.Duration) {
...
 func (c *IPAMContext) eniIPPoolReconcile(ipPool map[string]*datastore.AddressInfo, attachedENI awsutils.ENIMetadata, eni string) {
-	for _, localIP := range attachedENI.LocalIPv4s {
-		if localIP == c.primaryIP[eni] {
-			log.Debugf("Reconcile and skip primary IP %s on ENI %s", localIP, eni)
+	for _, privateIPv4 := range attachedENI.IPv4Addresses {
+		strPrivateIPv4 := aws.StringValue(privateIPv4.PrivateIpAddress)
+		if strPrivateIPv4 == c.primaryIP[eni] {
+			log.Debugf("Reconcile and skip primary IP %s on ENI %s", strPrivateIPv4, eni)
 			continue
 		}
```
mogren: Could this check be this instead?
```go
if aws.BoolValue(privateIPv4.Primary) {
	log.Debugf("Skip primary IP %s on ENI %s during reconcile", strPrivateIPv4, eni)
	continue
}
```

pkg/awsutils/awsutils.go
```diff
@@ -184,12 +184,21 @@ type ENIMetadata struct {
 	SubnetIPv4CIDR string
 
 	// The ip addresses allocated for the network interface
-	LocalIPv4s []string
+	IPv4Addresses []*ec2.NetworkInterfacePrivateIpAddress
```
mogren: 👍

pkg/awsutils/awsutils.go
```diff
@@ -442,20 +451,61 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(macStr string) (ENIMetadat
...
+			strPrivateIPv4 := aws.StringValue(privateIPv4.PrivateIpAddress)
+			if imdsIPv4 == strPrivateIPv4 {
+				found = true
+				break
+			}
+		}
+		if !found {
+			missingDNI = append(missingDNI, imdsIPv4)
+		}
+	}
```
mogren: Working with sets in go is somewhat painful... 😉

ipamd/ipamd.go
```diff
@@ -746,27 +746,24 @@ func (c *IPAMContext) tryAssignIPs() (increasedPool bool, err error) {
 // 2) set up linux ENI related networking stack.
 // 3) add all ENI's secondary IP addresses to datastore
 func (c *IPAMContext) setupENI(eni string, eniMetadata awsutils.ENIMetadata) error {
-	ec2Addrs, eniPrimaryIP, err := c.getENIaddresses(eni)
```
jaypipes: Note that this right here is the key to this patch. Removing the call to `getENIaddresses()` here eliminates one source of calls to DescribeNetworkInterfaces EC2 API. setupENI() is called from both `nodeInit()` *and* the reconcile loop, both of which pass in an ENIMetadata struct that has already been populated with the most up-to-date ENI information, so there is no need to call getENIAddresses() here to re-fetch that information.

pkg/awsutils/awsutils.go
jaypipes: indeed. there is a `sets` utility library in upstream k/k that makes working with this a bit easier, but wasn't worth pulling in the dep IMHO.

ipamd/ipamd.go
jaypipes: yes, I can change it to that in a followup. ++

ipamd/ipamd.go
mogren: Yes, this will be called every time an ENI gets attached.

pkg/awsutils/awsutils.go
mogren: @jaypipes Nick added "github.com/deckarep/golang-set" earlier.. :)

ipamd/ipamd.go
```diff
@@ -295,14 +295,21 @@ func (c *IPAMContext) nodeInit() error {
...
 	c.updateIPStats(numUnmanaged)
 
+	var pbVPCcidrs []string
+	vpcCIDRs := c.awsClient.GetVPCIPv4CIDRs()
+
+	for _, cidr := range vpcCIDRs {
+		pbVPCcidrs = append(pbVPCcidrs, *cidr)
+	}
+
 	_, vpcCIDR, err := net.ParseCIDR(c.awsClient.GetVPCIPv4CIDR())
```
M00nF1sh: `GetVPCIPv4CIDR` is called twice here, we should rely on the fact that it's cached in c.awsClient

ipamd/ipamd.go
```diff
@@ -315,26 +322,25 @@ func (c *IPAMContext) nodeInit() error {
 		retry := 0
 		for {
 			retry++
-			err = c.setupENI(eni.ENIID, eni)
+			if err = c.setupENI(eni.ENIID, eni); err == nil {
+				log.Infof("ENI %s set up.", eni.ENIID)
+				break
+			}
+
 			if retry > maxRetryCheckENI {
```
M00nF1sh: not related to this PR, seems if this branch `if retry > maxRetryCheckENI` triggered it will hide the real err?

jaypipes: agreed. can fix that in followup.

Issue #785, opened 2020-01-02
scripts/run-integration-tests.sh leaves improperly-modified config YAML file


jaypipes:
After running `scripts/run-integration-tests.sh`, the git working tree is dirty due to modification of the `config/v1.5/aws-k8s-cni.yaml` file:

```
jaypipes@thelio:~/go/src/github.com/aws/amazon-vpc-cni-k8s$ git status
On branch e2e-test-cleanup
Your branch is ahead of 'origin/master' by 2 commits.
  (use "git push" to publish your local commits)

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   config/v1.5/aws-k8s-cni.yaml

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	config/v1.5/aws-k8s-cni.yaml.bak
```

Instead of `sed`'ing the actual config YAML file (`config/v1.5/aws-k8s-cni.yaml`), we should instead make a *copy* of that file, put it into the `/tmp/cni-test/$TESTDIR` folder and modify that instead.

PR #813, opened 2020-01-29
make copy of base config template in e2e tests

jaypipes:
Do not inline edit (using sed) the base configuration YAML file to be
kubectl apply'd during e2e tests. Instead, create a copy of the base
configuration YAML file and modify the copy.

Fixes Issue #785 


mogren:
LGTM! Rebased on master and opened #827 instead.

Issue #814, opened 2020-01-29
flaky unit test in nodeInit() DescribeENI call argument order


jaypipes:
```
--- FAIL: TestNodeInit (0.00s)
    controller.go:150: Unexpected call to *mock_awsutils.MockAPIs.DescribeENI([eni-00000001]) at /go/src/github.com/{{ORG_NAME}}/{{REPO_NAME}}/pkg/awsutils/mocks/awsutils_mocks.go:102 because: 
        Expected call at /go/src/github.com/{{ORG_NAME}}/{{REPO_NAME}}/ipamd/ipamd_test.go:146 doesn't match the argument at index 0.
        Got: eni-00000001
        Want: is equal to eni-00000000
    panic.go:563: missing call(s) to *mock_awsutils.MockAPIs.DescribeENI(is equal to eni-00000000) /go/src/github.com/{{ORG_NAME}}/{{REPO_NAME}}/ipamd/ipamd_test.go:146
    panic.go:563: aborting test due to missing call(s)
```

Seeing this on some `make unit-test` runs.

jaypipes:
Was introduced with #809

jaypipes:
Note that I believe this to be some unit-test-specific bug. I've run numerous e2e functional tests against #809 and see no issues.

jaypipes:
Tracked down the source of the flakiness. The datastore struct stores IP pool information in a map:

https://github.com/aws/amazon-vpc-cni-k8s/blob/045f8f6f7b82bb18ca7efa08320ca027e5c01254/pkg/ipamd/datastore/data_store.go#L138

However, when calling `datastore.GetENINeedsIP()`, the code loops over map elements using `range`:

https://github.com/aws/amazon-vpc-cni-k8s/blob/045f8f6f7b82bb18ca7efa08320ca027e5c01254/pkg/ipamd/datastore/data_store.go#L430

This is not a deterministic behaviour and is the reason we're seeing random flakiness in this unit test.

PR #819, opened 2020-01-29
make datastore.GetENINeedsIPs deterministic

jaypipes:
The datastore used in ipamd uses a map[string]*IPPool to store ENI
information. The `datastore.GetENINeedsIP()` method was iterating over
the map using `range`, which is not deterministic since a map's elements
have no guaranteed order.

This was problematic for a unit test since a mock was expecting the
primary ENI to be returned from the datastore but every so often, the
secondary ENI was being returned due to the non-deterministic map
iteration. This patch fixes this by simply sorting the map's keys and
iterating over the sorted keys, accessing the ENI information by key
instead by iteration.

Fixes Issue #814 

### DIFF
--- a/pkg/ipamd/datastore/data_store.go
+++ b/pkg/ipamd/datastore/data_store.go
@@ -14,6 +14,7 @@
 package datastore
 
 import (
+	"sort"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/ipamd/datastore/data_store.go
```diff
@@ -427,11 +428,21 @@ func (e *ENIIPPool) hasPods() bool {
 
 // GetENINeedsIP finds an ENI in the datastore that needs more IP addresses allocated
 func (ds *DataStore) GetENINeedsIP(maxIPperENI int, skipPrimary bool) *ENIIPPool {
-	for _, eni := range ds.eniIPPools {
+	// NOTE(jaypipes): Some tests rely on key order so we iterate over the IP
+	// pool structs here in sorted key order.
+	// TODO(jaypipes): Don't use a map as the primary iterator vehicle.
+	// Instead, use a slice of *ENIPool and use a map for existence checks only
```
mogren: Created #820

Issue #511, opened 2019-06-20
Use a structured logging library


chiefy:
Feature request: use json logging, or something standard (syslog) because trying to get a logging shipper to parse multiline is a pita. Thanks!

For anyone using fluent-bit, I had to use `tail` input and this as a Multiline_Parser

```
    [PARSER]
        Name        aws-cni
        Format      regex
        Regex       ^(?<time>[^\[]+)\s*\[(?<level>[A-Z]*)\]\s(?<message>.*)$
        Time_Key    time
        Time_Format %Y-%m-%dT%H:%M:%SZ
```

max-rocket-internet:
> because trying to get a logging shipper to parse multiline is a pita

Or just don't log multi-line messages?

jaypipes:
We should standardize on either klog or logrus.

mogren:
Change merged, following upstream precedence:

https://godoc.org/sigs.k8s.io/controller-runtime/pkg/log/zap

PR #824, opened 2020-01-31
Zap logger

nithu0115:
*Issue #, if available:* #511  

*Description of changes:*
Adding zap logger 

- [X] Added zap logger
- [X] Parameterize config
- [X] Update Readme 
- [X] Unit tests

#### Outputs

Screen Shot 2020-03-16 at 2 11 07 PM

Screen Shot 2020-03-16 at 2 19 55 PM

Screen Shot 2020-03-16 at 2 19 23 PM


pkg/utils/logger/zaplogger.go
```diff
@@ -0,0 +1,173 @@
...
+		MaxSize:    100,
+		MaxBackups: 5,
+		MaxAge:     30,
+		Compress:   true,
+	}
+	return zapcore.AddSync(lumberJackLogger)
+}
+
+// NewZapLogger creates and returns a new global logger. It overwrites the
+// existing global logger if that has been previously defined.
```
mogren: Nit: This comment doesn't look right.

pkg/utils/logger/zaplogger.go
```diff
@@ -0,0 +1,173 @@
...
+
+	logFileLocation := getLogFileLocation(srvName)
+
+	if logFileLocation != "stdout" {
+		writer := getLogWriter(logFileLocation)
+		cores = append(cores, zapcore.NewCore(getEncoder(), writer, logLevel))
+	} else {
+		writer := zapcore.Lock(os.Stdout)
+		cores = append(cores, zapcore.NewCore(getEncoder(), writer, logLevel))
+	}
```
mogren: I think this would be easier to read:
```go
	var writer zapcore.WriteSyncer
	if logFileLocation == "stdout" {
		writer = zapcore.Lock(os.Stdout)
	} else {
		writer = getLogWriter(logFileLocation)
	}	
	cores = append(cores, zapcore.NewCore(getEncoder(), writer, logLevel))
```

pkg/utils/logger/zaplogger.go
```diff
@@ -0,0 +1,173 @@
...
+		if logFilePath == "" {
+			logFilePath = defaultPluginLogFilePath
+		}
+		return logFilePath
+	default:
+		return "stdout"
+	}
+}
+
+func newZapLogger(srvName string) Logger {
```
mogren: Does `srvName` mean "server name"? "service name"? Looking at the code, it seems to mean "component", would that be a better name?

pkg/utils/logger/logger.go
```diff
@@ -13,69 +13,115 @@
...
- <formats>
-  <format id="main" format="%%UTCDate(2006-01-02T15:04:05.000Z07:00) [%%LEVEL] %%t%%Msg%%n" />
- </formats>
-</seelog>
-`
+	envLogLevel              = "AWS_VPC_K8S_CNI_LOGLEVEL"
+	defaultIPamDLogFilePath  = "/host/var/log/aws-routed-eni/ipamd.log"
+	envIPamDLogFilePath      = "AWS_VPC_K8S_CNI_LOG_FILE"
+	defaultPluginLogFilePath = "/var/log/aws-routed-eni/plugin.log"
+	envPluginFilePath        = "AWS_VPC_K8S_PLUGIN_LOGL_FILE"
```
mogren: I would expect `AWS_VPC_K8S_PLUGIN_LOGL_FILE` to be passed to the plugin, just [like](https://github.com/aws/amazon-vpc-cni-k8s/blob/8340e95126da40e8489074c68052869c4cd098d2/scripts/entrypoint.sh#L76) the [`MTU`](https://github.com/aws/amazon-vpc-cni-k8s/blob/8340e95126da40e8489074c68052869c4cd098d2/cmd/routed-eni-cni-plugin/cni.go#L71). The ENV variables in the daemonset config is only available to ipamd.

Also, what does logging to `stdout` mean here? It will end up in the kubelet logs? Have we tested this config?

pkg/utils/logger/logger.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2017-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mogren: Nit: Just 2020?

pkg/k8sapi/discovery.go
```diff
@@ -256,7 +258,7 @@ func (d *Controller) handlePodUpdate(key string) error {
 		d.workerPodsLock.Lock()
 		defer d.workerPodsLock.Unlock()
 
-		log.Tracef("Update for pod %s: %+v, %+v", podName, pod.Status, pod.Spec)
+		log.Infof("Update for pod %s: %+v, %+v", podName, pod.Status, pod.Spec)
```
mogren: This will increase the amount of logging a bit, but not as much as before, since we have limited the watcher quite a lot.

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -397,13 +401,13 @@ func metricsListGrabAggregateConvert(t metricsTarget) (map[string]*dto.MetricFam
...
-	glog.Info("len(targetList)", len(targetList))
+	log.Infof("targetList: %v ", targetList)
+	log.Infof("len(targetList) %d", len(targetList))
 	for _, target := range targetList {
-		glog.Infof("Grab/Aggregate metrics from %v", target)
+		log.Infof("Grab/Aggregate metrics from %v", target)
 		rawOutput, err := t.grabMetricsFromTarget(target)
 		if err != nil {
-			glog.Errorf("Failed to getMetricsFromTarget: %v", err)
+			log.Errorf("Failed to getMetricsFromTarget: %v", err)
```
mogren: Should we make this a `log.Warnf()` instead? Ref to #592. (I think we can remove [this line](https://github.com/aws/amazon-vpc-cni-k8s/blob/8340e95126da40e8489074c68052869c4cd098d2/cmd/cni-metrics-helper/metrics/cni_metrics.go#L158) as well.)

pkg/ipamd/datastore/data_store.go
```diff
@@ -200,8 +201,8 @@ func (ds *DataStore) AddIPv4AddressToStore(eniID string, ipv4 string) error {
 	ds.lock.Lock()
 	defer ds.lock.Unlock()
 
-	log.Tracef("Adding ENI(%s)'s IPv4 address %s to datastore", eniID, ipv4)
-	log.Tracef("IP Address Pool stats: total: %d, assigned: %d", ds.total, ds.assigned)
+	log.Infof("Adding ENI(%s)'s IPv4 address %s to datastore", eniID, ipv4)
+	log.Infof("IP Address Pool stats: total: %d, assigned: %d", ds.total, ds.assigned)
```
mogren: Is there no trace option? This will cause a lot more logging, and we are not that interested in these lines by default. Even debug is too verbose here.

pkg/ipamd/ipamd.go
```diff
@@ -882,9 +884,9 @@ func (c *IPAMContext) nodeIPPoolTooLow() bool {
 	available := total - used
 	poolTooLow := available < c.maxIPsPerENI*c.warmENITarget || (c.warmENITarget == 0 && available == 0)
 	if poolTooLow {
-		log.Tracef("IP pool is too low: available (%d) < ENI target (%d) * addrsPerENI (%d)", available, c.warmENITarget, c.maxIPsPerENI)
+		log.Infof("IP pool is too low: available (%d) < ENI target (%d) * addrsPerENI (%d)", available, c.warmENITarget, c.maxIPsPerENI)
 	} else {
-		log.Tracef("IP pool is NOT too low: available (%d) >= ENI target (%d) * addrsPerENI (%d)", available, c.warmENITarget, c.maxIPsPerENI)
+		log.Infof("IP pool is NOT too low: available (%d) >= ENI target (%d) * addrsPerENI (%d)", available, c.warmENITarget, c.maxIPsPerENI)
```
mogren: Same here, this will generate a lot of unnecessary log lines.

pkg/utils/logger/logger_test.go
```diff
@@ -17,57 +17,66 @@ import (
...
+
+func TestGetLogFileLocationReturnsStdoutIfsrvNameDoesnotMatch(t *testing.T) {
+	srvName := "abcD"
+	path := "stdout"
+	assert.Equal(t, path, getLogFileLocation(srvName))
+}
+
+func TestGetLogFileLocationFailstoReturn(t *testing.T) {
+	srvName := "L-IPamD"
+	path := "/tmp/testin.log"
```
mogren: Nit: `testin.log`? 😄

pkg/ipamd/datastore/data_store.go
nithu0115: Unfortunately, zap logger does not have trace level. 

Do you have any suggestions here, other than info log level ?

mogren: Hmm, I guess `Debug` would work for the other cases, but these two I think we just need to remove. They don't add much information that is not already logged at `Info` level on line 223.

pkg/utils/logger/logger.go
nithu0115: > I would expect `AWS_VPC_K8S_PLUGIN_LOGL_FILE` to be passed to the plugin, just [like](https://github.com/aws/amazon-vpc-cni-k8s/blob/8340e95126da40e8489074c68052869c4cd098d2/scripts/entrypoint.sh#L76) the [`MTU`](https://github.com/aws/amazon-vpc-cni-k8s/blob/8340e95126da40e8489074c68052869c4cd098d2/cmd/routed-eni-cni-plugin/cni.go#L71). The ENV variables in the daemonset config is only available to ipamd.

I did not understand this point. Are you saying that the environment variable configured with `config.yaml` are only available for `L-IPamD` server ? I believe, to override the current `MTU` value, we still pass `AWS_VPC_ENI_MTU=9001`, right ? Please correct me if I am missing something here. 

> Also, what does logging to `stdout` mean here? It will end up in the kubelet logs? Have we tested this config?

`stdout` here write to default container logging i.e., `stdout/stderr`. Yes, I have tested it, but found another issue with the `entrypoint.sh` [script](https://github.com/aws/amazon-vpc-cni-k8s/blob/8340e95126da40e8489074c68052869c4cd098d2/scripts/entrypoint.sh#L87) where the logs which are written to stdout/stderr are piped to a file which is `cat`'ed out only upon `L-IPamD` initiation failure. (I will put in a new PR later for the `entrypoint.sh` script)

pkg/utils/logger/zaplogger.go
nithu0115: `srvName` means server name. The variable is used to differentiate between the log file paths for different binaries which we have. For example, if the server is `L-IPamD`, it written to `/host/var/log/aws-routed-eni/ipamd.log` and if the the server is `Plugin`, the logs are written to `/var/log/aws-routed-eni/plugin.log`.

pkg/ipamd/datastore/data_store.go
nithu0115: Got it, thanks for the clarification! Debug it is then...

pkg/utils/logger/logger_test.go
nithu0115: :D

cmd/cni-metrics-helper/metrics/metrics.go
nithu0115: Addressed.

pkg/utils/logger/zaplogger.go
mogren: I got how it's used, I was just confused that the plugin was called a server. It's a binary that gets executed 😄

pkg/utils/logger/logger.go
mogren: Yes, ENV variables set in `aws-k8s-cni.yaml` are only available to ipamd. The MTU values get set in the plugin config file so that the plugin can get it. See the linked examples in the previous comment.

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -130,27 +134,27 @@ func processPercentile(metric *dto.Metric, act *metricsAction) {
 }
 
 func processHistogram(metric *dto.Metric, act *metricsAction) {
-	glog.V(5).Info("processing HISTOGRAM:", metric.GetLabel())
-	glog.V(5).Info(metric.GetHistogram())
+	log.Infof("processing:", zap.Any("HISTOGRAM", metric.GetLabel()))
+	log.Infof("processing:", zap.Any("GETHISTOGRAM", metric.GetHistogram()))
```
jaypipes: Recommend removing the above log messages.

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -130,27 +134,27 @@ func processPercentile(metric *dto.Metric, act *metricsAction) {
...
 	histogram := metric.GetHistogram()
 
 	for _, bucket := range histogram.GetBucket() {
-		glog.V(10).Info("processing bucket:", bucket)
+		log.Infof("processing:", zap.Any("bucket", bucket))
```
jaypipes: Recommend removing above log message.

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -130,27 +134,27 @@ func processPercentile(metric *dto.Metric, act *metricsAction) {
...
 		existingBucket := false
 		for _, bucketInAct := range act.bucket.curBucket {
 			if bucket.GetUpperBound() == *bucketInAct.UpperBound {
 				// found the matching bucket
-				glog.V(10).Infof("Found the matching bucket with UpperBound: %f", *bucketInAct.UpperBound)
+				log.Infof("Found the matching bucket with UpperBound: %f", *bucketInAct.UpperBound)
 				act.actionFunc(bucketInAct.CumulativeCount, float64(bucket.GetCumulativeCount()))
-				glog.V(10).Infof("Found: *bucketInAct.CumulativeCount:%f, bucket.GetCumulativeCount():%f",
+				log.Infof("Found: *bucketInAct.CumulativeCount:%f, bucket.GetCumulativeCount():%f",
```
jaypipes: Recommend removing above log messages

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -130,27 +134,27 @@ func processPercentile(metric *dto.Metric, act *metricsAction) {
...
 					*bucketInAct.CumulativeCount, float64(bucket.GetCumulativeCount()))
 				existingBucket = true
 				break
 			}
 		}
 
 		if !existingBucket {
-			glog.V(10).Infof("Create a new bucket with upperBound:%f", bucket.GetUpperBound())
+			log.Infof("Create a new bucket with upperBound:%f", bucket.GetUpperBound())
```
jaypipes: Recommend moving this after line 163 and changing "Create" to "Created"

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -250,7 +254,7 @@ func postProcessingHistogram(convert metricsConvert) bool {
 }
 
 func processMetric(family *dto.MetricFamily, convert metricsConvert) (bool, error) {
-	glog.Info("Processing metric: ", family.GetName())
+	log.Infof("Processing: ", zap.String("metric", family.GetName()))
```
jaypipes: Recommend just removing the above log message

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -357,7 +361,7 @@ func produceCloudWatchMetrics(t metricsTarget, families map[string]*dto.MetricFa
 					cw.Publish(dataPoint)
 				}
 			case dto.MetricType_SUMMARY:
-				glog.Infof("Produce PERCENTILE metrics: %s, value: %f", action.cwMetricName, action.data.curSingleDataPoint)
+				log.Infof("Produce PERCENTILE metrics: %s, value: %f", action.cwMetricName, action.data.curSingleDataPoint)
```
jaypipes: I'm pretty sure the above log messages starting with "Produce..." are not needed and are just fluff.

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -397,13 +401,13 @@ func metricsListGrabAggregateConvert(t metricsTarget) (map[string]*dto.MetricFam
 	resetMetrics(interestingMetrics)
 
 	targetList := t.getTargetList()
-	glog.Info("targetList: ", targetList)
-	glog.Info("len(targetList)", len(targetList))
+	log.Infof("targetList: %v ", targetList)
+	log.Infof("len(targetList) %d", len(targetList))
```
jaypipes: The above log messages are not needed. They are really just temporary debugging prints and don't belong as log messages.

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -397,13 +401,13 @@ func metricsListGrabAggregateConvert(t metricsTarget) (map[string]*dto.MetricFam
...
 	for _, target := range targetList {
-		glog.Infof("Grab/Aggregate metrics from %v", target)
+		log.Infof("Grab/Aggregate metrics from %v", target)
 		rawOutput, err := t.grabMetricsFromTarget(target)
 		if err != nil {
-			glog.Errorf("Failed to getMetricsFromTarget: %v", err)
+			log.Warnf("Failed to getMetricsFromTarget: %v", err)
```
jaypipes: Warnings should be for things that should be addressed by an operator/admin. Is the above really something that should be addressed by an admin? Recommend using `Infof` instead.

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -397,13 +401,13 @@ func metricsListGrabAggregateConvert(t metricsTarget) (map[string]*dto.MetricFam
 	resetMetrics(interestingMetrics)
 
 	targetList := t.getTargetList()
-	glog.Info("targetList: ", targetList)
-	glog.Info("len(targetList)", len(targetList))
+	log.Infof("targetList: %v ", targetList)
+	log.Infof("len(targetList) %d", len(targetList))
 	for _, target := range targetList {
-		glog.Infof("Grab/Aggregate metrics from %v", target)
+		log.Infof("Grab/Aggregate metrics from %v", target)
```
jaypipes: Pretty sure the above isn't necessary. Recommend removing.

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -412,13 +416,13 @@ func metricsListGrabAggregateConvert(t metricsTarget) (map[string]*dto.MetricFam
 		origFamilies, err := parser.TextToMetricFamilies(bytes.NewReader(rawOutput))
 
 		if err != nil {
-			glog.Warning("Failed to parse metrics:", err)
+			log.Warnf("Failed to parse metrics:", zap.Error(err))
```
jaypipes: It's best practice to simply return the error and let the top-most caller (i.e. `main()`) determine whether to print a message to a log.

The reason for this best practice is to control what to do with errors (log them, aggregate them, forward them, etc) in a single place.

Read here for more information on this: https://dave.cheney.net/2015/11/05/lets-talk-about-logging

In particular the following:

>  If an error that cannot be handled occurs, it should bubble up main.main where the program terminates. The minor inconvenience of having to insert the FATAL prefix in front of the final log message, or writing directly to os.Stderr with fmt.Fprintf is not sufficient justification for a logging package growing a log.Fatal method.

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -412,13 +416,13 @@ func metricsListGrabAggregateConvert(t metricsTarget) (map[string]*dto.MetricFam
...
 			return nil, nil, true, err
 		}
 
 		families, err = filterMetrics(origFamilies, interestingMetrics)
 		if err != nil {
-			glog.Warning("Failed to filter metrics:", err)
+			log.Warnf("Failed to filter metrics:", zap.Error(err))
```
jaypipes: simply return the error, don't log a message here.

cmd/grpc-health-probe/main.go
```diff
@@ -35,10 +38,9 @@ const (
 )
 
 func init() {
-	log.SetFlags(0)
 	flag.StringVar(&remoteURL, "addr", "", "(required) tcp host:port to connect")
 	flag.StringVar(&serviceName, "service", "", "service name to check (default: \"\")")
-	flag.StringVar(&userAgent, "user-agent", "grpc-health-probe", "user-agent header value of health check requests")
+	flag.StringVar(&userAgent, "user-agent", "grpc_health_probe", "user-agent header value of health check requests")
```
jaypipes: The binary is now called `grpc-health-probe`

pkg/eniconfig/eniconfig.go
```diff
@@ -54,6 +53,9 @@ type ENIConfig interface {
 
 var ErrNoENIConfig = errors.New("eniconfig: eniconfig is not available")
 
+//log zaplogger
```
jaypipes: Please remove above comment.

pkg/ipamd/datastore/data_store.go
jaypipes: Agreed. Debug is appropriate here.

pkg/ipamd/introspect.go
```diff
@@ -80,7 +78,7 @@ func (c *IPAMContext) ServeIntrospection() {
 			}
 
 			once.Do(func() {
-				log.Error("Error running http API: ", err)
+				log.Errorf("Error running http API: %v", err)
```
jaypipes: Recommend removing this. It's not really useful to log an error when we're just going to retry the operation here. Better to handle the error in `main()`

pkg/ipamd/ipamd.go
```diff
@@ -302,14 +304,14 @@ func (c *IPAMContext) nodeInit() error {
 
 	_, vpcCIDR, err := net.ParseCIDR(c.awsClient.GetVPCIPv4CIDR())
 	if err != nil {
-		log.Error("Failed to parse VPC IPv4 CIDR", err.Error())
+		log.Errorf("Failed to parse VPC IPv4 CIDR %v", err.Error())
```
jaypipes: just return the error and handle printing of errors in main()

pkg/ipamd/ipamd.go
```diff
@@ -302,14 +304,14 @@ func (c *IPAMContext) nodeInit() error {
...
 		return errors.Wrap(err, "ipamd init: failed to retrieve VPC CIDR")
 	}
 
 	primaryIP := net.ParseIP(c.awsClient.GetLocalIPv4())
 	err = c.networkClient.SetupHostNetwork(vpcCIDR, vpcCIDRs, c.awsClient.GetPrimaryENImac(), &primaryIP)
 	if err != nil {
-		log.Error("Failed to set up host network", err)
+		log.Errorf("Failed to set up host network %v", err)
```
jaypipes: just return the error and handle printing of errors in main()

pkg/ipamd/ipamd.go
```diff
@@ -915,9 +917,9 @@ func (c *IPAMContext) shouldRemoveExtraENIs() bool {
 	// We need the +1 to make sure we are not going below the WARM_ENI_TARGET.
 	shouldRemoveExtra := available >= (c.warmENITarget+1)*c.maxIPsPerENI
 	if shouldRemoveExtra {
-		log.Tracef("It might be possible to remove extra ENIs because available (%d) >= (ENI target (%d) + 1) * addrsPerENI (%d): ", available, c.warmENITarget, c.maxIPsPerENI)
+		log.Infof("It might be possible to remove extra ENIs because available (%d) >= (ENI target (%d) + 1) * addrsPerENI (%d): ", available, c.warmENITarget, c.maxIPsPerENI)
 	} else {
-		log.Tracef("Its NOT possible to remove extra ENIs because available (%d) < (ENI target (%d) + 1) * addrsPerENI (%d): ", available, c.warmENITarget, c.maxIPsPerENI)
+		log.Infof("Its NOT possible to remove extra ENIs because available (%d) < (ENI target (%d) + 1) * addrsPerENI (%d): ", available, c.warmENITarget, c.maxIPsPerENI)
```
jaypipes: Above should be Debugf

pkg/ipamd/ipamd.go
```diff
@@ -1074,7 +1076,7 @@ func UseCustomNetworkCfg() bool {
 		if err == nil {
 			return parsedValue
 		}
-		log.Error("Failed to parse "+envCustomNetworkCfg+"; using default: false", err.Error())
+		log.Errorf("Failed to parse %s; using default: false, err: %v", envCustomNetworkCfg, err)
```
jaypipes: just return the error and handle printing of errors in main()

pkg/ipamd/ipamd.go
```diff
@@ -1148,7 +1150,7 @@ func (c *IPAMContext) ipTargetState() (short int, over int, enabled bool) {
 	// over is less than the warm IP target alone if it would imply reducing total IPs below the minimum target
 	over = max(min(over, total-c.minimumIPTarget), 0)
 
-	log.Tracef("Current warm IP stats: target: %d, total: %d, assigned: %d, available: %d, short: %d, over %d", c.warmIPTarget, total, assigned, available, short, over)
+	log.Infof("Current warm IP stats: target: %d, total: %d, assigned: %d, available: %d, short: %d, over %d", c.warmIPTarget, total, assigned, available, short, over)
```
jaypipes: Debugf

pkg/ipamd/metrics.go
```diff
@@ -39,14 +38,14 @@ func (c *IPAMContext) ServeMetrics() {
...
-	log.Info("Serving metrics on port ", metricsPort)
+	log.Infof("Serving metrics on port %d", metricsPort)
 	server := c.setupMetricsServer()
 	for {
 		once := sync.Once{}
 		_ = retry.RetryWithBackoff(retry.NewSimpleBackoff(time.Second, time.Minute, 0.2, 2), func() error {
 			err := server.ListenAndServe()
 			once.Do(func() {
-				log.Error("Error running http API: ", err)
+				log.Errorf("Error running http API: %v", err)
```
jaypipes: don't print the error here since we're just going to retry the operation.

pkg/networkutils/network.go
```diff
@@ -625,11 +625,11 @@ func getConnmark() uint32 {
 	if connmark := os.Getenv(envConnmark); connmark != "" {
 		mark, err := strconv.ParseInt(connmark, 0, 64)
 		if err != nil {
-			log.Error("Failed to parse "+envConnmark+"; will use ", defaultConnmark, err.Error())
+			log.Errorf("Failed to parse %s; will use %d, error: %v", envConnmark, defaultConnmark, err)
 			return defaultConnmark
 		}
 		if mark > math.MaxUint32 || mark <= 0 {
-			log.Error(""+envConnmark+" out of range; will use ", defaultConnmark)
+			log.Errorf("%s out of range; will use %s", envConnmark, defaultConnmark)
```
jaypipes: Use Infof here. The error is being "handled" and therefore you should simply log an informative message that it was handled instead of error.

pkg/networkutils/network.go
```diff
@@ -625,11 +625,11 @@ func getConnmark() uint32 {
 	if connmark := os.Getenv(envConnmark); connmark != "" {
 		mark, err := strconv.ParseInt(connmark, 0, 64)
 		if err != nil {
-			log.Error("Failed to parse "+envConnmark+"; will use ", defaultConnmark, err.Error())
+			log.Errorf("Failed to parse %s; will use %d, error: %v", envConnmark, defaultConnmark, err)
```
jaypipes: Use Infof here. The error is being "handled" and therefore you should simply log an informative message that it was handled instead of error.

pkg/publisher/publisher.go
```diff
@@ -86,14 +88,14 @@ func New(ctx context.Context) (Publisher, error) {
 	}
 	clusterID, err := ec2Client.GetClusterTag("CLUSTER_ID")
 	if err != nil || clusterID == "" {
-		glog.Errorf("Failed to obtain cluster-id, fetching name.  %v", err)
+		log.Errorf("Failed to obtain cluster-id, fetching name.  %v", err)
```
jaypipes: This isn't an error. Recommend just removing the above.

pkg/publisher/publisher.go
```diff
@@ -86,14 +88,14 @@ func New(ctx context.Context) (Publisher, error) {
...
 		clusterID, err = ec2Client.GetClusterTag("Name")
 		if err != nil || clusterID == "" {
-			glog.Errorf("Failed to obtain cluster-id or name, defaulting to 'k8s-cluster'.  %v", err)
+			log.Errorf("Failed to obtain cluster-id or name, defaulting to 'k8s-cluster'.  %v", err)
```
jaypipes: Use Infof here. The error is being "handled" and therefore you should simply log an informative message that it was handled instead of error.

pkg/publisher/publisher.go
```diff
@@ -155,7 +157,7 @@ func (p *cloudWatchPublisher) pushLocal() {
 
 func (p *cloudWatchPublisher) push(metricData []*cloudwatch.MetricDatum) {
 	if len(metricData) == 0 {
-		glog.Warning("Missing data for publishing CloudWatch metrics")
+		log.Warn("Missing data for publishing CloudWatch metrics")
```
jaypipes: Is this an actionable warning? Would an admin take some action if they saw this? If not, recommend just removing it.

pkg/publisher/publisher.go
```diff
@@ -170,7 +172,7 @@ func (p *cloudWatchPublisher) push(metricData []*cloudwatch.MetricDatum) {
 		// Publish data
 		err := p.send(input)
 		if err != nil {
-			glog.Errorf("Unable to publish CloudWatch metrics: %v", err)
+			log.Errorf("Unable to publish CloudWatch metrics: %v", err)
```
jaypipes: Infof since we're not doing anything about the "error"

pkg/utils/logger/logger.go
```diff
@@ -13,69 +13,114 @@
...
- <formats>
-  <format id="main" format="%%UTCDate(2006-01-02T15:04:05.000Z07:00) [%%LEVEL] %%t%%Msg%%n" />
- </formats>
-</seelog>
-`
+	envLogLevel              = "AWS_VPC_K8S_CNI_LOGLEVEL"
+	defaultIPamDLogFilePath  = "/host/var/log/aws-routed-eni/ipamd.log"
+	envIPamDLogFilePath      = "AWS_VPC_K8S_CNI_LOG_FILE"
+	defaultPluginLogFilePath = "/var/log/aws-routed-eni/plugin.log"
+	envPluginFilePath        = "AWS_VPC_K8S_PLUGIN_LOGL_FILE"
```
jaypipes: `LOGL`

pkg/utils/logger/logger.go
```diff
@@ -13,69 +13,114 @@
...
+		log := defaultZapLogger()
+		return &structuredLogger{
+			zapLogger: log,
+		}
 	}
+	return log
+}
+
+//InitZapLogger log  name generated by srvName
+func InitZapLogger(srvName string) Logger {
```
jaypipes: A more idiomatic name for this function would be just `New()`.

Don't leak implementation details (the fact that we're using the zap library) out of the primary public interface. There's no reason for callers to know or care that we're using zap behind the scenes.

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -98,26 +102,26 @@ func getMetricsFromPod(client clientset.Interface, podName string, namespace str
...
+		log.Infof("processing : ", zap.Float64("COUNTER", metric.GetCounter().GetValue()))
 	}
 	act.actionFunc(&act.data.curSingleDataPoint, metric.GetCounter().GetValue())
 }
 
 func processPercentile(metric *dto.Metric, act *metricsAction) {
 	var p99 float64
 
-	glog.V(10).Info("processing PERCENTILE: ", p99)
+	log.Infof("processing: ", zap.Float64("PERCENTILE", p99))
```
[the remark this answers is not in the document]
mogren: I agree, better to not log this.

pkg/k8sapi/discovery.go
```diff
@@ -113,7 +115,7 @@ func (d *Controller) GetCNIPods() []string {
 		cniPods = append(cniPods, k)
 	}
 
-	log.Info("GetCNIPods discovered", cniPods)
+	log.Infof("GetCNIPods discovered %v", cniPods)
```
mogren: Could you remove the `log.Infof()` and `fmt.Printf()` on lines 94 and 95 in `discovery.go` as well?

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -412,13 +390,13 @@ func metricsListGrabAggregateConvert(t metricsTarget) (map[string]*dto.MetricFam
 		origFamilies, err := parser.TextToMetricFamilies(bytes.NewReader(rawOutput))
 
 		if err != nil {
-			glog.Warning("Failed to parse metrics:", err)
+			log.Error("Failed to parse metrics:")
```
jaypipes: remove this. Just let the error bubble up to the caller and the caller can log a message if it wants.

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -412,13 +390,13 @@ func metricsListGrabAggregateConvert(t metricsTarget) (map[string]*dto.MetricFam
...
 			return nil, nil, true, err
 		}
 
 		families, err = filterMetrics(origFamilies, interestingMetrics)
 		if err != nil {
-			glog.Warning("Failed to filter metrics:", err)
+			log.Error("Failed to filter metrics:")
```
jaypipes: remove this. Just let the error bubble up to the caller and the caller can log a message if it wants.

cmd/grpc-health-probe/main.go
```diff
@@ -89,49 +91,49 @@ func main() {
...
-		log.Printf("connection establisted (took %v)", connDuration)
+		log.Infof("connection established (took %v)", connDuration)
 	}
 
 	rpcStart := time.Now()
 	rpcCtx, rpcCancel := context.WithTimeout(ctx, rpcTimeoutDur)
 	defer rpcCancel()
 	resp, err := healthpb.NewHealthClient(conn).Check(rpcCtx, &healthpb.HealthCheckRequest{Service: serviceName})
-	log.Print(resp)
+	log.Infof("response %q", resp)
```
jaypipes: Pretty sure this is not a useful log message, really. Recommend removing it.

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -69,6 +67,10 @@ type NetConf struct {
 
 	// MTU for eth0
 	MTU string `json:"mtu"`
+
+	PluginLogFile string `json:"pluginLogFile"`
+
+	PluginLogLevel string `json:"PluginLogLevel"`
```
jaypipes: Please set the json attribute name to `pluginLogLevel` to match the other names.

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -318,19 +319,16 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 }
 
 func main() {
-	logger.SetupLogger(logger.GetLogFileLocation(defaultLogFilePath))
-
+	log := logger.Default()
 	log.Infof("Starting CNI Plugin %s ...", version)
```
jaypipes: I would actually remove the above log line. It's not correct. :) The CNI plugin is not "started". Kubelet simply executes the CNI plugin, passing an ADD or DEL request to it. So, all the above log line does is clutter log output for kubelet, which will have logs that looks like this:

```
Starting CNI Plugin 1
Received CNI add request: ContainerID(blah) Netns(blah) IfName(eth0) Args(blah) Path(blah) argsStdinData(blah)
Starting CNI Plugin 1
Received CNI add request: ContainerID(blah) Netns(blah) IfName(eth0) Args(blah) Path(blah) argsStdinData(blah)
Starting CNI Plugin 1
Received CNI add request: ContainerID(blah) Netns(blah) IfName(eth0) Args(blah) Path(blah) argsStdinData(blah)
```

pkg/ipamd/ipamd.go
```diff
@@ -326,15 +323,15 @@ func (c *IPAMContext) nodeInit() error {
...
+				log.Warn("Unable to discover attached IPs for ENI from metadata service")
 				ipamdErrInc("waitENIAttachedMaxRetryExceeded")
 				break
 			}
 
 			log.Warnf("Error trying to set up ENI %s: %v", eni.ENIID, err)
 			if strings.Contains(err.Error(), "setupENINetwork: failed to find the link which uses MAC address") {
 				// If we can't find the matching link for this MAC address, there is no point in retrying for this ENI.
-				log.Errorf("Unable to match link for this ENI, going to the next one.")
+				log.Error("Unable to match link for this ENI, going to the next one.")
```
jaypipes: Change to Debugf

scripts/entrypoint.sh
```diff
@@ -72,8 +76,10 @@ cp portmap "$HOST_CNI_BIN_PATH"
 cp aws-cni "$HOST_CNI_BIN_PATH"
 cp aws-cni-support.sh "$HOST_CNI_BIN_PATH"
 
-sed -i s/__VETHPREFIX__/"${AWS_VPC_K8S_CNI_VETHPREFIX:-"eni"}"/g 10-aws.conflist
-sed -i s/__MTU__/"${AWS_VPC_ENI_MTU:-"9001"}"/g 10-aws.conflist
+sed -i s~VETHPREFIX~"${AWS_VPC_K8S_CNI_VETHPREFIX}"~g 10-aws.conflist
+sed -i s~MTU~"${AWS_VPC_ENI_MTU}"~g 10-aws.conflist
+sed -i s~PLUGINLOGFILE~"${AWS_VPC_K8S_PLUGIN_LOG_FILE}"~g 10-aws.conflist
+sed -i s~PLUGINLOGLEVEL~"${AWS_VPC_K8s_PLUGIN_LOG_LEVEL}"~g 10-aws.conflist
```
jaypipes: The above removed the `__` prefix and suffix on the variables being interpolated in the conflist file, so the sed won't work. Not sure why you changed the above from `s/__` to `s~`?

cmd/aws-k8s-agent/main.go
```diff
@@ -16,29 +16,28 @@ package main
...
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/eniconfig"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/k8sapi"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
 )
 
 const (
-	defaultLogFilePath = "/host/var/log/aws-routed-eni/ipamd.log"
+	defaultIPamDLogFilePath = "/host/var/log/aws-routed-eni/ipamd.log"
+	defaultIPamDLogLevel    = "Debug"
```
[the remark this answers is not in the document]
nithu0115: Got it, reverting back to original naming convention.

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -337,7 +320,7 @@ func produceCloudWatchMetrics(t metricsTarget, families map[string]*dto.MetricFa
 		for _, action := range convertMetrics.actions {
 			switch mType {
 			case dto.MetricType_COUNTER:
-				glog.Infof("Produce COUNTER metrics: %s, value: %f", action.cwMetricName, action.data.curSingleDataPoint)
+				log.Infof("Produce COUNTER metrics: %s, value: %f", action.cwMetricName, action.data.curSingleDataPoint)
```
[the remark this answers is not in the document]
nithu0115: Ack.

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -397,13 +378,10 @@ func metricsListGrabAggregateConvert(t metricsTarget) (map[string]*dto.MetricFam
...
 
 	targetList := t.getTargetList()
-	glog.Info("targetList: ", targetList)
-	glog.Info("len(targetList)", len(targetList))
 	for _, target := range targetList {
-		glog.Infof("Grab/Aggregate metrics from %v", target)
 		rawOutput, err := t.grabMetricsFromTarget(target)
 		if err != nil {
-			glog.Errorf("Failed to getMetricsFromTarget: %v", err)
+			log.Info("Failed to getMetricsFromTarget")
```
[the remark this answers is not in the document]
nithu0115: Ack,

cmd/cni-metrics-helper/metrics/metrics.go
nithu0115: Ack

cmd/cni-metrics-helper/metrics/metrics.go
nithu0115: Ack.

cmd/grpc-health-probe/main.go
nithu0115: Ack.

cmd/routed-eni-cni-plugin/cni.go
nithu0115: typo, sorry!

cmd/routed-eni-cni-plugin/cni.go
nithu0115: I agree, I will remove it.

pkg/ipamd/ipamd.go
nithu0115: Noted!

scripts/entrypoint.sh
nithu0115: The reason why i changed sed ~ is because of the below error, where AWS_VPC_K8S_PLUGIN_LOG_FILE=/var/log/aws-routed-eni/plugin.log 

sed -i s/__PLUGINLOGLEVEL__/"${AWS_VPC_K8S_PLUGIN_LOG_FILE}"/g test.txt
sed: -e expression #1, char 23: unknown option to `s

nithu0115: The reason why i changed sed ~ is because of the below error, where AWS_VPC_K8S_PLUGIN_LOG_FILE=/var/log/aws-routed-eni/plugin.log 

sed -i s/__PLUGINLOGLEVEL__/"${AWS_VPC_K8S_PLUGIN_LOG_FILE}"/g test.txt
sed: -e expression #1, char 23: unknown option to `s

nithu0115:
@jaypipes @mogren thanks for the feedback. I made modification as per your comments. PR is ready for your review :)

cmd/aws-k8s-agent/main.go
```diff
@@ -16,29 +16,27 @@ package main
...
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/eniconfig"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/k8sapi"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
 )
 
-const (
-	defaultLogFilePath = "/host/var/log/aws-routed-eni/ipamd.log"
-)
+const binaryName = "L-IPamD"
```
mogren: I think `ipamd` is a fine abbreviation of "IP Address Management Daemon". The Linux part is not needed, and neither the special casing.

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -97,27 +100,16 @@ func getMetricsFromPod(client clientset.Interface, podName string, namespace str
 }
 
 func processGauge(metric *dto.Metric, act *metricsAction) {
-	if act.logToFile {
-		glog.Infof("Label: %v, Value: %v ", metric.GetLabel(), metric.GetGauge().GetValue())
-	} else {
-		glog.V(10).Info("processing GAUGE: ", metric.GetGauge().GetValue())
-	}
```
mogren: 👍 Good cleanup!

cmd/grpc-health-probe/main.go
```diff
@@ -3,11 +3,12 @@ package main
 import (
 	"context"
 	"flag"
-	"log"
 	"os"
 	"os/signal"
 	"time"
 
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
+
```
mogren: I think it's fine that the gRPC health check is just using the standard go library. These logs are not essential. @jaypipes, any comments on this?

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -94,43 +94,53 @@ func init() {
...
-	}
-	if len(conf.VethPrefix) > 4 {
-		return errors.New("conf.VethPrefix can be at most 4 characters long")
-	}
-
-	// MTU
-	if conf.MTU == "" {
-		log.Debug("MTU not set, defaulting to 9001")
-		conf.MTU = "9001"
-	}
```
mogren: I think we need this default.

cmd/routed-eni-cni-plugin/driver/mocks/driver_mocks.go
```diff
@@ -1,4 +1,4 @@
-// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mogren: `// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.` is the correct header. We should remove all years.

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -1,4 +1,4 @@
-// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// Copyright 2017-2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
mogren: `// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.` is the correct header. We should remove all years.

cmd/aws-k8s-agent/main.go
jaypipes: Heh, TIL that the "L-" stands for "Linux"...

Agreed, though, that "ipamd" is a better and more Linux-appropriate daemon name.

If you make this change in a future PR, however, please remember to update https://github.com/awslabs/amazon-eks-ami/blob/master/log-collector-script/linux/eks-log-collector.sh

cmd/grpc-health-probe/main.go
jaypipes: meh, it's fine to leave it here I guess.

cmd/routed-eni-cni-plugin/cni.go
jaypipes: Yes, we do.

cmd/aws-k8s-agent/main.go
nithu0115: > I think `ipamd` is a fine abbreviation of "IP Address Management Daemon". The Linux part is not needed, and neither the special casing.

I wanted it to be consistent with the [log message](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/cmd/aws-k8s-agent/main.go#L43). I can change it to "ipamd".

nithu0115:
Updated, rebased as per your comments @jaypipes @mogren. I am planning to run a benchmark test and will post the results here.

README.md
```diff
@@ -299,7 +299,19 @@ Default: Unset
 
 Valid Values: `stdout` or a file path
 
-Specifies where to write the logging output. Either to stdout or to override the default file.
+Specifies where to write the logging output of L-IPamD. Either to stdout or to override the default file.
```
mogren: Nit: This is not really just "ipamd", is it? It's the plugin binary as well.

pkg/k8sapi/discovery.go
```diff
@@ -256,7 +254,7 @@ func (d *Controller) handlePodUpdate(key string) error {
 		d.workerPodsLock.Lock()
 		defer d.workerPodsLock.Unlock()
 
-		log.Tracef("Update for pod %s: %+v, %+v", podName, pod.Status, pod.Spec)
+		log.Debugf("Update for pod %s: %+v, %+v", podName, pod.Status, pod.Spec)
```
mogren: I think this might become too verbose, since it is printing the full `pod.Spec` on every pod update on the node. The info level log on line 267 is probably enough.

pkg/publisher/publisher.go
```diff
@@ -96,7 +86,16 @@ func New(ctx context.Context) (Publisher, error) {
...
+	clusterID, err := ec2Client.GetClusterTag("CLUSTER_ID")
+	if err != nil || clusterID == "" {
+		log.Infof("Failed to obtain cluster-id, fetching name.  %v", err)
+		clusterID, err = ec2Client.GetClusterTag("Name")
+		if err != nil || clusterID == "" {
+			log.Infof("Failed to obtain cluster-id or name, defaulting to 'k8s-cluster'.  %v", err)
+			clusterID = "k8s-cluster"
+		}
+	}
+	log.Infof("Using cluster ID %s", clusterID)
```
mogren: I think the changes from #856 have been accidentally reverted in this file. That's the lines 56-63, 99 and the `getClusterID` function.

README.md
nithu0115: You are right, thanks for catching that. I updated the readme.

pkg/publisher/publisher.go
nithu0115: Not sure what happened here. I reverted it back. Thanks for catching !

pkg/k8sapi/discovery.go
nithu0115: I would like to leave it as is for now as this will provide us some better info while troubleshooting.

nithu0115:
@mogren @jaypipes, the original 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.6.0 compressed size is (linux/amd64) `106.27 MB` and the new image which includes this commit is `90.19 MB`

jaypipes:
> @mogren @jaypipes, the original 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.6.0 compressed size is (linux/amd64) `106.27 MB` and the new image which includes this commit is `90.19 MB`

So, around a 15% reduction in image size. Awesome.

README.md
```diff
@@ -299,7 +299,29 @@ Default: Unset
...
+
+Default: Unset
+
+Valid Values: `stdout` or a file path
+
+Specifies where to write the logging output for `aws-cni` plugin. Either to stdout or to override the default file (i.e., `/var/log/aws-routed-eni/plugin.log`).
+
+---
+
+`AWS_VPC_K8s_PLUGIN_LOG_LEVEL`
```
jaypipes: lowercase "s" in the above :)

scripts/entrypoint.sh
```diff
@@ -36,9 +36,13 @@ if [ ! -f grpc-health-probe ]; then
...
-AGENT_LOG_PATH=${AGENT_LOG_PATH:-aws-k8s-agent.log}
-HOST_CNI_BIN_PATH=${HOST_CNI_BIN_PATH:-/host/opt/cni/bin}
-HOST_CNI_CONFDIR_PATH=${HOST_CNI_CONFDIR_PATH:-/host/etc/cni/net.d}
+AGENT_LOG_PATH=${AGENT_LOG_PATH:-"aws-k8s-agent.log"}
+HOST_CNI_BIN_PATH=${HOST_CNI_BIN_PATH:-"/host/opt/cni/bin"}
+HOST_CNI_CONFDIR_PATH=${HOST_CNI_CONFDIR_PATH:-"/host/etc/cni/net.d"}
+AWS_VPC_K8S_CNI_VETHPREFIX=${AWS_VPC_K8S_CNI_VETHPREFIX:-"eni"}
+AWS_VPC_ENI_MTU=${AWS_VPC_ENI_MTU:-"9001"}
+AWS_VPC_K8S_PLUGIN_LOG_FILE=${AWS_VPC_K8S_PLUGIN_LOG_FILE:-"/var/log/aws-routed-eni/plugin.log"}
+AWS_VPC_K8s_PLUGIN_LOG_LEVEL=${AWS_VPC_K8S_PLUGIN_LOG_LEVEL:-"Debug"}
```
mogren: Nit: `AWS_VPC_K8s_PLUGIN_LOG_LEVEL` here and on line 85 still has the lower case `s`.

PR #826, opened 2020-01-31
Test wait for update

mogren:
*Issue #, if available:*

*Description of changes:*
* Run integration tests before and after updating the CNI
* Run e2e tests from CircleCI (part of #686)
* Merge config files from release-1.6 branch

### DIFF
--- a/scripts/lib/cluster.sh
+++ b/scripts/lib/cluster.sh
@@ -1,15 +1,19 @@
 #!/usr/bin/env bash
 
 function down-test-cluster() {
+    if [[ -n "${CIRCLE_JOB:-}" ]]; then
+        $TESTER_PATH eks delete cluster --path $CLUSTER_CONFIG || (echo "failed!" && exit 1)
+    else
     echo -n "Deleting cluster $CLUSTER_NAME (this may take ~10 mins) ... "
-    $TESTER_PATH eks delete cluster --path $CLUSTER_CONFIG >>$CLUSTER_MANAGE_LOG_PATH 2>&1 ||
-        ( echo "failed. Check $CLUSTER_MANAGE_LOG_PATH." && exit 1 )
+    $TESTER_PATH eks delete cluster --path $CLUSTER_CONFIG >> $CLUSTER_MANAGE_LOG_PATH 2>&1 ||
+        (echo "failed. Check $CLUSTER_MANAGE_LOG_PATH." && exit 1)
     echo "ok."
+    fi
 }
 
 function up-test-cluster() {
+    echo -n "Configuring cluster $CLUSTER_NAME"
     ssh-keygen -q -P cni-test -f $SSH_KEY_PATH
-
     AWS_K8S_TESTER_EKS_CLUSTER_NAME=$CLUSTER_NAME \
         AWS_K8S_TESTER_EKS_KUBECONFIG_PATH=$KUBECONFIG_PATH \
         AWS_K8S_TESTER_EKS_KUBERNETES_VERSION=${K8S_VERSION%.*} \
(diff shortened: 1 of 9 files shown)


scripts/lib/cluster.sh
```diff
@@ -1,15 +1,19 @@
 #!/usr/bin/env bash
 
 function down-test-cluster() {
+    if [[ -n "${CIRCLE_JOB:-}" ]]; then
```
jaypipes: Why would you want to disable the cluster delete when running in CI?

scripts/lib/cluster.sh
```diff
@@ -28,9 +32,14 @@ function up-test-cluster() {
...
+        $TESTER_PATH eks create config --path $CLUSTER_CONFIG 1>&2
+
+    if [[ -n "${CIRCLE_JOB:-}" ]]; then
+        $TESTER_PATH eks create cluster --path $CLUSTER_CONFIG || (echo "failed!" && exit 1)
+    else
+        echo -n "Creating cluster $CLUSTER_NAME (this may take ~20 mins. details: tail -f $CLUSTER_MANAGE_LOG_PATH)... "
+        $TESTER_PATH eks create cluster --path $CLUSTER_CONFIG 2>> $CLUSTER_MANAGE_LOG_PATH 1>&2 ||
+            (echo "failed. Check $CLUSTER_MANAGE_LOG_PATH." && exit 1)
+        echo "ok."
+    fi
```
jaypipes: I'm confused why you're changing the above depending on whether it's running in CI or not?

scripts/run-integration-tests.sh
```diff
@@ -138,10 +138,25 @@ cp "$BASE_CONFIG_PATH" "$TEST_CONFIG_PATH"
...
+GO111MODULE=on go test -v -timeout 0 ./... --kubeconfig=$KUBECONFIG --ginkgo.focus="\[cni-integration\]" --ginkgo.skip="\[Disruptive\]" \
+    --assets=./assets
+TEST_PASS=$?
+popd
+
+echo "*******************************************************************************"
+echo "Updating CNI to image $IMAGE_NAME:$TEST_IMAGE_VERSION"
+$KUBECTL_PATH apply -f "$TEST_CONFIG_PATH"
+
+echo "Sleping for 110s"
```
jaypipes: "Sleeping"

scripts/run-integration-tests.sh
```diff
@@ -138,10 +138,25 @@ cp "$BASE_CONFIG_PATH" "$TEST_CONFIG_PATH"
...
+echo "TODO: Poll and wait for updates to complete instead!"
+sleep 110
```
jaypipes: why 110 seconds? is that a magic value? :)

scripts/lib/cluster.sh
mogren: Just making it more verbose. See comment below.

scripts/lib/cluster.sh
mogren: It's just changing the logging output. By default, CircleCI will terminate any test after 10 minutes fi there is no console output. That can be changed by setting `no_output_timeout` for the job, but when you have to wait 40 minutes it's more interesting to see what is going on.

scripts/run-integration-tests.sh
mogren: Yes, for now. The test cluster has 3 nodes, and the rolling update is using the default grace period of 30s, so it takes around 30s to shut down each node, and another 5-10 seconds to start up. It's ok to start the test once all old ipamd pods have shut down.

scripts/lib/cluster.sh
jaypipes: ack, makes sense. ++

PR #832, opened 2020-02-06
Clean up healthcheck

mogren:
*Description of changes:*
* Use `127.0.0.1` everywhere
* Renamed some variables
* Updated comments

### DIFF
--- a/pkg/ipamd/rpc_handler.go
+++ b/pkg/ipamd/rpc_handler.go
@@ -30,11 +30,12 @@ import (
 
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd/datastore"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/k8sapi"
-	pb "github.com/aws/amazon-vpc-cni-k8s/rpc"
+	"github.com/aws/amazon-vpc-cni-k8s/rpc"
 )
 
 const (
-	ipamdgRPCaddress = "127.0.0.1:50051"
+	ipamdgRPCaddress      = "127.0.0.1:50051"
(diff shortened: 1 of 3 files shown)


pkg/ipamd/rpc_handler.go
```diff
@@ -98,37 +99,38 @@ func (s *server) DelNetwork(ctx context.Context, in *pb.DelNetworkRequest) (*pb.
...
 		log.Errorf("Failed to listen gRPC port: %v", err)
 		return errors.Wrap(err, "ipamd: failed to listen to gRPC port")
 	}
-	s := grpc.NewServer()
-	pb.RegisterCNIBackendServer(s, &server{ipamContext: c})
-	hs := health.NewServer()
-	// TODO: Implement watch once the status is check is handled correctly.
-	hs.SetServingStatus("grpc.health.v1.aws-node", healthpb.HealthCheckResponse_SERVING)
-	healthpb.RegisterHealthServer(s, hs)
+	grpcServer := grpc.NewServer()
```
jaypipes: you don't like those short variable names, huh? :)

Personally, I follow the principle that the smaller the scope of the variable, the shorter the name should be. In this case, the scope of these variables is very small (single-function scope with the function only about ten lines) so I don't mind the variable names being one or two letters. For global or package-scoped variables or variables used in lengthy (20+ lines) functions and/or passed between multiple functions I usually insist on lengthy and explicit variable names

mogren: I mostly agree, and for for-loops and similar I use short variables. I guess I'm still slightly Java-institutionalized... 😄

Issue #775, opened 2019-12-28
cni-metrics-helper: Replace CLUSTER_ID with eks:cluster-name tag


groodt:
It seems that the "official" EKS tags applied automatically by EKS node-groups is `eks:cluster-name` and not `CLUSTER_ID` or `Name`.

Would you accept a PR to replace either or both of these values with `eks:cluster-name`?

If not, I believe that cni_metrics_helper is unusable with EKS managed nodegroups as they do not allow custom tags such as `CLUSTER_ID` or `Name` to be propagated to the EC2 instances.

mogren:
@groodt Thanks, sounds like an improvement we should do, and all PRs are welcome. 😄

manfredlift:
Any updates on this? We are considering using managed workers (EKS node groups) instead of unmanaged workers, but this makes the transition difficult if we don't want to give up CNI metrics.

groodt:
Hey @manfredlift 

I haven't had any time to look at this, but hope to soon. I'm also happy if somebody else wishes to pick it up. 

The location in the code is here I believe:
https://github.com/aws/amazon-vpc-cni-k8s/blob/50a1b0c279c731bd0323cc02ceef6ba069d8ea51/pkg/publisher/publisher.go#L87

There is also a chance that EKS allows us to propagate tags to instances in managed node groups in future, so it could also get "fixed" on that end. I think it would be cleaner if the cni-metrics-helper used the "standard" tags though, because then all you need to do is install the CNI plugin and cni metrics helper to an EKS cluster and it should "just" work.

mogren:
Fixed in #856

PR #856, opened 2020-02-28
Use eks:cluster-name as clusterId

groodt:
*Issue #775*

*Description of changes:*
Prioritise `eks:cluster-name` as clusterId for cni-metrics-helper. This is the tag that EKS node-groups is automatically adding.


pkg/publisher/publisher.go
```diff
@@ -207,6 +198,24 @@ func (p *cloudWatchPublisher) getCloudWatchMetricNamespace() *string {
...
+			clusterID, err = ec2Client.GetClusterTag("Name")
+			if err != nil || clusterID == "" {
+				glog.Errorf("Failed to determine cluster-id, defaulting to 'k8s-cluster'.  %v", err)
+				clusterID = "k8s-cluster"
+			}
+		}
+	}
+	glog.Info("Using cluster ID ", clusterID)
+	return clusterID
+}
```
jaypipes: Hi @groodt thanks very much for the contribution! May I suggest reworking the above to be a little cleaner? You can use a const to list the tag names to try, in order, and then loop through the tag names like so:

```go
const (
        defaultClusterID = "k8s-cluster"
        clusterIDTags = []string{
                "eks:cluster-name",
                "CLUSTER_ID",
                "Name",
        }
)
...
func getClusterID(ec2Client *ec2wrapper.EC2Wrapper) string {
        var clusterID string
        var err error
        for _, tag := range clusterIDTags {
	        clusterID, err = ec2Client.GetClusterTag(tag)
	        if err != nil && clusterID != "" {
                        break
		}
	}
        if clusterID == "" {
                clusterID = defaultClusterID
        }
	glog.Info("Using cluster ID ", clusterID)
	return clusterID
}
```

groodt: Thanks for the suggestion. That is cleaner. I'll make the changes ASAP.

groodt: I've made the change. Slices can't be `const`, so I've inlined the variables into the function. Open to alternatives. Just let me know what would be the preference.

jaypipes: Yes, forgot about that. You could have either just replaced `const` with `var`, but this is perfectly fine as well :)

pkg/publisher/publisher.go
```diff
@@ -207,6 +198,32 @@ func (p *cloudWatchPublisher) getCloudWatchMetricNamespace() *string {
 	return aws.String(cloudwatchMetricNamespace)
 }
 
+func getClusterId(ec2Client *ec2wrapper.EC2Wrapper) string {
```
jaypipes: Tiny nit... could you name this `getClusterID()` (capitalized ID) please?

groodt:
Changes in. PTAL.

groodt:
@jaypipes 

I just realised that I think the condition should be:
```
		if err == nil && clusterID != "" {
			break
		}
```

groodt:
Thanks @jaypipes 

Will this get merged by a collaborator automatically at some stage, or do I need to assign somebody specific?

PR #863, opened 2020-03-10
Cleanup code autogen

anguslees:
*Description of changes:*

Refactor/cleanup only.  Should have zero behavioural change.

- Add a `make generate` target that just runs `go generate`.

- Autogenerate protobuf files (`rpc/rpc.pb.go`).  For reference, I used:
  ```console
  % protoc --version
  libprotoc 3.0.0
  ```

- Change mockgen invocation to execute the version of mockgen that
  matches the version we will compile against.
  (ie: Use `go run github.com/golang/mock/mockgen ...`)

- Update to mockgen 1.4.1

Reviewer: I recommend reviewing this by individual commits.  The last commit is the result of rerunning `make generate`.


mogren:
Thanks @anguslees! It seems like the generated files have some different formatting, that's why the build failed:

`Makefile:139: recipe for target 'check-format' failed`

A `make format` on the generated commit should fix it.

PR #866, opened 2020-03-13
Preliminary support for adding multiple aws-cni interfaces to a single pod

hallsaa:
This does a few things:
1. It allows for a worker node to interpret multiple ENIConfigs
2. It then causes the worker node to attempt to bring up an ENI for each ENIConfig
3. It allows multus to attach multiple interfaces to a single pod
4. When installing a pod interface, allow for an ENIConfig to be specified to choose the IP to use.
5. Add routes for local subnets in the container.

*Issue #, if available:*

*Description of changes:*


jaypipes:
@hallsaa just FYI, we haven't forgotten about you :) We will do a full review on this PR probably next week (which should give you some time to track down the bugs appearing in the e2e tests).

Our top priority for this PR is to ensure it is safely adding new functionality (multi-veth support) while ensuring nothing breaks for the single-veth use cases.

pkg/ipamd/datastore/data_store.go
```diff
@@ -176,7 +182,17 @@ func NewDataStore(log logger.Logger) *DataStore {
 }
 
 // AddENI add ENI to data store
-func (ds *DataStore) AddENI(eniID string, deviceNumber int, isPrimary bool) error {
+func (ds *DataStore) AddENI(eniID string, deviceNumber int, isPrimary bool, eniConfigName string) error {
+	addr := net.IPNet{
+		IP:	net.ParseIP("0.0.0.0"),
+		Mask: net.CIDRMask(32, 32),
+	}
```
mogren: What is this `addr` needed for? A default CIDR mask?

pkg/ipamd/ipamd.go
```diff
@@ -644,12 +651,109 @@ func (c *IPAMContext) updateLastNodeIPPoolAction() {
...
 
-func (c *IPAMContext) tryAllocateENI() error {
+
+// For each ENIConfig we want to make sure there is at least
+// one matching interface attached
+func (c *IPAMContext) globalAllocateENI(enis []awsutils.ENIMetadata) error {
+
+	if c.eniConfig == nil {
+		return fmt.Errorf("eniConfig does not exist")
+	}
```
mogren: When would this be nil? Doesn't it get set to `"default"` by default?

pkg/ipamd/ipamd.go
```diff
@@ -644,12 +651,109 @@ func (c *IPAMContext) updateLastNodeIPPoolAction() {
...
+		log.Debugf("Existing ENI %s, subnet %s, sgAll %s", eni.ENIID, eni.SubnetId, securityGroups)
+		existingENIs = append(existingENIs,v1alpha1.ENIConfigSpec{
+			SecurityGroups: securityGroups,
+			Subnet:         eni.SubnetId,
+		})
+	}
+	log.Debugf("existingENIs %s", existingENIs)
+
+	// TODO - This needs to filter down to just the ENIConfigs tagged to this specific host
+	// for now it configures for all ENIConfigs present
```
mogren: Yes, we should test this on some large clusters.

pkg/ipamd/datastore/data_store.go
hallsaa: This is a default mask to provide backwards compatibility for the large number of callers to AddENI. I'll add a comment to make that clear.

pkg/ipamd/ipamd.go
hallsaa: Are there existing tests to help facilitate this?

hallsaa: And is additional filtering a merge blocker for this change?

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -131,12 +137,29 @@ func (createVethContext *createVethPairContext) run(hostNS ns.NetNS) error {
 		return errors.Wrap(err, "setup NS network: failed to add default gateway")
 	}
 
+
+
 	// Add a default route via dummy next hop(169.254.1.1). Then all outgoing traffic will be routed by this
 	// default route via dummy next hop (169.254.1.1).
-	if err = createVethContext.ip.AddDefaultRoute(gwNet.IP, contVeth); err != nil {
+	// If this is not the first interface, the route may already exist
+	if err = createVethContext.ip.AddDefaultRoute(gwNet.IP, contVeth); err != nil && err.Error() != "file exists"{
```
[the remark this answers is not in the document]
hallsaa: This is from the additional interfaces calling it again. I think the system only allows one "default" route to exist, even if the interface is different.

pkg/ipamd/datastore/data_store.go
hallsaa: Actually, this was only being used in tests at this point, I've updated them to bypass this shim function and deleted it.

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -242,10 +256,11 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 }
 
 // generateHostVethName returns a name to be used on the host-side veth device.
+// Note: the maximum length for linux interface name is 15
 func generateHostVethName(prefix, namespace, podname string) string {
 	h := sha1.New()
 	h.Write([]byte(fmt.Sprintf("%s.%s", namespace, podname)))
-	return fmt.Sprintf("%s%s", prefix, hex.EncodeToString(h.Sum(nil))[:11])
+	return fmt.Sprintf("%s%s", prefix, hex.EncodeToString(h.Sum(nil)))[:15]
```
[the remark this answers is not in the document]
hallsaa: I think keeping a cap here is useful to make sure some of the container id is part of the name to keep pods from having the same interface name. How long that limit should be, I'm not sure, so I didn't make a change here.

pkg/ipamd/ipamd.go
mogren: I don't think so, but I'll be sure to try running a custom network config setup on a somewhat large cluster to verify this first.

pkg/ipamd/ipamd.go
hallsaa: I don't think there is a good reason for this to happen today, I just put in the test as an additional sanity check.

hallsaa:
Replacing with https://github.com/aws/amazon-vpc-cni-k8s/pull/974 into a multi-attachment specific branch.

Issue #867, opened 2020-03-14
Calico config is using deprecated node selector.


Gmagella:
Looking at AWS instructions on how to install calico on EKS [here](https://docs.aws.amazon.com/eks/latest/userguide/calico.html) its pointing to amazon-vpc-cni-k8s/release-1.5/config/v1.5/calico.yaml which have the node selector as such:
`kind: DaemonSet
...
    spec:
      nodeSelector:
        beta.kubernetes.io/os: linux`
Which is deprecated, and should be replaced to 
`kind: DaemonSet
...
    spec:
      nodeSelector:
        kubernetes.io/os: linux`

jaypipes:
I agree that the `beta.kubernetes.io/os: linux` annotation is deprecated, however please keep in mind the version of the CNI does not correspond to the version of Kubernetes being targeted. Because we don't know which version of Kubernetes the CNI plugin is being installed in, it is safer to have both the deprecated beta and non-beta current annotation in a nodeAffinity selector and use an OR strategy like so:

```yaml
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
              - matchExpressions:
                 # Remove this match expression when we're sure the CNI isn't
                 # installed anywhere that uses kubelet's that label Node objects
                 # with the deprecated beta label...
                  - key: "beta.kubernetes.io/os"
                    operator: In
                    values:
                      - linux
                  - key: "beta.kubernetes.io/arch"
                    operator: In
                    values:
                      - amd64
                  - key: eks.amazonaws.com/compute-type
                    operator: NotIn
                    values:
                      - fargate
              - matchExpressions:
                  - key: "kubernetes.io/os"
                    operator: In
                    values:
                      - linux
                  - key: "beta.kubernetes.io/arch"
                    operator: In
                    values:
                      - amd64
                  - key: eks.amazonaws.com/compute-type
                    operator: NotIn
                    values:
                      - fargate
```

I believe both the `calico.yaml` and the `aws-k8s-cni.yaml` files in the config directories should be updated to the above and be consistent. I would welcome a pull request from you that made the above change, @Gmagella! :)

mogren:
Root cause was lack of resources, see https://github.com/aws/amazon-vpc-cni-k8s/pull/868#issuecomment-599286497

PR #868, opened 2020-03-14
Fixing node Selector on Calico.yml for 1.15 and 1.16 version of CNI

gmagella-ca:
*Issue #, if available:* 867

*Description of changes:* Just removing "beta" from node Selector. This use was deprecated according to: https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/

### DIFF
--- a/config/v1.5/calico.yaml
+++ b/config/v1.5/calico.yaml
@@ -21,7 +21,7 @@ spec:
     spec:
       priorityClassName: system-node-critical
       nodeSelector:
-        beta.kubernetes.io/os: linux
+        kubernetes.io/os: linux
       hostNetwork: true
       serviceAccountName: calico-node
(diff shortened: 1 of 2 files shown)


jaypipes:
@gmagella-ca Unfortunately, the version of the CNI is 1.5 and 1.6 (not 1.15 and 1.16). The CNI versions do *not* correspond to the version of Kubernetes intended as a target for installation. I think it's safer to wait for probably another 6-12 months before we remove the deprecated beta annotations. That would be the (rough) timeframe when EKS would no longer support versions of Kubernetes that used the deprecated beta annotation. Closing this PR for now.

gmagella-ca:
Understood. It makes sense. 
I thought the nodes were not getting the beta.kubernetes annotation but it turns out they're getting both, and my issue (#867) root cause was actually lack of resources on the cluster and not the annotation containing "beta."

Issue #872, opened 2020-03-17
IPAM connectivity failed when upgrading from v1.5.5 to 1.6.0


laghao:
I updated my eks cluster to 1.15.10 and that worked.
Then I tried to update the cni-k8s from v1.5.5 to v1.6.0 on my test k8s test nodes(2) and as it's a daemonset I had 1 aws-node running and the other having following error:
```
kubectl logs -f pod/aws-node-cjqwm -nkube-system
starting IPAM daemon in background ... ok.
checking for IPAM connectivity ...  failed.
timed out waiting for IPAM daemon to start.
```

I delete the pod but it's still having the same Error:
```
kubectl get po --all-namespaces

NAMESPACE     NAME                             READY   STATUS    RESTARTS   AGE
kube-system   aws-node-22mnl                   1/1     Running   0          15m
kube-system   aws-node-h6nrx                   0/1     Running   3          3m9s
```

More details:
```
kubectl describe po aws-node-h6nrx -nkube-system

Events:
  Type     Reason     Age                    From                                                   Message
  ----     ------     ----                   ----                                                   -------
  Normal   Scheduled  4m49s                  default-scheduler                                      Successfully assigned kube-system/aws-node-h6nrx to ip-10-1-46-183.eu-central-1.compute.internal
  Warning  Unhealthy  3m33s                  kubelet, ip-10-1-46-183.eu-central-1.compute.internal  Readiness probe errored: rpc error: code = Unknown desc = container not running (c542f67fbf22592a6840faa98cd3e9f1c774efeead2a6068319b0488570a903f)
  Warning  Unhealthy  2m39s                  kubelet, ip-10-1-46-183.eu-central-1.compute.internal  Liveness probe failed: timeout: failed to connect service ":50051" within 1s
  Normal   Pulling    2m18s (x4 over 4m48s)  kubelet, ip-10-1-46-183.eu-central-1.compute.internal  Pulling image "602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.6.0"
  Normal   Pulled     2m17s (x4 over 4m47s)  kubelet, ip-10-1-46-183.eu-central-1.compute.internal  Successfully pulled image "602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.6.0"
[...]

Issue #865, opened 2020-03-13
Pod restarting at node startup after upgrading to Kubernetes 1.15


nachomillangarcia:
`aws-node` pods restart 1 time in every node startup, with this message in logs:

```
starting IPAM daemon in background ... ok.
checking for IPAM connectivity ...  failed.
timed out waiting for IPAM daemon to start.
```

This behavior has started just after upgrading EKS cluster to `1.15.10`. Specifically, I think the problem is in `kube-proxy` version `1.15.10`. I downgraded `kube-proxy` to `1.14.7` and the issue stopped. After upgrading it again to `1.15.10`, it restarted.

In the IPAMD logs on the node, I see the service restarting always at this point:

```
2020-03-13T11:16:54.321Z [INFO] 	Setting up host network... 
2020-03-13T11:16:54.322Z [DEBUG] 	Trying to find primary interface that has mac : xx:xx:xx:xx:xx:xx
2020-03-13T11:16:54.322Z [DEBUG] 	Discovered interface: lo, mac: 
2020-03-13T11:16:54.322Z [DEBUG] 	Discovered interface: eth0, mac: xx:xx:xx:xx:xx:xx
2020-03-13T11:16:54.322Z [INFO] 	Discovered primary interface: eth0
2020-03-13T11:16:54.322Z [DEBUG] 	Setting RPF for primary interface: /proc/sys/net/ipv4/conf/eth0/rp_filter
2020-03-13T11:16:54.323Z [DEBUG] 	Found the Link that uses mac address xx:xx:xx:xx:xx:xx and its index is 2 (attempt 1/5)


2020-03-13T11:16:55.840Z [INFO] 	Starting L-IPAMD v1.6.0  ...
2020-03-13T11:16:56.160Z [INFO] 	Testing communication with server
2020-03-13T11:16:56.161Z [INFO] 	Running with Kubernetes cluster version: v1.15+. git version: v1.15.10-eks-bac369. git tree state: clean. commit: bac3690554985327ae4d13e42169e8b1c2f37226. platform: linux/amd64
```

Anybody getting the same issue after upgrading? Thanks!

nithu0115:
Similar issue #872

jaypipes:
I'm wondering if the kube-proxy thing might be a red herring. Perhaps we just need to up the timeout for k8s API connectivity from 30 seconds to something like 1 minute?

That line:

```
checking for IPAM connectivity ...  failed.
```

[...]

Issue #625, opened 2019-09-25
Cannot scale new nodes with CNI version 1.5.3


cgkades:
When trying to scale new nodes after creating an EKS cluster, the new nodes will not leave the NotReady state. The aws-node pod is in a CrashLoopBackOff state.

This seems to be from the file 10-aws.conflist not being placed on the node. It looks like the install script had a line removed that added the file, and it was then moved to main.go. The problem is that it never reaches the code to copy the file

```
bash-4.2# cat /host/var/log/aws-routed-eni/ipamd.log.2019-09-25-20
2019-09-25T20:01:32.800Z [INFO] Starting L-IPAMD v1.5.3  ...
2019-09-25T20:02:02.803Z [INFO] Testing communication with server
2019-09-25T20:02:32.803Z [INFO] Failed to communicate with K8S Server. Please check instance security groups or http proxy setting
2019-09-25T20:02:32.803Z [ERROR]        Failed to create client: error communicating with apiserver: Get https://172.20.0.1:443/version?timeout=32s: dial tcp 172.20.0.1:443: i/o timeout
2019-09-25T20:02:33.901Z [INFO] Starting L-IPAMD v1.5.3  ...
2019-09-25T20:03:03.903Z [INFO] Testing communication with server
```
I apologize but I wasn't able to dig in far enough to know why it needs the 10-aws.conflist in /etc/cni/net.d first, but that does seem to be the issue.

matthsmi:
To expand on what @cgkades is saying...

We've found is that `k8sapi.CreateKubeClient()` fails with `Failed to create client: error communicating with apiserver: Get https://172.20.0.1:443/version?timeout=32s: dial tcp 172.20.0.1:443: i/o timeout` (https://github.com/aws/amazon-vpc-cni-k8s/blob/release-1.5.3/main.go#L47-L51). 

As such, it never gets to copy over the `/etc/cni/10-aws.conflist`. If we manually create the `10-aws.conflist` all is right and aws-node starts up.

This behavior isn't in 1.5.1 and appears to be introduced in commit: 7fd7c93b5b3be2495b2b864d3a992840d4942a52

mogren:
@matthsmi @cgkades Thanks for opening a new ticket. (#283 is not related to this case)

[...]

PR #874, opened 2020-03-18
Remove timeout for ipamd startup

jaypipes:
Note: Edited by @mogren

Disable the timeout when checking the aws-k8s-agent (ipamd) startup in the entrypoint.sh script. 

Try to talk to ipamd once every second. Reopened and changed because of #1028. We now rely on the liveness probe instead.

Related: #625, #865 #872 


mogren:
I think we should remove the readiness probe instead. Also, kube-proxy should definitely be up within 30 seconds.

anguslees:
Why would this need to be configurable, rather than just picking a better (bigger?) hardcoded value?

mogren:
@anguslees I agree, I just reopened this old PR because of #1028. I'll change it to just keep retrying until the liveness probe kills it.

PR #892, opened 2020-04-02
Add missing permissions in typha-cpha sa (Calico)

marcincuber:
*Description of changes:*

Add missing permission to Calico service account typha-cpha. Fixes permissions issues visible in `calico-typha-horizontal-autoscaler` pod logs

### DIFF
--- a/config/v1.6/calico.yaml
+++ b/config/v1.6/calico.yaml
@@ -634,7 +634,7 @@ metadata:
 rules:
   - apiGroups: [""]
     resources: ["nodes"]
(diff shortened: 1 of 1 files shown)


marcincuber:
@mogren thanks for approving. `extensions` API is commonly used for deployments and daemonsets. I try to keep up with updates and I switched to using `apps/v1` while back. In this case you have to maintain permissions for both `extenstion` and `apps`. 

I can confirm that this calico config works with EKS 1.15

Issue #796, opened 2020-01-17
Run aws-node as unprivileged pod


mogren:
We currently set the `aws-node` pod to be `privileged`, but that might not be needed.

Check if  `CAP_NET_ADMIN` and `CAP_DAC_OVERRIDE` is enough to set the RDF check to loose and copy the binary and config file:

https://github.com/aws/amazon-vpc-cni-k8s/blob/1ee59a06c597c62aa93e85fc1d2eedef553707a3/pkg/networkutils/network.go#L233-L246

mogren:
Unfortunately, having the capabilities limited to `["NET_ADMIN", "DAC_OVERRIDE"]` is not enough to write to `/proc/sys/net/*`.  From the logs:
```
[DEBUG] Setting RPF for primary interface: /proc/sys/net/ipv4/conf/eth0/rp_filter
[ERROR] Failed to set up host networkfailed to configure eth0 RPF check: open /proc/sys/net/ipv4/conf/eth0/rp_filter: read-only file system
[ERROR] Initialization failure: ipamd init: failed to set up host network: failed to configure eth0 RPF check: open /proc/sys/net/ipv4/conf/eth0/rp_filter: read-only file system
```

SaranBalaji90:
There are few ways to achieve this,

1) Set the rp_filter through init container.
```
    initContainers:
      - command:
        - sh
        - -c
        - sysctl net.ipv4.conf.eth0.rp_filter=2
        image: golang:1.13-stretch
        name: rp_filter_setting
        securityContext:
          privileged: true
```
2) Set the rp_filter when the ec2 instance starts up. But the problem with this approach is, if anyone builds custom AMI then they have to make sure this change is added to their AMI build scripts.
3) Enabling unsafe sysctl - https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/#enabling-unsafe-sysctls but this doesn't help us, because https://github.com/kubernetes/kubernetes/blob/7f23a743e8c23ac6489340bbb34fa6f1d392db9d/pkg/kubelet/sysctl/whitelist.go#L89 kubelet rejects the pod allocation when net.* sysctl is used along with host network. 

This leave us with option-1 being more suitable for our usecase.

Proposed solution:
1) Add a flag that instructs aws-node pod whether to perform or skip setting rp_filter value for eth0.
2) Add init container as described above and remove "privileged"  mode for aws-node.
[...]

PR #901, opened 2020-04-06
Update rp_filter only when /proc is write accessible

SaranBalaji90:
*Issue #, if available:* https://github.com/aws/amazon-vpc-cni-k8s/issues/796

*Description of changes:*
This change is required to remove privileged permission for aws-node pods.

Once this is merged, in the next release we can remove privileged permission for aws-node and perform the operation through init container as described in the Issue link above. 

### DIFF
--- a/pkg/networkutils/network.go
+++ b/pkg/networkutils/network.go
@@ -243,10 +243,16 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDR *net.IPNet, vpcCIDRs []*string, 
 		primaryIntfRPFilter := "net/ipv4/conf/" + primaryIntf + "/rp_filter"
 		const rpFilterLoose = "2"
 
-		log.Debugf("Setting RPF for primary interface: %s", primaryIntfRPFilter)
-		err = n.procSys.Set(primaryIntfRPFilter, rpFilterLoose)
-		if err != nil {
-			return errors.Wrapf(err, "failed to configure %s RPF check", primaryIntf)
+		if n.procSys.IsPathWriteAccessible(primaryIntfRPFilter) {
+			// Setting RPF will be removed from aws-node when we bump our minor version.
+			log.Debugf("Setting RPF for primary interface: %s", primaryIntfRPFilter)
+			err = n.procSys.Set(primaryIntfRPFilter, rpFilterLoose)
+			if err != nil {
+				return errors.Wrapf(err, "failed to configure %s RPF check", primaryIntf)
+		    }
+		} else {
+			// when aws-node run as an un-privileged pod, /proc will be mounted as read only
+			log.Infof("Skip updating RPF for primary interface: %s", primaryIntfRPFilter)
 		}
 	}
 
--- a/pkg/procsyswrapper/procsys.go
+++ b/pkg/procsyswrapper/procsys.go
@@ -15,11 +15,14 @@ package procsyswrapper
 
 import (
 	"io/ioutil"
+
+	"golang.org/x/sys/unix"
 )
 
 type ProcSys interface {
 	Get(key string) (string, error)
 	Set(key, value string) error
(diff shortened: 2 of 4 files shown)


pkg/networkutils/network.go
```diff
@@ -243,10 +243,16 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDR *net.IPNet, vpcCIDRs []*string,
...
-			return errors.Wrapf(err, "failed to configure %s RPF check", primaryIntf)
+		if n.procSys.IsPathWriteAccessible(primaryIntfRPFilter) {
+			// Setting RPF will be removed from aws-node when we bump our major version.
+			log.Debugf("Setting RPF for primary interface: %s", primaryIntfRPFilter)
+			err = n.procSys.Set(primaryIntfRPFilter, rpFilterLoose)
+			if err != nil {
+				return errors.Wrapf(err, "failed to configure %s RPF check", primaryIntf)
+		    }
+		} else {
+			// if aws-node run as un-privileged pods then /proc will be mounted as read only.
```
mogren, line 254: Nit: `// If aws-node is running as an un-privileged pod, /proc will be mounted as read only.`

pkg/procsyswrapper/procsys.go
```diff
@@ -42,3 +45,8 @@ func (p *procSys) Get(key string) (string, error) {
 func (p *procSys) Set(key, value string) error {
 	return ioutil.WriteFile(p.path(key), []byte(value), 0644)
 }
+
+// IsPathWriteAccessible verifies if aws-node pod can write to the specified path
+func (p *procSys) IsPathWriteAccessible(key string) bool {
+	return unix.Access(p.prefix + key, unix.W_OK) != nil
```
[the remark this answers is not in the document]
SaranBalaji90: I just force pushed a change few mins back that should address this.

pkg/networkutils/network.go
```diff
@@ -243,10 +243,16 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDR *net.IPNet, vpcCIDRs []*string,
 		primaryIntfRPFilter := "net/ipv4/conf/" + primaryIntf + "/rp_filter"
 		const rpFilterLoose = "2"
 
-		log.Debugf("Setting RPF for primary interface: %s", primaryIntfRPFilter)
-		err = n.procSys.Set(primaryIntfRPFilter, rpFilterLoose)
-		if err != nil {
-			return errors.Wrapf(err, "failed to configure %s RPF check", primaryIntf)
+		if n.procSys.IsPathWriteAccessible(primaryIntfRPFilter) {
+			// Setting RPF will be removed from aws-node when we bump our major version.
```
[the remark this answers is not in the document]
SaranBalaji90: I meant when we do 1.7.0

pkg/procsyswrapper/procsys.go
```diff
@@ -42,3 +45,8 @@ func (p *procSys) Get(key string) (string, error) {
 func (p *procSys) Set(key, value string) error {
 	return ioutil.WriteFile(p.path(key), []byte(value), 0644)
 }
+
+// IsPathWriteAccessible verifies if aws-node pod can write to the specified path
+func (p *procSys) IsPathWriteAccessible(key string) bool {
```
anguslees, line 50: This encourages TOCTOU (time of check is not time of use) race attacks.  A more robust pattern is to just attempt the real `Set` and see if the error result is "permission denied" (`os/error.IsPermission(err)` in golang).

SaranBalaji90: oops looks like I missed this :/ Here TOCTOU is not possible because when aws-node starts up it will either have /proc accessible or not depending on privileged flag. It won't change during the execution. Please correct me if otherwise.

Issue #737, opened 2019-12-02
VPC CIDR is cached forever causing pod routing issues when secondary VPC CIDR ranges are created afterwards


nithu0115:
Problem statement:
As of today, VPC CIDR ranges are cached during initialization. If new CIDR ranges are added afterwards to address IP space issue, the CNI should be restarted to fetch new CIDR ranges to update the cache to add ip rules/routes to reach to other pods in the cluster with new subnet IP range. 

Solution:
Refresh VPC CIDR ranges cache every 2 seconds to avoid staleness. 

Steps to replicate the issue:
1) Create EKS cluster in a VPC which has just one CIDR range (10.10.0.0/16)
2) Create worker nodes in the above VPC CIDr range 
3) Add secondary CIDR range (100.10.0.0/16) to existing VPC 
4) Launch new worker nodes in the subnet which has 100.10.0.0/24 CIDR 
5) Pod 1 with 100.10.12.13 IP part of secondary VPC CIDR subnet cannot be talk to coreDNS pod with IP 10.10.23.45 which is part of primary VPC CIDR subnet.

jaypipes:
Every two seconds may be a bit excessive. Certainly we should refresh the VPC CIDR ranges periodically, but I think 15 or 30 seconds might be a better interval. Alternately, is there a way we can be notified of VPC subnets being created, deleted or modified instead of periodically refreshing our view?

nithu0115:
> Every two seconds may be a bit excessive. Certainly we should refresh the VPC CIDR ranges periodically, but I think 15 or 30 seconds might be a better interval. 

Agreed! I will go with 15 seconds then.  

>Alternately, is there a way we can be notified of VPC subnets being created, deleted or modified instead of periodically refreshing our view?

I did some research around adding a watch/poll to get notifications about VPC changes without making AWS API calls, but did not find a way.

0xlen:
Before getting the rollout for the fix in next couple versions, I would to post the couple suggestions/workarounds for users having the cache issue after adding secondary VPC CIDR:

- Restart the CNI Plugin should remedy this issue. This can be done by adding label or update ENV of CNI Plugin to trigger this update behavior.
- For production usage concern, it would be great if can launch new work node and so new running `aws-node` Pod should aware the secondary VPC CIDR, then replace the old one to ensure it won't impact existing environment.

M00nF1sh:
[...]

PR #903, opened 2020-04-07
refresh subnet/CIDR information periodically

nithu0115:
*Issue #, if available:* #737 

*Description of changes:* 
1) Added forever go-routine to update VPC CIDR blocks by checking EC2 instance metadata every 30 seconds
2) Added forever go-routine to update `ip rules` only if there is a new VPC CIDR added for existing pods to enable connectivity between pods running on VPC CIDR1(old) on eth1 to pods running on VPC CIDR2 (new) on eth1.

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -22,6 +22,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
@@ -163,14 +164,14 @@ type APIs interface {
 // EC2InstanceMetadataCache caches instance metadata
 type EC2InstanceMetadataCache struct {
 	// metadata info
-	securityGroups   []*string
+	securityGroups   StringSet
 	subnetID         string
 	cidrBlock        string
 	localIPv4        string
 	instanceID       string
 	instanceType     string
 	vpcIPv4CIDR      string
-	vpcIPv4CIDRs     []*string
+	vpcIPv4CIDRs     StringSet
 	primaryENI       string
 	primaryENImac    string
 	availabilityZone string
@@ -224,8 +225,47 @@ func prometheusRegister() {
 	}
 }
 
+//StringSet is a set of strings
+type StringSet struct {
+	sync.RWMutex
+	data 	sets.String
+}
+
+func (ss *StringSet) AWSStrings() []*string {
+	ss.RLock()
+	defer ss.RUnlock()
+	var dataSlice []*string
+	for key, _ := range ss.data {
+		dataSlice = append(dataSlice, aws.String(key))
+	}
+	return dataSlice
+}
+
+func (ss *StringSet) Set(items []string) {
+	ss.Lock()
+	defer ss.Unlock()
+	ss.data = sets.NewString(items...)
+}
+
+func (ss *StringSet) IsEmpty() bool {
+	if ss.data.Len() == 0 {
+		return true
+	}
+	return false
+}
+
+func (ss *StringSet) Difference (other *StringSet) *StringSet {
+	ss.RLock()
+	defer ss.RUnlock()
+	//example: s1 = {a1, a2, a3} s2 = {a1, a2, a4, a5} s1.Difference(s2) = {a3} s2.Difference(s1) = {a4, a5}
+	return &StringSet{data: ss.data.Difference(other.data)}
+}
+
 // New creates an EC2InstanceMetadataCache
 func New() (*EC2InstanceMetadataCache, error) {
+	//ctx is passed to initWithEC2Metadata func to cancel spawned go-routines when tests are run
+	ctx := context.Background()
+
 	// Initializes prometheus metrics
 	prometheusRegister()
 
@@ -250,7 +290,7 @@ func New() (*EC2InstanceMetadataCache, error) {
 
 	ec2SVC := ec2wrapper.New(sess)
 	cache.ec2SVC = ec2SVC
-	err = cache.initWithEC2Metadata()
+	err = cache.initWithEC2Metadata(ctx)
 	if err != nil {
 		return nil, err
 	}
@@ -262,7 +302,7 @@ func New() (*EC2InstanceMetadataCache, error) {
 }
 
 // InitWithEC2metadata initializes the EC2InstanceMetadataCache with the data retrieved from EC2 metadata service
-func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
+func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context) error {
 	// retrieve availability-zone
 	az, err := cache.ec2Metadata.GetMetadata(metadataAZ)
 	if err != nil {
@@ -315,20 +355,6 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
 		return errors.Wrap(err, "get instance metadata: failed to find primary ENI")
 	}
 
-	// retrieve security groups
-	metadataSGIDs, err := cache.ec2Metadata.GetMetadata(metadataMACPath + mac + metadataSGs)
-	if err != nil {
-		awsAPIErrInc("GetMetadata", err)
-		log.Errorf("Failed to retrieve security-group-ids data from instance metadata service, %v", err)
-		return errors.Wrap(err, "get instance metadata: failed to retrieve security-group-ids")
-	}
-	sgIDs := strings.Fields(metadataSGIDs)
-
-	for _, sgID := range sgIDs {
-		log.Debugf("Found security-group id: %s", sgID)
-		cache.securityGroups = append(cache.securityGroups, aws.String(sgID))
-	}
-
 	// retrieve sub-id
 	cache.subnetID, err = cache.ec2Metadata.GetMetadata(metadataMACPath + mac + metadataSubnetID)
 	if err != nil {
@@ -347,20 +373,92 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
 	}
 	log.Debugf("Found vpc-ipv4-cidr-block: %s ", cache.vpcIPv4CIDR)
 
-	// retrieve vpc-ipv4-cidr-blocks
+	// initSGIDs retrieves security groups
+	err = cache.refreshSGIDs(mac)
+	if err != nil {
+		return err
+	}
+
+	// initVPCIPv4CIDRs retrieves VPC IPv4 CIDR blocks
+	err = cache.refreshVPCIPv4CIDRs(mac)
+	if err != nil {
+		return err
+	}
+
+	// refresh security groups and VPC CIDR blocks in the background
+	// Ignoring errors since we will retry in 30s
+	go wait.Forever(func() { _ = cache.refreshSGIDs(mac) }, 30*time.Second)
+	go wait.Forever(func() { _ = cache.refreshVPCIPv4CIDRs(mac) }, 30*time.Second)
+
+	// We use the ctx here for testing, since we spawn go-routines above which will run forever.
+	select {
+	case <-ctx.Done():
+		return nil
+	default:
+	}
+
+	return nil
+}
+
+// refreshSGIDs retrieves security groups
+func (cache *EC2InstanceMetadataCache) refreshSGIDs(mac string) error {
+	metadataSGIDs, err := cache.ec2Metadata.GetMetadata(metadataMACPath + mac + metadataSGs)
+	if err != nil {
+		awsAPIErrInc("GetMetadata", err)
+		log.Errorf("Failed to retrieve security-group-ids data from instance metadata service")
+		return errors.Wrap(err, "get instance metadata: failed to retrieve security-group-ids")
+	}
+
+	sgIDs  := strings.Fields(metadataSGIDs)
+
+	newSGs := StringSet{}
+	newSGs.Set(sgIDs)
+	addedSGs   := newSGs.Difference(&cache.securityGroups)
+	deletedSGs := cache.securityGroups.Difference(&newSGs)
+
+	if !addedSGs.IsEmpty() {
+		for _, sg := range addedSGs.AWSStrings() {
+			log.Infof("Found %s, added to ipamd cache", *sg)
+		}
+	}
+	if !deletedSGs.IsEmpty() {
+		for _, sg := range deletedSGs.AWSStrings() {
+				log.Infof("Removed %s from ipamd cache", *sg)
+		}
+	}
+	cache.securityGroups.Set(sgIDs)
+
+	return nil
+}
+
+// refreshVPCIPv4CIDRs retrieves VPC IPv4 CIDR blocks
+func (cache *EC2InstanceMetadataCache) refreshVPCIPv4CIDRs(mac string) error {
 	metadataVPCIPv4CIDRs, err := cache.ec2Metadata.GetMetadata(metadataMACPath + mac + metadataVPCcidrs)
 	if err != nil {
 		awsAPIErrInc("GetMetadata", err)
 		log.Errorf("Failed to retrieve vpc-ipv4-cidr-blocks from instance metadata service")
-		return errors.Wrap(err, "get instance metadata: failed to retrieve vpc-ipv4-cidr-block data")
+		return errors.Wrap(err, "get instance metadata: failed to retrieve vpc-ipv4-cidr-blocks data")
 	}
 
 	vpcIPv4CIDRs := strings.Fields(metadataVPCIPv4CIDRs)
 
-	for _, vpcCIDR := range vpcIPv4CIDRs {
-		log.Debugf("Found VPC CIDR: %s", vpcCIDR)
-		cache.vpcIPv4CIDRs = append(cache.vpcIPv4CIDRs, aws.String(vpcCIDR))
+	newVpcIPv4CIDRs := StringSet{}
+	newVpcIPv4CIDRs.Set(vpcIPv4CIDRs)
+	addedVpcIPv4CIDRs   := newVpcIPv4CIDRs.Difference(&cache.securityGroups)
+	deletedVpcIPv4CIDRs := cache.securityGroups.Difference(&newVpcIPv4CIDRs)
+
+	if !addedVpcIPv4CIDRs.IsEmpty() {
+		for _, vpcIPv4CIDR := range addedVpcIPv4CIDRs.AWSStrings() {
+			log.Infof("Found %s, added to ipamd cache", *vpcIPv4CIDR)
+		}
 	}
+	if !deletedVpcIPv4CIDRs.IsEmpty() {
+		for _, vpcIPv4CIDR := range deletedVpcIPv4CIDRs.AWSStrings() {
+			log.Infof("Removed %s from ipamd cache", *vpcIPv4CIDR)
+		}
+	}
+	cache.vpcIPv4CIDRs.Set(vpcIPv4CIDRs)
+
 	return nil
 }
 
@@ -663,7 +761,7 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
 	eniDescription := eniDescriptionPrefix + cache.instanceID
 	input := &ec2.CreateNetworkInterfaceInput{
 		Description: aws.String(eniDescription),
-		Groups:      cache.securityGroups,
+		Groups:      cache.securityGroups.AWSStrings(),
 		SubnetId:    aws.String(cache.subnetID),
 	}
 
@@ -1270,7 +1368,7 @@ func (cache *EC2InstanceMetadataCache) GetVPCIPv4CIDR() string {
 
 // GetVPCIPv4CIDRs returns VPC CIDRs
 func (cache *EC2InstanceMetadataCache) GetVPCIPv4CIDRs() []*string {
-	return cache.vpcIPv4CIDRs
+	return cache.vpcIPv4CIDRs.AWSStrings()
 }
 
 // GetLocalIPv4 returns the primary IP address on the primary interface
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -17,23 +17,24 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"reflect"
 	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
 
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/awsutils"
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/eniconfig"
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd/datastore"
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/networkutils"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
-
-	"github.com/aws/amazon-vpc-cni-k8s/pkg/awsutils"
-	"github.com/aws/amazon-vpc-cni-k8s/pkg/eniconfig"
-	"github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd/datastore"
-	"github.com/aws/amazon-vpc-cni-k8s/pkg/networkutils"
 )
 
 // The package ipamd is a long running daemon which manages a warm pool of available IP addresses.
@@ -377,7 +378,7 @@ func (c *IPAMContext) nodeInit() error {
 			}
 
 			if retry > maxRetryCheckENI {
-				log.Warn("Unable to discover attached IPs for ENI from metadata service")
+				log.Warnf("Reached max retry: Unable to discover attached IPs for ENI from metadata service (attempted %d/%d): %v", retry, maxRetryCheckENI, err)
 				ipamdErrInc("waitENIAttachedMaxRetryExceeded")
 				break
 			}
@@ -397,6 +398,24 @@ func (c *IPAMContext) nodeInit() error {
 		return err
 	}
 
+	if err = c.configureIPRulesForPods(pbVPCcidrs); err != nil {
+		return err
+	}
+
+	// For a new node, attach IPs
+	increasedPool, err := c.tryAssignIPs()
+	if err == nil && increasedPool {
+		c.updateLastNodeIPPoolAction()
+	} else {
+		return err
+	}
+
+	//Spawning checkAndUpdateRules go-routine
+	go wait.Forever(func() { pbVPCcidrs = c.checkVPCCIDRsAndRules(pbVPCcidrs)}, 30*time.Second)
+	return nil
+}
+
+func (c *IPAMContext) configureIPRulesForPods(pbVPCcidrs []string) error {
 	rules, err := c.networkClient.GetRuleList()
 	if err != nil {
 		log.Errorf("During ipamd init: failed to retrieve IP rule list %v", err)
@@ -411,15 +430,23 @@ func (c *IPAMContext) nodeInit() error {
 
 		err = c.networkClient.UpdateRuleListBySrc(rules, srcIPNet, pbVPCcidrs, !c.networkClient.UseExternalSNAT())
 		if err != nil {
-			log.Errorf("UpdateRuleListBySrc in nodeInit() failed for IP %s: %v", info.IP, err)
+			log.Warnf("UpdateRuleListBySrc in nodeInit() failed for IP %s: %v", info.IP, err)
 		}
 	}
-	// For a new node, attach IPs
-	increasedPool, err := c.tryAssignIPs()
-	if err == nil && increasedPool {
-		c.updateLastNodeIPPoolAction()
+	return nil
+}
+
+func (c *IPAMContext) checkVPCCIDRsAndRules(oldVPCCidrs []string) []string {
+	var pbVPCCIDRs []string
+	newVPCCIDRs := c.awsClient.GetVPCIPv4CIDRs()
+	for _, cidr := range newVPCCIDRs {
+		pbVPCCIDRs = append(pbVPCCIDRs, *cidr)
 	}
-	return err
+
+	if len(oldVPCCidrs) != len(pbVPCCIDRs) || !reflect.DeepEqual(oldVPCCidrs, pbVPCCIDRs) {
+		_ = c.configureIPRulesForPods(pbVPCCIDRs)
+	}
+	return pbVPCCIDRs
 }
 
 func (c *IPAMContext) updateIPStats(unmanaged int) {
--- a/go.mod
+++ b/go.mod
@@ -41,7 +41,6 @@ require (
 	golang.org/x/sys v0.0.0-20191026070338-33540a1f6037
 	golang.org/x/time v0.0.0-20180412165947-fbb02b2291d2 // indirect
 	google.golang.org/grpc v1.29.0
-	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0
 	gopkg.in/yaml.v2 v2.2.7 // indirect
--- a/pkg/awsutils/awsutils_test.go
+++ b/pkg/awsutils/awsutils_test.go
@@ -14,9 +14,11 @@
 package awsutils
 
 import (
+	"context"
 	"errors"
 	"os"
 	"sort"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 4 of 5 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -339,21 +339,41 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
...
+}
 
-	vpcIPv4CIDRs := strings.Fields(metadataVPCIPv4CIDRs)
-
-	for _, vpcCIDR := range vpcIPv4CIDRs {
-		log.Debugf("Found VPC CIDR: %s", vpcCIDR)
-		cache.vpcIPv4CIDRs = append(cache.vpcIPv4CIDRs, aws.String(vpcCIDR))
+// initVPCIPv4CIDRs vpc-ipv4-cidr-blocks
+func (cache *EC2InstanceMetadataCache) initVPCIPv4CIDRs(mac string, waitCh chan string, errorCh chan error) {
+	err := retry.RetryNWithBackoff(retry.NewSimpleBackoff(500*time.Millisecond, time.Minute, 0.5, 3), 5, func() error {
```
mogren, line 359: Hmm, the retry is fancy, but for metadata I don't think we need it. We don't use it for any of the other `GetMetadata()` calls, and I can't recall ever seeing it failing on the logs.

pkg/awsutils/awsutils.go
```diff
@@ -339,21 +339,41 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
...
+		metadataVPCIPv4CIDRs, err := cache.ec2Metadata.GetMetadata(metadataMACPath + mac + metadataVPCcidrs)
+		if err != nil {
+			awsAPIErrInc("GetMetadata", err)
+			log.Errorf("Failed to retrieve vpc-ipv4-cidr-blocks from instance metadata service")
+			return errors.Wrap(err, "get instance metadata: failed to retrieve vpc-ipv4-cidr-blocks data %v")
+		}
+		vpcIPv4CIDRs := strings.Fields(metadataVPCIPv4CIDRs)
+		for _, vpcCIDR := range vpcIPv4CIDRs {
+			log.Debugf("Found VPC CIDR: %s", vpcCIDR)
+			cache.vpcIPv4CIDRs = append(cache.vpcIPv4CIDRs, aws.String(vpcCIDR))
```
mogren, line 369: Will this not generate duplicates on the second run? `cache.vpcIPv4CIDRs` won't be empty then I assume?

pkg/ipamd/ipamd.go
```diff
@@ -338,18 +338,41 @@ func (c *IPAMContext) nodeInit() error {
...
+	if err == nil && increasedPool {
+		c.updateLastNodeIPPoolAction()
+	} else {
+		return err
+	}
+
+	//Spawning checkAndUpdateRules go-routine
+	go wait.Forever(func() {
+		log.Debug("Spawning checkAndUpdateRules go-routine")
+		c.checkVPCCIDRsAndRules(pbVPCcidrs)
```
mogren, line 358: How does `pbVPCcidrs` get updated here? I assume we would have to return it from the function, something like:
```
pbVPCcidrs = c.checkVPCCIDRsAndRules(pbVPCcidrs)
```
If not, we will always compare with the initial VPC CIDRs from the first `nodeInit()` call.

pkg/ipamd/ipamd.go
```diff
@@ -15,6 +15,7 @@ package ipamd
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/util/wait"
```
mogren, line 18: Nit: This one seem to be out of place. `make format` should fix it.

pkg/awsutils/awsutils.go
```diff
@@ -343,21 +329,85 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
...
 	for _, vpcCIDR := range vpcIPv4CIDRs {
 		log.Debugf("Found VPC CIDR: %s", vpcCIDR)
-		cache.vpcIPv4CIDRs = append(cache.vpcIPv4CIDRs, aws.String(vpcCIDR))
+		cache.appendIfMissing(vpcCIDR)
+	}
+	waitCh <- "VPC CIDR blocks retrieved"
+}
+
+func (cache *EC2InstanceMetadataCache) appendIfMissing(input string) {
+	if strings.Contains(input, "sg-") {
```
mogren, line 386: I'm not sure this function is needed. Is it not easier to just set `cache.securityGroups` to whatever the metadata returns? Also, this is the eth0 (primary MAC) only. It is used in `createENI` when setting the SG to use. If we modify and read this in different threads, we need to lock it. Something similar to the `UnmanagedENISet`. Maybe we should rename that type to `StringSet` and use it for both?

pkg/awsutils/awsutils.go
```diff
@@ -343,21 +329,85 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
...
+		}
+
+		if len(cache.securityGroups) == 0 {
+			cache.securityGroups = append(cache.securityGroups, aws.String(input))
+		}
+	} else {
+		if len(cache.vpcIPv4CIDRs) != 0 {
+			for _, value := range cache.vpcIPv4CIDRs {
+				if aws.StringValue(value) != input {
+					cache.vpcIPv4CIDRs = append(cache.vpcIPv4CIDRs, aws.String(input))
```
mogren, line 402: `cache.vpcIPv4CIDRs` is read through `GetVPCIPv4CIDRs()` and used when generating the `AddNetworkReply`, so It will also need a lock.

pkg/ipamd/ipamd.go
```diff
@@ -384,18 +384,40 @@ func (c *IPAMContext) nodeInit() error {
...
+	if err == nil && increasedPool {
+		c.updateLastNodeIPPoolAction()
+	} else {
+		return err
+	}
+
+	//Spawning checkAndUpdateRules go-routine
+	go wait.Forever(func() {
+		pbVPCcidrs = c.checkVPCCIDRsAndRules(pbVPCcidrs)
+	}, time.Second)
```
mogren, line 404: Update every second sounds like a lot. Why not something a bit less frequent?

pkg/ipamd/ipamd.go
```diff
@@ -384,18 +384,40 @@ func (c *IPAMContext) nodeInit() error {
...
+		log.Warnf("During IPamD init, Failed to get Pod information from Kubernetes API Server %v", err)
+		// This can happens when L-IPAMD starts before kube-proxy.
 		return errors.Wrap(err, "failed to get running pods!")
 	}
 	log.Debugf("getLocalPodsWithRetry() found %d local pods", len(localPods))
 
 	rules, err := c.networkClient.GetRuleList()
 	if err != nil {
-		log.Errorf("During ipamd init: failed to retrieve IP rule list %v", err)
+		log.Errorf("During IPamD init: failed to retrieve IP rule list %v", err)
```
mogren, line 420: Nit: `ipamd` is fine 😄

pkg/ipamd/ipamd.go
```diff
@@ -384,18 +384,40 @@ func (c *IPAMContext) nodeInit() error {
...
+	//Spawning checkAndUpdateRules go-routine
+	go wait.Forever(func() {
+		pbVPCcidrs = c.checkVPCCIDRsAndRules(pbVPCcidrs)
+	}, time.Second)
+
+	return nil
+}
+
+func (c *IPAMContext) configureIPRulesForPods(pbVPCcidrs []string) error {
+	localPods, err := c.getLocalPodsWithRetry()
```
mogren, line 410: This calls the API server. It used to only be done once when a new node starts up, and it's not something we should have to do a lot.

pkg/ipamd/ipamd.go
```diff
@@ -423,12 +445,24 @@ func (c *IPAMContext) nodeInit() error {
...
+}
+
+func (c *IPAMContext) checkVPCCIDRsAndRules(oldVPCCidrs []string) []string {
+	var pbVPCCIDRs []string
+	newVPCCIDRs := c.awsClient.GetVPCIPv4CIDRs()
+	for _, cidr := range newVPCCIDRs {
+		pbVPCCIDRs = append(pbVPCCIDRs, *cidr)
+	}
+
+	if len(oldVPCCidrs) != len(pbVPCCIDRs) {
```
mogren, line 459: What if one CIDR was deleted and another one added to the VPC?

pkg/ipamd/ipamd.go
nithu0115: I am not sure if I understand this comment. I am relying on this same func to get running Pod IP address for me to re-configure `ip rules` when there is change.

pkg/awsutils/awsutils.go
```diff
@@ -343,21 +350,73 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
...
+	// spawing go routine and retrieving vpc-ipv4-cidr-block
+	waitSGCh  := make(chan string)
+	waitVPCCh := make(chan string)
+	errorCh := make(chan error)
+	go wait.Forever(func() {
+		cache.initSGIDs(mac, waitSGCh, errorCh)
+		cache.initVPCIPv4CIDRs(mac, waitVPCCh, errorCh)
+	}, 30*time.Second)
+
+	for {
```
mogren, line 362: This needs a short comment explaining the setup.

pkg/awsutils/awsutils.go
```diff
@@ -1225,7 +1288,13 @@ func (cache *EC2InstanceMetadataCache) GetVPCIPv4CIDR() string {
 
 // GetVPCIPv4CIDRs returns VPC CIDRs
 func (cache *EC2InstanceMetadataCache) GetVPCIPv4CIDRs() []*string {
-	return cache.vpcIPv4CIDRs
+	var vpcIPv4CIDRs []*string
```
mogren, line 1291: Nit: Rename to `vpcCIDRslice` and move before the for-loop.

pkg/awsutils/awsutils.go
```diff
@@ -1225,7 +1288,13 @@ func (cache *EC2InstanceMetadataCache) GetVPCIPv4CIDR() string {
...
+	cache.vpcIPv4CIDRs.RLock()
+	defer cache.vpcIPv4CIDRs.RUnlock()
+	for vpcIPv4CIDR := range cache.vpcIPv4CIDRs.data{
+		vpcIPv4CIDRs = append(vpcIPv4CIDRs, aws.String(vpcIPv4CIDR))
```
mogren, line 1295: I have a sense that this doesn't work. I suspect it will return the last CIDR in the set every time since we return the address of `vpcIPv4CIDR` and that value will be updated on each loop. Similar to [this code](https://github.com/aws/amazon-vpc-cni-k8s/blob/9c863340dac6fc959636c57ae4a1706d02748874/pkg/awsutils/awsutils.go#L500-L506).

pkg/awsutils/awsutils.go
```diff
@@ -220,6 +221,26 @@ func prometheusRegister() {
 	}
 }
 
+//stringSet is a set of strings
+type stringSet struct {
```
mogren, line 225: If we add the `reset()` function, we can replace `UnmanagedENISet` with this `stringSet`. Also, how will we ever remove a CIDR from the set otherwise?

pkg/awsutils/awsutils.go
nithu0115: Done!

pkg/awsutils/awsutils.go
nithu0115: > If we add the `reset()` function, we can replace `UnmanagedENISet` with this `stringSet`.

I will open a separate issue for this one and do a follow up once we merge this PR.

pkg/ipamd/ipamd.go
```diff
@@ -390,11 +391,33 @@ func (c *IPAMContext) nodeInit() error {
...
+		pbVPCcidrs = c.checkVPCCIDRsAndRules(pbVPCcidrs)
+	}, 30*time.Second)
+
+	return nil
+}
+
+func (c *IPAMContext) configureIPRulesForPods(pbVPCcidrs []string) error {
+	localPods, err := c.getLocalPodsWithRetry()
+	if err != nil {
+		log.Warnf("During IPamD init, Failed to get Pod information from Kubernetes API Server %v", err)
```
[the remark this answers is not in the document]
nithu0115: I changed it a generic warning message. func `configureIPRulesForPods` is invoked after init as well with this PR as we to fetch Pod IPs from API server if there is a change to VPC CIDR block

pkg/awsutils/awsutils.go
nithu0115: Thanks for catching this 🙏

pkg/awsutils/awsutils.go
```diff
@@ -343,21 +356,71 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
...
+		case errs := <-errorCh:
+			return errs
+		}
+		select {
+		case <- waitVPCCh:
+			break
+		case errs := <-errorCh:
+			return errs
+		}
+		break
```
mogren, line 382: This infinite for-loop still doesn't look right. For example, the `break` on line 371 will exit the for-loop even if the one on line 378 has not yet exited. Options could be to have an [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup?tab=doc#example-Group-Parallel) or use something similar to [this  `WaitGroup` example](https://golangcode.com/errors-in-waitgroups/).

I think an easier solution would be to remove the channels completely, call the functions sequentially, once each without a go routine in the `init()` function, then set up the background go routines after. What do you think?

pkg/awsutils/awsutils.go
```diff
@@ -1225,7 +1294,15 @@ func (cache *EC2InstanceMetadataCache) GetVPCIPv4CIDR() string {
 
 // GetVPCIPv4CIDRs returns VPC CIDRs
 func (cache *EC2InstanceMetadataCache) GetVPCIPv4CIDRs() []*string {
-	return cache.vpcIPv4CIDRs
+	cache.vpcIPv4CIDRs.RLock()
+	defer cache.vpcIPv4CIDRs.RUnlock()
+	var vpcIPv4CIDRslice []*string
+	for vpcIPv4CIDR := range cache.vpcIPv4CIDRs.data{
+		//In Golang, the loop iterator variable is a single variable that takes different values in each loop iteration.
```
mogren, line 1301: Nit: `// In go, the loop...`

pkg/awsutils/awsutils.go
nithu0115: > This infinite for-loop still doesn't look right. For example, the `break` on line 371 will exit the for-loop even if the one on line 378 has not yet exited. 

Hmm, AFAIT, it will exit the first `select` and moves on to the second one, but not the `for-loop` itself. 

>Options could be to have an [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup?tab=doc#example-Group-Parallel) or use something similar to [this `WaitGroup` example](https://golangcode.com/errors-in-waitgroups/).

I tried both`errgroup` and `waitgroup` packages but had a difficult with `wait.Forever`

> I think an easier solution would be to remove the channels completely, call the functions sequentially, once each without a go routine in the `init()` function, then set up the background go routines after. What do you think?

Generally, I would like to avoid `init()` function but like the idea to call to function and spawn separate `go-routines` for both of them and don't worry about channels. I am open to ideas, let me know :)

pkg/awsutils/awsutils.go
```diff
@@ -343,19 +358,61 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
...
+	go wait.Forever(func() { cache.initVPCIPv4CIDRs(mac) }, 30*time.Second)
+
+	select {
+	case <-ctx.Done():
+		return nil
+	}
+}
+
+// initSGIDs retrieves security groups
+func (cache *EC2InstanceMetadataCache) initSGIDs(mac string) error {
```
jaypipes, line 384: Instead of `initXXX` can we name these functions `refreshXXX` to better indicate that they are not just run on initialization of the cache?

pkg/awsutils/awsutils.go
```diff
@@ -343,19 +358,61 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
...
+	}
+
+	// initVPCIPv4CIDRs retrieves VPC IPv4 CIDR blocks
+	err = cache.initVPCIPv4CIDRs(mac)
+	if err != nil {
+		return err
+	}
+
+	// refresh security groups and VPC CIDR blocks in the background
+	go wait.Forever(func() { cache.initSGIDs(mac) }, 30*time.Second)
```
jaypipes, line 374: Not handling errors inside the forever routines?

pkg/awsutils/awsutils.go
```diff
@@ -343,19 +358,61 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
...
 		log.Errorf("Failed to retrieve vpc-ipv4-cidr-blocks from instance metadata service")
-		return errors.Wrap(err, "get instance metadata: failed to retrieve vpc-ipv4-cidr-block data")
+		return errors.Wrap(err, "get instance metadata: failed to retrieve vpc-ipv4-cidr-blocks data")
 	}
 
 	vpcIPv4CIDRs := strings.Fields(metadataVPCIPv4CIDRs)
-
+	cache.vpcIPv4CIDRs.reset()
 	for _, vpcCIDR := range vpcIPv4CIDRs {
 		log.Debugf("Found VPC CIDR: %s", vpcCIDR)
```
jaypipes, line 414: The above is now going to be output every 30 seconds, one for each subnet associated with the node. Instead of just calling `reset()` on the StringSet above, you could consider determining what the set difference is between the values you fetched from IMDS and then only outputting a log message (or changing the value of the StringSet) if something changed.

pkg/awsutils/awsutils.go
```diff
@@ -657,9 +714,15 @@ func (cache *EC2InstanceMetadataCache) attachENI(eniID string) (string, error) {
 // return ENI id, error
 func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string, subnet string) (string, error) {
 	eniDescription := eniDescriptionPrefix + cache.instanceID
+	var securityGroups []*string
+	for securityGroup := range cache.securityGroups.data {
+		//In go, the loop iterator variable is a single variable that takes different values in each loop iteration.
```
jaypipes, line 719: if you just want the keys in the `data` `map[string]bool`, use the double-valued range iterator:

```go
securityGroups make([]*string, 0, len(cache.securityGroups.data))
for key, _ range cache.securityGroups.data {
    securityGroups = append(securityGroups, aws.String(key))
}
```

Also consider making the above into a `AWSStrings()` function of `StringSet` that simply returns a slice of string pointers.

pkg/awsutils/awsutils.go
```diff
@@ -1225,7 +1288,15 @@ func (cache *EC2InstanceMetadataCache) GetVPCIPv4CIDR() string {
...
-	return cache.vpcIPv4CIDRs
+	cache.vpcIPv4CIDRs.RLock()
+	defer cache.vpcIPv4CIDRs.RUnlock()
+	var vpcIPv4CIDRslice []*string
+	for vpcIPv4CIDR := range cache.vpcIPv4CIDRs.data{
+		//In go, the loop iterator variable is a single variable that takes different values in each loop iteration.
+		tmpVar := vpcIPv4CIDR
+		vpcIPv4CIDRslice = append(vpcIPv4CIDRslice, aws.String(tmpVar))
+	}
+	return vpcIPv4CIDRslice
```
jaypipes, line 1299: and then the above would need to change to just this:

```go
return cache.vpcIPv4CIDRs.AWSStrings()
```

since you would place the RLock()/RUnlock() within the `StringSet.AWSStrings()` implementation

pkg/awsutils/awsutils.go
nithu0115: > Instead of `initXXX` can we name these functions `refreshXXX` to better indicate that they are not just run on initialization of the cache?

I like the idea but we call these functions once during init and then spawn a go-routine. Instead, how does  something more generic like `fetchSGIDs` sound ?

pkg/awsutils/awsutils.go
nithu0115: Yes, I thought about it. But, it does not make sense as we will retry the same call again in 30 seconds.

nithu0115:
@jaypipes, addressed your comments.

pkg/awsutils/awsutils.go
```diff
@@ -346,20 +356,106 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
...
-		cache.vpcIPv4CIDRs = append(cache.vpcIPv4CIDRs, aws.String(vpcCIDR))
+	if cache.vpcIPv4CIDRs.data.Len() == 0 {
+		cache.vpcIPv4CIDRs.data = make(sets.String)
+	    for _, vpcCIDR := range vpcIPv4CIDRs {
+			log.Debugf("Found VPC CIDR: %s", vpcCIDR)
+			cache.vpcIPv4CIDRs.data.Insert(vpcCIDR)
+		}
 	}
+
+	if cache.vpcIPv4CIDRs.data.Len() != 0 {
```
jaypipes, line 443: this should be an else block, similar comment to above...

pkg/awsutils/awsutils.go
```diff
@@ -223,8 +224,31 @@ func prometheusRegister() {
...
 
+type AWSStrings struct {
+
+}
+//StringSet is a set of strings
+type StringSet struct {
+	sync.RWMutex
+	data 	sets.String
+	// dataNew var is only used to check difference
+	dataNew sets.String
```
jaypipes, line 235: this field isn't needed.

the StringSet should be comparable to another StringSet object instead of having a StringSet store an "old" and "new" set of strings.

pkg/awsutils/awsutils.go
```diff
@@ -346,20 +356,106 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
...
+			log.Debugf("Deleted Security Group: %s", delSG)
+		}
+		for addSG, _ := range cache.securityGroups.dataNew.Difference(cache.securityGroups.data) {
+			cache.securityGroups.data.Insert(addSG)
+			log.Debugf("Added Security Group: %s", addSG)
+		}
+	}
+
+	return nil
+}
```
jaypipes, line 419: The above code from 403-419 is just setting the `cache.securityGroups.data` value to the set of security groups in the `sgIDs` variable. I don't think that's what you wanted is it?

pkg/awsutils/awsutils.go
```diff
@@ -224,8 +225,66 @@ func prometheusRegister() {
...
+
+func (ss *StringSet) compareAddDel (items []string) {
+	newSet := StringSet{data: sets.String{}}
+	newSet.Set(items)
+	//For example: s1 = {a1, a2, a3} s2 = {a1, a2, a4, a5} s1.Difference(s2) = {a3} s2.Difference(s1) = {a4, a5}
+	setDiff := ss.data.Difference(newSet.data)
+	if setDiff.Len() != 0 {
+		ss.Lock()
+		defer ss.Unlock()
```
mogren, line 267: I would expect the `defer ss.Unlock()` here to not run until the whole `compareAddDel()` function exits, so if we both add and delete an item, this would deadlock. Probably better to just do it once in the beginning of this function.

Maybe worth having a test that does this? Create a set of 3 strings, `["a1", "a2", "a3"}`, then call it with `["a2", "a3", "a4"]`.

pkg/awsutils/awsutils.go
```diff
@@ -224,8 +225,66 @@ func prometheusRegister() {
...
+
+func (ss *StringSet) IsEmpty() bool {
+	if ss.data.Len() == 0 {
+		ss.data = sets.NewString()
+		return true
+	}
+	return false
+}
+
+func (ss *StringSet) compareAddDel (items []string) {
```
mogren, line 260: Nit: Extra space after `compareAddDel`. Also not sure about the name.. Is it more like `updateSet()`, since it doesn't really compare with a slice, but updates the actual set and log the changes.

pkg/awsutils/awsutils.go
```diff
@@ -224,8 +225,66 @@ func prometheusRegister() {
...
+	newSet := StringSet{data: sets.String{}}
+	newSet.Set(items)
+	//For example: s1 = {a1, a2, a3} s2 = {a1, a2, a4, a5} s1.Difference(s2) = {a3} s2.Difference(s1) = {a4, a5}
+	setDiff := ss.data.Difference(newSet.data)
+	if setDiff.Len() != 0 {
+		ss.Lock()
+		defer ss.Unlock()
+		for delItem, _ := range setDiff {
+			ss.data.Delete(delItem)
+			log.Infof("Removed %s from ipamd cache", delItem)
```
mogren, line 270: Maybe we should just call it a set and quote the string?
```go
log.Infof("Removed '%s' from StringSet", delItem)
```
Same for the add log line.

pkg/awsutils/awsutils.go
```diff
@@ -224,8 +225,66 @@ func prometheusRegister() {
...
+			ss.data.Insert(addItem)
+			log.Infof("Found new %s, added to ipamd cache", addItem)
+		}
+	}
+}
+
 // New creates an EC2InstanceMetadataCache
 func New() (*EC2InstanceMetadataCache, error) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
mogren, line 287: Why is the `defer cancel()` needed here? It is probably worth a comment.

pkg/awsutils/awsutils.go
```diff
@@ -347,20 +392,66 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
...
+		return errors.Wrap(err, "get instance metadata: failed to retrieve security-group-ids")
+	}
+
+	sgIDs := strings.Fields(metadataSGIDs)
+
+	if cache.securityGroups.IsEmpty() {
+		cache.securityGroups.Set(sgIDs)
+	} else {
+		cache.securityGroups.compareAddDel(sgIDs)
+	}
```
mogren, line 433: Just calling `cache.securityGroups.compareAddDel(sgIDs)` directly should also work, or am I missing something?

pkg/awsutils/awsutils.go
```diff
@@ -347,20 +392,66 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
...
 
 	vpcIPv4CIDRs := strings.Fields(metadataVPCIPv4CIDRs)
 
-	for _, vpcCIDR := range vpcIPv4CIDRs {
-		log.Debugf("Found VPC CIDR: %s", vpcCIDR)
-		cache.vpcIPv4CIDRs = append(cache.vpcIPv4CIDRs, aws.String(vpcCIDR))
+	if cache.vpcIPv4CIDRs.IsEmpty() {
+		cache.vpcIPv4CIDRs.Set(vpcIPv4CIDRs)
+	} else {
+		cache.vpcIPv4CIDRs.compareAddDel(vpcIPv4CIDRs)
```
mogren, line 452: Same here

pkg/awsutils/awsutils.go
```diff
@@ -224,8 +225,66 @@ func prometheusRegister() {
...
+	ss.Lock()
+	defer ss.Unlock()
+	for _, item := range items {
+		ss.data.Insert(item)
+	}
+}
+
+func (ss *StringSet) IsEmpty() bool {
+	if ss.data.Len() == 0 {
+		ss.data = sets.NewString()
```
jaypipes, line 254: A readonly method like `IsEmpty()` should never mutate state. Please remove the line:

```go
ss.data = sets.NewString()
```

pkg/awsutils/awsutils.go
```diff
@@ -224,8 +225,66 @@ func prometheusRegister() {
...
+		dataSlice = append(dataSlice, aws.String(key))
+	}
+	return dataSlice
+}
+
+func (ss *StringSet) Set(items []string) {
+	ss.Lock()
+	defer ss.Unlock()
+	for _, item := range items {
+		ss.data.Insert(item)
```
jaypipes, line 248: This needs to handle the case when `data` is `nil`. Also, it isn't actually "setting" the value of the `StringSet`s `data` field. This method should look like this instead:

```go
func (ss *StringSet) Set(items []string) {
    ss.Lock()
    defer ss.Unlock()
    ss.data = sets.NewString(items...)    
}
```

pkg/awsutils/awsutils.go
jaypipes, line 260: This method should be called `Difference` and simply return a thread-safe result of the underlying `ss.data.Difference()` method:

```go
func (ss *StringSet) Difference(other *StringSet) *StringSet {
    ss.Lock()
    defer ss.Unlock()
    return &StringSet{data: ss.data.Difference(other.data)}
}
```

pkg/awsutils/awsutils.go
jaypipes: No, these log messages should not be in the `StringSet` implementation. See below for where they should be.

pkg/awsutils/awsutils.go
jaypipes: Actually, the above should just be:

```go
newSGs := &StringSet{}.Set(sgIDs...)
deleted := cache.securityGroups.Difference(newSGs)
added := newSGs.Difference(cache.securiyGroups)
if !added.IsEmpty() {
    for _, sg := range added.AWSStrings() {
        log.Infof("Found new %s, added to ipamd cache", *sg)
    }
}
if !deleted.IsEmpty() {
    for _, sg := range deleted.AWSStrings() {
        log.Infof("Removed %s from ipamd cache", *sg)
    }
}
cache.securityGroups.Set(sgIDs)
```

pkg/awsutils/awsutils.go
```diff
@@ -347,20 +373,90 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
...
+	// initVPCIPv4CIDRs retrieves VPC IPv4 CIDR blocks
+	err = cache.refreshVPCIPv4CIDRs(mac)
+	if err != nil {
+		return err
+	}
+
+	// refresh security groups and VPC CIDR blocks in the background
+	// Ignoring errors since we will retry in 30s
+	go wait.Forever(func() { _ = cache.refreshSGIDs(mac) }, 30*time.Second)
+	go wait.Forever(func() { _ = cache.refreshVPCIPv4CIDRs(mac) }, 30*time.Second)
```
mogren, line 391: Nit: Just for visibility... We already create a background go routine in the `New()` function on line 299 to clean up leaked ENIs. Maybe we can move these there as well, to make them more visible. To make it obvious that when you create an 'awsutils.New()`, you get 3 background routines.

pkg/awsutils/awsutils.go
```diff
@@ -224,8 +225,47 @@ func prometheusRegister() {
...
+}
+
+func (ss *StringSet) Set(items []string) {
+	ss.Lock()
+	defer ss.Unlock()
+	ss.data = sets.NewString(items...)
+}
+
+func (ss *StringSet) IsEmpty() bool {
+	if ss.data.Len() == 0 {
```
jaypipes, line 251: nit.. this assumes a non-nil `data` field. Plus, you could simplify this method to:

```go
return ss.data != nil && ss.data.Len() == 0
```

pkg/awsutils/awsutils.go
```diff
@@ -347,20 +373,92 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
 	}
 	log.Debugf("Found vpc-ipv4-cidr-block: %s ", cache.vpcIPv4CIDR)
 
-	// retrieve vpc-ipv4-cidr-blocks
+	// initSGIDs retrieves security groups
```
jaypipes, line 376: I don't think initSGIDs is the name of the method any more?

pkg/awsutils/awsutils.go
```diff
@@ -347,20 +373,92 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
...
+	err = cache.refreshSGIDs(mac)
+	if err != nil {
+		return err
+	}
+
+	// initVPCIPv4CIDRs retrieves VPC IPv4 CIDR blocks
```
jaypipes, line 382: I don't think initVPCIPv4CIDRs is the name of the method any more?

pkg/ipamd/ipamd.go
```diff
@@ -411,15 +430,23 @@ func (c *IPAMContext) nodeInit() error {
...
 		}
 	}
-	// For a new node, attach IPs
-	increasedPool, err := c.tryAssignIPs()
-	if err == nil && increasedPool {
-		c.updateLastNodeIPPoolAction()
+	return nil
+}
+
+func (c *IPAMContext) checkVPCCIDRsAndRules(oldVPCCidrs []string) []string {
```
jaypipes, line 439: nit: I prefer to name functions that mutate an object's state (like this one does for the `configureIPRulesForPods()` call when the new rules don't equal the old rules) with something an action that indicates the function may change state.

Perhaps something like `updateCIDRsRulesOnChange()` might work?

pkg/awsutils/awsutils.go
mogren: Fixed in #1056

pkg/awsutils/awsutils.go
```diff
@@ -347,20 +373,92 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata() error {
 	}
 	log.Debugf("Found vpc-ipv4-cidr-block: %s ", cache.vpcIPv4CIDR)
 
-	// retrieve vpc-ipv4-cidr-blocks
+	// initSGIDs retrieves security groups
+	err = cache.refreshSGIDs(mac)
```
SaranBalaji90, line 377: This is not invoked for CustomNetworking right?

nithu0115: Right, it is not!

Issue #887, opened 2020-03-27
Excessive DescribeNetworkInterfaces AWS API calls


jbnjohnathan:
We have a AWS account where we have one EKS cluster. The cluster has about 7 nodes.
We are using amazon-k8s-cni:v1.6.0
The cluster has about 100 pods in total.

Before we had 4 more clusters in the same account and started getting Throttle issues from AWS API.
I imported API calls metrics into CloudWatch and found that something was consistently making around 500 calls DescribeNetworkInterfaces per 5 minutes.
When we deleted 4 clusters this number dropped to about 80-100.
Again, that is only DescribeNetworkInterfaces.
I can see from the CloudTrail that it is coming from our EKS clusters.

A saw that there was an issue open for reducing the amount of calls (https://github.com/aws/amazon-vpc-cni-k8s/issues/873)
I don't have data further back so I cannot say, but I perceive this problem of Throttling to start after we upgraded the CNI plugin from 1.5.5 to 1.6.0

I find the number of API calls excessive since they are at a constant level even when no new deploys to the cluster is ongoing.

With 1.6.0 we have also seen this in the event log
`
Failed create pod sandbox: rpc error: code = Unknown desc = failed to set up sandbox container "7e3f4bcf7367353f7ec38373c8055a4a3d3ffaf224384351ae0997b06cfc3f87" network for pod "prometheus-stateful-set-v3-0": NetworkPlugin cni failed to set up pod "prometheus-stateful-set-v3-0_scheduled-system-test" network: add cmd: failed to assign an IP address to container
`

Before I finished writing this ticket I downgraded from 1.6.0 to 1.5.5 and saw a dramatic drop of DescribeNetworkInterfaces from 100 per 5min to between 0-1. (zero - one).

So something big changed between 1.5.5 and 1.6.0 in regard to this behavior.

jbnjohnathan:
An update. I just now noticed that there was a note in the 1.6.0 release notes about 
`
Note! If you edit an existing aws-node deployment, make sure to mount /var/run/dockershim.sock correctly!
`

We did not do this. Can this be the cause of this issue?

jbnjohnathan:
We re-deployed with the /var/run/dockershim.sock mounted and using 1.6.0 but we still saw a big increase in DescribeNetworkInterfaces compared with 1.5.5

mogren:
[...]

PR #909, opened 2020-04-11
Reduce number of calls to EC2 API

mogren:
*Issue #, if available:* #887

*Description of changes:*
* Remove Tags from the metadata struct, since tags are only available by calling EC2
* Remove `currentENIs` value since it's not used anywhere
* Made a separate function to get IPv4 addresses for one ENI from EC2
* Made another function to only get the ENI `AttachmentID` from EC2
* Batch the call to check all attached ENIs into one API request
* Added tests for `no_manage`
* Verified that `no_manage` tags on both restarts and adding new ENIs to a node

Test showing number of `DescribeNetworkInterfaces` calls running 5000 pods on 100 nodes:
ipamd-ec2-api-calls

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -114,8 +114,11 @@ type APIs interface {
 	// GetAttachedENIs retrieves eni information from instance metadata service
 	GetAttachedENIs() (eniList []ENIMetadata, err error)
 
-	// DescribeENI returns the IPv4 addresses of ENI interface, tags, and the ENI attachment ID
-	DescribeENI(eniID string) (addrList []*ec2.NetworkInterfacePrivateIpAddress, tags map[string]string, attachemdID *string, err error)
+	// GetIPv4sFromEC2 returns the IPv4 addresses for a given ENI
+	GetIPv4sFromEC2(eniID string) (addrList []*ec2.NetworkInterfacePrivateIpAddress, err error)
+
+	// DescribeAllENIs calls EC2 and returns the ENIMetadata and a tag map for each ENI
+	DescribeAllENIs() ([]ENIMetadata, map[string]TagMap, error)
 
 	// AllocIPAddress allocates an IP address for an ENI
 	AllocIPAddress(eniID string) error
@@ -165,9 +168,6 @@ type EC2InstanceMetadataCache struct {
 	region           string
 	accountID        string
 
-	// dynamic
-	currentENIs int
-
 	ec2Metadata ec2metadata.EC2Metadata
 	ec2SVC      ec2wrapper.EC2
 }
@@ -188,9 +188,6 @@ type ENIMetadata struct {
 
 	// The ip addresses allocated for the network interface
 	IPv4Addresses []*ec2.NetworkInterfacePrivateIpAddress
-
-	// Tags are the tags associated with this ENI in AWS
-	Tags map[string]string
 }
 
 func (eni ENIMetadata) PrimaryIPv4Address() string {
@@ -201,6 +198,8 @@ func (eni ENIMetadata) PrimaryIPv4Address() string {
 	}
 	return ""
 }
+
+type TagMap map[string]string
 
 // msSince returns milliseconds since start.
 func msSince(start time.Time) float64 {
@@ -370,7 +369,6 @@ func (cache *EC2InstanceMetadataCache) setPrimaryENI() error {
 	}
 	eniMACs := strings.Fields(metadataENImacs)
 	log.Debugf("Discovered %d interfaces.", len(eniMACs))
-	cache.currentENIs = len(eniMACs)
 
 	// retrieve the attached ENIs
 	for _, eniMAC := range eniMACs {
@@ -411,7 +409,7 @@ func (cache *EC2InstanceMetadataCache) setPrimaryENI() error {
 		if cache.primaryENImac == result[0] {
 			//primary interface
 			cache.primaryENI = eni
-			log.Debugf("Found ENI %s is a primary ENI", eni)
+			log.Debugf("%s is the primary ENI of this instance", eni)
 			return nil
 		}
 	}
@@ -429,7 +427,6 @@ func (cache *EC2InstanceMetadataCache) GetAttachedENIs() (eniList []ENIMetadata,
 	}
 	macsStrs := strings.Fields(macs)
 	log.Debugf("Total number of interfaces found: %d ", len(macsStrs))
-	cache.currentENIs = len(macsStrs)
 
 	var enis []ENIMetadata
 	// retrieve the attached ENIs
@@ -446,7 +443,7 @@ func (cache *EC2InstanceMetadataCache) GetAttachedENIs() (eniList []ENIMetadata,
 func (cache *EC2InstanceMetadataCache) getENIMetadata(macStr string) (ENIMetadata, error) {
 	eniMACList := strings.Split(macStr, "/")
 	eniMAC := eniMACList[0]
-	log.Debugf("Found ENI mac address : %s", eniMAC)
+	log.Debugf("Found ENI MAC address: %s", eniMAC)
 
 	eni, deviceNum, err := cache.getENIDeviceNumber(eniMAC)
 	if err != nil {
@@ -458,42 +455,17 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(macStr string) (ENIMetadat
 	if err != nil {
 		return ENIMetadata{}, errors.Wrapf(err, "get ENI metadata: failed to retrieve IPs and CIDR for ENI: %s", eniMAC)
 	}
-	privateIPv4s, tags, _, err := cache.DescribeENI(eni)
-	if err != nil {
-		return ENIMetadata{}, errors.Wrapf(err, "get ENI metadata: failed to describe ENI: %s, %v", eniMAC, err)
-	}
-	// getIPsAndCIDR() queries IMDS for IPv4 addresses attached to the ENI.
-	// DescribeENI() calls the DescribeNetworkInterfaces AWS API call, which
-	// technically should be the source of truth and contain the freshest
-	// information. Let's just do a quick scan here and output some diagnostic
-	// messages if we find stale info in the IMDS result.
-	imdsIPv4Set := sets.NewString(imdsIPv4s...)
-	privateIPv4Set := sets.String{}
-	for _, privateIPv4 := range privateIPv4s {
-		privateIPv4Set.Insert(aws.StringValue(privateIPv4.PrivateIpAddress))
-	}
-	missingIMDS := privateIPv4Set.Difference(imdsIPv4Set).List()
-	missingDNI := imdsIPv4Set.Difference(privateIPv4Set).List()
-	if len(missingIMDS) > 0 {
-		strMissing := strings.Join(missingIMDS, ",")
-		log.Debugf("getENIMetadata: DescribeNetworkInterfaces(%s) yielded private IPv4 addresses %s that were not yet found in IMDS.", eni, strMissing)
-	}
-	if len(missingDNI) > 0 {
-		strMissing := strings.Join(missingDNI, ",")
-		log.Debugf("getENIMetadata: IMDS query yielded stale IPv4 addresses %s that were not found in DescribeNetworkInterfaces(%s).", strMissing, eni)
-	}
 	return ENIMetadata{
 		ENIID:          eni,
 		MAC:            eniMAC,
 		DeviceNumber:   deviceNum,
 		SubnetIPv4CIDR: cidr,
-		IPv4Addresses:  privateIPv4s,
-		Tags:           tags,
+		IPv4Addresses:  imdsIPv4s,
 	}, nil
 }
 
 // getIPsAndCIDR return list of IPs, CIDR, error
-func (cache *EC2InstanceMetadataCache) getIPsAndCIDR(eniMAC string) ([]string, string, error) {
+func (cache *EC2InstanceMetadataCache) getIPsAndCIDR(eniMAC string) ([]*ec2.NetworkInterfacePrivateIpAddress, string, error) {
 	start := time.Now()
 	cidr, err := cache.ec2Metadata.GetMetadata(metadataMACPath + eniMAC + metadataSubnetCIDR)
 	awsAPILatency.WithLabelValues("GetMetadata", fmt.Sprint(err != nil)).Observe(msSince(start))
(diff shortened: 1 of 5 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -506,17 +472,28 @@ func (cache *EC2InstanceMetadataCache) getIPsAndCIDR(eniMAC string) ([]string, s
...
+	ipv4s := make([]*ec2.NetworkInterfacePrivateIpAddress, 0, len(ipv4Strs))
+	// network/interfaces/macs/mac/public-ipv4s	The public IP address or Elastic IP addresses associated with the interface.
+	// There may be multiple IPv4 addresses on an instance. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html
+	isFirst := true
+	for _, ipv4 := range ipv4Strs {
+		// TODO: Verify that the first IP is always the primary
+		primary := isFirst
+		ip := ipv4
+		ipv4s = append(ipv4s, &ec2.NetworkInterfacePrivateIpAddress{PrivateIpAddress: &ip, Primary: &primary})
+		isFirst = false
```
mogren, line 494: I checked this on 10 nodes scaling up and down by also doing an EC2 call for each ENI, and every single time the first IP was the same as the one tagged as Primary in the EC2 API response.

pkg/awsutils/awsutils.go
```diff
@@ -458,42 +449,17 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(macStr string) (ENIMetadat
...
-	missingIMDS := privateIPv4Set.Difference(imdsIPv4Set).List()
-	missingDNI := imdsIPv4Set.Difference(privateIPv4Set).List()
-	if len(missingIMDS) > 0 {
-		strMissing := strings.Join(missingIMDS, ",")
-		log.Debugf("getENIMetadata: DescribeNetworkInterfaces(%s) yielded private IPv4 addresses %s that were not yet found in IMDS.", eni, strMissing)
-	}
-	if len(missingDNI) > 0 {
-		strMissing := strings.Join(missingDNI, ",")
-		log.Debugf("getENIMetadata: IMDS query yielded stale IPv4 addresses %s that were not found in DescribeNetworkInterfaces(%s).", strMissing, eni)
-	}
```
jaypipes, line 484: The above code existed in order to determine that IMDS had stale information. `cache.DescribeENI()` is the only thing that actually ends up calling the DescribeNetworkInterfaces EC2 API call. By removing the call to EC2's DescribeNetworkInterfaces API call here, the caller of the cache `getENIMetadata()` call will now only look at potentially stale IMDS information. The only caller of `getENIMetadata()` happens to be the cache's `GetAttachedENIs()` method, which itself is called during ipamd's `nodeInit()` method:

https://github.com/aws/amazon-vpc-cni-k8s/blob/599fed32fdc8466b74e9468901a7d1bf4c228b42/pkg/ipamd/ipamd.go#L277

That method, unfortunately, MUST have the most up-to-date information returned to it, which means that DescribeNetworkInterfaces EC2 API call should be made, not IMDS. This is unfortunately what I was dealing with when I wrote the above code.

The other place that `GetAttachedENIs()` is called is in ipamd's `waitENIAttached()` method:

https://github.com/aws/amazon-vpc-cni-k8s/blob/599fed32fdc8466b74e9468901a7d1bf4c228b42/pkg/ipamd/ipamd.go#L799

In *this* case, we actually DO want to only check IMDS and not the EC2 DescribeNetworkInterfaces API call. So, I think the appropriate thing to do is to break up `GetAttachedENIs()` into an EC2-calling method and an IMDS-calling method and call one or the other depending on whether it is in `nodeInit()` or `waitENIAttached()`.

Otherwise, we really don't know whether the node is being initialized with stale data or not.

Thoughts?

pkg/awsutils/awsutils.go
SaranBalaji90: nit: you can toggle primary value without isFirst.

pkg/ipamd/ipamd.go
```diff
@@ -165,19 +165,22 @@ var (
...
+	awsClient           awsutils.APIs
+	dataStore           *datastore.DataStore
+	k8sClient           k8sapi.K8SAPIs
+	useCustomNetworking bool
+	eniConfig           eniconfig.ENIConfig
+	criClient           cri.APIs
+	networkClient       networkutils.NetworkAPIs
+	maxIPsPerENI        int
+	maxENI              int
+	// unmanagedENIs is a list of ENIs tagged with
```
SaranBalaji90, line 177: nit: missing "node.k8s.amazonaws.com/no_manage" at the end of this comment?

pkg/ipamd/ipamd_test.go
```diff
@@ -372,9 +372,7 @@ func TestNodeIPPoolReconcile(t *testing.T) {
 		},
 	}, nil)
 
-	mockAWS.EXPECT().GetPrimaryENI().Return(primaryENIid)
-
-	mockAWS.EXPECT().GetPrimaryENI().Return(primaryENIid)
+	mockAWS.EXPECT().GetPrimaryENI().AnyTimes().Return(primaryENIid)
```
SaranBalaji90, line 375: If its just 2 times, can you use .Times(2)?

pkg/ipamd/ipamd.go
```diff
@@ -1102,18 +1100,66 @@ func getMinimumIPTarget() int {
 	return noMinimumIPTarget
 }
 
-func filterUnmanagedENIs(enis []awsutils.ENIMetadata) ([]awsutils.ENIMetadata, int) {
-	numFiltered := 0
+// filterUnmanagedENIs filters out ENIs marked with the "node.k8s.amazonaws.com/no_manage" tag
+func (c *IPAMContext) filterUnmanagedENIs(enis []awsutils.ENIMetadata) []awsutils.ENIMetadata {
```
SaranBalaji90, line 1104: Can you add some tests for this?

pkg/ipamd/ipamd.go
```diff
@@ -1102,18 +1100,66 @@ func getMinimumIPTarget() int {
...
+	noManageENIs := make([]string, 0, len(enis))
+	// Update filter if needed
+	if c.needTagRefresh() {
```
SaranBalaji90, line 1107: Isn't there is a race condn if we invoke nodeIpPoolReconcile and if the ENI is just attached then CNI will try to add/remove IPs from the one which it shouldnt manage right?

pkg/awsutils/awsutils.go
mogren: Great points @jaypipes. I know the reason for that diff and I did test a version where I had a "callEC2" bool, to disable it for the `waitENIAttached()` function. That said, I think there are two improvements to be made here. One is to move the tag check out from metadata, as has been done already, since that data is not available from IMDS. In addition we should make a separate `getENIInfoFromEC2()` function that does a batch-describe call of all attached ENIs to EC2 instead of calling them one by one, to be used by the init function to verify the status of all ENIs when the ipamd pod starts up. Will update the PR.

pkg/ipamd/ipamd.go
mogren: Yes, will have to rewrite most of them after these latest changes.

pkg/ipamd/ipamd.go
mogren: Yes. I've rewritten this to not allow any tag refreshing. The ENIs will have to be tagged before they are attached. Another option would be to actively filter for the ones we tag ourselves.. 🤔

pkg/awsutils/awsutils.go
mogren: Actually not, since I use the address of it. If I do, they will all refer to the same false boolean.

SaranBalaji90: ah missed the later part.

pkg/awsutils/awsutils.go
```diff
@@ -895,23 +899,101 @@ func (cache *EC2InstanceMetadataCache) DescribeENI(eniID string) ([]*ec2.Network
...
 	}
-	tags := make(map[string]string, len(result.NetworkInterfaces[0].TagSet))
-	for _, tag := range result.NetworkInterfaces[0].TagSet {
-		if tag.Key == nil || tag.Value == nil {
-			log.Errorf("nil tag on ENI: %v", eniID)
-			continue
+	return result.NetworkInterfaces[0].PrivateIpAddresses, nil
+}
+
+// DescribeAllENIs returns the ENIMetadata and tags for all attached ENIs
```
jaypipes, line 912: Please update the docstring here to mention that this method calls EC2 DescribeNetworkInterfaces API call.

pkg/awsutils/awsutils.go
```diff
@@ -895,23 +899,101 @@ func (cache *EC2InstanceMetadataCache) DescribeENI(eniID string) ([]*ec2.Network
...
+		ec2IPv4Set.Insert(aws.StringValue(privateIPv4.PrivateIpAddress))
+		if aws.BoolValue(privateIPv4.Primary) {
+			ec2IPv4PrimaryIP = aws.StringValue(privateIPv4.PrivateIpAddress)
+		}
+	}
+	missingIMDS := ec2IPv4Set.Difference(imdsIPv4Set).List()
+	missingDNI := imdsIPv4Set.Difference(ec2IPv4Set).List()
+	if len(missingIMDS) > 0 {
+		strMissing := strings.Join(missingIMDS, ",")
+		log.Errorf("compareIPLists: DescribeNetworkInterfaces(%s) yielded private IPv4 addresses %s that were not yet found in IMDS.", eniID, strMissing)
```
jaypipes, line 988: here and below, should be `log.Debugf` (or potentially `log.Infof`) not `log.Errorf` since a) we're not returning an error b) the message is merely informationl and c) there's nothing the user/operator can do about it. :)

pkg/awsutils/awsutils.go
```diff
@@ -895,23 +899,101 @@ func (cache *EC2InstanceMetadataCache) DescribeENI(eniID string) ([]*ec2.Network
...
+			tagMap[eniMetadata.ENIID] = tags
 		}
-		tags[*tag.Key] = *tag.Value
 	}
+	return allENIs, tagMap, nil
+}
 
-	return result.NetworkInterfaces[0].PrivateIpAddresses, tags, result.NetworkInterfaces[0].Attachment.AttachmentId, nil
+// compareIPLists compares the IP and metadata returned by IMDS and the EC2 API DescribeNetworkInterfaces calls
+func compareIPLists(eniID string, imdsIPv4s, ec2IPv4s []*ec2.NetworkInterfacePrivateIpAddress) {
```
jaypipes, line 964: maybe call this function `logStaleData()` or `logOutOfSyncState()` or something like that to make it clear that the function is looking for stale/out-of-sync information and just logging it.

pkg/ipamd/ipamd.go
```diff
@@ -187,6 +188,38 @@ type IPAMContext struct {
 	terminating            int32 // Flag to warn that the pod is about to shut down.
 }
 
+// UnmanagedENISet keeps a set of ENI IDs for ENIs tagged with "node.k8s.amazonaws.com/no_manage"
+type UnmanagedENISet struct {
+	set  map[string]bool
+	lock sync.RWMutex
```
jaypipes, line 194: note that you can "inherit" from `sync.RWMutex` here instead of making a named `lock` field:

```go
type UnmanagedENISet struct {
    sync.RWMutex
    enis map[string]bool
}
```

and then you can call `Lock()` directly, potentially from a mutator method of `*UnamanagedENISet`) like so:

```go
func (u *UnmanagedENISet) add(eniID string) {
    u.Lock()
    defer u.Unlock()
    if len(u.data) == 0 {
        u.data = make(map[string]bool)
    }
    u.data[eniID] = true
}
```

pkg/ipamd/ipamd.go
```diff
@@ -187,6 +188,38 @@ type IPAMContext struct {
...
+	if len(c.unmanagedENIs.set) == 0 {
+		c.unmanagedENIs.set = make(map[string]bool)
+	}
+	for eniID, tags := range tagMap {
+		if tags[eniNoManageTagKey] == "true" {
+			if eniID == c.awsClient.GetPrimaryENI() {
+				log.Debugf("Ignoring no_manage tag on primary ENI %s", eniID)
+			} else {
+				log.Debugf("Marking ENI %s tagged with %s as being unmanaged", eniID, eniNoManageTagKey)
+				c.unmanagedENIs.set[eniID] = true
```
jaypipes, line 217: and then here, just do:

```go
c.unmanagedENIs.add(eniID)
```

pkg/ipamd/ipamd.go
```diff
@@ -187,6 +188,38 @@ type IPAMContext struct {
...
+
+func (c *IPAMContext) addUnmanagedENIs(tagMap map[string]awsutils.TagMap) {
+	if len(tagMap) == 0 {
+		return
+	}
+	c.unmanagedENIs.lock.Lock()
+	defer c.unmanagedENIs.lock.Unlock()
+	if len(c.unmanagedENIs.set) == 0 {
+		c.unmanagedENIs.set = make(map[string]bool)
+	}
```
jaypipes, line 210: and now you can remove the above since you've hidden the implementation details of the lock and map init in the `*UnmanagedENISet.add()` method.

pkg/ipamd/ipamd.go
```diff
@@ -187,6 +188,38 @@ type IPAMContext struct {
 	terminating            int32 // Flag to warn that the pod is about to shut down.
 }
 
+// UnmanagedENISet keeps a set of ENI IDs for ENIs tagged with "node.k8s.amazonaws.com/no_manage"
+type UnmanagedENISet struct {
+	set  map[string]bool
```
jaypipes, line 193: how about "data" instead of "set" as a field name?

pkg/ipamd/ipamd.go
```diff
@@ -723,8 +757,8 @@ func (c *IPAMContext) tryAssignIPs() (increasedPool bool, err error) {
 				return false, errors.Wrap(err, fmt.Sprintf("failed to allocate one IP addresses on ENI %s, err: %v", eni.ID, err))
 			}
 		}
-
-		ec2Addrs, _, err := c.getENIaddresses(eni.ID)
+		// This call to EC2 is needed to verify which IPs got attached to this ENI.
```
jaypipes, line 760: :+1:

pkg/awsutils/awsutils.go
mogren: `logOutOfSyncState()` sounds good, updated.

pkg/ipamd/ipamd.go
mogren: Much nicer, thanks!

PR #837, opened 2020-02-11
Support architecture targeted builds

jahkeup:
*Issue #, if available:*

Related: #687 

*Description of changes:*

These changes clean up the Makefile in order to widely update references for building cross-arch artifacts and images in some configurations. I've added many docstrings to the updated sections as well as the left in place sections, hopefully this a welcome addition - please let me know if I've misunderstood or misinterpreted usages in any way!

With these changes in place I have been able to build container images for both `aarch64` and `x86_64` (on the same host using QEMU as `binfmt` handlers!) as well as the bare executables by providing the `GOARCH` styled architecture names as `ARCH`:

```bash
# Build container images for aarch64 (either *on* aarch64 or using `binfmt` method)
make all ARCH=arm64
# and the executables:
make build-linux ARCH=arm64

# Build container images for x86_64 explicitly
make all ARCH=amd64
# and the executables:
make build-linux ARCH=amd64
```

The updated targets also take into account the currently running host and enforce the use of the native build where required (such as testing with `-race` enabled).

I've updated the Circle CI builds to utilize these code paths that build for `aarch64` and `x86_64`, though I have not yet tested this and would like help to validate this proposed change.

cc: @mogren 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #912, opened 2020-04-17
Ensure makefile exits with non-zero code when code is unformatted

mogren:
*Description of changes:*
#837 caused the `check-format` target to no longer exit with an error on unformatted files. This PR adds a check of how many files are not formatted, and if it is not `0`, we exit with an error code showing how many files that failed the check. Verified that `make format` still works correctly and formats the code.


mogren:
@astrived Without this change, having incorrectly formatted files passes the format test. This check worked before, but got disabled in #837

jaypipes:
Merging since does not affect integration testing...

PR #913, opened 2020-04-17
Add code coverage reports

mogren:
*Description of changes:*
Initial support for Codcov to show unit test coverage. 

Sample output: https://codecov.io/gh/mogren/amazon-vpc-cni-k8s/branch/add-code-cov


.codecov.yml
```diff
@@ -0,0 +1,37 @@
+# To validate:
+#   cat codecov.yml | curl --data-binary @- https://codecov.io/validate
+
+codecov:
+  notify:
+    require_ci_to_pass: no
```
astrived, line 6: Have we tried this change with ci_to_pass yes ?

mogren: Yes, but then we won't get the coverage update until after the integration tests have been completed. Since they are gated, and then also take around 3.5h, it will just delay the report a lot.

astrived: Got it.. this is an useful change, approving it.

Issue #914, opened 2020-04-17
segfault after node rebooted


kgtw:
We use kured to reboot nodes for security patches, after a node got rebooted the cni pod consistently failed to start due to a segfault.

Kubernetes: v1.16.7
CNI Plugin: 0.7.5-1
AWS CNI: 1.5.5

```
====== Installing AWS-CNI ======
====== Starting amazon-k8s-agent ======
2020-04-17T10:55:47.126Z [INFO]	Starting L-IPAMD v1.5.5  ...
2020-04-17T10:55:47.161Z [INFO]	Testing communication with server
2020-04-17T10:55:47.162Z [INFO]	Running with Kubernetes cluster version: v1.16. git version: v1.16.7. git tree state: clean. commit: be3d344ed06bff7a4fc60656200a93c74f31f9a4. platform: linux/amd64
2020-04-17T10:55:47.162Z [INFO]	Communication with server successful
2020-04-17T10:55:47.162Z [INFO]	Starting Pod controller
2020-04-17T10:55:47.162Z [INFO]	Waiting for controller cache sync
2020-04-17T10:55:47.163Z [DEBUG]	Discovered region: eu-central-1
2020-04-17T10:55:47.164Z [DEBUG]	Found availability zone: eu-central-1a
2020-04-17T10:55:47.164Z [DEBUG]	Discovered the instance primary ip address: 172.21.82.254
2020-04-17T10:55:47.165Z [DEBUG]	Found instance-id: i-057d745cb8e27e423
2020-04-17T10:55:47.165Z [DEBUG]	Found instance-type: r5.2xlarge
2020-04-17T10:55:47.166Z [DEBUG]	Found primary interface's MAC address: 02:f5:dc:34:cd:30
2020-04-17T10:55:47.166Z [DEBUG]	Discovered 2 interfaces.
2020-04-17T10:55:47.167Z [DEBUG]	Found device-number: 1
2020-04-17T10:55:47.167Z [DEBUG]	Found account ID: 737873494095
2020-04-17T10:55:47.168Z [DEBUG]	Found eni: eni-0208359d78f41cd6f
2020-04-17T10:55:47.168Z [DEBUG]	Found device-number: 0
2020-04-17T10:55:47.169Z [DEBUG]	Found eni: eni-0b8b4a6f654608f8c
2020-04-17T10:55:47.169Z [DEBUG]	Found ENI eni-0b8b4a6f654608f8c is a primary ENI
[...]

PR #915, opened 2020-04-17
ensure non-nil Attachment in getENIAttachmentID

jaypipes:
Practice good code safety in the `EC2MetadataCache.getENIAttachmentID()` method
by not assuming that either the `DescribeNetworkInterfacesOutput`
struct's `NetworkInterfaces` field is not empty and that the first
`NetworkInterface` struct in that collection has a non-nil `Attachment`
field.

Fixes Issue #914

### DIFF
--- a/pkg/awsutils/awsutils_test.go
+++ b/pkg/awsutils/awsutils_test.go
@@ -305,6 +305,68 @@ func TestAWSGetFreeDeviceNumberNoDevice(t *testing.T) {
 	assert.Error(t, err)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/awsutils/awsutils_test.go
```diff
@@ -305,6 +305,68 @@ func TestAWSGetFreeDeviceNumberNoDevice(t *testing.T) {
 	assert.Error(t, err)
 }
 
+func TestGetENIAttachmentID(t *testing.T) {
```
mogren, line 308: Nice!

PR #916, opened 2020-04-17
Bump containernetworking dependencies only

mogren:
*Description of changes:*
* Bump of containernetworking dependencies
* The `ns` package has moved to the `plugins` package.
* Bumped portmap version


pkg/ipamd/ipamd.go
```diff
@@ -150,13 +150,13 @@ var (
...
 			Name: "awscni_add_ip_req_count",
-			Help: "The number of add IP address request",
+			Help: "The number of add IP address requests",
 		},
 	)
 	delIPCnt = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
 			Name: "awscni_del_ip_req_count",
-			Help: "The number of delete IP address request",
+			Help: "The number of delete IP address requests",
```
astrived, line 159: Is this change primarily for readability ?

pkg/cninswrapper/ns.go
```diff
@@ -13,7 +13,7 @@
 
 package cninswrapper
 
-import "github.com/containernetworking/cni/pkg/ns"
+import "github.com/containernetworking/plugins/pkg/ns"
```
astrived, line 16: Have all these packages moved under plugins directory now ? Looking at this PR in isolation is not giving me that insight. Can we still continue to use cni directory packages ?

pkg/ipamd/ipamd.go
mogren: Yes, sentence was not correct.

pkg/cninswrapper/ns.go
mogren: The upstream package was moved in [v0.6.0 back in August 2017](https://github.com/containernetworking/cni/releases/tag/v0.6.0), so we have to import the files from where they are now.

astrived: Thanks for clarification. Lets run integration changes post this at least once again before the release.

go.mod
```diff
@@ -15,7 +16,7 @@ require (
 	github.com/google/gofuzz v0.0.0-20170612174753-24818f796faf // indirect
 	github.com/googleapis/gnostic v0.2.0 // indirect
 	github.com/gregjones/httpcache v0.0.0-20190212212710-3befbb6ad0cc // indirect
-	github.com/hashicorp/golang-lru v0.0.0-20180201235237-0fb14efe8c47 // indirect
```
jaypipes, line 18: \o/ always good to get rid of older non-semver deps.

PR #919, opened 2020-04-20
Use -buildmode=pie for binaries

mogren:
*Description of changes:*
* Enable `-buildmode=pie`
* Override for cross-compiles since we have native dependencies

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,8 +55,7 @@ DOCKER_ARCH = $(lastword $(subst :, ,$(filter $(ARCH):%,amd64:amd64 arm64:arm64v
 IMAGE_ARCH_SUFFIX = $(addprefix -,$(filter $(ARCH),arm64))
 # GOLANG_IMAGE is the building golang container image used.
 GOLANG_IMAGE = golang:1.13-stretch
-# For the requseted build, these are the set of Go specific build environment
-# variables.
+# For the requested build, these are the set of Go specific build environment variables.
 export GOARCH ?= $(ARCH)
 export GOOS = linux
 export CGO_ENABLED = 0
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -94,7 +93,8 @@ dist: all
 	docker save $(METRICS_IMAGE_NAME) | gzip > $(METRICS_IMAGE_DIST)
 
 # Build the VPC CNI plugin agent using the host's Go toolchain.
-build-linux: BUILD_FLAGS = -ldflags '-s -w $(LDFLAGS)'
+BUILD_MODE ?= -buildmode=pie
+build-linux: BUILD_FLAGS = $(BUILD_MODE) -ldflags '-s -w $(LDFLAGS)'
```
astrived, line 97: Apart from build flags is there a way you are validating the the buildmode=pie is effective ?

mogren: It is visible in multiple ways. First, from the build logs:
```
Step 13/18 : RUN make build-linux
 ---> Running in 224a2ad9101a
go build -buildmode=pie -ldflags '-s -w -X main.version=v1.6.1-pie' -o aws-k8s-agent     ./cmd/aws-k8s-agent
# github.com/aws/amazon-vpc-cni-k8s/cmd/aws-k8s-agent
loadinternal: cannot find runtime/cgo
go build -buildmode=pie -ldflags '-s -w -X main.version=v1.6.1-pie' -o aws-cni           ./cmd/routed-eni-cni-plugin
# github.com/aws/amazon-vpc-cni-k8s/cmd/routed-eni-cni-plugin
loadinternal: cannot find runtime/cgo
go build -buildmode=pie -ldflags '-s -w -X main.version=v1.6.1-pie' -o grpc-health-probe ./cmd/grpc-health-probe
# github.com/aws/amazon-vpc-cni-k8s/cmd/grpc-health-probe
loadinternal: cannot find runtime/cgo
```
(Note, the warnings are spurious and tracked in https://github.com/golang/go/issues/31544. We are not using `cgo`.)

It can also be verified on the worker nodes:

**Without PIE:**
```
[ec2-user@ip-10-10-xx-xxx bin]$ file aws-cni
aws-cni: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, stripped
```
**With PIE:**
```
[ec2-user@ip-10-10-xx-xxx bin]$ file aws-cni
aws-cni: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked (uses shared libs), for GNU/Linux 2.6.32, BuildID[sha1]=4a17a0b6e5655e3bc52e789a3ae8ff1ab75ebac8, stripped
```
`shared object` means PIE, and the output has been fixed in more recent versions of the `file` command (It was added in 5.36 of `file`, we have 5.11 in AL2)

astrived: great!

Issue #158, opened 2018-08-17
Copy Containernetworking CNI plugins in an init container


micahhausler:
Right now we rely on the CNI binaries being present on the host, but we should add an init container to our DaemonSet that copies CNI binaries to the host. Our [AMI currently bakes these in](https://github.com/awslabs/amazon-eks-ami/blob/9736e73/install-worker.sh#L78-L86), but this CNI provider should provide these itself

Essentially the init container Dockerfile would look need to look like 

```Dockerfile
FROM amazonlinux

ENV CNI_STAGE_DIR /opt/stage/cni
ENV CNI_BIN_DIR /opt/cni/bin
RUN mkdir -p /opt/cni/bin

ENV CNI_VERSION v0.6.0
RUN wget https://github.com/containernetworking/cni/releases/download/${CNI_VERSION}/cni-amd64-${CNI_VERSION}.tgz \
    && tar -xvf cni-amd64-${CNI_VERSION}.tgz -C $CNI_STAGE_DIR \
    && rm cni-amd64-${CNI_VERSION}.tgz

ENV CNI_PLUGIN_VERSION v0.7.1 
RUN wget https://github.com/containernetworking/plugins/releases/download/${CNI_PLUGIN_VERSION}/cni-plugins-amd64-${CNI_PLUGIN_VERSION}.tgz \
    && tar -xvf cni-plugins-amd64-${CNI_PLUGIN_VERSION}.tgz -C $CNI_STAGE_DIR \
    && rm cni-plugins-amd64-${CNI_PLUGIN_VERSION}.tgz

ENTRYPOINT ["/usr/bin/cp", "-r", "$CNI_STAGE_DIR", "/opt/cni/bin"]
```

mogren:
@micahhausler Fixed, only took 23 months! 😄

PR #935, opened 2020-04-30
Fix order of file copies in entrypoint.sh

dthorsen:
Fixes this error when you upgrade the CNI because the CNI starts prior to portmap
is copied and portmap is then sometimes in use by the time the cp commands execute

*Issue #, if available:*

*Description of changes:*
This fixes the following error from causing the CNI pod to enter crashloopbackoff:
`cp: cannot create regular file '/host/opt/cni/bin/portmap': Text file busy`

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -60,7 +60,15 @@ wait_for_ipam() {
 
 echo -n "Copying portmap binary... "
 
-cp portmap "$HOST_CNI_BIN_PATH"
+if [[ -f "$HOST_CNI_BIN_PATH" ]]
+then
+    mv "$HOST_CNI_BIN_PATH" "$HOST_CNI_BIN_PATH.old"
+    cp portmap "$HOST_CNI_BIN_PATH"
+    rm "$HOST_CNI_BIN_PATH.old"
+else
+    cp portmap "$HOST_CNI_BIN_PATH"
+fi
+
 
 echo -n "Starting IPAM daemon in the background ... "
 ./aws-k8s-agent | tee -i "$AGENT_LOG_PATH" 2>&1 &
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -58,11 +58,9 @@ wait_for_ipam() {
     return 1
 }
 
-echo -n "Copying CNI plugin binaries and config files ... "
+echo -n "Copying portmap binary... "
 
 cp portmap "$HOST_CNI_BIN_PATH"
-cp aws-cni "$HOST_CNI_BIN_PATH"
-cp aws-cni-support.sh "$HOST_CNI_BIN_PATH"
 
 echo -n "Starting IPAM daemon in the background ... "
 ./aws-k8s-agent | tee -i "$AGENT_LOG_PATH" 2>&1 &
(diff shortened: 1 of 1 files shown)


mogren:
Hi @dthorsen, good catch finding that cp error message! In this case, we would like to only copy the portmap and not the CNI binary and config before ipamd starts, since that might make a new node ready even if it cannot start. Could you update the PR to only copy the portmap early?

That said, I'm working on a PR to [add an init container](https://github.com/mogren/amazon-vpc-cni-k8s/blob/init-container/scripts/dockerfiles/Dockerfile.init) instead, to make the setup part more explicit. (Issue #158)

mogren:
Taking a second look at this, I wonder if we should check if the file exists, if so rename it to `portmap.old`, copy in the (potentially new) `portmap` then delete `portmap.old`. That seems like a safer way to do this update in case some process is currently executing. Thanks again for noticing this, it has been hard to track down.

dthorsen:
Thanks @mogren! That makes sense I think. I moved only the portmap binary to copy early.

dthorsen:
Oh that might make sense. We can totally do that.

dthorsen:
I think this does what you are asking. The timing difference here is probably milliseconds though. I am not certain it buys much over simply copying over the file. I suspect that if something was actually executing portmap it would be cached in RAM anyway and I don't think those few extra milliseconds will protect from much.

scripts/entrypoint.sh
```diff
@@ -58,6 +58,18 @@ wait_for_ipam() {
     return 1
 }
 
+echo -n "Copying portmap binary... "
+
+if [[ -f "$HOST_CNI_BIN_PATH" ]]
+then
+    mv "$HOST_CNI_BIN_PATH" "$HOST_CNI_BIN_PATH.old"
```
mogren, line 65: This will rename the whole bin directory. How about something like
```bash
HOST_PORTMAP="$HOST_CNI_BIN_PATH/portmap"
if [[ -f "$HOST_PORTMAP" ]]
then
    mv "$HOST_PORTMAP" "$HOST_PORTMAP.old"
```

dthorsen: Ah yes, of course. Thanks

scripts/entrypoint.sh
```diff
@@ -58,6 +58,19 @@ wait_for_ipam() {
...
+
+HOST_PORTMAP="$HOST_CNI_BIN_PATH/portmap"
+if [[ -f "$HOST_PORTMAP" ]]
+then
+    mv "$HOST_PORTMAP" "$HOST_PORTMAP.old"
+    cp portmap "$HOST_CNI_BIN_PATH"
+    rm "$HOST_PORTMAP.old"
+else
+    cp portmap "$HOST_CNI_BIN_PATH"
+fi
```
mogren, line 71: Having taken [another look](https://superuser.com/a/1139861/82514), I think this could be simplified even more. What do you think about:
```
HOST_PORTMAP="$HOST_CNI_BIN_PATH/portmap"
if [[ -f "$HOST_PORTMAP" ]]; then
    rm "$HOST_PORTMAP"
fi
cp portmap "$HOST_CNI_BIN_PATH"
```

jaypipes: Agreed @mogren. ++

dthorsen: Simplification engaged

PR #936, opened 2020-04-30
Disable ENI sync flag

haouc:
*Description of changes:*

If the all resources are provisioned upfront, based on the new flag, we can stop doing calls to EC2 to get ENI and IP.

* Add a new flag `DISABLE_NETWORK_RESOURCE_PROVISIONING` to control IP update check.

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -109,6 +109,10 @@ const (
 	// eniNoManageTagKey is the tag that may be set on an ENI to indicate ipamd
 	// should not manage it in any form.
 	eniNoManageTagKey = "node.k8s.amazonaws.com/no_manage"
+
+	// disableENIProvisioning is used to specify that ENI doesn't need to be synced during initializing a pod.
+	envDisableENIProvisioning = "DISABLE_NETWORK_RESOURCE_PROVISIONING"
+	noDisableENIProvisioning = false
 )
 
 var log = logger.Get()
@@ -186,6 +190,7 @@ type IPAMContext struct {
 	// so that we don't reconcile and add it back too quickly if IMDS lags behind reality.
 	reconcileCooldownCache ReconcileCooldownCache
 	terminating            int32 // Flag to warn that the pod is about to shut down.
+	disableENIProvisioning bool
 }
 
 // UnmanagedENISet keeps a set of ENI IDs for ENIs tagged with "node.k8s.amazonaws.com/no_manage"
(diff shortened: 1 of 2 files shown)


pkg/ipamd/ipamd.go
```diff
@@ -502,8 +508,12 @@ func (c *IPAMContext) getLocalPodsWithRetry() ([]*k8sapi.K8SPodInfo, error) {
 func (c *IPAMContext) StartNodeIPPoolManager() {
 	sleepDuration := ipPoolMonitorInterval / 2
 	for {
-		time.Sleep(sleepDuration)
-		c.updateIPPoolIfRequired()
+		waitLongEnough := waitLongEnoughSinceLastAction(c.lastNodeIPPoolAction)
+		// TODO: adding another condition of is_new_ENI_available
+		if !c.disableENIProvisioning && waitLongEnough {
```
mogren, line 513: I don't think this is needed. The check for new ENIs happens in `c.nodeIPPoolReconcile`

pkg/ipamd/ipamd.go
```diff
@@ -947,6 +957,15 @@ func ipamdErrInc(fn string) {
...
 
+func waitLongEnoughSinceLastAction(lastTime time.Time) bool {
+	return time.Now().Sub(lastTime) > nodeIPPoolReconcileInterval
+}
+
+func (c *IPAMContext) hasNewENIAvailable() bool {
+	// TODO: check if attachENI is not in ENIPool yet.
+	return false
+}
+
```
mogren, line 968: Not needed here.

pkg/ipamd/ipamd.go
```diff
@@ -1148,6 +1167,21 @@ func getMinimumIPTarget() int {
 	return noMinimumIPTarget
 }
 
+func disablingENIProvisioning() bool {
```
mogren, line 1170: This function could be simplified using :

https://github.com/aws/amazon-vpc-cni-k8s/blob/8ffcd8a6168295df4ab2fd62d0cc3e3b5e4bae3d/pkg/ipamd/introspect.go#L203

pkg/ipamd/ipamd.go
```diff
@@ -502,8 +508,10 @@ func (c *IPAMContext) getLocalPodsWithRetry() ([]*k8sapi.K8SPodInfo, error) {
 func (c *IPAMContext) StartNodeIPPoolManager() {
 	sleepDuration := ipPoolMonitorInterval / 2
 	for {
-		time.Sleep(sleepDuration)
-		c.updateIPPoolIfRequired()
+		if !c.disableENIProvisioning{
+			time.Sleep(sleepDuration)
+			c.updateIPPoolIfRequired()
+		}
```
mogren, line 514: If the `time.Sleep()` call is inside the if-statement, the `nodeIPPoolReconcile()` will be called twice as often. It's a bit weird, but since that should not trigger EC2 API calls we should be fine.

haouc: hmmm, that is a good point. Let me make a compensate on the interval in the whitelist code change. Thanks!

PR #941, opened 2020-05-01
Adding whitelist to filter accountId which is not included in given list.

haouc:
*Issue #, if available:*
We want to limit tracking ENI calls to ec2.
*Description of changes:*
Added a whitelist that created base don given accounts. The current owner account id will be checked and only whitelisted id can keep tracking ENI and IPs.

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -154,6 +154,9 @@ type APIs interface {
 
 	// GetPrimaryENImac returns the mac address of the primary ENI
 	GetPrimaryENImac() string
+
+	// GetAccountId returns the account id/owner id
+	GetAccountId() string
 }
 
 // EC2InstanceMetadataCache caches instance metadata
@@ -1232,4 +1235,9 @@ func (cache *EC2InstanceMetadataCache) GetPrimaryENI() string {
 // GetPrimaryENImac returns the mac address of primary eni
 func (cache *EC2InstanceMetadataCache) GetPrimaryENImac() string {
 	return cache.primaryENImac
+}
+
+// GetAccountId returns the account id/owner id
+func (cache *EC2InstanceMetadataCache) GetAccountId() string {
+	return cache.accountID
 }
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -190,6 +190,7 @@ type IPAMContext struct {
 	// so that we don't reconcile and add it back too quickly if IMDS lags behind reality.
 	reconcileCooldownCache ReconcileCooldownCache
 	terminating            int32 // Flag to warn that the pod is about to shut down.
+	accountWhitelist     []string
 }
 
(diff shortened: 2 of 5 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -1233,3 +1236,8 @@ func (cache *EC2InstanceMetadataCache) GetPrimaryENI() string {
 func (cache *EC2InstanceMetadataCache) GetPrimaryENImac() string {
 	return cache.primaryENImac
 }
+
+// GetAccountId returns the account id/owner id
+func (cache *EC2InstanceMetadataCache) GetAccountId() string {
```
haouc, line 1241: I think this info should be reliable?

pkg/ipamd/ipamd.go
```diff
@@ -501,14 +507,30 @@ func (c *IPAMContext) getLocalPodsWithRetry() ([]*k8sapi.K8SPodInfo, error) {
...
+	// If owner id is not set, or not in the whitelist, don't do ENI tracking.
+	shouldWhitelistTheAccount :=
+		c.accountWhitelist == nil || contains(c.accountWhitelist, c.awsClient.GetAccountId())
+
 	for {
 		time.Sleep(sleepDuration)
-		c.updateIPPoolIfRequired()
+		if shouldWhitelistTheAccount {
+			c.updateIPPoolIfRequired()
+		}
```
mogren, line 519: I think this is not the right place to check this. Maybe we need a more clear specification of the ask here. All node local updates still need to happen, but calls to the EC2 API should be avoided, so the checks need to be more granular.

pkg/ipamd/ipamd.go
```diff
@@ -109,6 +109,10 @@ const (
 	// eniNoManageTagKey is the tag that may be set on an ENI to indicate ipamd
 	// should not manage it in any form.
 	eniNoManageTagKey = "node.k8s.amazonaws.com/no_manage"
+
+	// whitelistedAccountIds is the account ids which are allowed to track ENI.
+	// account ids should be delimited by comma
+	envWhitelistedAccountIds = "ALLOWED_ENI_TRACKING_IDS"
```
mogren, line 115: The comment was a bit hard to read. How about:
```go
	// envWhitelistedAccountIds is the account IDs which are allowed to call EC2.
	// Account IDs should be a comma separated list.
	envWhitelistedAccountIds = "ALLOWED_ENI_TRACKING_IDS"
```

pkg/ipamd/ipamd_test.go
```diff
@@ -14,6 +14,7 @@
 package ipamd
 
 import (
+	_ "fmt"
```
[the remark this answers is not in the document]
haouc: Ah, thanks for the good catch. A debugging log had leftover that should be removed.

pkg/ipamd/ipamd.go
haouc: Will update. Thanks.

mogren:
I don't think this is needed, and if it is we would still have to rewrite it because of the changes that have been merged so far. Closing for now.

Issue #64, opened 2018-05-02
AWS CNI should install loopback cni plugin


nckturner:
This is not required by AWS CNI but kubelet fails without it.

worzel666:
Hey, we're experiencing this issue as well, @nckturner did you find a fix?

mogren:
Not needed.

scrothers:
@mogren, 

  Warning  FailedCreatePodSandBox  110s (x37 over 5m2s)   kubelet, ip-10-1-131-66.us-east-2.compute.internal  (combined from similar events): Failed to create pod sandbox: rpc error: code = Unknown desc = failed to set up sandbox container "b576e7cf5b25584b510ca25a0a97ef39590b1eb688c2a7075879bc744b97837b" network for pod "default-http-backend-67cf578fc4-l4sv4": networkPlugin cni failed to set up pod "default-http-backend-67cf578fc4-l4sv4_ingress-nginx" network: failed to find plugin "loopback" in path [/opt/cni/bin]

Seems to be needed.

PR #955, opened 2020-05-05
Add init container

mogren:
*Issue #, if available:*
#158, #64, 

*Description of changes:*
Add an init container to the CNI. 

* Add init container to copy `portmap` and `loopback` binaries and set up `rp_filter`
* Remove privileged from `aws-node` and leave `"NET_ADMIN"` capability only
* Add example "master" config
* Using `sysctl` from `procps-ng` to update the IPv4 config.
* Tested that the default config, no `AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER=false`, still works

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@
 
 .PHONY: all dist check clean \
 		lint format check-format vet docker-vet \
-		build-linux docker \
+		build-linux docker init \
 		unit-test unit-test-race build-docker-test docker-func-test \
 		build-metrics docker-metrics \
 		metrics-unit-test docker-metrics-test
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -108,7 +118,14 @@ docker:
 		.
 	@echo "Built Docker image \"$(IMAGE_NAME)\""
 
-# Run the built cni container image to use in functional testing
+init:
```
jaypipes, line 121: can we call this target `docker-init` or `docker-build-init-image` to align with the other docker build-related targets?

config/master/aws-k8s-cni-cn.yaml
```diff
@@ -137,7 +139,9 @@
           "requests":
             "cpu": "10m"
         "securityContext":
-          "privileged": true
```
jaypipes, line 140: :+1:

Makefile
mogren: Renamed the target to `docker-init` since I had to rebase anyway.

PR #956, opened 2020-05-06
Add support for multiple IPs per pod and IPv6

anguslees:
- Add new ASSIGN_IPV4 and ASSIGN_IPV6 CNI config env vars, that default
  to true, false resp.

- Various additions to include obvious IPv6 equivalents alongside IPv4
  parameters.


anguslees:
Note! Almost no testing, because I think that requires boiling the ocean first :(
My primary goal here is to introduce the various internal golang changes without changing/breaking the existing IPv4 behaviour (which _is_ covered by existing tests).  We can fix any bugs in actual IPv6 functionality later.
Happy to follow some different path if you would prefer.  Eg: I could do single addr -> addrs array, but remove all the IPv6 bits from this PR.

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -69,13 +70,31 @@ type NetConf struct {
...
 // K8sArgs is the valid CNI_ARGS used for Kubernetes
 type K8sArgs struct {
 	types.CommonArgs
 
 	// IP is pod's ip address
 	IP net.IP
 
+	// IPS are pod's IP addresses. It is an error if both IP and
+	// IPS are defined, and IP is not one of the addresses in IPS.
+	IPS IPList
```
mogren, line 96: Nit: This is called `IPs` in the [containernetworking codebase](https://github.com/containernetworking/cni/blob/master/pkg/types/current/types.go#L128)

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -69,13 +70,31 @@ type NetConf struct {
 	PluginLogLevel string `json:"pluginLogLevel"`
 }
 
+// Implements encoding.TextUnmarshaler from comma-separated string.
+type IPList []net.IP
+
+func (l *IPList) UnmarshalText(data []byte) error {
```
mogren, line 76: Used by [LoadArgs](https://github.com/containernetworking/cni/blob/master/pkg/types/args.go#L101) when parsing arguments to the CNI through reflection, so never called directly.

test/integration/integration_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/config"
 	. "github.com/onsi/gomega"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
mogren, line 12: Nit: Not sure why the formatter does this.

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -311,47 +326,46 @@ func addContainerRule(netLink netlinkwrapper.NetLink, isToContainer bool, addr *
...
+func tearDownNS(addrs []*net.IP, table int, netLink netlinkwrapper.NetLink, log logger.Logger) error {
+	if len(addrs) == 0 {
 		return errors.New("can't tear down network namespace with no IP address")
 	}
-	// Remove to-pod rule
-	toContainerRule := netLink.NewRule()
-	toContainerRule.Dst = addr
-	toContainerRule.Priority = toContainerRulePriority
-	err := netLink.RuleDel(toContainerRule)
+	for _, addr := range addrs {
```
mogren, line 338: The `nil` check of `addr` is important. Calling the netLink wrapper with `nil` by accident will silently and without error delete the first `ip rule`. This is what caused all the issues with CNI v1.5.4 last year.

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -145,6 +164,35 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
...
+		k8sArgs.IP = k8sArgs.IPS[0]
+	}
+
+	// Default the host-side veth prefix to 'eni'.
+	if conf.VethPrefix == "" {
+		conf.VethPrefix = "eni"
+	}
+	if len(conf.VethPrefix) > 4 {
+		return errors.New("conf.VethPrefix can be at most 4 characters long")
+	}
```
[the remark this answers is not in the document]
anguslees: Oops, bad merge. Removed.

cmd/routed-eni-cni-plugin/cni.go
anguslees: Yes, but all the other fields in this struct are uppercase.
(I don't see why this struct needs to match the containernetworking Result struct - am I missing a connection here?)

test/integration/integration_test.go
anguslees: I don't know either.   We're probably using different versions of goformat, since there's nothing to stop that.

cmd/routed-eni-cni-plugin/driver/driver.go
mogren: I'm even considering switching out the netlink wrapper.. I'll do some tests with https://github.com/mdlayher/netlink

anguslees: .. or a language that has references rather than just pointers :trollface:

anguslees:
(Pushed update to patch and replaced PR description to fit current state.)

pkg/ipamd/datastore/data_store.go
```diff
@@ -432,50 +497,52 @@ func (ds *DataStore) DelIPv4AddressFromStore(eniID string, ipv4 string, force bo
 	// Prometheus gauge
 	totalIPs.Set(float64(ds.total))
 
-	delete(curENI.IPv4Addresses, ipv4)
+	curENI.removeAddress(idx)
```
mogren, line 500: Hmm.. So calling this with either the IPv4 or the IPv6 address will remove both from the datastore.

pkg/ipamd/ipamd.go
```diff
@@ -1147,9 +1252,36 @@ func GetConfigForDebug() map[string]interface{} {
 		envWarmIPTarget:     getWarmIPTarget(),
 		envWarmENITarget:    getWarmENITarget(),
 		envCustomNetworkCfg: UseCustomNetworkCfg(),
+		envAssignIPv4:       getAssignIPv4(),
+		envAssignIPv6:       getAssignIPv6(),
+	}
+}
+
+func getEnvBool(name string, def bool) bool {
```
mogren, line 1260: There is a `getBoolEnvVar(name string, defaultValue bool)` in network.go already, maybe we should make some kind of utility function for this?

PR #959, opened 2020-05-07
Add a match expresison for 1.12 clusters

mogren:
*Description of changes:*

Applying the sample config to a 1.12 cluster requires the `beta.kubernetes.io` node labels. Adding a separate match expression to support older clusters.


jaypipes:
@mogren this needs to be applied to the Helm Chart too: https://github.com/aws/eks-charts/blob/master/stable/aws-vpc-cni/templates/daemonset.yaml#L30-L44

PR #962, opened 2020-05-08
Remove unused IPv4-specific fields from protobuf and K8sArgs

anguslees:
Several `IPv4Addr` fields in protobuf messages were only used for log
messages on the other side.  Remove these fields.

CNI arg `IP` is not populated by Kubernetes' CNI shim (afaics?), and was only
used for one of these informational-only protobuf fields.  Remove.


cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -73,9 +73,6 @@ type NetConf struct {
 type K8sArgs struct {
 	types.CommonArgs
 
-	// IP is pod's ip address
-	IP net.IP
```
mogren, line 77: Yes, noticed that the IP field was never used when looking at your IPv6 PR. :)

mogren:
Integration tests passed, rebasing and merging

PR #965, opened 2020-05-08
Add bash command instead of orb

mogren:
*Description of changes:*

Adding the bash command recommended by https://codecov.io instead of using the CircleCI orb, since it's "unmanaged". Will add the badge in a follow up PR since we want to generate the reports first.


mogren:
See #913 and #920 for details.

PR #972, opened 2020-05-12
Persist IPAM state to local file and use across restarts

anguslees:
Persist IPAM state to a file in /var/run (by default), and use this to
recover state across a restart.  Note no state needs to be preserved
across a reboot, since all containers are also restarted.

Removes need[*] for docker/CRI and Kubernetes API from ipamd.

[*] But not "use" of docker/CRI :(
- CRI is necessary for this release to handle upgrades from earlier
  versions - without requiring a reboot of the node.  We can
  drop CRI for real in release after release that contains this PR.
- The CNI K8S_POD_* arguments are still passed in the gRPC request(s)
  to ipamd.  It is expected that pod name/namespace _will_ be necessary
  at some point (to fetch pod annotations).


pkg/ipamd/rpc_handler.go
```diff
@@ -44,13 +42,10 @@ type server struct {
...
-	log.Infof("Received AddNetwork for NS %s, Pod %s, NameSpace %s, Sandbox %s, ifname %s",
-		in.Netns, in.K8S_POD_NAME, in.K8S_POD_NAMESPACE, in.K8S_POD_INFRA_CONTAINER_ID, in.IfName)
+	log.Infof("Received AddNetwork for NS %s, Sandbox %s, ifname %s",
+		in.Netns, in.K8S_POD_INFRA_CONTAINER_ID, in.IfName)
 
-	addr, deviceNumber, err := s.ipamContext.dataStore.AssignPodIPv4Address(&k8sapi.K8SPodInfo{
-		Name:      in.K8S_POD_NAME,
-		Namespace: in.K8S_POD_NAMESPACE,
-		Sandbox:   in.K8S_POD_INFRA_CONTAINER_ID})
+	addr, deviceNumber, err := s.ipamContext.dataStore.AssignPodIPv4Address(in.K8S_POD_INFRA_CONTAINER_ID)
```
mogren, line 48: I like this a lot...

pkg/ipamd/rpc_handler.go
```diff
@@ -80,21 +75,12 @@ func (s *server) AddNetwork(ctx context.Context, in *rpc.AddNetworkRequest) (*rp
...
-		Namespace: in.K8S_POD_NAMESPACE,
-		Sandbox:   in.K8S_POD_INFRA_CONTAINER_ID})
+	ip, deviceNumber, err := s.ipamContext.dataStore.UnassignPodIPv4Address(in.K8S_POD_INFRA_CONTAINER_ID)
 
-	if err != nil && err == datastore.ErrUnknownPod {
-		// If L-IPAMD restarts, the pod's IP address are assigned by only pod's name and namespace due to kubelet's introspection.
-		ip, deviceNumber, err = s.ipamContext.dataStore.UnassignPodIPv4Address(&k8sapi.K8SPodInfo{
-			Name:      in.K8S_POD_NAME,
-			Namespace: in.K8S_POD_NAMESPACE})
-	}
```
mogren, line 97: I always felt that this was super hacky. It was in my long backlog to clean it up, so thanks for working on this.

Can it ever be that `k8sArgs.K8S_POD_INFRA_CONTAINER_ID` is not set? I was referring to that old comment:

     If L-IPAMD restarts, the pod's IP address are assigned by only pod's name and namespace due to kubelet's introspection.

pkg/ipamd/datastore/data_store.go
```diff
@@ -166,15 +161,119 @@ func prometheusRegister() {
...
+		addr := eni.IPv4Addresses[allocation.IPv4]
+
+		ds.assignPodIPv4AddressUnsafe(allocation.SandboxID, eni, addr)
+	}
+
+	ds.log.Debugf("Completed reading ipam state from %s", ds.backingStore)
+	return nil
+}
+
+func (ds *DataStore) writeBackingStoreUnsafe() error {
```
mogren, line 237: When will this be called, and do we need to lock the datastore until this is persisted? What happens if we generate the `data` struct, but before it's serialized and flushed to disk we free or use one IP? Or if we get an error back for some reason, what should we do If the aws-node pod restarts to rebuild the sandbox state?

pkg/ipamd/rpc_handler.go
anguslees: I think that comment refers to the way the aws-node internals work.  In the current version of this code, it calls this function _without_ container ID to indicate ipam cold-restart discovery and _with_ container ID to indicate regular CNI-driven pod creation.

Fwiw Kubernetes' dockershim [*always*](https://github.com/kubernetes/kubernetes/blob/6acceaeb2dbfc4e4b4528eeee0b034de999213a6/pkg/kubelet/dockershim/network/cni/cni.go#L395-L406) sets K8S_POD_INFRA_CONTAINER_ID to the same as CNI (non-k8s) ContainerID arg.  I haven't audited other CRI implementations (I should do so).  I think we can rely on always using CNI ContainerID arg (that's the point), so I might switch to that.

pkg/ipamd/datastore/data_store.go
anguslees: Not glued in yet.  I'm going to call it while holding the datastore lock, during every new IP allocation.

anguslees:
I think this is ready for review fwiw.

Open questions:
- upgrade migration.  Assuming we don't want to reboot the node, then we need a way to migrate the existing containerID<->IPs allocations into the new persistent store (or some other strategy).
- Verify what other CRIs do with CNI ContainerID arg vs K8S_POD_INFRA_CONTAINER_ID.  I suspect we should switch to the former.

anguslees:
cc #714

anguslees:
Ok, there's now *zero* references to the CNI `K8S_POD_*` args in the codebase, and the IPAM state is stored against (network name, container ID, interface name) as recommended by CNI spec.

Also: I've audited the entire internet (dockershim, containerd, cri-o) and verified that every implementation of k8s/CNI sets "Container ID" to the same as "K8S_POD_INFRA_CONTAINER_ID".  I [asked](https://kubernetes.slack.com/archives/C09QYUH5W/p1589961250114400) on #sig-network and found no surprises there either.

I think this leaves:
- upgrade.  I'll have to think about this.  It might mean restoring the CRI code for at least another minor release.

anguslees:
Sigh.

Ok I've added back passing K8S_* from CNI to ipamd, even though they are then ignored.  It's highly likely that we'll need to fetch (eg) pod annotations in the future, and this will make that future upgrade easier.

Also upgrade is now done by restoring (some) of the previous CRI code.  Now, if the ipamd checkpoint file is missing (ie: not-found -> upgrade or reboot) then we fallback to querying CRI to discover any in-use Pod IPs.  In the _next_ release after this we'll be able to remove that and only rely on the checkpoint file (ie: not-found -> reboot only).  For future reference, the code that should be removed is: everything covered by `BackfillMissingCheckpointFromCRI` constant, and the CRI socket references in the k8s manifest.

I have also updated PR description at the top to match current PR.

pkg/ipamd/datastore/data_store.go
```diff
@@ -93,57 +101,117 @@ var (
...
+// InCoolingPeriod checks whether an addr is in addressCoolingPeriod
+func (addr AddressInfo) inCoolingPeriod() bool {
+	return time.Since(addr.UnassignedTime) <= addressCoolingPeriod
+}
+
+// ENIPool is a collection of ENIIPPool, keyed by ENI ID
+type ENIPool map[string]*ENIIPPool
+
+// AssignedIPv4Addresses is the number of IP addresses already assigned
+func (p *ENIPool) AssignedIPv4Addresses() int {
```
mogren, line 177: Nit: Slightly confusing to have two functions named `AssignedIPv4Addresses()`, one on `ENIIPPool` and one on `ENIPool`. Even the names of those structs are a bit confusing I think...

pkg/ipamd/datastore/data_store.go
```diff
@@ -93,57 +101,117 @@ var (
 	prometheusRegistered = false
 )
 
+// IPAMKey is the IPAM primary key.  Quoting CNI spec:
+//   Plugins that store state should do so using a primary key of
+//   (network name, CNI_CONTAINERID, CNI_IFNAME).
+type IPAMKey struct {
```
mogren, line 107: 👍

config/master/manifests.jsonnet
```diff
@@ -35,18 +35,8 @@ local awsnode = {
...
-        apiGroups: [""],
-        resources: ["pods", "nodes", "namespaces"],
-        verbs: ["list", "watch", "get"],
-      },
-      {
-        apiGroups: ["extensions"],
-        resources: ["daemonsets"],
-        verbs: ["list", "watch"],
+        resources: ["eniconfigs"],
+        verbs: ["get", "list", "watch"],
```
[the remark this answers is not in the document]
anguslees: Thumbs-up this comment if you want me to break it out into a separate PR.  This whole PR basically started from being surprised by the k8s manifest, and I worked backwards :P

pkg/ipamd/datastore/data_store.go
anguslees: Yes.  There are a few function names that are duplicated - with the idea that there are three nested levels `Address` < ENI (`ENIIPPool`) < all ENIs (`ENIPool`) and the methods sum/aggregate the similar method across the lower level.

I completely agree that ENIIPPool and ENIPool are terribly confusing names.  I was trying to avoid changing the original type names, even though I knew I should do so :P  I will rename `ENIIPPool` to `ENI` so it clearly suggests a single ENI (which is what it represents).

pkg/ipamd/datastore/data_store.go
```diff
@@ -166,13 +235,128 @@ func prometheusRegister() {
...
+		data = CheckpointData{
+			Version:     CheckpointFormatVersion,
+			Allocations: entries,
+		}
+
+	} else if os.IsNotExist(err) {
+		// Assume that no file == no containers are
+		// currently in use, eg a fresh reboot just
+		// cleared everything out.  This is ok, and a
+		// no-op.
```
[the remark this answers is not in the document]
anguslees: Sure .. but at some point we need to stop trying to outcompete malicious users.  (What if we continue checking CRI, but they remove the CRI socket or replace kubelet binary with a modified one that does something odd?)

After the upgrade to this CNI, I think the only way to be out of sync is if the kubelet is behaving incorrectly (or deliberate user intervention, as you suggest).  In tje worst case, a reboot will reset everything (kubelet, CRI, and ipam.json) to empty.

If you're still concerned, we should just drop most of this PR (I'm ok with that).  There's no point checkpointing to disk _and_ still querying CRI every time.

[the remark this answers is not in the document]
mogren: True. I'd really like to get rid of having to rely on the CRI. It's definitely better not to treat worker nodes as pets. 😄

pkg/ipamd/datastore/data_store.go
```diff
@@ -263,79 +445,62 @@ func (ds *DataStore) DelIPv4AddressFromStore(eniID string, ipv4 string, force bo
...
 
-func incrementAssignedCount(ds *DataStore, eni *ENIIPPool, addr *AddressInfo) {
+// It returns the assigned IPv4 address, device number
+func (ds *DataStore) assignPodIPv4AddressUnsafe(ipamKey IPAMKey, eni *ENI, addr *AddressInfo) (string, int) {
+	ds.log.Infof("AssignPodIPv4Address: Assign IP %v to sandbox %s",
+		addr.Address, ipamKey)
+
+	if addr.Assigned() {
+		panic("addr already assigned")
+	}
```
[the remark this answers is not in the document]
anguslees: I can:
- return an error at runtime, which (eventually) results in the pod restarting as well.
- ignore it and overwrite an existing pod ip
- ignore it and incorrectly set up the new pod
- remove the check so we don't see a `panic()` in the code, document the expected invariant, and trust that callers never invoke this function for an already assigned ipamKey
- (current version) assert the expected invariant explicitly in code, and thus know that no callers invoke this function for an already assigned ipamKey.

Which would you like?

pkg/ipamd/datastore/data_store.go
```diff
@@ -166,13 +235,128 @@ func prometheusRegister() {
...
+		// cleared everything out.  This is ok, and a
+		// no-op.
+		return nil
+	} else if err != nil {
+		return fmt.Errorf("datastore: error reading backing store: %v", err)
+	}
+
+	if data.Version != CheckpointFormatVersion {
+		return fmt.Errorf("datastore: unknown backing store format (%s != %s) - wrong CNI/ipamd version? (Rebooting this node will restart local pods and probably help)", data.Version, CheckpointFormatVersion)
+	}
```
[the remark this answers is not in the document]
anguslees: Yes, but that means we're at war with our future selves.

The point of this format version check is so we can introduce a future incompatible change, and make sure that we don't silently ignore whatever that incompatibility is.  If we _do_ silently ignore it here, then we'll need to invent _another_ mechanism to make sure we don't.  Repeat.  (Incidentally, if CRI is sufficient/acceptable, then we never needed the checkpoint in the first place).

pkg/ipamd/datastore/data_store.go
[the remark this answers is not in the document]
mogren: The safest would be to panic for sure, since we should never be in this state. It would give a strong signal that something is very wrong in the current state of the aws-node pod.

pkg/ipamd/ipamd.go
```diff
@@ -113,6 +111,10 @@ const (
 	// disableENIProvisioning is used to specify that ENI doesn't need to be synced during initializing a pod.
 	envDisableENIProvisioning = "DISABLE_NETWORK_RESOURCE_PROVISIONING"
 	noDisableENIProvisioning  = false
+
+	// Specify where ipam should persist its current IP<->container allocations.
+	envBackingStorePath     = "AWS_VPC_K8S_CNI_BACKING_STORE"
+	defaultBackingStorePath = "/var/run/aws-routed-eni/ipam.json"
```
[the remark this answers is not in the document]
anguslees: Changed.

pkg/ipamd/introspect.go
```diff
@@ -84,7 +84,6 @@ func (c *IPAMContext) setupIntrospectionServer() *http.Server {
 	serverFunctions := map[string]func(w http.ResponseWriter, r *http.Request){
 		"/v1/enis":                      eniV1RequestHandler(c),
 		"/v1/eni-configs":               eniConfigRequestHandler(c),
-		"/v1/pods":                      podV1RequestHandler(c),
```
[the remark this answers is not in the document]
anguslees: https://github.com/awslabs/amazon-eks-ami/pull/487 - in draft until this PR is finalised.

cmd/aws-k8s-agent/main.go
```diff
@@ -45,15 +45,12 @@ func _main() int {
 		return 1
 	}
 
-	discoverController := k8sapi.NewController(kubeClient)
-	go discoverController.DiscoverLocalK8SPods()
-
```
mogren, line 50: Getting rid of this, and just reading it [from the CRI socket](https://github.com/aws/amazon-vpc-cni-k8s/pull/972/files#diff-716bb85864c637ee87fb1ae9ca36fad5R276) will resolve #711  and replace #738 as well...

uthark:
@mogren will you cherry-pick it to 1.6.x branch?

mogren:
@uthark I think this change is a bit too big for a patch release, so instead planning to get #903 updated and merged, then start preparing a `v1.7.0-rc1` release for testing. The config files are all updated because of #955 and #986, so a lot of changes coming.

PR #974, opened 2020-05-14
1.6 multi eni

hallsaa:
*Description of changes:*

This does a few things:

    It allows for a worker node to interpret multiple ENIConfigs
    It then causes the worker node to attempt to bring up an ENI for each ENIConfig
    It allows multus to attach multiple interfaces to a single pod
    When installing a pod interface, allow for an ENIConfig to be specified to choose the IP to use.
    Add routes for local subnets in the container.


hallsaa:
This is adapting https://github.com/aws/amazon-vpc-cni-k8s/pull/866 to the new multi-attachment branch.

infinitydon:
@mogren @hallsaa - Please is this feature now generally available? I can't find the reference in the docs.

PR #987, opened 2020-05-20
Limit scope of logs writable by ipamd container

anguslees:
Reduce the logs exposed to ipamd container to just
`/var/log/aws-routed-eni/` rather than all of `/var/log` (!)

Also correct documented log file defaults:

- `AWS_VPC_K8S_PLUGIN_LOG_FILE` defaults to
`/var/log/aws-routed-eni/plugin.log` in scripts/entrypoint.sh#L44

- `AWS_VPC_K8S_CNI_LOG_FILE` defaults to
`/host/var/log/aws-routed-eni/ipamd.log` in utils/logger/config.go#L47


anguslees:
(This PR is based on #986 - will be rebased/delayed/etc, depending on what happens with that PR)

Note the default for  AWS_VPC_K8S_CNI_LOG_FILE is a bit confusing, since /host/foo assumes a particular container volume setup.  Perhaps we should take the opportunity to change the container default, rather than document the existing behaviour.  Either way it wouldn't affect anyone using our provided manifests.

mogren:
>  Perhaps we should take the opportunity to change the container default, rather than document the existing behaviour. Either way it wouldn't affect anyone using our provided manifests.

Yes, I totally agree. I would probably prefer to switch to log to `stdout` by default, since the [eks-log-collector.sh](https://github.com/awslabs/amazon-eks-ami/blob/master/log-collector-script/linux/eks-log-collector.sh) script already fetches the pod logs.

anguslees:
> I would probably prefer to switch to log to stdout by default

I could also change the code/image to default to stdout, but then point it back to the existing files in the manifest (so no change for people using the manifest).

Let me know what you want me to do here (if anything).

mogren:
I guess it's safer to just keep the default file for now, and just clean up the mounts. Just needs a rebase to make the changes a bit easier to see. 😄

anguslees:
(Rebased)

config/master/manifests.jsonnet
```diff
@@ -182,7 +183,12 @@ local awsnode = {
           volumes: [
             {name: "cni-bin-dir", hostPath: {path: "/opt/cni/bin"}},
             {name: "cni-net-dir", hostPath: {path: "/etc/cni/net.d"}},
-            {name: "log-dir", hostPath: {path: "/var/log"}},
+            {name: "log-dir",
+              hostPath: {
+                path: "/var/log/aws-routed-eni",
+                type: "DirectoryOrCreate",
```
mogren, line 189: 👍

config/master/manifests.jsonnet
```diff
@@ -172,7 +173,7 @@ local awsnode = {
               volumeMounts: [
                 {mountPath: "/host/opt/cni/bin", name: "cni-bin-dir"},
                 {mountPath: "/host/etc/cni/net.d", name: "cni-net-dir"},
-                {mountPath: "/host/var/log", name: "log-dir"},
+                {mountPath: "/var/log/aws-routed-eni", name: "log-dir"},
```
mogren, line 176: Can we keep the `/host/` part here so that both log files end up in the same host folder? With this config, `ipamd.log` will be inside the container only, in `/host/var/log/aws-routed-eni/`, and the `plugin.log` file from the host is visible as `/var/log/aws-routed-eni/plugin.log`. Makes it more clear that it's actually a host mounted directory.

anguslees: Yep, reverted this part of the PR to continue using an explicit `/host` prefix.

Fwiw, I changed the `AWS_VPC_K8S_CNI_LOG_FILE` and `mountPath` in this PR to the same value so `ipamd.log` _was_ still writing to the same real host `/var/log/aws-routed-eni/`.  I was trying to _reduce_ confusion by using identical paths inside and outside the container, but obviously that didn't work :stuck_out_tongue:

mogren: Ahh... Thanks. Was too confusing for me for sure 😆

Issue #463, opened 2019-05-10
Capability to create constraints on bandwidth for each pod 


charan-amzn:
Hi,

In use-case such as blockchain workloads where there are multiple pods in 100s of nodes, it will be good to have capability to create constraints on the bandwidth for each pod, that is, say this pod's maximum bandwidth is 1mbit/s, that one is 100mbit/s etc. 

As I understand we can do this within CNI [Alpha feature] - https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/ 

In EKS, when we get to 1.14 version, can we have this capability implemented for our CNI ?

ksemaev:
Guys, we need that implementation also, as large traffic spikes for one pod cause outage for the whole cluster, which is really not prod-ready solution

jweite:
+1.  Creating a platform for running end-user submitted code on EKS and need to minimize the impact a badly behaved submission can have on its peers.  Ability to incorporate the [Bandwidth plugin](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/#support-traffic-shaping) would be key.

cs50student2:
We use K8s as PaaS to provide CS students with a cloud-based IDE where each student's IDE is backed by a single-container pod and it would be really useful to be able to limit each IDE's bandwidth per https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/#support-traffic-shaping.

kzidane:
We use K8s as PaaS to provide CS students with a cloud-based IDE where each student's IDE is backed by a single-container pod and it would be really useful to be able to limit each IDE's bandwidth per https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/#support-traffic-shaping.

jansony1:
+1

mogren:
I opened #1029 that adds some foundations for this. Still some usability issues before we can close this issue though, even after that PR gets merged.

willejs:
It would be useful to be able to customize the content of 10-aws.conflist in general to specify other plugins such as the tuning plugin. I think this issue could/should cover that too?

mogren:
Completing this is currently blocked on containerd/cri#1544

jayanthvn:
Bandwidth plugin support is merged to master. Regarding customizing the content of 10-aws.conflist we will look into.

PR #990, opened 2020-05-23
Chain interface to other CNI plugins

theMagicalKarp:
## What
I'd like to take advantage of the [bandwidth CNI plugin](https://github.com/containernetworking/plugins/tree/master/plugins/meta/bandwidth). To do so I had to add the following lines to make sure the bandwidth plugin was aware of the `Interfaces` created by this CNI plugin.

This is partially related to https://github.com/aws/amazon-vpc-cni-k8s/issues/463.


cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -232,6 +232,9 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 
 	result := &current.Result{
 		IPs: ips,
+		Interfaces: []*current.Interface{
+			&current.Interface{Name: hostVethName},
+		},
```
mogren, line 237: Looking at the [CNI SPEC](https://github.com/containernetworking/cni/blob/master/SPEC.md#ips), it states that if an `interfaces` list is returned, the `ip` should have an `interface` index to match. In this case, it would mean adding `Interface: 0,` after line 229, since we only return one interface name.

theMagicalKarp: I've mostly been doing manually testing, but this is the config I'm running with at the moment. (which I copied from the docs https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/#support-traffic-shaping)

```JSON
{
  "cniVersion": "0.3.1",
  "name": "aws-cni",
  "plugins": [
    {
      "name": "aws-cni",
      "type": "aws-cni",
      "vethPrefix": "eni"
    },
    {
      "type": "portmap",
      "capabilities": {"portMappings": true},
      "snat": true
    },
    {
      "type": "bandwidth",
      "capabilities": {
        "bandwidth": true
      }
    }
  ]
}
```

Also maybe unrelated, but I'm using Amazon Linux 2 and I've had to patch this line out, https://github.com/containernetworking/plugins/blob/master/plugins/meta/bandwidth/ifb_creator.go#L34

Because the MTU it's trying to set here is 9001, and the highest it'll accept is 1500. Otherwise seems to work with these annotations on pods. (maybe you understand more?)

```YAML
kubernetes.io/egress-bandwidth: 100M
kubernetes.io/ingress-bandwidth: 100M
```

theMagicalKarp:
I've mostly been doing manually testing, but this is the config I'm running with at the moment. (which I copied from the docs https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/#support-traffic-shaping)

```JSON
{
  "cniVersion": "0.3.1",
  "name": "aws-cni",
  "plugins": [
    {
      "name": "aws-cni",
      "type": "aws-cni",
      "vethPrefix": "eni"
    },
    {
      "type": "portmap",
      "capabilities": {"portMappings": true},
      "snat": true
    },
    {
      "type": "bandwidth",
      "capabilities": {
        "bandwidth": true
      }
    }
  ]
}
```

Also maybe unrelated, but I'm using Amazon Linux 2 and I've had to patch this line out, https://github.com/containernetworking/plugins/blob/master/plugins/meta/bandwidth/ifb_creator.go#L34

Because the MTU it's trying to set here is 9001, and the highest it'll accept is 1500. Otherwise seems to work with these annotations on pods. (maybe you understand more?)

```YAML
kubernetes.io/egress-bandwidth: 100M
kubernetes.io/ingress-bandwidth: 100M
```

mogren:
Thanks for the details! Just so you know, you can set the environment variable `AWS_VPC_ENI_MTU` to `"1500"` instead of the default of `9001`. Then you won't have to comment out that line.

mogren:
Hi @theMagicalKarp, I'd like to get this merged, but by now it needs a rebase first.

mogren:
Rebased this PR in #1029

PR #992, opened 2020-05-25
Update run-integration-tests.sh to default to config/master

anguslees:
Default to testing upgrades to `config/master` manifests in
`run-integration-tests.sh`.  Follow-on to #986.

Correct internal comment re starting version: we test upgrades from
whatever version is installed by `eks create cluster`.

Also: Switch to shell `${foo:=bar}` idiom to reduce duplication (and
chance for typos) in `foo=${foo:-bar}`.

### DIFF
--- a/scripts/run-integration-tests.sh
+++ b/scripts/run-integration-tests.sh
@@ -11,12 +11,12 @@ source $DIR/lib/cluster.sh
 
 OS=$(go env GOOS)
 ARCH=$(go env GOARCH)
-AWS_REGION=${AWS_REGION:-us-west-2}
-K8S_VERSION=${K8S_VERSION:-1.14.6}
-PROVISION=${PROVISION:-true}
-DEPROVISION=${DEPROVISION:-true}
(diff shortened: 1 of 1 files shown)


scripts/run-integration-tests.sh
```diff
@@ -11,12 +11,12 @@ source $DIR/lib/cluster.sh
...
-PROVISION=${PROVISION:-true}
-DEPROVISION=${DEPROVISION:-true}
-BUILD=${BUILD:-true}
-RUN_CONFORMANCE=${RUN_CONFORMANCE:-false}
+: ${AWS_REGION:=us-west-2}
+: ${K8S_VERSION:=1.14.6}
+: ${PROVISION:=true}
+: ${DEPROVISION:=true}
+: ${BUILD:=true}
+: ${RUN_CONFORMANCE:=false}
```
mogren, line 19: Nice trick!

I had never seen this, found an explanations here https://stackoverflow.com/a/28085062/4992

PR #1017, opened 2020-06-08
Cache go packages.

bnapolitan:
Cache go packages via CircleCI caching that are previously downloaded on every run.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,6 +65,7 @@ jobs:
           profile-name: awstester
       - restore_cache:
           keys:
+            - dependency-packages-store-{{ checksum "test/integration/go.mod" }}
             - dependency-packages-store
       - k8s/install-kubectl:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -63,13 +63,21 @@ jobs:
...
+            - dependency-packages-store
       - k8s/install-kubectl:
           # requires 1.14.9 for k8s testing, since it uses log api.
           kubectl-version: v1.14.9
       - run:
           name: Run the integration tests
           command: ./scripts/run-integration-tests.sh
           no_output_timeout: 15m
+      - save_cache:
+          key: dependency-packages-store
```
mogren, line 77: Maybe we should use a checksum on go.mod here? Something like this:
```
      - save_cache:
          key: dependency-packages-store-{{ checksum "go.mod" }}
```
And same in the restore of course. From the docs https://circleci.com/docs/2.0/caching/#managing-caches

PR #1028, opened 2020-06-15
Update probe settings

mogren:
*Description of changes:*
* Reduce readiness probe startup delay to 1s
* Increase liveness probe startup delay to 60, since it will only fail if the CNI fails to start, or has crashed. 
* Reduce shutdown grace period to 10 seconds

### DIFF
--- a/config/master/aws-k8s-cni-cn.yaml
+++ b/config/master/aws-k8s-cni-cn.yaml
@@ -118,7 +118,7 @@
             "command":
             - "/app/grpc-health-probe"
             - "-addr=:50051"
-          "initialDelaySeconds": 35
+          "initialDelaySeconds": 60
         "name": "aws-node"
         "ports":
         - "containerPort": 61678
@@ -128,7 +128,7 @@
             "command":
             - "/app/grpc-health-probe"
             - "-addr=:50051"
-          "initialDelaySeconds": 35
+          "initialDelaySeconds": 1
         "resources":
           "requests":
             "cpu": "10m"
@@ -161,6 +161,7 @@
           "name": "cni-bin-dir"
       "priorityClassName": "system-node-critical"
       "serviceAccountName": "aws-node"
+      "terminationGracePeriodSeconds": 10
       "tolerations":
       - "operator": "Exists"
       "volumes":
--- a/config/master/manifests.jsonnet
+++ b/config/master/manifests.jsonnet
@@ -100,6 +100,7 @@ local awsnode = {
         },
         spec: {
           priorityClassName: "system-node-critical",
+          terminationGracePeriodSeconds: 10,
           affinity: {
             nodeAffinity: {
               requiredDuringSchedulingIgnoredDuringExecution: {
@@ -143,9 +144,11 @@ local awsnode = {
                 exec: {
                   command: ["/app/grpc-health-probe", "-addr=:50051"],
                 },
-                initialDelaySeconds: 35,
+                initialDelaySeconds: 1,
               },
-              livenessProbe: self.readinessProbe,
+              livenessProbe: self.readinessProbe + {
+                initialDelaySeconds: 60,
+              },
               env_:: {
                 AWS_VPC_ENI_MTU: "9001",
                 AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER: "false",
--- a/config/master/aws-k8s-cni-us-gov-east-1.yaml
+++ b/config/master/aws-k8s-cni-us-gov-east-1.yaml
@@ -118,7 +118,7 @@
             "command":
             - "/app/grpc-health-probe"
             - "-addr=:50051"
-          "initialDelaySeconds": 35
+          "initialDelaySeconds": 60
         "name": "aws-node"
         "ports":
         - "containerPort": 61678
@@ -128,7 +128,7 @@
             "command":
             - "/app/grpc-health-probe"
             - "-addr=:50051"
-          "initialDelaySeconds": 35
+          "initialDelaySeconds": 1
         "resources":
           "requests":
             "cpu": "10m"
@@ -161,6 +161,7 @@
           "name": "cni-bin-dir"
       "priorityClassName": "system-node-critical"
       "serviceAccountName": "aws-node"
+      "terminationGracePeriodSeconds": 10
       "tolerations":
       - "operator": "Exists"
       "volumes":
--- a/config/master/aws-k8s-cni-us-gov-west-1.yaml
+++ b/config/master/aws-k8s-cni-us-gov-west-1.yaml
@@ -118,7 +118,7 @@
             "command":
             - "/app/grpc-health-probe"
             - "-addr=:50051"
-          "initialDelaySeconds": 35
+          "initialDelaySeconds": 60
         "name": "aws-node"
         "ports":
         - "containerPort": 61678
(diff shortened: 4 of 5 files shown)


mogren:
Oh, you are right! We might have to revisit the time it takes though, rolling updates take a really long time currently.

anguslees:
> We might have to revisit the time it takes though

How are the new values arrived at?
(So we can point back at this PR and our original assumptions, when we next want to change these values)

config/master/aws-k8s-cni-cn.yaml
```diff
@@ -154,6 +155,7 @@
           "name": "cni-bin-dir"
       "priorityClassName": "system-node-critical"
       "serviceAccountName": "aws-node"
+      "terminationGracePeriodSeconds": 10
```
mogren, line 158: The default `terminationGracePeriodSeconds` is 30 seconds. The CNI does not really need any time to shut down, so reducing this to `10s`.

config/master/aws-k8s-cni-cn.yaml
```diff
@@ -121,7 +122,7 @@
             "command":
             - "/app/grpc-health-probe"
             - "-addr=:50051"
-          "initialDelaySeconds": 35
+          "initialDelaySeconds": 10
```
mogren, line 125: Unless the CNI has a lot of pods and ENIs attached, the pod will get ready in just a few seconds. Having the initial delay be 35 seconds is not needed for the readiness probe.

config/master/aws-k8s-cni-cn.yaml
```diff
@@ -112,6 +112,7 @@
             - "/app/grpc-health-probe"
             - "-addr=:50051"
           "initialDelaySeconds": 35
+          "periodSeconds": 60
```
mogren, line 115: If the CNI pod starts up, we will always return `HealthCheckResponse_SERVING`. If the calls to EC2 API gets throttled, it can take a few minutes for the pod to come up. Increasing the period from the default of 10 seconds will give it some more time to get out of throttling and start up before it gets killed and restarted. 

Question: Is it better to just bump the initial delay? What is a good time between checks? By default, 3 failed checks are required before the pod gets killed and restarted.

config/master/aws-k8s-cni-cn.yaml
anguslees: Why not zero (no initialDelay) ?

config/master/aws-k8s-cni-cn.yaml
anguslees: why not zero (exit immediately)?

config/master/aws-k8s-cni-cn.yaml
anguslees: Right, it sounds like this disagrees with the [above](https://github.com/aws/amazon-vpc-cni-k8s/pull/1028#discussion_r441826154) `initialDelaySeconds` "the pod will get ready in just a few seconds" statement.

If it's only start time, then yes, I would increase `initialDelaySeconds` instead.  Increasing `periodSeconds` means that we're also going to take longer to respond to an issue _after_ startup, which is not great.

Note k8s 1.16 also has a new `startupProbe` which is similar to just increasing initialDelaySeconds, but overkill in this case imo.

config/master/aws-k8s-cni-cn.yaml
mogren: Here I did set it at 10s on purpose to avoid leaking ENIs. There is a [shutdown hook]() and that flag is checked before [trying to free](https://github.com/aws/amazon-vpc-cni-k8s/blob/e38a4d8a8a9358f9b1e24f465a8d42a95a837533/pkg/ipamd/ipamd.go#L480-L483) or [add](https://github.com/aws/amazon-vpc-cni-k8s/blob/e38a4d8a8a9358f9b1e24f465a8d42a95a837533/pkg/ipamd/ipamd.go#L569-L572) an ENI. If the shutdown happens just when ipamd has detached an ENI, but before it has deleted it, we would "leak" it. It usually takes around 2 seconds for the detach to complete, but p99 is 10s according to EC2. Same for creating ENIs, we want the create to complete and the ENI to be attached to the node (with the delete on termination flag set) so that it gets cleaned up if the node terminates.

That said, i don't think we have done enough testing on this.

config/master/aws-k8s-cni-cn.yaml
mogren: True, we might as well set it to 0. 😄

config/master/aws-k8s-cni-cn.yaml
mogren: Yes, better to increase the initial delay. I agree that it sounds like two opposing statements, but I just think it's two different cases. It is true that ipamd usually starts fast, but not if it gets throttled by EC2. In that case we want to keep retrying without the pod getting killed before it is even ready.

config/master/aws-k8s-cni-cn.yaml
mogren: Hmm.. Maybe 1? I've never seen it start faster than that.. 😀

config/master/aws-k8s-cni-cn.yaml
```diff
@@ -121,7 +121,7 @@
             "command":
             - "/app/grpc-health-probe"
             - "-addr=:50051"
-          "initialDelaySeconds": 35
+          "initialDelaySeconds": 1
```
mogren, line 124: Helped a lot! I tested on a cluster with 50 pods on 3 nodes:
```
kube-system   aws-node-2ph8t              0/1     ContainerCreating   0          1s      
kube-system   aws-node-2ph8t              1/1     Running             0          3s 
```

config/master/aws-k8s-cni-cn.yaml
anguslees: (:rocket: This is exactly the sort of analysis I was hoping to capture)

config/master/manifests.jsonnet
```diff
@@ -143,9 +144,14 @@ local awsnode = {
...
                   command: ["/app/grpc-health-probe", "-addr=:50051"],
                 },
-                initialDelaySeconds: 35,
+                initialDelaySeconds: 1,
+              },
+              livenessProbe: {
+                exec: {
+                  command: ["/app/grpc-health-probe", "-addr=:50051"],
+                },
+                initialDelaySeconds: 60,
```
anguslees, line 153: The PR description says "Increase liveness polling period".  I think this PR increases the liveness _initial delay_ period however - which one was intended? (I suspect this should have been `periodSeconds=60`)

(Separately) jsonnet style tip:  You can "merge" values with `+`.  Since liveness/readiness probes are very similar, I would do this as:
```jsonnet
livenessProbe: self.readinessProbe + {
    initialDelaySeconds: 60, // If desired, but see above comment
    periodSeconds: 60,
},
```

mogren: I'll fix the comment. I think, like you said in https://github.com/aws/amazon-vpc-cni-k8s/pull/1028#discussion_r441854922, that we should just increase the initial timeout, but keep the period at the default in order to catch issues that happen after startup faster. That should give the aws-node about 90 seconds to get started, after a 60 s startup delay and three 10 s liveness probes.

mogren:
We will need to bump this time as well:

https://github.com/aws/amazon-vpc-cni-k8s/blob/a5d340bf92a56c5391dccd1efe6fcc546a04cd28/scripts/entrypoint.sh#L49-L61

PR #1035, opened 2020-06-16
Refactor ENI limit struct

mogren:
*Description of changes:*

Refactored the two static ENI limit lists to a map from instance type to `InstanceTypeLimits`. This is a preparation for adding additional limits. Also added some better logging when generating the file and the option to generate a matching `eni-max-pods.txt` file for the EKS AMI repo by using `make generate-max-pods`.


pkg/awsutils/awsutils.go
```diff
@@ -199,6 +199,13 @@ type ENIMetadata struct {
 	IPv4Addresses []*ec2.NetworkInterfacePrivateIpAddress
 }
 
+// ENILimit keeps track of ENI limits per instance type
+type ENILimit struct {
```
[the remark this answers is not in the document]
mogren: I agree, good point. :)

pkg/awsutils/awsutils.go
```diff
@@ -199,6 +199,13 @@ type ENIMetadata struct {
...
+	InstanceType string
```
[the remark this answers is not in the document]
mogren: It's not needed at all, just kept it because it was used when printing the string. Found a better way.

pkg/awsutils/awsutils.go
```diff
@@ -1168,23 +1174,26 @@ func (cache *EC2InstanceMetadataCache) GetENILimit() (int, error) {
...
+				ENILimit:  eniLimit,
+				IPv4Limit: ipv4Limit,
+			}
+			InstanceNetworkingLimits[instanceType] = eniLimits
 		} else {
 			return 0, errors.New(fmt.Sprintf("%s: %s", UnknownInstanceType, cache.instanceType))
 		}
 	}
-	return eniLimit, nil
+	return int(eniLimits.ENILimit), nil
```
anguslees, line 1189: We should pick an int size for this and use it consistently (`int` or `int64`) - to avoid overflow issues.  I imagine `int` is more than sufficient for the number of ENIs (typically O(10)), and IP addresses (which is either going to be O(100) or effectively limitless (eg IPv6 /64 subnet)).

Makefile
```diff
@@ -192,10 +192,15 @@ generate:
...
+# Generate limit file go code
 generate-limits: GOOS=
 generate-limits:
-	go run pkg/awsutils/gen_vpc_ip_limits.go
+	go run scripts/gen_vpc_ip_limits.go
+
+# Generate eni-max-pods.txt file for EKS AMI
+generate-max-pods: GOOS=
+generate-max-pods: 
+	GENERATE_MAX_PODS=true go run scripts/gen_vpc_ip_limits.go
```
anguslees, line 203: Too fancy.  Just (re)generate both output files always.

mogren: Hmm... The issue here is that the `eni-max-pods.txt` file actually [belongs to the EKS AMI package](https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt)... I just added the functionality here to generate it, using the same code that already generated the CNI limits. Since these are flags to the Kubelet on startup, it's not as easy to fetch the limits there.

anguslees: It's just a loop over the results of a single API call - any benefit isn't worth carrying extra unused code.  Either we invoke (and therefore exercise) both code paths all the time, or the code path that this repo doesn't care about should be removed.

I haven't dug deeply enough to understand how this glues into the AMI build process, but afaics it's one of:
- we don't need eni-max-pods.txt from this binary: remove the code branch
- we need to generate eni-max-pods.txt from this binary during AMI build: generate it always, and .gitignore it from this repo (with a comment).
- we need to pull the already-generated eni-max-pods.txt from this repo during AMI build: generate it always and commit it to this repo (with a comment).

pkg/awsutils/awsutils.go
mogren: I guess the main issue here is that the EC2 API returns these values..
```
eniLimit := aws.Int64Value(info.NetworkInfo.MaximumNetworkInterfaces)
ipv4Limit := aws.Int64Value(info.NetworkInfo.Ipv4AddressesPerInterface)
```
I would prefer using `int` everywhere, but I guess the question is where we should convert. 😃

anguslees: Oh right, the in64 came from the API :(
In that case, we should probably convert to int as soon as possible (so it's unambiguous after that point):

```golang
eniLimit := int(aws.Int64Value(info.NetworkInfo.MaximumNetworkInterfaces))
ipv4Limit := int(aws.Int64Value(info.NetworkInfo.Ipv4AddressesPerInterface))
```

Makefile
mogren: @anguslees We already have a dependency the other way, where the CNI will [download the debug script from the EKS AMI repo](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/Makefile#L212-L221). I've changed this PR to use `int` and to generate both files, checking in `misc/eni-max-pods.txt`.

pkg/awsutils/awsutils.go
```diff
@@ -1144,19 +1150,19 @@ func (cache *EC2InstanceMetadataCache) AllocIPAddress(eniID string) error {
...
 }
 
-// GetENIipLimit return IP address limit per ENI based on EC2 instance type
-func (cache *EC2InstanceMetadataCache) GetENIipLimit() (int, error) {
-	ipLimit, ok := InstanceIPsAvailable[cache.instanceType]
+// GetENIIPv4Limit return IP address limit per ENI based on EC2 instance type
+func (cache *EC2InstanceMetadataCache) GetENIIPv4Limit() (int, error) {
+	eniLimits, ok := InstanceNetworkingLimits[cache.instanceType]
 	if !ok {
 		log.Errorf("Failed to get ENI IP limit due to unknown instance type %s", cache.instanceType)
```
anguslees, line 1157: This code doesn't fallback to `DescribeInstanceTypes` ?
Do we want to add that in this PR or a followup? (I'm ok with either - this PR doesn't make it _worse_)

pkg/awsutils/awsutils.go
```diff
@@ -1144,19 +1150,19 @@ func (cache *EC2InstanceMetadataCache) AllocIPAddress(eniID string) error {
...
 		return 0, errors.New(UnknownInstanceType)
 	}
-	return ipLimit - 1, nil
+	return eniLimits.IPv4Limit - 1, nil
```
anguslees, line 1160: Add a comment here, so the -1 is less surprising.

pkg/awsutils/awsutils.go
mogren: Well, this is really a bit hacky, but in `GetENILimit()` we add the missing instance type to the `InstanceNetworkingLimits` map, so as long as that function is called first, it always works. That said, we should probably move that part out to a separate function. I'll make a follow up PR.

PR #1048, opened 2020-06-22
Default to random-fully

### LINKED ISSUE #1040: Make AWS_VPC_K8S_CNI_RANDOMIZESNAT=prng the default
First of all, thanks for all the hard work you do on this project.

I think `AWS_VPC_K8S_CNI_RANDOMIZESNAT=prng` should be the default, to avoid intermitted connection timeouts.

When moving from Kops to EKS, we noticed connection timeouts every "few minutes" to central services like Redis. Because we use PHP, which doesn't come with a connection pool built in, we create lots of connections at the same time. Here the symptom we saw in the logs (so PHP users can find this issue):
```
CRITICAL: Uncaught PHP Exception Uncaught PHP Exception RedisException: "Connection timed out" at srcKernelProdContainer.php line
```

[This blog post](https://tech.xing.com/a-reason-for-unexplained-connection-timeouts-on-kubernetes-docker-abd041cf7e02) from XING described the problem we’re seeing with a solution. [Another blog post](https://engineering.prezi.com/https-engineering-prezi-com-a-kubernetes-crime-story-2e8d75a77630) from Prezi translated the solution to AWS EKS. The fix was to set the config option `AWS_VPC_K8S_CNI_RANDOMIZESNAT=prng` (see [docs](https://docs.aws.amazon.com/eks/latest/userguide/cni-env-vars.html)). After deploying this fix the symptom was gone.

What do you think? Is there any disadvantage to changing the default?

mogren:
*Issue #, if available:*
Resolves #1040

*Description of changes:*
* Default `AWS_VPC_K8S_CNI_RANDOMIZESNAT` to `"prng"`, meaning `--random-fully` for SNAT.

Ping @mikestef9

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,17 +184,17 @@ private subnet and connected to the internet through an AWS NAT Gateway or anoth
 
 Type: String
 
-Default: `hashrandom`
+Default: `prng`
 
 Valid Values: `hashrandom`, `prng`, `none`
 
-Specifies whether the SNAT `iptables` rule should randomize the outgoing ports for connections\. This should be used when
-`AWS_VPC_K8S_CNI_EXTERNALSNAT=false`. When enabled (`hashrandom`) the `--random` flag will be added to the SNAT `iptables`
-rule\. To use pseudo random number generation rather than hash based (i.e. `--random-fully`) use `prng` for the environment
-variable. For old versions of `iptables` that do not support `--random-fully` this option will fall back to `--random`.
-Disable (`none`) this functionality if you rely on sequential port allocation for outgoing connections.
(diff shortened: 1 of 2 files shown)


jqmichael:
That means the iptables version of kube-proxy needs to be at least 1.16.2?

mogren:
@jqmichael Not really, this is independent of kube-proxy. That base image should have at least iptables 1.6.2 though, or they will get some warnings and the fallback "random" will be used. Also, the plan is to back port https://github.com/kubernetes/kubernetes/pull/82966 to Kubernetes v1.16 as well, to allow kube-proxy to use iptables 1.8 or later.

PR #1056, opened 2020-06-24
Minor nits fast follow up to PR 903

nithu0115:
*Issue #, if available:* N/A

*Description of changes:*
This is a fast follow up to #903. This has clean up and minor nits.


pkg/awsutils/awsutils.go
```diff
@@ -248,10 +248,7 @@ func (ss *StringSet) Set(items []string) {
 }
 
 func (ss *StringSet) IsEmpty() bool {
-	if ss.data.Len() == 0 {
-		return true
-	}
-	return false
+	return ss.data != nil && ss.data.Len() == 0
```
anguslees, line 251: I think this method needs to hold the lock too (there are potential races between the two references to `ss.data`, and (possibly) internally in `Len()`)

pkg/awsutils/awsutils.go
```diff
@@ -248,10 +248,7 @@ func (ss *StringSet) Set(items []string) {
...
 }
 
 func (ss *StringSet) Difference (other *StringSet) *StringSet {
```
anguslees, line 254: I think this method needs to hold the `other` lock too.  In this simple case/type, I think that's fine, but be aware that could lead to fun deadlocks if we get more complicated interactions between two StringSets.

anguslees: Fwiw, I think several of the uses of Difference below could instead use a regular (non-threadsafe) `sets.String`, and avoid thinking about locking.

nithu0115: Make sense.  I thought about using `sets.String` but after several discussions, I went with `StringSet` approach which I thought was much cleaner.

pkg/awsutils/awsutils.go
nithu0115: Updated to hold a Rlock here.

pkg/awsutils/awsutils.go
nithu0115: > I think this method needs to hold the `other` lock too. In this simple case/type, I think that's fine, but be aware that could lead to fun deadlocks if we get more complicated interactions between two StringSets.

Added `other.RLock`

PR #1063, opened 2020-06-28
Some refresh cleanups

mogren:
*Issue #, if available:*

Related to #903 and #1056 

*Description of changes:*

* Made sure the unit race tests pass
* Fix logging error for `make unit-test-race`
* Removed two unused constants
* Use `[]string` instead of `[]*string` for getting the strings from the set
* Ran `make format`, so please review diff with "Ignore white space"

Ping  @nithu0115 

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -142,8 +142,8 @@ type APIs interface {
 	// GetVPCIPv4CIDR returns VPC's 1st CIDR
 	GetVPCIPv4CIDR() string
 
-	// GetVPCIPv4CIDRs returns VPC's CIDRs
-	GetVPCIPv4CIDRs() []*string
+	// GetVPCIPv4CIDRs returns VPC's CIDRs from instance metadata
+	GetVPCIPv4CIDRs() []string
 
 	// GetLocalIPv4 returns the primary IP address on the primary ENI interface
 	GetLocalIPv4() string
@@ -227,17 +227,14 @@ func prometheusRegister() {
 //StringSet is a set of strings
 type StringSet struct {
 	sync.RWMutex
-	data 	sets.String
+	data sets.String
 }
 
-func (ss *StringSet) AWSStrings() []*string {
+func (ss *StringSet) SortedList() []string {
 	ss.RLock()
 	defer ss.RUnlock()
-	var dataSlice []*string
-	for key, _ := range ss.data {
-		dataSlice = append(dataSlice, aws.String(key))
-	}
-	return dataSlice
+	// sets.String.List() returns a sorted list
+	return ss.data.List()
 }
 
 func (ss *StringSet) Set(items []string) {
@@ -246,13 +243,7 @@ func (ss *StringSet) Set(items []string) {
 	ss.data = sets.NewString(items...)
 }
 
-func (ss *StringSet) IsEmpty() bool {
-	ss.RLock()
-	defer ss.RUnlock()
-	return ss.data != nil && ss.data.Len() == 0
-}
-
-func (ss *StringSet) Difference (other *StringSet) *StringSet {
+func (ss *StringSet) Difference(other *StringSet) *StringSet {
 	ss.RLock()
 	other.RLock()
 	defer ss.RUnlock()
@@ -280,9 +271,7 @@ func New() (*EC2InstanceMetadataCache, error) {
 	cache.region = region
 	log.Debugf("Discovered region: %s", cache.region)
 
-	sess, err := session.NewSession(
-		&aws.Config{Region: aws.String(cache.region),
-			MaxRetries: aws.Int(15)})
+	sess, err := session.NewSession(&aws.Config{Region: aws.String(cache.region), MaxRetries: aws.Int(15)})
 	if err != nil {
 		log.Errorf("Failed to initialize AWS SDK session %v", err)
 		return nil, errors.Wrap(err, "instance metadata: failed to initialize AWS SDK session")
@@ -385,7 +374,7 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context) 
 		return err
 	}
 
-	// refresh security groups and VPC CIDR blocks in the background
+	// Refresh security groups and VPC CIDR blocks in the background
 	// Ignoring errors since we will retry in 30s
 	go wait.Forever(func() { _ = cache.refreshSGIDs(mac) }, 30*time.Second)
 	go wait.Forever(func() { _ = cache.refreshVPCIPv4CIDRs(mac) }, 30*time.Second)
@@ -396,7 +385,6 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context) 
 		return nil
 	default:
 	}
-
 	return nil
 }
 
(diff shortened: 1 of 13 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -445,21 +437,20 @@ func (cache *EC2InstanceMetadataCache) refreshVPCIPv4CIDRs(mac string) error {
 
 	newVpcIPv4CIDRs := StringSet{}
 	newVpcIPv4CIDRs.Set(vpcIPv4CIDRs)
-	addedVpcIPv4CIDRs   := newVpcIPv4CIDRs.Difference(&cache.securityGroups)
-	deletedVpcIPv4CIDRs := cache.securityGroups.Difference(&newVpcIPv4CIDRs)
+	addedVpcIPv4CIDRs := newVpcIPv4CIDRs.Difference(&cache.vpcIPv4CIDRs)
+	deletedVpcIPv4CIDRs := cache.vpcIPv4CIDRs.Difference(&newVpcIPv4CIDRs)
```
mogren, line 441: I missed this earlier, sorry about that.

pkg/awsutils/awsutils_test.go
```diff
@@ -103,14 +103,14 @@ func TestInitWithEC2metadata(t *testing.T) {
 	assert.Equal(t, localIP, ins.localIPv4)
 	assert.Equal(t, ins.instanceID, instanceID)
 	assert.Equal(t, ins.primaryENImac, primaryMAC)
-	assert.Equal(t, len(ins.securityGroups.data), 2)
+	assert.Equal(t, len(ins.securityGroups.SortedList()), 2)
 	assert.Equal(t, subnetID, ins.subnetID)
 	assert.Equal(t, vpcCIDR, ins.vpcIPv4CIDR)
-	assert.Equal(t, len(ins.vpcIPv4CIDRs.data), 2)
+	assert.Equal(t, len(ins.vpcIPv4CIDRs.SortedList()), 2)
```
mogren, line 109: Using `data` directly here triggered the race detector.

pkg/awsutils/awsutils.go
```diff
@@ -228,17 +228,13 @@ func prometheusRegister() {
...
-func (ss *StringSet) AWSStrings() []*string {
+func (ss *StringSet) SortedList() []string {
 	ss.RLock()
 	defer ss.RUnlock()
-	var dataSlice []*string
-	for key, _ := range ss.data {
-		dataSlice = append(dataSlice, aws.String(key))
-	}
-	return dataSlice
+	return ss.data.List()
```
anguslees, line 237: Can we leave a comment here saying that sets.String.List() returns a sorted list?
(I had to go look up the docs for that sets.String to see if this line was reasonable or not)

pkg/awsutils/awsutils.go
```diff
@@ -410,25 +403,24 @@ func (cache *EC2InstanceMetadataCache) refreshSGIDs(mac string) error {
...
-	sgIDs  := strings.Fields(metadataSGIDs)
+	sgIDs := strings.Fields(metadataSGIDs)
 
 	newSGs := StringSet{}
 	newSGs.Set(sgIDs)
-	addedSGs   := newSGs.Difference(&cache.securityGroups)
+	addedSGs := newSGs.Difference(&cache.securityGroups)
 	deletedSGs := cache.securityGroups.Difference(&newSGs)
 
 	if !addedSGs.IsEmpty() {
```
anguslees, line 413: Nit: I think we should remove this `IsEmpty()` check, and the similar one a few lines below.  I think the inner `for` loop will already do the right thing (nothing) if the list is empty, and fewer branches == simpler code.

pkg/awsutils/awsutils.go
```diff
@@ -445,21 +437,20 @@ func (cache *EC2InstanceMetadataCache) refreshVPCIPv4CIDRs(mac string) error {
 
 	newVpcIPv4CIDRs := StringSet{}
 	newVpcIPv4CIDRs.Set(vpcIPv4CIDRs)
-	addedVpcIPv4CIDRs   := newVpcIPv4CIDRs.Difference(&cache.securityGroups)
-	deletedVpcIPv4CIDRs := cache.securityGroups.Difference(&newVpcIPv4CIDRs)
+	addedVpcIPv4CIDRs := newVpcIPv4CIDRs.Difference(&cache.vpcIPv4CIDRs)
+	deletedVpcIPv4CIDRs := cache.vpcIPv4CIDRs.Difference(&newVpcIPv4CIDRs)
 
 	if !addedVpcIPv4CIDRs.IsEmpty() {
```
anguslees, line 443: Nit: Same `IsEmpty()` comment.  I think we should remove these checks.

pkg/networkutils/network_test.go
```diff
@@ -360,8 +358,8 @@ func TestSetupHostNetworkWithExcludeSNATCIDRs(t *testing.T) {
 
 	mockProcSys.EXPECT().Set("net/ipv4/conf/lo/rp_filter", "2").Return(nil)
 
-	var vpcCIDRs []*string
-	vpcCIDRs = []*string{aws.String("10.10.0.0/16"), aws.String("10.11.0.0/16")}
+	var vpcCIDRs []string
+	vpcCIDRs = []string{"10.10.0.0/16", "10.11.0.0/16"}
```
anguslees, line 362: Nit: `:=` (unless there's some stylistic reason to keep the declaration separate from initialisation here?)

pkg/ipamd/ipamd.go
```diff
@@ -344,7 +342,7 @@ func (c *IPAMContext) nodeInit() error {
 	vpcCIDRs := c.awsClient.GetVPCIPv4CIDRs()
 
 	for _, cidr := range vpcCIDRs {
```
anguslees, line 344: I think this `for` loop can be removed, remove `pbVPCcidrs`, and just use `vpcCIDRs` throughout the rest of this function.

pkg/ipamd/ipamd.go
```diff
@@ -440,7 +438,7 @@ func (c *IPAMContext) updateCIDRsRulesOnChange(oldVPCCidrs []string) []string {
 	var pbVPCCIDRs []string
 	newVPCCIDRs := c.awsClient.GetVPCIPv4CIDRs()
 	for _, cidr := range newVPCCIDRs {
```
anguslees, line 440: I think this `for` loop can be removed, remove `pbVPCCIDRs`, and just use `newVPCCIDRs` throughout the rest of this function.

pkg/ipamd/rpc_handler.go
```diff
@@ -55,8 +55,8 @@ func (s *server) AddNetwork(ctx context.Context, in *rpc.AddNetworkRequest) (*rp
 
 	var pbVPCcidrs []string
 	for _, cidr := range s.ipamContext.awsClient.GetVPCIPv4CIDRs() {
```
anguslees, line 57: I think this `for` loop can be simplified: to just a debug print loop:
```go
pbVPCcidrs := s.ipamContext.awsClient.GetVPCIPv4CIDRs()
for _, cidr := range pbVPCcidrs {
	log.Debugf("VPC CIDR %s", cidr)
}

```

pkg/ipamd/rpc_handler_test.go
```diff
@@ -82,7 +81,7 @@ func TestServer_AddNetwork(t *testing.T) {
 
 		var expectedCIDRs []string
 		for _, cidr := range tc.vpcCIDRs {
```
anguslees, line 83: This `for` loop can be simplified to just `expectedCIDRs := tc.vpcCIDRs`

.. but wait. Isn't `expectedCIDRs` overwritten on the next line?  Maybe these few lines do nothing (other than an expensive way to pre-allocate the array capacity) and should just be removed entirely?

pkg/awsutils/awsutils_test.go
anguslees: This was indeed a race (if the other goroutine was still running at this point). Yay for race detector :)

pkg/networkutils/network_test.go
mogren: Yeah, `staticcheck` warned me about these. Should be fixed for sure :)

pkg/ipamd/rpc_handler_test.go
mogren: Heh, yeah, they do nothing. `staticcheck` found this as well, but I didn't want to change it in #1064 because I had changed the return values in this PR to be `[]string` instead of `[]*string`.

pkg/awsutils/awsutils.go
mogren: Since `IsEmpty()` is not used anywhere any more, I'll just remove it.

PR #1068, opened 2020-06-30
Introduce automated performance testing.

bnapolitan:
Scale pods up and down on different instance types and track how long it takes to propagate.


bnapolitan:
Still in progress because a communal bucket for test results is still undetermined.

scripts/run-integration-tests.sh
```diff
@@ -211,6 +216,22 @@ if [[ $TEST_PASS -eq 0 && "$RUN_CONFORMANCE" == true ]]; then
...
+    $KUBECTL_PATH delete -f deploy-130-pods.yaml
+
+    $KUBECTL_PATH apply -f deploy-730-pods.yaml
+    run_performance_test_730_pods
+    $KUBECTL_PATH delete -f deploy-730-pods.yaml
+
+    $KUBECTL_PATH apply -f deploy-5000-pods.yaml
+    run_performance_test_5000_pods
+    $KUBECTL_PATH delete -f deploy-5000-pods.yaml
+    PERFORMANCE_DURATION=$((SECONDS - START))
```
mogren, line 232: I think we want to keep these timings separate, so 3 distinct values. One is pod density (730 pods on one node), and the 5000 on 100 nodes is more cluster size and EC2 calls.

test/integration/README.md
```diff
@@ -0,0 +1,34 @@
+## Conformance test duration log
+
+Design document link - https://quip-amazon.com/BoJLAC3IpIfW
```
mogren, line 3: This link doesn't work outside Amazon, we should only refer to documents in this repo or the public docs at AWS.

test/integration/README.md
```diff
@@ -0,0 +1,34 @@
...
+* May 29: 2h 59min 30s
+    * Cache dependencies when testing default CNI
+    * Timelines:
+        * Docker build: 3.583 min
+        * Up test cluster: 31.4 min
+        * Default CNI: 50s
+        * Updating CNI image: 92s
+        * Current image integration: 17s
+        * Conformance tests: 113.8 min (1.9 hrs)
+        * Down cluster: 30.417 min
```
mogren, line 23: We can round these to whole minutes.

test/integration/README.md
```diff
@@ -0,0 +1,34 @@
...
+* June 5: 1h 24min 9s
+    * Parallel execution of conformance tests
+    * Timelines:
+        * Docker build: 3.617 min
+        * Up test cluster: 31.3 min
+        * Default CNI: 52s
+        * Updating CNI image: 92s
+        * Current image integration: 18s
+        * Conformance tests: 16.317 min
+        * Down cluster: 29.95 min
```
mogren, line 33: I think down cluster should be faster now. 😄

scripts/run-integration-tests.sh
bnapolitan: I did record per-test durations in the common.sh functions. Those numbers are printed when the test finishes and then uploaded to s3. But I could also print them at the end of the script if you think that'd be helpful!

mogren: Hmm, maybe. This is just the cluster delete time anyway. 😄

scripts/lib/performance_tests.sh
```diff
@@ -0,0 +1,209 @@
...
+    now="pod-130-Test#${TEST_ID}-$(date +"%m-%d-%Y-%T").csv"
+    echo $now
+
+    echo $(date +"%m-%d-%Y-%T") >> $now
+    echo $((SCALE_UP_DURATION_ARRAY[0])), $((SCALE_DOWN_DURATION_ARRAY[0])) >> $now
+    echo $((SCALE_UP_DURATION_ARRAY[1])), $((SCALE_DOWN_DURATION_ARRAY[1])) >> $now
+    echo $((SCALE_UP_DURATION_ARRAY[2])), $((SCALE_DOWN_DURATION_ARRAY[2])) >> $now
+
+    cat $now
+    aws s3 cp $now s3://cni-performance-test-data
```
mogren, line 69: I think this bucket needs to be a configurable setting. And if it's not set, we should skip the upload. Something like
```bash
if [[ -n "${S3_PERF_TEST_BUCKET:-}" ]]; then
    aws s3 cp $filename "$S3_PERF_TEST_BUCKET"
else 
    echo "No S3 bucket name given, not uploading results"
fi
```

scripts/lib/performance_tests.sh
```diff
@@ -0,0 +1,209 @@
...
+    while [ $INDEX -lt ${#SCALE_DOWN_DURATION_ARRAY[@]} ]
+    do
+        echo "${SCALE_DOWN_DURATION_ARRAY[$INDEX]} seconds"
+        INDEX=$((INDEX + 1))
+    done
+    echo ""
+    DEPLOY_DURATION=$((SECONDS - DEPLOY_START))
+
+    now="pod-130-Test#${TEST_ID}-$(date +"%m-%d-%Y-%T").csv"
+    echo $now
```
mogren, line 61: Can we call this `filename` instead?

scripts/lib/performance_tests.sh
```diff
@@ -0,0 +1,209 @@
...
+    
+    DEPLOY_START=$SECONDS
+
+    SCALE_UP_DURATION_ARRAY=()
+    SCALE_DOWN_DURATION_ARRAY=()
+    while [ ${#SCALE_UP_DURATION_ARRAY[@]} -lt 3 ]
+    do
+        ITERATION_START=$SECONDS
+        $KUBECTL_PATH scale -f ./testdata/deploy-5000-pods.yaml --replicas=5000
+        sleep 100
```
mogren, line 155: What is 100 seconds based on here?

scripts/run-integration-tests.sh
```diff
@@ -213,6 +218,22 @@ if [[ $TEST_PASS -eq 0 && "$RUN_CONFORMANCE" == true ]]; then
   echo "TIMELINE: Conformance tests took $CONFORMANCE_DURATION seconds."
 fi
 
+if [[ "$RUN_PERFORMANCE_TESTS" == true ]]; then
+    START=$SECONDS
+    $KUBECTL_PATH apply -f ./testdata/deploy-130-pods.yaml
+    run_performance_test_130_pods
+    $KUBECTL_PATH delete -f ./testdata/deploy-130-pods.yaml
```
mogren, line 225: Why are the testdata apply functions not done inside the test functions?

scripts/lib/performance_tests.sh
jayanthvn: Should we do the check for S3 bucket before the test?

PR #1069, opened 2020-07-02
create a temporary config file to be manipulated by sed in dockerfile

willejs:
Currently, you can't easily provide your own config file for the cni. Mounting a file from a configmap is readonly, and the magical find replace that sed does is inline against the existing file fails. Instead, I am proposing that we copy the file to a tempfile and modify that before we copy it to its destination.

I thought about perhaps having an environment variable for a custom template file, flag, or similar, but I think that actually that could cause more complexity in the entrypoint script, and I didn't want to change the behavior that much.

Sorry for the 'ergh' next to the sed comment, I wanted to show that I wasn't a fan of how this works, the sedding of config files 🤣. I appreciate why this is done here though.

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -95,7 +95,7 @@ install aws-cni "$HOST_CNI_BIN_PATH"
 
 # create a temporary file to be manipulated by sed (ergh).
 # this enables you to mount your own config file from a configmap that is read only
-temp_config_file=$(mktmp)
+temp_config_file=$(mktemp)
 cp 10-aws.conflist "$temp_config_file"
 sed -i s~__VETHPREFIX__~"${AWS_VPC_K8S_CNI_VETHPREFIX}"~g "$temp_config_file"
 sed -i s~__MTU__~"${AWS_VPC_ENI_MTU}"~g "$temp_config_file"
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -93,17 +93,13 @@ log_in_json info "Copying CNI plugin binary and config file ... "
 
 install aws-cni "$HOST_CNI_BIN_PATH"
 
-# create a temporary file to be manipulated by sed (ergh).
-# this enables you to mount your own config file from a configmap that is read only
-temp_config_file=$(mktemp)
-cp 10-aws.conflist "$temp_config_file"
-sed -i s~__VETHPREFIX__~"${AWS_VPC_K8S_CNI_VETHPREFIX}"~g "$temp_config_file"
-sed -i s~__MTU__~"${AWS_VPC_ENI_MTU}"~g "$temp_config_file"
-sed -i s~__PLUGINLOGFILE__~"${AWS_VPC_K8S_PLUGIN_LOG_FILE}"~g "$temp_config_file"
-sed -i s~__PLUGINLOGLEVEL__~"${AWS_VPC_K8S_PLUGIN_LOG_LEVEL}"~g "$temp_config_file"
-cp "$temp_config_file" "$HOST_CNI_CONFDIR_PATH"
-# clean up temp file
-rm "$temp_config_file"
+# modify the static config to populate it with the env vars
+sed \
+  -e s~__VETHPREFIX__~"${AWS_VPC_K8S_CNI_VETHPREFIX}"~g \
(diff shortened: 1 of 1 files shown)


willejs:
@jaypipes could you review this please?

willejs:
@mogren can you check this out too please?

scripts/entrypoint.sh
```diff
@@ -93,11 +93,17 @@ log_in_json info "Copying CNI plugin binary and config file ... "
...
 
-sed -i s~__VETHPREFIX__~"${AWS_VPC_K8S_CNI_VETHPREFIX}"~g 10-aws.conflist
-sed -i s~__MTU__~"${AWS_VPC_ENI_MTU}"~g 10-aws.conflist
-sed -i s~__PLUGINLOGFILE__~"${AWS_VPC_K8S_PLUGIN_LOG_FILE}"~g 10-aws.conflist
-sed -i s~__PLUGINLOGLEVEL__~"${AWS_VPC_K8S_PLUGIN_LOG_LEVEL}"~g 10-aws.conflist
-cp 10-aws.conflist "$HOST_CNI_CONFDIR_PATH"
+# create a temporary file to be manipulated by sed (ergh).
+# this enables you to mount your own config file from a configmap that is read only
+temp_config_file=$(mktemp)
+cp 10-aws.conflist "$temp_config_file"
```
anguslees, line 99: I agree the original code was already leading that way, but `cp` then `sed`-in-place is a bit weird given that sed is already a fine "stream editor" :P

May I suggest:
```
temp_config_file="$HOST_CNI_CONFDIR_PATH/10-aws.conflist.tmp"
 
sed \
  -e s~__VETHPREFIX__~"${AWS_VPC_K8S_CNI_VETHPREFIX}"~g \
  -e s~__MTU__~"${AWS_VPC_ENI_MTU}"~g \
  -e s~__PLUGINLOGFILE__~"${AWS_VPC_K8S_PLUGIN_LOG_FILE}"~g \
  -e s~__PLUGINLOGLEVEL__~"${AWS_VPC_K8S_PLUGIN_LOG_LEVEL}"~g \
  10-aws.conflist > "$temp_config_file"

mv "$temp_config_file" "$HOST_CNI_CONFDIR_PATH/10-aws.conflist"
```

In addition, the above code writes the tempfile to the same filesystem as the final file, and then atomically renames it into place.

scripts/entrypoint.sh
```diff
@@ -93,11 +93,17 @@ log_in_json info "Copying CNI plugin binary and config file ... "
...
+sed -i s~__VETHPREFIX__~"${AWS_VPC_K8S_CNI_VETHPREFIX}"~g "$temp_config_file"
+sed -i s~__MTU__~"${AWS_VPC_ENI_MTU}"~g "$temp_config_file"
+sed -i s~__PLUGINLOGFILE__~"${AWS_VPC_K8S_PLUGIN_LOG_FILE}"~g "$temp_config_file"
+sed -i s~__PLUGINLOGLEVEL__~"${AWS_VPC_K8S_PLUGIN_LOG_LEVEL}"~g "$temp_config_file"
+cp "$temp_config_file" "$HOST_CNI_CONFDIR_PATH"
+# clean up temp file
+rm "$temp_config_file"
```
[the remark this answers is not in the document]
willejs: Thanks for the review. This makes sense, I didn't realise that I missed off the filename extension.

scripts/entrypoint.sh
willejs: Thanks for the review! I thought about something like this, but didn't want to change too much, and hadn't read the cni code to ensure it only reads *.conflist. I just did and it does, so this approach is much nicer!

willejs: I have removed the tempfile madness and used it as a "stream editor" and redirected the contents into the destination file. I think this makes more sense!

anguslees: It's much shorter/cleaner, I agree :)

My concern is that now it might fail midway (or race with kubelet logic) and the kubelet will end up reading half the file.  .. Now that I think about it, I think that's ok because it's json and it won't be valid until the last closing brace is seen anyway - the worst that will happen is that the kubelet will try to start some pods that will fail-retry a few times until the CNI pod is restarted and fixes the config.

PR #1071, opened 2020-07-03
Fix device number and update table name the device index

mogren:
*Description of changes:*
When we attach new ENIs, we store the `deviceNumber`. We used to add +1 to all except the primary (device number 0). This was done because of how we named the route tables. This PR does the following:
* Store the actual device number value in `deviceNumber` in the `ENI` struct (visible through introspect endpoint)
* Have a `tableNumber` that matches the existing number used by previous versions
* Set a hard upper limit of 100 for the device number to avoid clashes with vlan table names.

### DIFF
--- a/cmd/routed-eni-cni-plugin/driver/driver.go
+++ b/cmd/routed-eni-cni-plugin/driver/driver.go
@@ -46,8 +46,8 @@ const (
 
 // NetworkAPIs defines network API calls
 type NetworkAPIs interface {
-	SetupNS(hostVethName string, contVethName string, netnsPath string, addr *net.IPNet, table int, vpcCIDRs []string, useExternalSNAT bool, mtu int, log logger.Logger) error
-	TeardownNS(addr *net.IPNet, table int, log logger.Logger) error
+	SetupNS(hostVethName string, contVethName string, netnsPath string, addr *net.IPNet, deviceNumber int, vpcCIDRs []string, useExternalSNAT bool, mtu int, log logger.Logger) error
+	TeardownNS(addr *net.IPNet, deviceNumber int, log logger.Logger) error
 	SetupPodENINetwork(hostVethName string, contVethName string, netnsPath string, addr *net.IPNet, vlanID int, eniMAC string,
 		subnetGW string, parentIfIndex int, mtu int, log logger.Logger) error
 	TeardownPodENINetwork(vlanID int, log logger.Logger) error
@@ -173,12 +173,12 @@ func (createVethContext *createVethPairContext) run(hostNS ns.NetNS) error {
 }
 
 // SetupNS wires up linux networking for a pod's network
-func (os *linuxNetwork) SetupNS(hostVethName string, contVethName string, netnsPath string, addr *net.IPNet, table int, vpcCIDRs []string, useExternalSNAT bool, mtu int, log logger.Logger) error {
-	log.Debugf("SetupNS: hostVethName=%s, contVethName=%s, netnsPath=%s, table=%d, mtu=%d", hostVethName, contVethName, netnsPath, table, mtu)
-	return setupNS(hostVethName, contVethName, netnsPath, addr, table, vpcCIDRs, useExternalSNAT, os.netLink, os.ns, mtu, log, os.procSys)
+func (os *linuxNetwork) SetupNS(hostVethName string, contVethName string, netnsPath string, addr *net.IPNet, deviceNumber int, vpcCIDRs []string, useExternalSNAT bool, mtu int, log logger.Logger) error {
+	log.Debugf("SetupNS: hostVethName=%s, contVethName=%s, netnsPath=%s, deviceNumber=%d, mtu=%d", hostVethName, contVethName, netnsPath, deviceNumber, mtu)
+	return setupNS(hostVethName, contVethName, netnsPath, addr, deviceNumber, vpcCIDRs, useExternalSNAT, os.netLink, os.ns, mtu, log, os.procSys)
 }
 
-func setupNS(hostVethName string, contVethName string, netnsPath string, addr *net.IPNet, table int, vpcCIDRs []string, useExternalSNAT bool,
+func setupNS(hostVethName string, contVethName string, netnsPath string, addr *net.IPNet, deviceNumber int, vpcCIDRs []string, useExternalSNAT bool,
 	netLink netlinkwrapper.NetLink, ns nswrapper.NS, mtu int, log logger.Logger, procSys procsyswrapper.ProcSys) error {
 
 	hostVeth, err := setupVeth(hostVethName, contVethName, netnsPath, addr, netLink, ns, mtu, procSys, log)
@@ -213,22 +213,24 @@ func setupNS(hostVethName string, contVethName string, netnsPath string, addr *n
 	log.Infof("Added toContainer rule for %s", addr.String())
 
 	// add from-pod rule, only need it when it is not primary ENI
-	if table > 0 {
+	if deviceNumber > 0 {
+		// To be backwards compatible, we will have to keep this off-by one setting
+		tableNumber := deviceNumber + 1
 		if useExternalSNAT {
 			// add rule: 1536: from <podIP> use table <table>
-			err = addContainerRule(netLink, false, addr, table)
+			err = addContainerRule(netLink, false, addr, tableNumber)
 			if err != nil {
 				log.Errorf("Failed to add fromContainer rule for %s err: %v", addr.String(), err)
 				return errors.Wrap(err, "add NS network: failed to add fromContainer rule")
 			}
-			log.Infof("Added rule priority %d from %s table %d", fromContainerRulePriority, addr.String(), table)
+			log.Infof("Added rule priority %d from %s table %d", fromContainerRulePriority, addr.String(), tableNumber)
 		} else {
 			// add rule: 1536: list of from <podIP> to <vpcCIDR> use table <table>
 			for _, cidr := range vpcCIDRs {
 				podRule := netLink.NewRule()
 				_, podRule.Dst, _ = net.ParseCIDR(cidr)
 				podRule.Src = addr
-				podRule.Table = table
+				podRule.Table = tableNumber
 				podRule.Priority = fromContainerRulePriority
 
 				err = netLink.RuleAdd(podRule)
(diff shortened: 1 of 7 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -588,21 +588,21 @@ func (cache *EC2InstanceMetadataCache) getIPsAndCIDR(eniMAC string) ([]*ec2.Netw
 }
 
 // getENIDeviceNumber returns ENI ID, device number, error
-func (cache *EC2InstanceMetadataCache) getENIDeviceNumber(eniMAC string) (string, int, error) {
+func (cache *EC2InstanceMetadataCache) getENIDeviceNumber(eniMAC string) (eniID string, deviceNumber int, err error) {
```
jaypipes, line 591: you name these return parameters here but don't actually use the named variables.

pkg/networkutils/network.go
```diff
@@ -662,20 +662,20 @@ func LinkByMac(mac string, netLink netlinkwrapper.NetLink, retryInterval time.Du
 }
 
 // SetupENINetwork adds default route to route table (eni-<eni_table>)
-func (n *linuxNetwork) SetupENINetwork(eniIP string, eniMAC string, eniTable int, eniSubnetCIDR string) error {
-	return setupENINetwork(eniIP, eniMAC, eniTable, eniSubnetCIDR, n.netLink, retryLinkByMacInterval, retryRouteAddInterval, n.mtu)
+func (n *linuxNetwork) SetupENINetwork(eniIP string, eniMAC string, deviceNumber int, eniSubnetCIDR string) error {
+	return setupENINetwork(eniIP, eniMAC, deviceNumber, eniSubnetCIDR, n.netLink, retryLinkByMacInterval, retryRouteAddInterval, n.mtu)
```
jaypipes, line 666: :+1:  on the above method parameter renames.

SaranBalaji90:
This change is failing in our internal pipeline. Ip rule table number and actual route table table numbers were different.

SaranBalaji90:
Does eth1 gets route table number: 1 ?because I usually see different route table for eth1

pkg/awsutils/awsutils.go
```diff
@@ -639,38 +639,40 @@ func (cache *EC2InstanceMetadataCache) getIPsAndCIDR(eniMAC string) ([]*ec2.Netw
...
-		return eni, 0, nil
+	if cache.primaryENI == eniID {
+		log.Debugf("Using device number 0 for primary ENI: %s", eniID)
+		if deviceNumber != 0 {
+			log.Warnf("Device number of primary ENI is %d! It was expected to be 0", deviceNumber)
+		}
+		return eniID, deviceNumber, nil
 	}
-	// 0 is reserved for primary ENI, the rest of them has to +1 to avoid collision at 0
-	return eni, int(deviceNum + 1), nil
```
mogren, line 673: @SaranBalaji90 Before this PR, `eth1` gets `table 2` because of this code, since it has device number 1. `eth2` gets `table 3` and so on.

SaranBalaji90: Are you sure? Because on a c5.18xlarge eth1 got 15 but rest of them got similar number as device index - for eg eth2 got route table number 2 and so on.

mogren: Was the device index of `eth1` 14 at that time? The device index does not always match the interface name, especially when deleting and attaching a lot of ENIs. I have also seen `eth4` get `table 3` since it had the device number 2.

SaranBalaji90: I don't have the node now but I think looking at the code it should be 14. I wonder how is this number populated? Because of pod ENI feature we don't want this number to go beyond 100.

SaranBalaji90: Looks like for eth1 for some reason device number is always at the end. Rest of them gets (interface device index - 1) as device number.

pkg/awsutils/awsutils.go
```diff
@@ -639,38 +639,40 @@ func (cache *EC2InstanceMetadataCache) getIPsAndCIDR(eniMAC string) ([]*ec2.Netw
...
+	// 0 is reserved for primary ENI
+	return eniID, deviceNumber, nil
```
SaranBalaji90, line 675: This has to be deviceNumber +1 to maintain backward compatibility otherwise how would this work on existing clusters? existing clusters might already have route for deviceNumber.

mogren: You are right that we can't change the table name now and stay backwards compatible on nodes that are already running. We are stuck using the name we currently have. 

I have a different suggestion on how to solve it though. Let's use `tableNumber := deviceNumber + 1` where we set up the routes, and stop having a variable called `deviceNumber` that doesn't actually store the device number. I'll update the PR.

pkg/networkutils/network.go
```diff
@@ -698,21 +698,19 @@ func linkByMac(mac string, netLink netlinkwrapper.NetLink, retryInterval time.Du
...
 	retryLinkByMacInterval time.Duration, retryRouteAddInterval time.Duration, mtu int) error {
-
-	if eniTable == 0 {
-		log.Debugf("Skipping set up ENI network for primary interface")
-		return nil
+	if deviceNumber == 0 {
+		return errors.New("setupENINetwork should never be called on the primary ENI")
 	}
-
+	tableNumber := deviceNumber + 1
```
SaranBalaji90, line 711: Instead of adding +1 everywhere, why not change method name from saying getDeviceNumber to getRoutingTableNumber? Or is this method being used at other places too?

mogren: Yes, it's used in other places and defined in the ENI struct:

https://github.com/aws/amazon-vpc-cni-k8s/blob/c19ef536c2d8a8f2723003eed140c316ff4be221/pkg/ipamd/datastore/data_store.go#L136-L150

Also, we really just add the +1 in two places, the first one where we set up the ENI, and then when we create the route. The delete is [just for logging](), since we delete by IP.

pkg/awsutils/awsutils.go
anguslees: :100: If we're stuck with the +1 thing, we should absolutely stop calling it deviceNumber

pkg/awsutils/awsutils.go
anguslees: > Because of pod ENI feature we don't want this number to go beyond 100

I know we're assuming that currently, but do we enforce that in the code anywhere?  It sounds like this warrants a nice loud/obvious error message - if we ever accidentally trigger it.

Fwiw, I think [the relevant code](https://github.com/aws/amazon-vpc-cni-k8s/blob/561190a6f982dbdefea6d233939960366a10397c/pkg/awsutils/awsutils.go#L676-L716) currently always looks for the lowest unoccupied DeviceNumber, so it will continue to reuse the same low numbers (bounded indirectly by max-eni).  In addition, I note the relevant bit of code imposes an [artifical max of 128](https://github.com/aws/amazon-vpc-cni-k8s/blob/561190a6f982dbdefea6d233939960366a10397c/pkg/awsutils/awsutils.go#L709), for no particular reason (the code uses a fixed size array, whereas it could have dynamically resized the array, or used a sparse hash).  We should probably lower `maxENIs` to 100, with a comment explaining why 100 is actually a real limit we have.

SaranBalaji90: Good point Gus. For some reason I missed that we were setting the device index. I agree we should change it to 100 and raise error if it goes beyond 100.

mogren: Good call, set 100 as a hard limit.

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -447,14 +449,15 @@ func tearDownNS(addr *net.IPNet, table int, netLink netlinkwrapper.NetLink, log
...
-	if table > 0 {
+	if deviceNumber > 0 {
 		// remove from-pod rule only for non main table
 		err := deleteRuleListBySrc(*addr)
 		if err != nil {
 			log.Errorf("Failed to delete fromContainer for %s %v", addr.String(), err)
 			return errors.Wrapf(err, "delete NS network: failed to delete fromContainer rule for %s", addr.String())
 		}
-		log.Infof("Delete fromContainer rule for %s in table %d", addr.String(), table)
+		tableNumber := deviceNumber + 1
```
SaranBalaji90, line 459: nit: do we need this variable or just do devNo +1 in logs?

pkg/awsutils/awsutils.go
```diff
@@ -639,38 +640,41 @@ func (cache *EC2InstanceMetadataCache) getIPsAndCIDR(eniMAC string) ([]*ec2.Netw
...
-	if cache.primaryENI == eni {
-		log.Debugf("Using device number 0 for primary eni: %s", eni)
-		return eni, 0, nil
+	if cache.primaryENI == eniID {
+		log.Debugf("Using device number 0 for primary ENI: %s", eniID)
+		if deviceNumber != 0 {
+			log.Errorf("Device number of primary ENI is %d! It was expected to be 0", deviceNumber)
+			// TODO: Can this happen? Is the only option to panic?
+		}
+		return eniID, deviceNumber, nil
```
SaranBalaji90, line 674: Should we return 0 instead?

cmd/routed-eni-cni-plugin/driver/driver.go
mogren: I agree that we don't really need it, but I was trying to be consistent and show why we log device number as table number.

pkg/awsutils/awsutils.go
mogren: Yes, better to always return `0` here, together with an error log.

mogren:
@SaranBalaji90 Updated a running cluster both from v1.6.3, and from v1.7.1 to this build. Table numbers match.

pkg/networkutils/network_test.go
```diff
@@ -132,12 +132,12 @@ func TestSetupENINetworkMACFail(t *testing.T) {
 	assert.Errorf(t, err, "simulated failure")
 }
 
-func TestSetupENINetworkPrimary(t *testing.T) {
+func TestSetupENINetowrkErrorOnPrimaryENI(t *testing.T) {
```
SaranBalaji90, line 135: why is this test failing now?

pkg/networkutils/network.go
```diff
@@ -764,22 +760,23 @@ func setupENINetwork(eniIP string, eniMAC string, eniTable int, eniSubnetCIDR st
...
+	log.Debugf("Setting up ENI's default gateway %v, table %d, linkIndex %d", gw, tableNumber, linkIndex)
 	routes := []netlink.Route{
 		// Add a direct link route for the host's ENI IP only
 		{
-			LinkIndex: deviceNumber,
+			LinkIndex: linkIndex,
 			Dst:       &net.IPNet{IP: gw, Mask: net.CIDRMask(32, 32)},
 			Scope:     netlink.SCOPE_LINK,
-			Table:     eniTable,
+			Table:     tableNumber,
```
SaranBalaji90, line 771: This didn't fail any test is it? Because now the value that is passed to this method and table number are different, I would expect mock to fail?

pkg/awsutils/awsutils.go
```diff
@@ -639,38 +640,42 @@ func (cache *EC2InstanceMetadataCache) getIPsAndCIDR(eniMAC string) ([]*ec2.Netw
...
+			// Can this even happen? To be backwards compatible, we will always return 0 here and log an error.
+			log.Errorf("Device number of primary ENI is %d! It was expected to be 0", deviceNumber)
+			return eniID, 0, nil
+		}
+		return eniID, deviceNumber, nil
 	}
-	// 0 is reserved for primary ENI, the rest of them has to +1 to avoid collision at 0
-	return eni, int(deviceNum + 1), nil
+	// 0 is reserved for primary ENI
+	return eniID, deviceNumber, nil
```
SaranBalaji90, line 678: This change should have failed some unit test right? For eg instead of returning x now it will be returning x-1

mogren: @SaranBalaji90 Yes, this test had to be updated:

https://github.com/aws/amazon-vpc-cni-k8s/pull/1071/files/6d847db2bcd471c1b180fd8d7e81712ec6dbd942#diff-52f7ba503946a19ecad23c70607d5112R825

SaranBalaji90:
Thanks @mogren. Will definitely help in future when reading the code.

PR #1076, opened 2020-07-08
Create automated KOPS release test.

bnapolitan:
Create cluster using KOPS and run conformance tests against that cluster.


scripts/run-integration-tests.sh
```diff
@@ -140,13 +150,16 @@ mkdir -p "$REPORT_DIR"
 mkdir -p "$TEST_CLUSTER_DIR"
 mkdir -p "$TEST_CONFIG_DIR"
 
-if [[ "$PROVISION" == true ]]; then
-    START=$SECONDS
+START=$SECONDS
+if [[ "$PROVISION" == true && "$RUN_KOPS_TEST" == false ]]; then
```
mogren, line 154: I think we need to switch this logic, because if you set `PROVISION=false`, you will get a Kops cluster. 
```bash
if [[ "$PROVISION" == true ]]; then
  up-test-cluster
else if [[ "$RUN_KOPS_TEST" == true ]]; then
  up-kops-cluster
fi
```

scripts/lib/integration.sh
```diff
@@ -0,0 +1,32 @@
...
+    do
+        sleep 5
+        echo "Waiting for daemonset update"
+    done
+    echo "Updated!"
+
+    go install github.com/onsi/ginkgo/ginkgo
+    wget -qO- https://dl.k8s.io/v$K8S_VERSION/kubernetes-test.tar.gz | tar -zxvf - --strip-components=4 -C /tmp  kubernetes/platforms/linux/amd64/e2e.test
+
+    ginkgo -p --focus="Conformance"  --failFast --flakeAttempts 2 \
```
mogren, line 17: Can we use the full path for ginko here? `$GOPATH/bin/ginko`

scripts/lib/cluster.sh
```diff
@@ -41,3 +41,43 @@ function up-test-cluster() {
...
+    --kubernetes-version ${K8S_VERSION} \
+    ${CLUSTER_NAME}
+    kops update cluster --name ${CLUSTER_NAME} --yes
+    sleep 100
+    while [[ ! $(kops validate cluster | grep "is ready") ]]
+    do
+        sleep 5
+        echo "Waiting for cluster validation"
+    done
+    kubectl apply -f https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/release-1.6.3/config/v1.6/cni-metrics-helper.yaml
```
mogren, line 76: I think this should be a matching version of the metrics agent. Something like
```
"$DIR/../config/$MANIFEST_CNI_VERSION/cni-metrics-helper.yaml"
```

scripts/run-integration-tests.sh
```diff
@@ -157,8 +170,12 @@ sed -i'.bak' "s,:$MANIFEST_IMAGE_VERSION,:$TEST_IMAGE_VERSION," "$TEST_CONFIG_PA
 sed -i'.bak' "s,602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni-init,$INIT_IMAGE_NAME," "$TEST_CONFIG_PATH"
 sed -i'.bak' "s,:$MANIFEST_IMAGE_VERSION,:$TEST_IMAGE_VERSION," "$TEST_CONFIG_PATH"
 
-export KUBECONFIG=$KUBECONFIG_PATH
-ADDONS_CNI_IMAGE=$($KUBECTL_PATH describe daemonset aws-node -n kube-system | grep Image | cut -d ":" -f 2-3 | tr -d '[:space:]')
+if [[ $RUN_KOPS_TEST != true ]]; then
+    export KUBECONFIG=$KUBECONFIG_PATH
+    ADDONS_CNI_IMAGE=$($KUBECTL_PATH describe daemonset aws-node -n kube-system | grep Image | cut -d ":" -f 2-3 | tr -d '[:space:]')
```
mogren, line 175: Why is this not done for the kops cluster?

PR #1088, opened 2020-07-13
Added config map for env vars

nithu0115:
*Issue #, if available:* N/A

*Description of changes:* 
Moved all env vars to a configmap to persist these variable during CNI update 

### DIFF
--- a/config/master/aws-k8s-cni-configmap.yaml
+++ b/config/master/aws-k8s-cni-configmap.yaml
@@ -1,0 +1,27 @@
+---
+"apiVersion": "v1"
+"data":
+  "ADDITIONAL_ENI_TAGS": "{}"
+  "AWS_VPC_CNI_NODE_PORT_SUPPORT": "true"
+  "AWS_VPC_ENI_MTU": "9001"
+  "AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER": "false"
+  "AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG": "false"
+  "AWS_VPC_K8S_CNI_EXTERNALSNAT": "false"
+  "AWS_VPC_K8S_CNI_LOGLEVEL": "DEBUG"
+  "AWS_VPC_K8S_CNI_LOG_FILE": "/host/var/log/aws-routed-eni/ipamd.log"
+  "AWS_VPC_K8S_CNI_RANDOMIZESNAT": "prng"
+  "AWS_VPC_K8S_CNI_VETHPREFIX": "eni"
+  "AWS_VPC_K8S_PLUGIN_LOG_FILE": "/var/log/aws-routed-eni/plugin.log"
+  "AWS_VPC_K8S_PLUGIN_LOG_LEVEL": "DEBUG"
+  "DISABLE_INTROSPECTION": "false"
+  "DISABLE_METRICS": "false"
+  "DISABLE_TCP_EARLY_DEMUX": "false"
+  "ENABLE_POD_ENI": "false"
+  "WARM_ENI_TARGET": "1"
+"kind": "ConfigMap"
+"metadata":
+  "labels":
+    "k8s-app": "aws-node-cm"
+  "name": "aws-k8s-cni-configmap"
+  "namespace": "kube-system"
+...
--- a/config/master/manifests.jsonnet
+++ b/config/master/manifests.jsonnet
@@ -27,6 +27,39 @@ local regions = {
   },
 };
 
+local awsnodecm = {
+  configMap: {
+    apiVersion: "v1",
+    kind: "ConfigMap",
+    metadata: {
+      name: "aws-k8s-cni-configmap",
+      namespace: "kube-system",
+      labels: {
+        "k8s-app": "aws-node-cm",
+      },
+    },
+    data: {
+      AWS_VPC_CNI_NODE_PORT_SUPPORT: "true",
+      AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG: "false",
+      AWS_VPC_ENI_MTU: "9001",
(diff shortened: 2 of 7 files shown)


jaypipes (dismissed):
:+1: nice cleanup, @nithu0115, thanks!

config/master/manifests.jsonnet
```diff
@@ -390,3 +409,4 @@ byRegion("cni-metrics-helper", metricsHelper);
   [kv[0] + ".yaml"]: std.manifestYamlStream(objectValues(kv[1]))
   for kv in objectItems(output)
 }
+{"aws-k8s-cni-configmap.yaml": std.manifestYamlStream(objectValues(awsnodecm))}
```
anguslees, line 412: (My previous attempt at this comment disappeared?)

Nit: Instead of this, append to `output` above.  Something like:
```jsonnet
local output = ... + {"aws-k8s-cni-configmap": awsnodecm};
```

nithu0115: (Sorry for the delay), changed the jsonnet file as per your suggestion, thanks @anguslees :)

config/master/aws-k8s-cni-configmap.yaml
```diff
@@ -0,0 +1,27 @@
+---
+"apiVersion": "v1"
+"data":
```
mogren, line 3: Initially, do we want to keep this ConfigMap in the same file as the daemonset config, or do we want to keep them in separate files? There are upsides and downsides with both I guess.

This will also be related to how we handle it in eks-charts.

config/master/manifests.jsonnet
```diff
@@ -27,6 +27,39 @@ local regions = {
...
+      AWS_VPC_K8S_CNI_LOGLEVEL: "DEBUG",
+      AWS_VPC_K8S_CNI_LOG_FILE: "/host/var/log/aws-routed-eni/ipamd.log",
+      AWS_VPC_K8S_PLUGIN_LOG_FILE: "/var/log/aws-routed-eni/plugin.log",
+      AWS_VPC_K8S_PLUGIN_LOG_LEVEL: "DEBUG",
+      DISABLE_INTROSPECTION: "false",
+      DISABLE_METRICS: "false",
+      AWS_VPC_K8S_CNI_VETHPREFIX: "eni",
+      ADDITIONAL_ENI_TAGS: "{}",
+      AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER: "false",
+      DISABLE_TCP_EARLY_DEMUX: "false",
```
mogren, line 57: `DISABLE_TCP_EARLY_DEMUX` is only used by the initContainer, not IPAMD. It's probably better not to have it here, unless we decide to give the full config to the init container as well?

mikestef9: I would think we only want users setting variables in one place. Do you see any downside to giving the full config map to the init container?

config/master/aws-k8s-cni-configmap.yaml
nithu0115: @mogren I like to have configmap as separate file. By having it as a separate file, we will not undo all the changes made by the customer in their previous deployments if we have it in a single file.

config/master/manifests.jsonnet
nithu0115: @mikestef19 good point. What I was aiming with this PR is to avoid any manual intervention to avoid env overrides made by customer and to run ‘kubectl apply -f ./..yaml’ during upgrades

mikestef9: Oh ya I totally agree with the premise of the PR, as do many customers https://github.com/aws/containers-roadmap/issues/865

nithu0115: Got it. I don't see any downside in giving full configMap to init container as there is no sensitive information, would like to hear @mogren @jayanthvn thoughts on this ? Happy to amend my PR to include that.

mogren: @nithu0115 Oh, I don't mind that either, but then we need to add
```
        "envFrom":
          "value": "1"	        - "configMapRef":
            "name": "aws-k8s-cni-configmap"
```
to the init-container as well, that's what I meant, line 168 in your config.

dschunack:
Hi,

whats the current state here? In the last month it was pushed only release to release.

TBBle:
As well as rebasing, this PR probably needs to have the Helm chart support added, which has appeared in this repo via #1271 since this PR was originally created.

jayanthvn:
Hi @TBBle.

Yes once we add the config map support we need to fix the helm for config map.

@dschunack,
With config maps along + DaemonSet there is no vanilla k8s support for rolling upgrade when config map is modified. We are working on this and should have some updates in the following weeks.

julianxhokaxhiu:
What's the current milestone for this PR? This gets even more urgent now that for eg. in order to allow nodes having more Pods we need to enable an environment variable ( see https://aws.amazon.com/blogs/containers/amazon-vpc-cni-increases-pods-per-node-limits/ ).

Having a ConfigMap would help here instead of running a custom `kubectl set env` command on each installation ( and which also can't be automated via Terraform... ).

Thanks!

dschunack:
Hi,

What's the current state for this PR?

mlschindler:
Any updates on this issue? It is rather cumbersome for us to use hacky config management with Terraform (kubectl provider, local null resources, etc) and not to mention needing to bounce nodes when this style of management fails (as it does a little too frequently).

dschunack:
Any Update?

dschunack:
Still waiting on this

jdn5126:
Using ConfigMap to source environment variables for CNI is being tracked as a v2.x task, as rolling update with ConfigMap + daemonset pod needs more investigation. In the meantime, `kubectl set env` is the recommended mechanism.

This PR will not merge as is, but will be left open for those tracking. #2102 is the associated the issue, so please track this going forward.

PR #1092, opened 2020-07-15
Reverting chaining fix since for containerd we need to investigate fu…

jayanthvn:
…rther

*Issue #, if available:*

*Description of changes:*
With chaining introduced by #1029 and containerd, pods are failing to get IP. Pods will be stuck in container creating. So reverting the fix and will investigate further.

### DIFF
--- a/cmd/routed-eni-cni-plugin/cni.go
+++ b/cmd/routed-eni-cni-plugin/cni.go
@@ -215,18 +215,15 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 		return errors.Wrap(err, "add command: failed to setup network")
 	}
 
-	interfaceIndex := 0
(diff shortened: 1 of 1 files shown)


cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -215,18 +215,15 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
...
 		{
 			Version:   "4",
 			Address:   *addr,
-			Interface: &interfaceIndex,
 		},
 	}
 
 	result := &current.Result{
 		IPs:        ips,
-		Interfaces: []*current.Interface{{Name: hostVethName}},
```
mogren, line 229: For some still unknown reason, returning the interface here breaks the CNI when using containerd. Reverting until we have root caused the issue.

PR #1097, opened 2020-07-17
Cover Warm IP and Warm ENI targets release tests.

bnapolitan:
Create automated tests for warm IP and warm ENI release tests by setting environment variables RUN_WARM_IP_TEST=true and RUN_WARM_ENI_TEST=true respectively.

### DIFF
--- a/scripts/lib/common.sh
+++ b/scripts/lib/common.sh
@@ -28,3 +28,16 @@ function display_timelines() {
     echo "TIMELINE: Conformance tests took $CONFORMANCE_DURATION seconds."
     echo "TIMELINE: Down processes took $DOWN_DURATION seconds."
 }
+
+function run_warm_ip_test() {
+    $KUBECTL_PATH set env ds aws-node -n kube-system WARM_IP_TARGET=2
+    $KUBECTL_PATH set env ds aws-node -n kube-system MINIMUM_IP_TARGET=10
+    #Sleep a couple seconds to ensure propogation
+    sleep 2
+    KUBECTL_WARM_IP_TARGET=$(kubectl describe ds -n kube-system | grep WARM_IP_TARGET)
+    KUBECTL_MINIMUM_IP_TARGET=$(kubectl describe ds -n kube-system | grep MINIMUM_IP_TARGET)
+    if [[ $KUBECTL_WARM_IP_TARGET != *"2"* || $KUBECTL_MINIMUM_IP_TARGET != *"10"* ]]; then
+        echo "WARM_IP_TARGET not propogated!"
+        on_error
+    fi
+}
--- a/scripts/lib/common.sh
+++ b/scripts/lib/common.sh
@@ -41,3 +41,14 @@ function run_warm_ip_test() {
         on_error
     fi
 }
+
+function run_warm_eni_test() {
+    $KUBECTL_PATH set env ds aws-node -n kube-system WARM_ENI_TARGET=0
+    #Sleep a couple seconds to ensure propogation
+    sleep 2
+    KUBECTL_WARM_ENI_TARGET=$(kubectl describe ds -n kube-system | grep WARM_ENI_TARGET)
+    if [[ $KUBECTL_WARM_ENI_TARGET != *"0" ]]; then
+        echo "WARM_ENI_TARGET not propogated!"
+        on_error
+    fi
+}
--- a/scripts/run-integration-tests.sh
+++ b/scripts/run-integration-tests.sh
@@ -19,6 +19,7 @@ ARCH=$(go env GOARCH)
 : "${DEPROVISION:=true}"
 : "${BUILD:=true}"
 : "${RUN_CONFORMANCE:=false}"
+: "${RUN_WARM_IP_TEST:=false}"
 
 __cluster_created=0
 __cluster_deprovisioned=0
@@ -185,7 +186,9 @@ sleep 110
 CNI_IMAGE_UPDATE_DURATION=$((SECONDS - START))
 echo "TIMELINE: Updating CNI image took $CNI_IMAGE_UPDATE_DURATION seconds."
 
-$KUBECTL_PATH set env ds aws-node -n kube-system WARM_IP_TARGET=2
+if [[ RUN_WARM_IP_TEST == true ]]; then
+    run_warm_ip_test
+fi
 
 echo "*******************************************************************************"
 echo "Running integration tests on current image:"
--- a/scripts/run-integration-tests.sh
+++ b/scripts/run-integration-tests.sh
@@ -20,6 +20,7 @@ ARCH=$(go env GOARCH)
 : "${BUILD:=true}"
 : "${RUN_CONFORMANCE:=false}"
 : "${RUN_WARM_IP_TEST:=false}"
+: "${RUN_WARM_ENI_TEST:=false}"
 
 __cluster_created=0
 __cluster_deprovisioned=0
@@ -188,6 +189,9 @@ echo "TIMELINE: Updating CNI image took $CNI_IMAGE_UPDATE_DURATION seconds."
 
 if [[ RUN_WARM_IP_TEST == true ]]; then
     run_warm_ip_test
+fi
+if [[ RUN_WARM_ENI_TEST == true ]]; then
+    run_warm_eni_test
 fi
 
 echo "*******************************************************************************"
--- a/scripts/run-integration-tests.sh
+++ b/scripts/run-integration-tests.sh
@@ -185,6 +185,8 @@ sleep 110
 CNI_IMAGE_UPDATE_DURATION=$((SECONDS - START))
 echo "TIMELINE: Updating CNI image took $CNI_IMAGE_UPDATE_DURATION seconds."
 
+$KUBECTL_PATH set env ds aws-node -n kube-system WARM_IP_TARGET=2
+
 echo "*******************************************************************************"
 echo "Running integration tests on current image:"
 echo ""


scripts/lib/common.sh
```diff
@@ -28,3 +28,27 @@ function display_timelines() {
     echo "TIMELINE: Conformance tests took $CONFORMANCE_DURATION seconds."
     echo "TIMELINE: Down processes took $DOWN_DURATION seconds."
 }
+
+function run_warm_ip_test() {
+    $KUBECTL_PATH set env ds aws-node -n kube-system WARM_IP_TARGET=2
+    $KUBECTL_PATH set env ds aws-node -n kube-system MINIMUM_IP_TARGET=10
+    #Sleep a couple seconds to ensure propogation
+    sleep 2
```
mogren, line 36: Nit: "propagation"

Also, setting these requires the `aws-node` pods to restart, and that probably takes up to 20seconds per node, meaning we need to wait for 1 minute or more.

scripts/lib/common.sh
```diff
@@ -28,3 +28,27 @@ function display_timelines() {
...
+    KUBECTL_WARM_IP_TARGET=$(kubectl describe ds -n kube-system | grep WARM_IP_TARGET)
+    KUBECTL_MINIMUM_IP_TARGET=$(kubectl describe ds -n kube-system | grep MINIMUM_IP_TARGET)
+    if [[ $KUBECTL_WARM_IP_TARGET != *"2"* || $KUBECTL_MINIMUM_IP_TARGET != *"10"* ]]; then
+        echo "WARM_IP_TARGET not propogated!"
+        on_error
+    fi
```
mogren, line 42: The ENV variables will be set on the daemonset instantly, but not be available in the pods until they have restarted. I guess we should either check that all pods have restarted recently, or that all of them have the ENV variables set.

scripts/lib/common.sh
```diff
@@ -28,3 +28,27 @@ function display_timelines() {
...
+}
+
+function run_warm_eni_test() {
+    $KUBECTL_PATH set env ds aws-node -n kube-system WARM_ENI_TARGET=0
+    #Sleep a couple seconds to ensure propogation
```
mogren, line 47: Same here, also "propagated".

scripts/lib/common.sh
```diff
@@ -28,3 +28,68 @@ function display_timelines() {
...
     echo "TIMELINE: Down processes took $DOWN_DURATION seconds."
 }
+
+function run_warm_ip_test() {
+    echo "Setting WARM_IP_TARGET=2 and MINUMUM_IP_TARGET=10"
+    echo "Running warm ip test"
+    $KUBECTL_PATH set env ds aws-node -n kube-system WARM_IP_TARGET=2
+    $KUBECTL_PATH set env ds aws-node -n kube-system MINIMUM_IP_TARGET=10
+    #Sleep a couple seconds to ensure propagation
+    sleep 2
```
jaypipes, line 38: Pretty sure there is no delay for propogating the setting of an env var on the Daemonset resource... since `kubectl set env` simply reads the Daemonset's desired state and replaces the desired state with whatever was supplied, and writes the resource back to etcd...

scripts/lib/common.sh
```diff
@@ -28,3 +28,68 @@ function display_timelines() {
...
+    KUBECTL_WARM_IP_TARGET=$($KUBECTL_PATH describe ds -n kube-system | grep WARM_IP_TARGET)
+    KUBECTL_MINIMUM_IP_TARGET=$($KUBECTL_PATH describe ds -n kube-system | grep MINIMUM_IP_TARGET)
+    if [[ $KUBECTL_WARM_IP_TARGET != *"2"* || $KUBECTL_MINIMUM_IP_TARGET != *"10"* ]]; then
+        echo "WARM_IP_TARGET not propogated in daemonset!"
+        on_error
+    fi
+
+    sleep 140
```
jaypipes, line 46: Maybe *this* is what you actually were waiting to propogate. In other words, waiting for the *Pods* that are contained in the Daemonset definition to get the desired changed environment vars.

jaypipes: That said, 140 seconds seems a bit long to wait, especially since you are doing a `while` loop below.

scripts/lib/common.sh
```diff
@@ -28,3 +28,68 @@ function display_timelines() {
...
+    while [[ $($KUBECTL_PATH describe pod $FIRST_DS_POD_NAME -n=kube-system | grep WARM_IP_TARGET) != *"2"* || \
+        $($KUBECTL_PATH describe pod $FIRST_DS_POD_NAME -n=kube-system | grep MINIMUM_IP_TARGET) != *"10"* || \
+        $($KUBECTL_PATH describe pod $SECOND_DS_POD_NAME -n=kube-system | grep WARM_IP_TARGET) != *"2"* || \
+        $($KUBECTL_PATH describe pod $SECOND_DS_POD_NAME -n=kube-system | grep MINIMUM_IP_TARGET)  != *"10"* ]]
+    do
+        echo "Waiting for daemonset pod propagation..."
+        $KUBECTL_PATH get pods -n kube-system
+        FIRST_DS_POD_NAME=$($KUBECTL_PATH get pods -n kube-system | grep aws-node | sed -n '1 p' | awk '{print $1}')
+        SECOND_DS_POD_NAME=$($KUBECTL_PATH get pods -n kube-system | grep aws-node | sed -n '2 p' | awk '{print $1}')
+        sleep 5
```
jaypipes, line 62: this `while` loop has no timeout. It will effectively just run forever as long as the Pod's env var didn't propogate.

scripts/lib/common.sh
```diff
@@ -28,3 +28,68 @@ function display_timelines() {
     echo "TIMELINE: Conformance tests took $CONFORMANCE_DURATION seconds."
     echo "TIMELINE: Down processes took $DOWN_DURATION seconds."
 }
+
+function run_warm_ip_test() {
```
jaypipes, line 32: little nit, but the `lib/common.sh` file was really for general-purpose Bash functions, not necessarily for functions that run a specific test, etc. Consider placing this and the below function into a separate `lib/warm_tests.sh` file? Also, consider putting a comment above each function explaining roughly what the purpose of the test is... In this case, the test is merely checking the propogation of environment variables in Kubernetes Daemonset -> Pod, but isn't actually testing the **functionality** of the warm IP pool.

scripts/lib/common.sh
```diff
@@ -28,3 +28,68 @@ function display_timelines() {
...
+        FIRST_DS_POD_NAME=$($KUBECTL_PATH get pods -n kube-system | grep aws-node | sed -n '1 p' | awk '{print $1}')
+        SECOND_DS_POD_NAME=$($KUBECTL_PATH get pods -n kube-system | grep aws-node | sed -n '2 p' | awk '{print $1}')
+        sleep 5
+    done
+    echo "WARM_IP_TARGET and MINIMUM_IP_TARGET successfully propagated!"
+    echo $WARM_IP_VALUE1
+    echo $MINIMUM_IP_VALUE1
+}
+
+function run_warm_eni_test() {
```
jaypipes, line 69: again, this isn't testing anything to do with the warm ENI pool. it's just testing Kubernetes Daemonset -> Pod env var propogation :)

bnapolitan: Yes, this is mainly there just to make sure that everything runs properly and all integration and conformance testing is passed when the env variables are set. I suppose a more accurate name would be "setup_warm_eni_test" as the functions aren't testing anything yet :)

scripts/lib/integration.sh
```diff
@@ -29,4 +29,74 @@ function run_kops_conformance() {
...
+        echo "WARM_IP_TARGET not propogated in daemonset!"
+        on_error
+    fi
+
+    FIRST_DS_POD_NAME=$($KUBECTL_PATH get pods -n kube-system | grep aws-node | sed -n '1 p' | awk '{print $1}')
+    SECOND_DS_POD_NAME=$($KUBECTL_PATH get pods -n kube-system | grep aws-node | sed -n '2 p' | awk '{print $1}')
+    WARM_IP_VALUE1=""
+    WARM_IP_VALUE2=""
+    MINIMUM_IP_VALUE1=""
+    MINIMUM_IP_VALUE2=""
```
jaypipes, line 51: AFAICT, the variables above starting with WARM_ and MINIMUM_ are not used anywhere.

scripts/lib/integration.sh
```diff
@@ -29,4 +29,74 @@ function run_kops_conformance() {
...
+    sleep 2
+    KUBECTL_WARM_ENI_TARGET=$($KUBECTL_PATH describe ds -n kube-system | grep WARM_ENI_TARGET)
+    if [[ $KUBECTL_WARM_ENI_TARGET != *"0" ]]; then
+        echo "WARM_ENI_TARGET not propogated in daemonset!"
+        on_error
+    fi
+    FIRST_DS_POD_NAME=$($KUBECTL_PATH get pods -n kube-system | grep aws-node | sed -n '1 p' | awk '{print $1}')
+    SECOND_DS_POD_NAME=$($KUBECTL_PATH get pods -n kube-system | grep aws-node | sed -n '2 p' | awk '{print $1}')
+    WARM_ENI_VALUE1=""
+    WARM_ENI_VALUE2=""
```
jaypipes, line 86: Same comment about unused WARM_ variables.

bnapolitan: Thanks for pointing that out, have been removed now!

scripts/lib/integration.sh
```diff
@@ -29,4 +29,68 @@ function run_kops_conformance() {
...
+        SECOND_DS_POD_NAME=$($KUBECTL_PATH get pods -n kube-system | grep aws-node | sed -n '2 p' | awk '{print $1}')
+        sleep 5
+        if [[ $((SECONDS - START)) -gt 500 ]]; then
+            echo "Propogation for WARM_IP_TARGET timed out!"
+            on_error
+        fi
+    done
+    echo "WARM_IP_TARGET and MINIMUM_IP_TARGET successfully propagated!"
+    echo $WARM_IP_VALUE1
+    echo $MINIMUM_IP_VALUE1
```
jaypipes, line 65: unused vars `echo`'d out above :)

scripts/lib/integration.sh
```diff
@@ -29,4 +29,68 @@ function run_kops_conformance() {
...
+        FIRST_DS_POD_NAME=$($KUBECTL_PATH get pods -n kube-system | grep aws-node | sed -n '1 p' | awk '{print $1}')
+        SECOND_DS_POD_NAME=$($KUBECTL_PATH get pods -n kube-system | grep aws-node | sed -n '2 p' | awk '{print $1}')
+        sleep 5
+        if [[ $((SECONDS - START)) -gt 500 ]]; then
+            echo "Propogation for WARM_IP_TARGET timed out!"
+            on_error
+        fi
+    done
+    echo "WARM_ENI_TARGET successfully propagated!"
+    echo $WARM_ENI_VALUE1
```
jaypipes, line 95: unused vars `echo`'d out above :)

PR #1102, opened 2020-07-24
[WIP] Use docker/CRI to discover pods at node init

fawadkhaliq:
*Issue* https://github.com/aws/amazon-vpc-cni-k8s/issues/711

*Description of changes:*
ipamd `nodeInit` relies on pod discovery watcher to get the list of running pods. The watcher and ipamd are launched in parallel. This can lead to a race condition where ipamd `getLocalPodsWithRetry` kicks in before the watcher is done with listing all the running pods on a node. If this happens, ipamd may end up allocating in-use IP addr to new pods

*Changes*
- Changed ipamd to query docker/CRI to discover the list of node local pods instead Kubernetes API
- Add support in the cri.go to get IP address of a pod and added `IP` field to `cri.SandboxInfo`
- Datastore inputs changed from `k8sapi.K8SPodInfo` to `cri.SandboxInfo`

*Alternative considered*
- Instead of using the watcher, make a blocking list pods call to the Kubernetes API server directly. This approach does not guarantee accurate state. On an overloaded system, API server might lag behind the node local state returning. While this approach would solve the original race condition, it can still return a different view of the running pods (eventually consistent)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ aws-k8s-agent
 aws-cni
 verify-aws
 verify-network
+coverage.txt
 *~
 *.swp
 .idea/
--- a/Makefile
+++ b/Makefile
@@ -114,10 +114,12 @@ docker-func-test: docker
 		"$(IMAGE_NAME)"
 
 # Run unit tests
+unit-test: export AWS_VPC_K8S_CNI_LOG_FILE=stdout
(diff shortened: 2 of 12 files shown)


fawadkhaliq:
We have decided to go with 1.7.x to have the fix

Issue #1089, opened 2020-07-14
envLogLevel is not getting env var


wguto:
https://github.com/aws/amazon-vpc-cni-k8s/blob/d070fcae5ee2b2292ae7da9e0d2af75ae35e849b/pkg/utils/logger/config.go#L24

I am passing this ENV VAR to the deployment, I can see the env var is in the container and all but it is still spamming info logs of this exact line:

https://github.com/aws/amazon-vpc-cni-k8s/blob/d070fcae5ee2b2292ae7da9e0d2af75ae35e849b/cmd/cni-metrics-helper/main.go#L105

any reason why it is not loading the env var?
I added the deployment.yaml to my gist [here](https://gist.github.com/wguto/034a1fb5accc728dd7cfe244211f539b)

mogren:
Hi @wguto!

This look like an issue with the log config for cni-metrics-helper. Thanks for reporting, we will fix it.

mogren:
Added reading `AWS_VPC_K8S_CNI_LOGLEVEL` environment variable in the metrics helper in #1104.

PR #1104, opened 2020-07-27
Enable log config for the metrics agent

mogren:
*Issue #, if available:*
#1089

*Description of changes:*
The cni-metrics-helper should check the same logging config env variable as ipamd does to get the log level. Will still log to stdout after this change.
* `cni-metrics-helper` will get the LogLevel from env variable `AWS_VPC_K8S_CNI_LOGLEVEL`
* Lowered the log level of full aggregated CloudWatch data to `Debug`
* Default empty `logFilePath` to `stdout` in Zap config

### DIFF
--- a/cmd/cni-metrics-helper/metrics/metrics.go
+++ b/cmd/cni-metrics-helper/metrics/metrics.go
@@ -18,19 +18,14 @@ import (
 	"bytes"
 	"fmt"
 
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/publisher"
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
 	"github.com/aws/aws-sdk-go/aws"
-
 	"github.com/aws/aws-sdk-go/service/cloudwatch"
-
 	dto "github.com/prometheus/client_model/go"
 	"github.com/prometheus/common/expfmt"
 	clientset "k8s.io/client-go/kubernetes"
-
-	"github.com/aws/amazon-vpc-cni-k8s/pkg/publisher"
-	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
 )
-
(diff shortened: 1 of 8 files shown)


cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -121,7 +117,7 @@ func processPercentile(metric *dto.Metric, act *metricsAction) {
 	act.actionFunc(&act.data.curSingleDataPoint, p99)
 }
 
-func processHistogram(metric *dto.Metric, act *metricsAction) {
+func processHistogram(metric *dto.Metric, act *metricsAction, log logger.Logger) {
```
jaypipes, line 120: this is fine. might be a bit cleaner, though, to make this function a method on `metricsTarget` struct versus adding a `log` parameter to all these functions. :)

cmd/cni-metrics-helper/metrics/metrics_test.go
```diff
@@ -17,16 +17,20 @@ import (
...
 )
 
 type testMetricsTarget struct {
 	metricFile         string
 	interestingMetrics map[string]metricsConvert
 }
 
+func (target *testMetricsTarget) getLogger() logger.Logger {
+	return logger.DefaultLogger()
+}
```
jaypipes, line 32: you wouldn't need this interface method if you just made all those functions instead be methods on metricsTarget.

wguto:
any progress on this pr?

cmd/cni-metrics-helper/metrics/metrics.go
mogren: I agree, but since `metricsTarget` is an interface, it will be a somewhat bigger change. Will check in a follow up PR how that would look like. 😄

mogren:
@wguto I was out last week, but now I'm back. 🙂

PR #1109, opened 2020-07-30
Tag ENI with creation timestamp and avoid cleanup if created in last …

jayanthvn:
…5 mins

*Issue #, if available:*

*Description of changes: Tag EKS created ENI with creation timestamp. While checking for leaked ENIs we prevent deletes of ENIs created in the last 5 minutes. Also for scenarios where Cx has one cluster with 1.5. x and another with 1.6.x - The cleanup function in 1.6.x looks for ENIs that are created by the CNI, but not attached to any instance. In 1.6.0 we tag the ENI with the instance name after it has been attached. In 1.5.x, we tag the nodes first, then attach them. If we find any such ENIs then we update with new time preventing those to be deleted.


pkg/awsutils/awsutils.go
```diff
@@ -1325,9 +1379,28 @@ func (cache *EC2InstanceMetadataCache) getFilteredListOfNetworkInterfaces() ([]*
...
+			continue
+		}
+		// Check that it's not a newly created ENI
+		tags := getTags(networkInterface, aws.StringValue(networkInterface.NetworkInterfaceId))
+
+		if value, ok := tags[eniCreatedAtTagKey]; ok {
+			parsedTime, err := time.Parse(time.RFC3339, value)
+			if err != nil {
+
+			}
```
mogren, line 1392: If we fail to parse the time here, should we just log a warning showing the tag, then tag it correctly?

pkg/awsutils/awsutils.go
```diff
@@ -1325,9 +1379,28 @@ func (cache *EC2InstanceMetadataCache) getFilteredListOfNetworkInterfaces() ([]*
...
+			if time.Since(parsedTime) < (5 * time.Minute) {
+				log.Infof("Found ENI created less than 5 mins so not cleaning up")
```
mogren, line 1394: Nit: `"Found an ENI created less than 5 minutes ago, so not cleaning it up"`

pkg/awsutils/awsutils.go
```diff
@@ -1325,9 +1379,28 @@ func (cache *EC2InstanceMetadataCache) getFilteredListOfNetworkInterfaces() ([]*
...
+				continue
+			}
+			log.Debugf("%v", value)
+		} else {
+			//Set time if we didn't have one. This is to catch the v1.5.x or earlier CNI versions.
```
mogren, line 1399: Nit: 
```
// Set a time if we didn't find one. This is to prevent accidentally deleting ENIs that are in the 
// process of being attached by CNI versions v1.5.x or earlier.
````

File: pkg/awsutils/awsutils.go [hunk 4]
```diff
@@ -1296,6 +1318,38 @@ func (cache *EC2InstanceMetadataCache) cleanUpLeakedENIs() {
...
+		Tags: tags,
+	}
+
+	_ = retry.RetryNWithBackoff(retry.NewSimpleBackoff(500*time.Millisecond, maxBackoffDelay, 0.3, 2), 5, func() error {
+		start := time.Now()
+		_, err := cache.ec2SVC.CreateTagsWithContext(context.Background(), input, userAgent)
+		awsAPILatency.WithLabelValues("CreateTags", fmt.Sprint(err != nil)).Observe(msSince(start))
+		if err != nil {
+			awsAPIErrInc("CreateTags", err)
+			log.Warnf("Failed to tag the newly created ENI %s:", eniID)
```
mogren, line 1345: Nit: Log message is not right, and won't log the error. 

Maybe `log.Warnf("Failed to add tag to ENI %s: %v", eniID, err)`

pkg/awsutils/awsutils.go
jayanthvn: Overlooked this. Yeah I can do that.

pkg/awsutils/awsutils.go
```diff
@@ -1314,9 +1368,31 @@ func (cache *EC2InstanceMetadataCache) getFilteredListOfNetworkInterfaces() ([]*
...
+		// Check that it's not a newly created ENI
+		tags := getTags(networkInterface, aws.StringValue(networkInterface.NetworkInterfaceId))
+
+		if value, ok := tags[eniCreatedAtTagKey]; ok {
+			parsedTime, err := time.Parse(time.RFC3339, value)
+			if err != nil {
+				log.Warnf("ParsedTime format %s is wrong so retagging with current TS", parsedTime)
+				cache.tagENIcreateTS(aws.StringValue(networkInterface.NetworkInterfaceId), maxENIBackoffDelay)
+			}
+			if time.Since(parsedTime) < (5 * time.Minute) {
```
mogren, line 1383: Extract to a constant, something like `eniDeleteCooldownTime` or something

PR #1111, opened 2020-07-31
Remove references to unused metadata `owner-id`

anguslees:
By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1110, opened 2020-07-31
Remove old pre-1.3 migration code

anguslees:
Also removes only use of IMDS vpc-ipv4-cidr-block

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #1023, opened 2020-06-10
Log a warning if the primary ENI has DeleteOnTermination false


mogren:
Since v1.6.0, most issues with leaked ENIs seem to come from launch templates that have `Attachment.DeleteOnTermination` set to `false`. We should log a Warning in that case.

In Launch Template -> Network Interfaces, `delete_on_termination` should be true.

PR #1113, opened 2020-08-03
Avoid caching dynamic VPC CIDR info

anguslees:
We expect the list of VPC CIDRs to change.  Rather than cache the VPC CIDRs and then asynchronously update that cache, just don't cache :) and fetch the latest VPC CIDRs on demand.

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -129,7 +129,7 @@ type APIs interface {
 	DeallocIPAddresses(eniID string, ips []string) error
 
 	// GetVPCIPv4CIDRs returns VPC's CIDRs from instance metadata
-	GetVPCIPv4CIDRs() []string
+	GetVPCIPv4CIDRs() ([]string, error)
 
 	// GetLocalIPv4 returns the primary IP address on the primary ENI interface
 	GetLocalIPv4() net.IP
@@ -164,7 +164,6 @@ type EC2InstanceMetadataCache struct {
 	localIPv4           net.IP
 	instanceID          string
 	instanceType        string
-	vpcIPv4CIDRs        StringSet
 	primaryENI          string
 	primaryENImac       string
 	availabilityZone    string
@@ -401,16 +400,9 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context) 
 		return err
 	}
 
-	// retrieve VPC IPv4 CIDR blocks
-	err = cache.refreshVPCIPv4CIDRs(mac)
-	if err != nil {
-		return err
-	}
-
 	// Refresh security groups and VPC CIDR blocks in the background
 	// Ignoring errors since we will retry in 30s
 	go wait.Forever(func() { _ = cache.refreshSGIDs(mac) }, 30*time.Second)
-	go wait.Forever(func() { _ = cache.refreshVPCIPv4CIDRs(mac) }, 30*time.Second)
 
 	// We use the ctx here for testing, since we spawn go-routines above which will run forever.
 	select {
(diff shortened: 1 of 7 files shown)


anguslees:
Note to reviewer: this PR anticipates the merge of two cleanup PRs #1110 and #1111 (and includes them in the commits in _this_ PR).  I suggest they be reviewed/merged first, and (if necessary) rebase this PR.

pkg/ec2metadata/client.go
```diff
@@ -22,6 +24,7 @@ import (
 // EC2Metadata wraps the methods from the amazon-sdk-go's ec2metadata package
 type EC2Metadata interface {
 	GetMetadata(path string) (string, error)
+	GetMetadataWithContext(ctx context.Context, path string) (string, error)
```
mogren, line 27: Ah... We already set the default config to have `WithMaxRetries(10)`...

pkg/awsutils/awsutils.go
```diff
@@ -286,314 +285,125 @@ func New() (*EC2InstanceMetadataCache, error) {
...
+			PrivateIpAddress: aws.String(ip4.String()),
+		}
 	}
 
-	if cache.primaryENI == eni {
-		log.Debugf("Using device number 0 for primary eni: %s", eni)
-		return eni, 0, nil
-	}
-	// 0 is reserved for primary ENI, the rest of them has to +1 to avoid collision at 0
-	return eni, int(deviceNum + 1), nil
```
anguslees, line 596: Note: I removed the +1 thing here, since we're already assuming that 0 == primary (and the previous code did too).

I think (with high confidence) that that's ok.  I think (with much less confidence) that's it's also ok across upgrade.  Upgrade testing/failures should look here for the cause...

pkg/awsutils/imds.go
```diff
@@ -0,0 +1,216 @@
...
+
+func (imds TypedIMDS) getInt(ctx context.Context, key string) (int, error) {
+	data, err := imds.GetMetadataWithContext(ctx, key)
+	if err != nil {
+		return 0, err
+	}
+	return strconv.Atoi(data)
+}
+
+// GetDeviceNumber returns the unique device number associated with an interface.  The primary interface is 0.
```
mogren, line 86: Is the primary interface always 0? I have kind of assumed it is, but is that documented somewhere?

pkg/awsutils/awsutils.go
```diff
@@ -44,16 +42,16 @@ import (
...
-	metadataSGs          = "/security-group-ids/"
-	metadataSubnetID     = "/subnet-id/"
-	metadataVPCcidrs     = "/vpc-ipv4-cidr-blocks/"
-	metadataDeviceNum    = "/device-number/"
-	metadataInterface    = "/interface-id/"
+	metadataSGs          = "/security-group-ids"
+	metadataSubnetID     = "/subnet-id"
+	metadataVPCcidrs     = "/vpc-ipv4-cidr-blocks"
+	metadataDeviceNum    = "/device-number"
+	metadataInterface    = "/interface-id"
```
mogren, line 54: I guess we don't use any of these any more, since they are all hardcoded in `imds.go`?

pkg/awsutils/awsutils.go
mogren: I started to do something similar in #1071, but was not 100% sure that the primary ENI is always at index 0: https://github.com/aws/amazon-vpc-cni-k8s/pull/1071/files#diff-b55f09f925e40eba84053fdae2096d9cR648

pkg/awsutils/imds.go
anguslees: Not explicitly afaics :(  Primary ENI is assumed to be always "eth0" in all our scripts/docs/tutorials and the metadata docs for device-number say "eg eth3 is device-number 3" (*).  Primary ENI is also the first/only interface attached at boot, so assuming indices go up then the first is also always 0.

(*) .. which is obviously nonsense in the general case, because you can't know how the OS is going to do device enumeration - but I believe this is the fantasy that the original Xen API was based around.

I can remove this comment if you want.  The rest of our code already assumes 0==primary (eg [here](https://github.com/aws/amazon-vpc-cni-k8s/blob/be9ca94f2b304b9359f4961d1e20e08d9aee5e94/pkg/awsutils/awsutils.go#L1063) and [here](https://github.com/aws/amazon-vpc-cni-k8s/blob/be9ca94f2b304b9359f4961d1e20e08d9aee5e94/pkg/awsutils/awsutils.go#L562)) and this hasn't caused us any problems - yet.

Mostly I wanted to be more consistent across our codebase instead of using a mix of `.Primary` and `.DeviceIndex==0`, and here seemed a reasonable place to declare this assumption.  I can certainly soften it to say "vpc-cni assumes that primary interface is always 0"

pkg/awsutils/awsutils.go
anguslees: Right.  A bunch of them are still used in test code, so I should at least move them there.  (I was sort of hoping that unused code tool I used would tell me this, but it seems it doesn't see `const`s)

pkg/awsutils/imds.go
mogren: This is slightly different:
```
		// TODO: Verify that the first IP is always the primary
		primary := isFirst
```
This is checking with IP should not be used by pods. I've checked in the [aws/amazon-ec2-net-util](https://github.com/aws/amazon-ec2-net-utils/blob/f42f94acf498d21ef5e216fae56a5bd2601ea2a8/ec2net-functions#L97-L100) repo where they also use index 0 to be the "attchment IP". We should definitely see if we can get rid of this limitation though, and just reserve the "node IP".

The `aws.Int64Value(ec2res.Attachment.DeviceIndex) == 0` check to add that log line was to help customers who don't have `delete_on_termination` set for the primary ENI. #1023

anguslees: (Oh yes, you're right, one of my examples was primary *IP* not primary ENI)

So I'm happy to remove this, and instead make the code track primary ENI separately, and allow any DeviceNumber (including 0) for any ENI.

What I _don't_ want to do is have a function called `EC2InstanceMetadataCache.getENIDeviceNumber()` that does _not_ return the actual device number :P    Shall I put in a PR ahead of this one that tracks "primary-ness" explicitly?

anguslees:
Rebased.  With the additional changes, I think this is too big now.  I'm going to split out the first commit into a new PR..

anguslees:
Drastically reduced the scope of this PR, and reworded the PR description accordingly.

pkg/awsutils/awsutils.go
```diff
@@ -1454,8 +1416,21 @@ func (cache *EC2InstanceMetadataCache) getFilteredListOfNetworkInterfaces() ([]*
...
+		return nil, err
+	}
+
+	// TODO: keep as net.IPNet and remove this round-trip to/from string
+	asStrs := make([]string, len(ipnets))
+	for i, ipnet := range ipnets {
+		asStrs[i] = ipnet.String()
+	}
+
+	return asStrs, nil
```
mogren, line 1433: This is the main change. Just always fetch the CIDR blocks from the metadata, instead of having a loop to update a cache.

mogren (approved):
One less go-routine... Nice! 

Triggered another e2e test: https://github.com/aws/amazon-vpc-cni-k8s/runs/1152850334?check_suite_focus=true

jayanthvn (approved):
Nice, as @mogren mentioned one less thread :)

Issue #186, opened 2018-09-28
Update documentation for CNI environment variables and provide an example


jicowan:
The documentation says the following about the environment variables you can add to the pod manifest for the CNI plugin: 
```
AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG
Type: Boolean
Default: false
```

```
WARM_ENI_TARGET
Type: Integer
Default: 1
```

Both of these should be of `Type: String`

A better description of how the custom networking for pods works would also be helpful.  When I reconfigured the existing nodes in my cluster to use custom networking, the daemonsets were assigned secondary IPs from the worker node's primary ENI instead of the ENI associated with the subnet specified in the CRD (custom network config).  Once I annotated the node with the CRD, pods scheduled onto that node were assigned secondary IPs from the custom network config range.  Also, please explain how pods, including daemonsets, can be assigned IPs from the custom range during scale out operations, prior to the node being annotated with a custom network config.

mogren:
The custom network part is tracked in https://github.com/awsdocs/amazon-eks-user-guide/pull/72, will fix the types.

marceloboeira:
it seems the issue is solved, should it be closed @jicowan / @mogren ?

mogren:
Thanks @marceloboeira, you are right! 😄

PR #1116, opened 2020-08-04
Changed data type for variables: AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG a…

abhinavmpandey08:
…nd WARM_ENI_TARGET to String

*Issue #, if available:* #186 

*Description of changes:* Updated variable type of `AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG` from `Boolean` to `String` and `WARM_ENI_TARGET` from `Integer` to `String`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ The following environment variables are available, and all of them are optional.
 
 `AWS_VPC_CNI_NODE_PORT_SUPPORT`
 
-Type: Boolean
+Type: Boolean as a String
 
 Default: `true`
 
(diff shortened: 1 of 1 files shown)


mogren:
Hmm... Technically, looking at the [env settings in the sample config](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/master/aws-k8s-cni.yaml#L109-L143), they are all strings, but contains different types of values. `AWS_VPC_ENI_MTU` should clearly be a number for example, but it is quoted in the config. Should we specify it as something else? Like `"quoted integer"` or `Type: Boolean as a string`?

abhinavmpandey08:
That's a good suggestion! I can do that. Should I change other variables' type to reflect this too or just the ones mentioned in  #[186](https://github.com/aws/amazon-vpc-cni-k8s/issues/186)?

mogren:
@abhinavmpandey08 I think we should fix the documentation for all ENV variables at the same time. :)

PR #1119, opened 2020-08-06
Prepare for v1.6.4-rc1 release

### LINKED ISSUE #711: Duplicate IP Addresses for pods w/o host network

mogren:
*Issue #, if available:*
Release #1118 to resolve #711

*Description of changes:*
* Updated changelog 
* Updsted config sample
* Fixed TravisCI config

### DIFF
--- a/config/v1.6/aws-k8s-cni.yaml
+++ b/config/v1.6/aws-k8s-cni.yaml
@@ -99,7 +99,7 @@ spec:
       tolerations:
         - operator: Exists
       containers:
-        - image: 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.6.3
+        - image: 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.6.4-rc1
           imagePullPolicy: Always
           ports:
             - containerPort: 61678
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ language: go
 # Any older versions are considered deprecated.
 go:
   - "1.13.x"
-  - "1.12.x"
 
 # Only clone the most recent commit.
 git:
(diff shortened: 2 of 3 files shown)


config/v1.6/aws-k8s-cni.yaml
```diff
@@ -99,7 +99,7 @@ spec:
       tolerations:
         - operator: Exists
       containers:
-        - image: 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.6.3
+        - image: 602401143452.dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.6.4-rc1
```
jayanthvn, line 102: Should we add cii-metrics helper rc1?

mogren: For `v1.6.4` final we should, but since no metrics have changed in this release candidate, `v1.6.3` works for testing. Same with the other configs, we will update once they are built and published.

PR #1123, opened 2020-08-07
Improve CRI->checkpoint logic in the face of downgrades

anguslees:
Commit e38a4d8a8a9358f9b1e24f465a8d42a95a837533 introduced a method of
checkpointing IPAM state to disk and avoiding CRI access.  As written,
that commit failed to handle upgrade/downgrade/upgrade: on the second
upgrade the checkpoint file would exist but contain stale information.

This change improves the logic to use a more robust "expand/contract"
rollout:

Migration phase0 (CNI 1.6): Read/write[*] from CRI only.
Migration phase1 (CNI 1.7): Read from CRI.  Write to CRI+file.
Migration phase2 (CNI 1.8?): Read from file. Write to CRI+file.
Migration phase3 (hypothetical): Read/write from file only.

This (and the previous) commit implements phase1.

[*] There is no "write" to CRI directly, but the logic assumes it
happens indirectly as a consequence of a successful CNI add/delete
operation. Phase2 and phase3 are thus equivalent in this migration.

Note this supports upgrades/downgrades across a single version
only (without a node reboot).


pkg/ipamd/datastore/data_store.go
```diff
@@ -52,10 +52,28 @@ const (
...
+// ourself in a file (similar to host-ipam CNI plugin).
+//
+// Because we don't want to require a node restart during CNI
+// upgrade/downgrade, we need an "expand/contract" style upgrade to
+// keep the two stores in sync:
+//
+// Migration phase0 (CNI 1.6): Read/write from CRI only.
+// Migration phase1 (CNI 1.7): Read from CRI.  Write to CRI+file.
+// Migration phase2 (CNI 1.8?): Read from file. Write to CRI+file.
+// Migration phase3 (hypothetical): Read/write from file only.
```
mogren, line 69: Thanks a lot for the clear documentation. I'll update the readme as well in a follow up PR

pkg/ipamd/ipamd_test.go
```diff
@@ -503,7 +504,9 @@ func TestIPAMContext_nodeIPPoolTooLow(t *testing.T) {
 }
 
 func testDatastore() *datastore.DataStore {
-	return datastore.NewDataStore(log, datastore.NewTestCheckpoint(datastore.CheckpointData{Version: datastore.CheckpointFormatVersion}))
+	ds := datastore.NewDataStore(log, datastore.NewTestCheckpoint(datastore.CheckpointData{Version: datastore.CheckpointFormatVersion}))
+	ds.CheckpointMigrationPhase = 2
```
mogren, line 508: Since `CheckpointMigrationPhase` is still `1` in the default code, we should have a `TestNodeInitPhase1` as well. I'll add another test to cover this.

pkg/ipamd/datastore/data_store.go
```diff
@@ -701,7 +731,16 @@ func (ds *DataStore) UnassignPodIPv4Address(ipamKey IPAMKey) (ip string, deviceN
...
+	if addr == nil {
+		// This `if` block should be removed when the CRI
+		// migration code is finally removed.  Leaving a
+		// compile dependency here to make that obvious :P
+		var _ = CheckpointMigrationPhase
+
+		// If the entry was discovered by querying CRI at
+		// restart rather than by observing an ADD operation
+		// directly, then we won't have captured the true
+		// networkname/ifname.
```
sramabad1, line 743: Can we have unified key for in-memory data store and File ? If so we can avoid the above logic. Also AssignIPv4Address doesnt account for key that has been built during restore(i.e.,  const BackfillNetworkName = "_migrated-from-cri" and const BackfillNetworkIface = "unknown")

anguslees: > Can we have unified key for in-memory data store and File

We do: `IPAMKey` is used for both in-memory and file, as you can see in the call to `FindAddressForSandbox` 2 lines above this diff chunk.

I think the placeholder values are not relevant for `AssignIPv4Address`?  `AssignIPv4Address` is invoked for _new_ CNI invocations, and the placeholder values are only used for assignments that were already created as part of a previous ipamd (and learned through CRI).

Oh, if you're referring to the ["return existing assignment for existing IPAMKey"](https://github.com/aws/amazon-vpc-cni-k8s/blob/170d3c12adcad0d56cfe7ced0f8ebf9a6b177ceb/pkg/ipamd/datastore/data_store.go#L480-L484) logic in `AssignPodIPv4Address`, then yes.  We can extend that to also include this _migrated-from-cri fallback, or I would rather just delete those lines tbh.  There is no legitimate reason that that function should ever be called twice for the same IPAMKey (without an intervening DEL) - I just postponed this conversation in my earlier PR...

sramabad1: By unified key, i meant, can we populate entire key into file and restore it back so that we don't have to use const BackfillNetworkName = "_migrated-from-cri" and const BackfillNetworkIface = "unknown ?

anguslees: We can't.  The only case where we ever use these placeholder values is the CRI->file upgrade case where we don't _know_ the correct values (because CRI doesn't record/expose it).

sramabad1: Thanks Gus for the detailed note. If i understand your explanation, CRI api called during nodeInit doesn't seem to provide all the information required to form complete IPAM key. If this is the case, then CNI spec guideline that you have quoted above cannot be met using current CRI API. I would say operational correctness should win over spec conformance and we can revisit conformance once CRI API can provide everything required to form the key. If CNI_CONTAINERID is guaranteed to be unique enough, we could just have this as part of the key

Issue #208, opened 2018-10-22
Configurable subnet/security group per pod?


mumoshu:
Hey! Thanks as always for providing us the world's best cloud 😉 

This is initially a question but can be turned into a feature request according to further discussions.

#165 has added the feature that works like "configurable pod subnet/security group per node". Would it be beneficial and feasible to extend it to be able to configure it per pod?

My use-case for such feature is minimizing EC2 instance cost by further putting pods workloads together. What we have today works by creating an ASG of nodes per `ENIConfig`(=desired set of pod subnet/security groups), which results in either reduced intensity of workloads per node or increased complexity of the cluster setup due to varying EC2 instance types.

My guess was that you would suggest varying EC2 instance types. That is, just create an `ENIConfig` per a dedicated pod subnet+security group set, while also using smaller instance types for less used pod subnet+security groups set. But I thought it would still worth clarifying by writing here as an issue anyway.

asevans:
I also would greatly prefer the ability to scope security groups to pods.

mumoshu:
Your point about flexibly modifying the relationship between workloads, perhaps by not forcing us to manually provision ASG per ENIConfig as I understood, makes sense.
Thanks for chiming in!

dada941:
Configuration could be done by a set of annotations on the pod itself.

Similar how containerd/CRI does to switch between container runtime (kata/gvisor etc) : https://github.com/containerd/cri/blob/master/pkg/annotations/annotations.go

garo:
A pod could have an annotation describing the required security group id. Based on this the CNI would create a new ENI with the appropriate security group and assign the pod an ip from this ENI.

As a single ENI can support multiple IPs (depending on instance type from 4 to 50) If another pod requires the same security group the pod could get an IP from the same ENI, but in other case the ENI would just be not filled completely and this would be an acceptable tradeoff.

I guess that this would need to be taken in account with the pod scheduling system (Pod Capacity) so that a pod wouldn't be scheduled into an node which could not give an ip due to the security group settings.

ofiliz:
Thank you all for the feedback! We are always working on improving container networking in EC2. We'll have more to share in the future.

s1rc:
Subscribing. Definitely looking forward to being able to assign security groups to pods.

C0en:
Apart from being able to assign a seperate config to a pod I think the most useful option would be to be able to assign a eniconfig per namespace.

pwdebruin1:
[...]

PR #1125, opened 2020-08-07
Add support to setup pod network using VLANs

SaranBalaji90:
***Description of changes:***

Issue:
https://github.com/aws/containers-roadmap/issues/177, https://github.com/aws/amazon-vpc-cni-k8s/issues/208, https://github.com/aws/containers-roadmap/issues/398

Changes included in this PR are as follows,
* Creates vlan for pod requesting unique security group.
* Skips SNAT for pkts flowing out of Vlan tags, this is to prevent outbound connections from bypassing the egress check.
* Supports NodePort for pod backed using Vlan dev.
* Adds packet verifier binary to validate the packet flow as part of integration tests.

High level design diagram

image

**Follow-up PRs**
* IPAMD changes to read the details of vlan and send it back to CNI plugin.


cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -175,45 +180,10 @@ func (os *linuxNetwork) SetupNS(hostVethName string, contVethName string, netnsP
...
-	}
-
-	createVethContext := newCreateVethPairContext(contVethName, hostVethName, addr, mtu)
-	if err := ns.WithNetNSPath(netnsPath, createVethContext.run); err != nil {
-		log.Errorf("Failed to setup NS network %v", err)
-		return errors.Wrap(err, "setupNS network: failed to setup NS network")
-	}
 
-	hostVeth, err := netLink.LinkByName(hostVethName)
+	hostVeth, err := setupVeth(hostVethName, contVethName, netnsPath, addr, netLink, ns, mtu, procSys, log)
```
mogren, line 184: Unfortunate diff.. This just moves the veth setup code into it's own function.

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -282,6 +252,151 @@ func setupNS(hostVethName string, contVethName string, netnsPath string, addr *n
...
+// SetupPodENINetwork sets up the network ns for pods requesting its own security group
+func (os *linuxNetwork) SetupPodENINetwork(hostVethName string, contVethName string, netnsPath string, addr *net.IPNet,
+	vlanId int, eniMAC string, subnetGW string, parentIfIndex int, mtu int, log logger.Logger) error {
+
+	hostVeth, err := setupVeth(hostVethName, contVethName, netnsPath, addr, os.netLink, os.ns, mtu, os.procSys, log)
+	if err != nil {
+		return errors.Wrapf(err, "SetupPodENINetwork failed to setup veth pair.")
+	}
+
+	vlanTableId := vlanId + 100
```
mogren, line 310: Table names are currently based on Device Index for attached interfaces. Since the maximum number of ENIs for an instance is currently 50, this should be safe.

pkg/networkutils/network.go
```diff
@@ -338,8 +338,9 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDRs []string, primaryMAC string, pr
 			}})
 	}
 
-	// Prepare the Desired Rule for SNAT Rule
-	snatRule := []string{"-m", "comment", "--comment", "AWS, SNAT",
+	// Prepare the Desired Rule for SNAT Rule for non-pod ENIs
+	snatRule := []string{"!", "-o", "vlan+",
+		"-m", "comment", "--comment", "AWS, SNAT",
```
mogren, line 343: This is important, we can't SNAT traffic from pods with Branch ENIs.

test/cmd/packet-verifier/packet-verifier.go
```diff
@@ -0,0 +1,272 @@
...
+	device      string
+
+	// vlan ID to monitor on the interfaces
+	vlanIDToMonitor int
+
+	// pcap parameters (requires libpcap-devel to be installed on the host)
+	snapshot_len int32 = 1024
+	promiscuous        = false
+	timeout            = 30 * time.Second
+	handle       *pcap.Handle
```
anguslees, line 38: Can we move the non-flag variables into `main()`?

In particular `enis` `handle` are probably going to be mutated at runtime and would be easier to think about if they were more narrowly scoped.

test/cmd/packet-verifier/packet-verifier.go
```diff
@@ -0,0 +1,272 @@
...
+	var dstPacketsProcessed int
+
+	// Use the handle as a packet source to process all packets
+	packetSource := gopacket.NewPacketSource(handle, handle.LinkType())
+	for packet := range packetSource.Packets() {
+		fmt.Printf("packet: %v\n", packet)
+
+		network := packet.Layer(layers.LayerTypeIPv4)
+		if network != nil {
+			if iface.shouldCheckSrc && packet.NetworkLayer().NetworkFlow().Src().String() != ipAddress &&
```
anguslees, line 214: parse `ipAddress` earlier and use `net.IP.Equal()` here.  It will be (much) faster, and avoid non-canonical strings causing spurious differences.

(Same throughout this function)

test/cmd/packet-verifier/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+ENV GOPROXY=direct
+ENV GOOS=linux
+ENV GOARCH=amd64
+COPY . /
+WORKDIR /
+RUN yum install -y git golang libpcap-devel
+RUN go build -o packet-verifier packet-verifier.go
+ 
+FROM amazonlinux:2
+RUN yum install -y libpcap-devel
```
anguslees, line 14: Do we need all of `libpcap-devel` here, or will just the runtime libraries in `libpcap` do?

pkg/networkutils/network.go
anguslees: Right, what does this affect?  Does this mean we can't send traffic outside the immediate VPC from these pods? (that seems bad)


Aside: my quick search just turned up https://docs.aws.amazon.com/eks/latest/userguide/external-snat.html which says:
> SNAT is necessary because the internet gateway only knows how to translate between the primary private and public or elastic IP address assigned to the primary elastic network interface of the Amazon EC2 instance node that pods are running on.

Huh, I wouldn't have guessed this limitation, and I couldn't find more info in the VPC docs after a brief search.  Is this an IGW limitation, or some side effect of routes that we configure for secondary addresses within CNI?  (as in: how will this affect vlan interfaces, if they are configured with the vpc subnet gateway?)

SaranBalaji90: Ingress and Egress permission on the security group is applied to the traffic outside the instance, which means that we can't do any NAT within the instance for these pods. This will bypass the security group check.

If these pods uses private subnet which has NAT gateway associated then traffic will follow out to the internet. Otherwise it can talk only to resources within the VPC.

Makefile
```diff
@@ -240,7 +240,7 @@ lint:
 
 # Run go vet on source code.
 vet:
-	go vet ./...
+	go vet $(ALLPKGS)
```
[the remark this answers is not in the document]
SaranBalaji90: I was seeing the following errors in pcap.go, so going to follow up separately on this with upstream.

```
# github.com/google/gopacket/pcap
../../../../pkg/mod/github.com/google/gopacket@v1.1.17/pcap/pcap.go:30:22: undefined: pcapErrorNotActivated
../../../../pkg/mod/github.com/google/gopacket@v1.1.17/pcap/pcap.go:52:17: undefined: pcapTPtr
../../../../pkg/mod/github.com/google/gopacket@v1.1.17/pcap/pcap.go:64:10: undefined: pcapPkthdr
../../../../pkg/mod/github.com/google/gopacket@v1.1.17/pcap/pcap.go:102:7: undefined: pcapBpfProgram
../../../../pkg/mod/github.com/google/gopacket@v1.1.17/pcap/pcap.go:103:7: undefined: pcapPkthdr
../../../../pkg/mod/github.com/google/gopacket@v1.1.17/pcap/pcap.go:261:33: undefined: pcapErrorActivated
../../../../pkg/mod/github.com/google/gopacket@v1.1.17/pcap/pcap.go:262:33: undefined: pcapWarningPromisc
../../../../pkg/mod/github.com/google/gopacket@v1.1.17/pcap/pcap.go:263:33: undefined: pcapErrorNoSuchDevice
../../../../pkg/mod/github.com/google/gopacket@v1.1.17/pcap/pcap.go:264:33: undefined: pcapErrorDenied
../../../../pkg/mod/github.com/google/gopacket@v1.1.17/pcap/pcap.go:265:33: undefined: pcapErrorNotUp
```

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -34,6 +34,8 @@ import (
 )
 
 const (
+	// vlan rule priority
+	vlanRulePriority = 1
```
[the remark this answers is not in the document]
SaranBalaji90: Sounds good.

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -282,6 +252,151 @@ func setupNS(hostVethName string, contVethName string, netnsPath string, addr *n
...
+	if err != nil {
+		return errors.Wrapf(err, "SetupPodENINetwork failed to setup veth pair.")
+	}
+
+	vlanTableId := vlanId + 100
+	vlanLink := buildVlanLink(vlanId, parentIfIndex, eniMAC)
+
+	// 1. clean up if vlan already exists (necessary when trunk ENI changes).
+	if oldVlan, err := os.netLink.LinkByName(vlanLink.Name); err == nil {
+		if err = os.netLink.LinkDel(oldVlan); err != nil {
```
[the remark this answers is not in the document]
SaranBalaji90: If err != nil it will proceed further. Do you mean to say proceed only if err != nil and is`NotFound`? I can add but thinking it might not be necessary.

cmd/routed-eni-cni-plugin/driver/driver_test.go
```diff
@@ -459,3 +471,116 @@ func TestTearDownPodNetwork(t *testing.T) {
...
+	}
+
+	actualRule := &netlink.Rule{}
+	m.netlink.EXPECT().NewRule().Return(actualRule)
+
+	expectedRule := &netlink.Rule{
+		Priority: vlanRulePriority,
+		Table:    101,
+	}
+	gomock.InOrder(
```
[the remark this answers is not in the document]
SaranBalaji90: First one needs to happen for sure, Getting link by name. Link deletion and rule deletion can happen in parallel too. But this is the closest I could get to to catch this ordering issue in the code.

test/cmd/packet-verifier/packet-verifier.go
SaranBalaji90: Will update this, thanks for the suggestion.

test/cmd/packet-verifier/Dockerfile
SaranBalaji90: I initially tried with just libpcap but that didn't help. Had to pull in couple of dependency from libpcap-devel as well.

SaranBalaji90:
> lgtm, nice code.
> 
> Some minor comments, but nothing serious.

Thank you :)

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -356,6 +471,35 @@ func tearDownNS(addr *net.IPNet, table int, netLink netlinkwrapper.NetLink, log
...
+	vlanRule.Priority = vlanRulePriority
+
+	for {
+		if err := os.netLink.RuleDel(vlanRule); err != nil {
+			if !containsNoSuchRule(err) {
+				return errors.Wrapf(err, "TeardownPodENINetwork: failed to delete container rule for %d", vlanId)
+			}
+			break
+		}
+	}
```
mogren, line 499: This is slightly tricky and it would be nice to have a comment explaining how it works. Basically, we create a new route table for each vlan in` SetupPodENINetwork()`. The table only has two rules, defined in `buildRoutesForVlan()`. When we tear down, we want to delete both those rules. Calling `netLink.RuleDel` with only the table name and priority will delete the first rule matching that, that's why we have the for-loop.

pkg/networkutils/network.go
anguslees: Thanks, that makes sense.  Does that mean we could do the same (NAT at ngw) for secondary IPs, and remove the iptables snat?

SaranBalaji90: Yes we can remove if worker nodes subnets have NAT attached, then we don't have to do this SNAT at the host.

mogren: If they have a NAT Gateway, they can set the wonderfully named `AWS_VPC_K8S_CNI_EXTERNALSNAT=true`, and then `aws-node` won't set up any SNAT rules ([Through this tricky code...](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/networkutils/network.go#L314)).

Issue #989, opened 2020-05-21
Elevated rate of `Client.PrivateIpAddressLimitExceeded` errors after 1.5.x to 1.6.1 upgrade


uthark:
After we upgraded to 1.6.1 version, we see increased rate of `Client.PrivateIpAddressLimitExceeded` errors for `AssignPrivateIpAddresses` call.

Error message:

```
"errorCode": "Client.PrivateIpAddressLimitExceeded",
    "errorMessage": "Number of private addresses will exceed limit.",
    "requestParameters": {
        "networkInterfaceId": "eni-XXXX",
        "privateIpAddressesSet": {},
        "secondaryPrivateIpAddressCount": 29
    },
``` 

We use `m5.12xlarge` which allows to have 30 IP Addresses per interface.

<img width="572" alt="Screen Shot 2020-05-21 at 12 07 58 PM" src="https://user-images.githubusercontent.com/176118/82596196-c5bc4280-9b5b-11ea-8ecc-cf77cb3d983e.png">

mogren:
@uthark Interesting, thanks for reporting! Are you using `WARM_IP_TARGET` or is this when attaching a new ENI with 29 new secondary IPs? A log dump would be most helpful to figure out what the issue is.

uthark:
we don't use `WARM_IP_TARGET `. 
We use `WARM_ENI_TARGET` with value 2.
Looking for logs and check if those are new instances. will post update later.

uthark:
This is on new instances, please see screenshot with CloudTrail events.
<img width="752" alt="Screen Shot 2020-05-21 at 12 51 28 PM" src="https://user-images.githubusercontent.com/176118/82599834-cb1c8b80-9b61-11ea-86e8-6904bc46d415.png">

ipamd.log:

```
cat ipamd.log
{"level":"error","ts":"2020-05-21T18:36:13.918Z","caller":"ipamd/ipamd.go:354","msg":"ipt.NewChain error for chain [AWS-SNAT-CHAIN-0]: running [/usr/sbin/iptables -t nat -N AWS-SNAT-CHAIN-0 --wait]: exit status 4: iptables: Resource temporarily unavailable.\n"}
[...]

PR #1131, opened 2020-08-13
Wait for ENI and secondary IPs

mogren:
*Issue #, if available:*
#989, #1148 

*Description of changes:*

The log statements helping us find #989 were added earlier this year in #909 and #939, but the issue has probably been there longer. The problem is detailed in #1148, we need to wait for the Secondary IPv4 addresses to show up in the EC2 response. This PR does the following:

* Move `waitENIAttached()` from `ipamd.go` to `awsutils.go`
* Added an additional argument for how many secondary IPs to wait for
* Use retry with backoff instead of for-loop with sleep.
* If we find at least 1 secondary IP, but less than we wanted, we don't return an error, but count it as a partial success
* Added unit tests 

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -77,8 +77,11 @@ const (
 var (
 	// ErrENINotFound is an error when ENI is not found.
 	ErrENINotFound = errors.New("ENI is not found")
-	// ErrNoNetworkInterfaces occurs when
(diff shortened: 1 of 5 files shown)


nithu0115:
Thanks, @mogren. LGTM 👍

mogren:
Replaced by #1174

PR #1132, opened 2020-08-13
Update CodeCov config to be more reasonable

mogren:
*Description of changes:*
* Only check coverage if tests succeed
* Reduce required coverage to 75%
* Only show total coverage, not patch coverage.

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -3,12 +3,12 @@
 
 codecov:
   notify:
-    require_ci_to_pass: no
+    require_ci_to_pass: yes
 
 coverage:
   precision: 2
   round: down
-  range: "70...100"
+  range: "50...75"
 
   status:
     project:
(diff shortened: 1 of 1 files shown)


.github/.codecov.yml
```diff
@@ -3,12 +3,12 @@
...
 codecov:
   notify:
-    require_ci_to_pass: no
+    require_ci_to_pass: yes
 
 coverage:
   precision: 2
   round: down
-  range: "70...100"
+  range: "50...75"
```
sramabad1, line 12: Sorry am missing some context. Any reason why we reduced coverage range and what are the implications ?

mogren: This is range is just setting the [color of the range](https://docs.codecov.io/docs/coverage-configuration), not changing the approval in any way. With this change, any file with at least 75% coverage would be marked as green.

PR #1133, opened 2020-08-13
go 1.15 was just released. We should at least bump to 1.14

mogren:
*Description of changes:*
[go 1.15 was just released](https://golang.org/doc/go1.15). Only the [last two releases get security fixes](https://golang.org/doc/devel/release.html#policy), so we need to bump the version to at least 1.14.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ orbs:
 references:
   build_job: &build
     docker:
-      - image: circleci/golang:1.13-stretch
+      - image: circleci/golang:1.14-stretch
     working_directory: /go/src/github.com/{{ORG_NAME}}/{{REPO_NAME}}
     environment: &env
       GO111MODULE: "on"
@@ -54,7 +54,7 @@ jobs:
 
   integration_test:
     docker:
-      - image: circleci/golang:1.13-stretch
+      - image: circleci/golang:1.14-stretch
     working_directory: /go/src/github.com/{{ORG_NAME}}/{{REPO_NAME}}
     environment:
       <<: *env
@@ -85,7 +85,7 @@ jobs:
 
   performance_test:
     docker:
-      - image: circleci/golang:1.13-stretch
+      - image: circleci/golang:1.14-stretch
     working_directory: /go/src/github.com/{{ORG_NAME}}/{{REPO_NAME}}
     environment:
       <<: *env
@@ -118,7 +118,7 @@ jobs:
 
   kops_test:
     docker:
-      - image: circleci/golang:1.13-stretch
+      - image: circleci/golang:1.14-stretch
     working_directory: /go/src/github.com/{{ORG_NAME}}/{{REPO_NAME}}
     environment:
       <<: *env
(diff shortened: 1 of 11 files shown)


scripts/dockerfiles/Dockerfile.release
```diff
@@ -1,5 +1,5 @@
 ARG docker_arch
-ARG golang_image=golang:1.13-stretch
+ARG golang_image=golang:1.14-stretch
```
saiteja313, line 2: We pass `DOCKER_BUILD_FLAGS` (includes `golang_image`) as part of make rules - [here](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/Makefile#L115)

Do you think we need to cleanup hard coded value for `golang_image` here and use the one from build args?

Can be a GitHub issue for future fix.

.circleci/config.yml
```diff
@@ -8,7 +8,7 @@ orbs:
 references:
   build_job: &build
     docker:
-      - image: circleci/golang:1.13-stretch
+      - image: circleci/golang:1.14-stretch
```
saiteja313, line 11: I think we can use [setting-an-environment-variable-in-a-project](https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-project) feature in Circle CI, merge all redundant values for golang Docker image to one place.

This means single var for all jobs :D  any thoughts?

scripts/dockerfiles/Dockerfile.release
mogren: Thanks for catching this, will fix!

.circleci/config.yml
mogren: I think it is best to have it set in the Makefile, since these CircleCI images might not match, and we need the version set when we build the images locally during development.

scripts/dockerfiles/Dockerfile.release
mogren:
```
> make docker

docker build --build-arg GOARCH="amd64" --build-arg docker_arch="amd64" --build-arg golang_image="golang:1.14-stretch" --network=host  \
		-f scripts/dockerfiles/Dockerfile.release \
		-t "amazon/amazon-k8s-cni:v1.6.3-fix-script-v1-44-gbd140061" \
		.
Sending build context to Docker daemon  77.46MB
Step 1/18 : ARG docker_arch
Step 2/18 : ARG golang_image
Step 3/18 : FROM $golang_image as builder
 ---> ad99badbea6f
.
.
.
```

anguslees:
Just a note for general awareness: go1.14 makes [some runtime changes](https://golang.org/doc/go1.14#runtime) that result in more syscalls returning EINTR than previously.  If we see any errors/stacktraces mentioning this, then it's a sign that the underlying code is not retrying syscalls when it should and we should fix/update the relevant code (we might see this from our netlink library dependency, for example.)

(Kubernetes upstream had some challenges around this, hence increasing awareness.)

mogren:
Yes, I remember seeing this back in January, so we stayed off 1.14 back then. https://github.com/golang/go/issues/36752 is the go issue. Kubernetes switched to 1.14.4 o Jun 22 https://github.com/kubernetes/kubernetes/pull/88638/. That said 1.15 does look a lot better... 😄

PR #1134, opened 2020-08-13
Handle version mismatch between CNI and IPAMD

jayanthvn:
*Issue #, if available:

*Description of changes: During upgrades/downgrades we can have a situation where ipamd is upgraded but the CNI binary is still old. So to handle such race conditions, we will copy the CNI binary first and then wait for IPAMD to come up.

### DIFF
--- a/pkg/ipamd/rpc_handler.go
+++ b/pkg/ipamd/rpc_handler.go
@@ -120,6 +120,10 @@ func (s *server) AddNetwork(ctx context.Context, in *rpc.AddNetworkRequest) (*rp
 		}
 	}
 	if addr == "" {
+		if in.ContainerID == "" || in.IfName == "" || in.NetworkName == "" {
+			log.Errorf("Unable to generate IPAMKey from %+v", in)
+			return &failureResponse, nil
+		}
 		ipamKey := datastore.IPAMKey{
 			ContainerID: in.ContainerID,
 			IfName:      in.IfName,
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -78,6 +78,9 @@ if [[ "$AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER" != "false" ]]; then
     done
 fi
 
+log_in_json info "Install CNI binary.."
+install aws-cni "$HOST_CNI_BIN_PATH"
+
 log_in_json info "Starting IPAM daemon in the background ... "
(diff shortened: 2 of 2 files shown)


scripts/entrypoint.sh
```diff
@@ -91,7 +94,7 @@ fi
 
 log_in_json info "Copying CNI plugin binary and config file ... "
 
-install aws-cni "$HOST_CNI_BIN_PATH"
+#install aws-cni "$HOST_CNI_BIN_PATH"
```
sramabad1: I think you should remove this line. If you still have reasons to keep it, add some comment so that reader has context.

pkg/ipamd/rpc_handler.go
```diff
@@ -120,6 +120,10 @@ func (s *server) AddNetwork(ctx context.Context, in *rpc.AddNetworkRequest) (*rp
 		}
 	}
 	if addr == "" {
+		if in.ContainerID == "" {
+			log.Errorf("ContainerID is empty")
+			return &failureResponse, nil
+		}
```
sramabad1, line 126: With the code changes in  scripts/entrypoint.sh, i guess we wouldn't need this. If you still want to keep this sanity check , should we also account for validity of other fields that form the key ?

scripts/entrypoint.sh
jayanthvn: Removed it :)

pkg/ipamd/rpc_handler.go
jayanthvn: Handled the checks for all fields in the IPAMKey

pkg/ipamd/rpc_handler.go
```diff
@@ -120,6 +120,10 @@ func (s *server) AddNetwork(ctx context.Context, in *rpc.AddNetworkRequest) (*rp
 		}
 	}
 	if addr == "" {
+		if in.ContainerID == "" || in.IfName == "" || in.NetworkName == "" {
+			log.Errorf("One of the IPAMKey is empty containerID %s, IfName %s, NetworkName %s", in.ContainerID, in.IfName, in.NetworkName)
```
mogren, line 124: Nit: How about:
```go
log.Errorf("Unable to generate IPAMKey from %+v", in)
```

jayanthvn: Yeah this is better. Will update it :)

PR #1136, opened 2020-08-14
Logger

nithu0115:
*Issue #, if available:* N/A

*Description of changes:*
Logging fix. 

```
[root@ip-192-168-70-65 aws-routed-eni]# grep -i "L-IpamD" ipamd.log
{"level":"info","ts":"2020-08-14T03:47:46.179Z","caller":"aws-k8s-agent/main.go:30","msg":"Starting L-IPAMD v1.7.0-rc1-20-gc18641ac-dirty  ..."}
```

### DIFF
--- a/cmd/routed-eni-cni-plugin/cni.go
+++ b/cmd/routed-eni-cni-plugin/cni.go
@@ -91,7 +91,6 @@ func LoadNetConf(bytes []byte) (*NetConf, logger.Logger, error) {
 
 	//logConfig
 	logConfig := logger.Configuration{
-		BinaryName:  conf.Name,
 		LogLevel:    conf.PluginLogLevel,
 		LogLocation: conf.PluginLogFile,
 	}
--- a/pkg/utils/logger/logger.go
+++ b/pkg/utils/logger/logger.go
@@ -13,8 +13,6 @@
 
 package logger
 
-const ipamdBinaryName = "aws-k8s-agent"
-
 //Log is global variable so that log functions can be directly accessed
 var log Logger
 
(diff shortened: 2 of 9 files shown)


pkg/utils/logger/logger.go
```diff
@@ -13,6 +13,9 @@
 
 package logger
 
+// "aws-k8s-agent" string passed as an input by cmd/aws-k8s-agent/main.go.
+// This is then compared to load binary specific configuration in New() method
+// Do not make any changes to "aws-k8s-agent" string.
```
anguslees, line 18: This relies on humans reading this comment and maintaining consistency across the two files, so is a bit fragile :(

Can we make the compiler check this for us?  Maybe:

1. In this file, export the constant:  `const IpamdBinaryName = "aws-k8s-agent"`
2. In `cmd/aws-k8s-agent/main.go` do `const binaryName = logger.IpamdBinaryName`

I think this still deserves a comment like yours above on the const in aws-k8s-agent/main.go to indicate that this specific binaryname value is magical in `logger.New()`.

(If we _really_ wanted to clean this up, we should just change `New()` to take a more useful argument directly instead of implicitly via BinaryName, and then pass the ipamd-special logging config directly from ipamd calling code.)

nithu0115: Thanks for the feedback, @anguslees. I went ahead made modification to remove Binary name and pass logger config during binary initialization.

nithu0115:
(Note to reviewers: There is more clean up to do here. I will have to pass the logger context down which is instantiated in main function to other packages instead of calling logger.Get() method. As this is a bigger change, it want to leave it out from this release)

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -91,7 +91,6 @@ func LoadNetConf(bytes []byte) (*NetConf, logger.Logger, error) {
 
 	//logConfig
 	logConfig := logger.Configuration{
-		BinaryName:  conf.Name,
 		LogLevel:    conf.PluginLogLevel,
 		LogLocation: conf.PluginLogFile,
```
mogren, line 95: The plugin will get right path passed in through `PluginLogFile`

nithu0115: Yup, through aws-conf file :)

PR #1143, opened 2020-08-14
Update image tag to support multi-arch images

mogren:
*Description of changes:*
* Update image tag to use multi-arch manifest in order to enable arm64 images.

### DIFF
--- a/config/v1.6/aws-k8s-cni-cn.yaml
+++ b/config/v1.6/aws-k8s-cni-cn.yaml
@@ -77,6 +77,7 @@ spec:
                     operator: In
                     values:
                       - amd64
+                      - arm64
                   - key: "eks.amazonaws.com/compute-type"
                     operator: NotIn
                     values:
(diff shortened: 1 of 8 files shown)


config/v1.5/aws-k8s-cni-cn.yaml
```diff
@@ -77,6 +77,7 @@ spec:
                     operator: In
                     values:
                       - amd64
+                      - arm64
```
nprab428, line 80: Why are you updating to `arm64` in this file? Unless `961992271922.dkr.ecr.cn-northwest-1.amazonaws.com.cn/amazon-k8s-cni:v1.5.7` is actually multi-arch we shouldn't be doing this.

config/v1.5/aws-k8s-cni.yaml
```diff
@@ -77,6 +77,7 @@ spec:
                     operator: In
                     values:
                       - amd64
+                      - arm64
```
nprab428, line 80: Same comment as the one in `aws-k8s-cni-cn.yaml`

PR #1152, opened 2020-08-19
return only on error for tryAssignIPs method

nithu0115:
*Issue #, if available:* N/A

*Description of changes:*
When testing #903, for an edge case scenario where the EC2 instance cannot assign IP addresse when we call tryAssignIPs   during node init, which increasedPool == false and error == nil. Even though there was no error, we prematurely think this is returning error which was blocking spawning of go-routine which never updated ip rule table.   

With this change, we are explicitly checking if the error is not nil and then return error. 

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -447,7 +447,7 @@ func (c *IPAMContext) nodeInit() error {
 	increasedPool, err := c.tryAssignIPs()
 	if err == nil && increasedPool {
 		c.updateLastNodeIPPoolAction()
-	} else {
+	} else if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/ipamd/ipamd.go
```diff
@@ -447,7 +447,7 @@ func (c *IPAMContext) nodeInit() error {
 	increasedPool, err := c.tryAssignIPs()
 	if err == nil && increasedPool {
 		c.updateLastNodeIPPoolAction()
-	} else {
+	} else if err != nil {
 		return err
 	}
```
SaranBalaji90, line 452: styling nit: instead of checking both err == nil and err != nill

if err != nil { return err } 
else if increasedPool {..}

Issue #1159, opened 2020-08-22
Calico deletes routes when using CNI v1.7.0


kgtw:
Recently applied 1.7.0 manifests in one of our clusters and we are seeing new pods are having interfaces created with no ip address associated with them on the host.

```
$ kubectl get pod -o wide | grep debug
shell-node-debug     1/1     Running   0     6s     172.30.164.149   ip-172-30-165-219.eu-central-1.compute.internal   <none> 
 <none>
```
```
$ tail /var/log/aws-routed-eni/plugin.log
{"level":"info","ts":"2020-08-22T11:41:26.166Z","caller":"routed-eni-cni-plugin/cni.go:117","msg":"Received CNI add request: ContainerID(7864f0bbebd4eb32871d4a42414e7e216f4614203d59c3ec1d7fbbe532975064) Netns(/proc/17287/ns/net) IfName(eth0) Args(IgnoreUnknown=1;K8S_POD_NAMESPACE=kube-system;K8S_POD_NAME=shell-node-debug;K8S_POD_INFRA_CONTAINER_ID=7864f0bbebd4eb32871d4a42414e7e216f4614203d59c3ec1d7fbbe532975064) Path(/opt/cni/bin/) argsStdinData({\"cniVersion\":\"0.3.1\",\"mtu\":\"9001\",\"name\":\"aws-cni\",\"pluginLogFile\":\"/var/log/aws-routed-eni/plugin.log\",\"pluginLogLevel\":\"Debug\",\"type\":\"aws-cni\",\"vethPrefix\":\"eni\"})"}
{"level":"debug","ts":"2020-08-22T11:41:26.166Z","caller":"routed-eni-cni-plugin/cni.go:117","msg":"MTU value set is 9001:"}
{"level":"info","ts":"2020-08-22T11:41:26.168Z","caller":"routed-eni-cni-plugin/cni.go:117","msg":"Received add network response for container 7864f0bbebd4eb32871d4a42414e7e216f4614203d59c3ec1d7fbbe532975064 interface eth0: Success:true IPv4Addr:\"172.30.164.149\" UseExternalSNAT:true VPCcidrs:\"172.30.160.0/20\" "}
[...]

PR #1166, opened 2020-08-24
Revert veth name generation due to Calico dependency

jayanthvn:
*Issue #1159 

*Description of changes: In 1.7.0 `generateHostVethName` was updated in #972 but this isn't compatible with the value expected by calico. Hence calico used to identify the existing ENIs are unexpected routes and delete them.

New format eni name -> enibe59f971f52  
```
192.168.39.120  0.0.0.0         255.255.255.255 UH    0      0        0 enibe59f971f52
```
Calico generated eni -> eni8478779f379
```
2020-08-24 21:52:24.310 [INFO][38] int_dataplane.go 1034: Received *proto.WorkloadEndpointUpdate update from calculation graph msg=id:<orchestrator_id:"k8s" workload_id:"default/my-nginx-    86b7cfc89-6lr6v" endpoint_id:"eth0" > endpoint:<state:"active" name:"eni8478779f379" profile_ids:"kns.default" profile_ids:"ksa.default.default" ipv4_nets:"192.168.39.120/32" >

2020-08-24 21:52:25.441 [INFO][38] route_table.go 556: Syncing routes: found unexpected route; ignoring    due to grace period. dest=192.168.48.72/32 ifaceName="enic9f368fa7c4" ipVersion=0x4

2020-08-24 21:52:33.223 [INFO][38] route_table.go 561: Syncing routes: removing old route. dest=192.168.   48.72/32 ifaceName="enic9f368fa7c4" ipVersion=0x4 routeProblems=[]string{"unexpected route"}
```

Reverting the eni generation, fixed the issue.

```
192.168.47.127  0.0.0.0         255.255.255.255 UH    0      0        0 enib4f6d4780c5

2020-08-24 22:38:42.227 [INFO][39] int_dataplane.go 1034: Received *proto.WorkloadEndpointUpdate update from calculation graph msg=id:<orchestrator_id:"k8s" workload_id:"default/my-nginx-    86b7cfc89-4hmrr" endpoint_id:"eth0" > endpoint:<state:"active" name:"enib4f6d4780c5" profile_ids:"kns.default" profile_ids:"ksa.default.default" ipv4_nets:"192.168.47.127/32" >
```

Here is the mapping in calico -

https://github.com/projectcalico/felix/blob/bbb79439123dce009b43d0b99f0b03d651bc142a/fv/workload/workload.go#L96
https://github.com/projectcalico/libcalico-go/blob/677e62cafc19cb9e47aec345c5451bf265c8979f/lib/backend/k8s/conversion/workload_endpoint_default.go#L40

### DIFF
--- a/cmd/routed-eni-cni-plugin/cni.go
+++ b/cmd/routed-eni-cni-plugin/cni.go
@@ -182,14 +182,14 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 	}
 
 	if r.PodVlanId != 0 {
-		hostVethName := generateHostVethName("vlan", conf.Name, args.ContainerID, args.IfName)
+		hostVethName := generateHostVethName("vlan", string(k8sArgs.K8S_POD_NAMESPACE), string(k8sArgs.K8S_POD_NAME))
 
 		err = driverClient.SetupPodENINetwork(hostVethName, args.IfName, args.Netns, addr, int(r.PodVlanId), r.PodENIMAC,
 			r.PodENISubnetGW, int(r.ParentIfIndex), mtu, log)
 	} else {
 		// build hostVethName
 		// Note: the maximum length for linux interface name is 15
-		hostVethName := generateHostVethName(conf.VethPrefix, conf.Name, args.ContainerID, args.IfName)
+		hostVethName := generateHostVethName(conf.VethPrefix, string(k8sArgs.K8S_POD_NAMESPACE), string(k8sArgs.K8S_POD_NAME))
 
 		err = driverClient.SetupNS(hostVethName, args.IfName, args.Netns, addr, int(r.DeviceNumber), r.VPCcidrs, r.UseExternalSNAT, mtu, log)
 	}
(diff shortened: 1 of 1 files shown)


fawadkhaliq:
What's the cost of reverting this? Is there _any_ impact this change have on the VPC CNI? Any consistencies/incompatibility on clusters that have this change (different) already?

mogren:
@fawadkhaliq The AWS VPC CNI does not care about the interface name. The reason it was changed in the first place was because we changed the key we use to identify what pod has what IP from just being pod `name` and `namespace` to include the container ID as well. (In #972)

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -237,9 +237,11 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 }
 
 // generateHostVethName returns a name to be used on the host-side veth device.
-func generateHostVethName(prefix, netname, containerid, ifname string) string {
+// The veth name is generated such that it aligns with the value expected
+// by Calico for NetworkPolicy enforcement.
```
fawadkhaliq, line 241: Is there a standard being followed in this naming convention? If not, we should treat this a stop-gap solution with a TODO followup as Calico or any other plugins (that might integrate in similar fashion) may break without a warning

mogren: Definitely agree. This kind of hidden coupling is something we should really try to avoid.

Apparently kind has to do something similar, see https://github.com/kubernetes-sigs/kind/issues/842 and the code in  https://github.com/cybozu/neco-containers/blob/cddc3696f72fbc80580d088f338ef42d8f4ae7cf/kind-node/src/ptp/ptp.go#L414

Issue #1172, opened 2020-08-25
Bandwidth plugin is bundled, but not actually usable


ianhe8x:
I'm interested in using bandwidth plugin in our cluster since it has been bundled with the aws-vpc-cni, and tried to enable it manually.
Then an error occurred because aws-vpc-cni doesn't pass the pod's network interface to cni plugins.

mogren:
Closing this issue in favor of #463

jayanthvn:
Currently this is blocked on containerd/cri#1544

PR #1173, opened 2020-08-25
enable bandwidth plugin (#1172)

ianhe8x:
*Issue #1172

*Description of changes:*
* include Interfaces into the result so it can be used by bandwidth plugin
* enable bandwidth plugin by default


jayanthvn:
Hi @ianhe8x ,

Thanks for opening the PR but currently we cannot support this because the bandwidth plugin needs interface chaining but that will prevent pod creation with containerd since current expectation is to always populate eth0 in the result. Kindly check this #1092 and https://github.com/containerd/cri/issues/1544.

Thanks.

ianhe8x:
Thanks, good to know that. I think we will create a custom build anyway.

Issue #1148, opened 2020-08-17
Wait for SecondaryIPs in waitENIAttached()


mogren:
We have seen an increase of `PrivateIpAddressLimitExceeded` since we started logging these errors back in May. #989

The issue seems to be that `waitENIAttached()` only check for the newly created ENI to be attached, not that the Secondary IPs have actually been assigned to that ENI. Because of that, the ENI gets added to the ipamd data store with no IPs, and on the next reconcile we think is empty, try to call `AllocIPAddresses()` on the same ENI again, which fails with `PrivateIpAddressLimitExceeded` that we ignore, and then go on to add those existing IPs to the data store.

We should instead wait for the secondary IPs to actually show up before adding the ENI to the data store.

naveenb29:
what CNI version does this affect ?

mogren:
@naveenb29  All versions seem to have had this issue since before 1.0.0. We did not log this until May, just quietly ignoring. The issue is in this check:

https://github.com/aws/amazon-vpc-cni-k8s/blob/f2ad9bac9c805e1de85b6986e7c69e6153f8a43c/pkg/ipamd/ipamd.go#L787-L792

Just getting the ENI attached is not enough, we also need to get the secondary IPv4 addresses. Added in #1174 

One reason it was not found earlier is that in the first call to create a new ENI and attach it, we fail to add any new IPs to the datastore. When we 5s later try again, we find the ENI that the CNI think is empty, the AssignPrivateIPAddresses call fails since they are already attached. We ignore that error since we now have the IPs, and just add them to the datastore and make them available for pods.

mogren:
Fix merged

PR #1174, opened 2020-08-26
Wait for ENI and secondary IPs

mogren:
Issue #, if available:
#989, #1148 

*Description of changes:*

The log statements helping us find #989 were added earlier this year in #909 and #939, but the issue has probably been there longer. The problem is detailed in #1148, we need to wait for the Secondary IPv4 addresses to show up in the EC2 response. This PR does the following:

* Move `waitENIAttached()` from `ipamd.go` to `awsutils.go`
* Added an additional argument for how many secondary IPs to wait for
* Use retry with backoff instead of for-loop with sleep.
* If we find at least 1 secondary IP, but less than we wanted, we don't return an error, but count it as a partial success
* Added unit tests 

(Replacing #1131 with a new PR)

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -79,8 +79,11 @@ const (
 var (
 	// ErrENINotFound is an error when ENI is not found.
 	ErrENINotFound = errors.New("ENI is not found")
-	// ErrNoNetworkInterfaces occurs when
-	// DesribeNetworkInterfaces(eniID) returns no network interfaces
+	// ErrAllSecondaryIPsNotFound is returned when not all secondary IPs on an ENI have been assigned
+	ErrAllSecondaryIPsNotFound = errors.New("All secondary IPs not found")
+	// ErrNoSecondaryIPsFound is returned when not all secondary IPs on an ENI have been assigned
+	ErrNoSecondaryIPsFound = errors.New("No secondary IPs have been assigned to this ENI")
+	// ErrNoNetworkInterfaces occurs when DescribeNetworkInterfaces(eniID) returns no network interfaces
 	ErrNoNetworkInterfaces = errors.New("No network interfaces found for ENI")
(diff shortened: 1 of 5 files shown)


mogren:
Integration tests passed https://app.circleci.com/pipelines/github/mogren/amazon-vpc-cni-k8s/726/workflows/243c1da5-ab60-423f-8692-ba5edca888b4/jobs/1454

pkg/awsutils/awsutils.go
```diff
@@ -1305,12 +1311,13 @@ func (cache *EC2InstanceMetadataCache) AllocIPAddresses(eniID string, numIPs int
 	output, err := cache.ec2SVC.AssignPrivateIpAddressesWithContext(context.Background(), input, userAgent)
 	awsAPILatency.WithLabelValues("AssignPrivateIpAddresses", fmt.Sprint(err != nil)).Observe(msSince(start))
 	if err != nil {
-		log.Errorf("Failed to allocate a private IP addresses on ENI %v: %v", eniID, err)
-		awsAPIErrInc("AssignPrivateIpAddresses", err)
 		if containsPrivateIPAddressLimitExceededError(err) {
-			log.Debug("AssignPrivateIpAddresses returned PrivateIpAddressLimitExceeded")
+			log.Debug("AssignPrivateIpAddresses returned PrivateIpAddressLimitExceeded, probably because the data store was out of sync." +
```
jayanthvn, line 1315: This error is seen by a ec2 call and when the Cx runs out of Ip's right? Should we change this to something like "Check IP usage limits" ?

mogren: This doesn't happen when the customer runs out of IPs, but when we call `AssignPrivateIpAddresses` asking for more secondary IPs than the ENI can support. I'll reword this log message. Also, after this change we should not store "empty" ENIs in the datastore.

PR #1186, opened 2020-09-01
Mount /run/xtables.lock to prevent unwanted race conditions.

kgtw:
**What type of PR is this?**
bug

**Which issue(s) this PR fixes**:
https://github.com/aws/amazon-vpc-cni-k8s/issues/989#issuecomment-683392635

**What this PR does / why we need it**:
As initially mentioned in https://github.com/aws/amazon-vpc-cni-k8s/issues/989#issuecomment-683392635 we should mount `/run/xtables.lock` to prevent race conditions where multiple programs are manipulating iptables at the same time.

The `--wait` on iptables commands isn't actually respected unless we mount the iptables lock file in to the container.

The downside of this is that it introduces a blocking action where the cni will wait until the lock is released.

I'm not sure how the kubernetes manifests under `v1.x` are generated, can copy by hand if thats the case..

**If issue # is not available please add repro steps and logs from IPAMD/CNI showing the issue**:

**Testing**:
Running `flock -e /run/xtables.lock sleep 180` on the host before starting the container you will observe that the cni will wait 3 minutes until the lock is released. In theory this could be indefinitely according to the iptables manual.

Trying to replicate the exact bug I mentioned in the comment is extremely difficult because you have to be updating iptables at the exact same time as the cni is. We have only seen this bug observed in our production environments a handful of times so far.

**Automation added to e2e**:
n/a

**Will this break upgrades and downgrades. Has it been tested?**:
Will not break upgrades/downgrades.

**Does this require config only upgrades?**:
yes?

**Does this PR introduce a user-facing change?**:
yes

```release-note
Users are recommended to mount `/run/xtables.lock` to prevent unwanted race conditions.
```

### DIFF
--- a/config/master/aws-k8s-cni-cn.yaml
+++ b/config/master/aws-k8s-cni-cn.yaml
@@ -189,6 +189,8 @@
           "name": "run-dir"
         - "mountPath": "/var/run/dockershim.sock"
           "name": "dockershim"
+        - "mountPath": "/run/xtables.lock"
+          "name": "xtables-lock"
       "hostNetwork": true
       "initContainers":
       - "image": "961992271922.dkr.ecr.cn-northwest-1.amazonaws.com.cn/amazon-k8s-cni-init:latest"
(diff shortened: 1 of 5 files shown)


anguslees (approved):
(Thanks!)

SaranBalaji90:
Thanks @kgtw for the fix. This will help closing some of the issues. @mogren or @jayanthvn we might have to backport this to 1.7 release branch.

PR #1191, opened 2020-09-03
Update faulty comment in eni-max-pods.txt

mogren:
**What type of PR is this?**
Documentation

**Which issue(s) this PR fixes**:
The current `eni-max-pods.txt` has a comment that says:
```
# * 2 additional host-networking pods (AWS ENI and kube-proxy) are accounted for
```
It should say CNI
```
# * +2 for for the pods that use host-networking (AWS CNI and kube-proxy)
```
Also removed some instance types (`c5ad`) that are available in the EC2 API.

**What this PR does / why we need it**:
The documentation is currently wrong.

**Automation added to e2e**:
Not needed

**Will this break upgrades and downgrades. Has it been tested?**:
No

**Does this require config upgrades?**:
no

**Does this PR introduce a user-facing change?**:
Just clearer documentation


jayanthvn (approved):
LGTM :)

PR #1205, opened 2020-09-09
No step to update since we removed the /ok-to-test

mogren:
**What type of PR is this?**

cleanup

A manual workflow issue

We don't have the /ok-to-test command, so no workflow to update.

https://github.com/aws/amazon-vpc-cni-k8s/runs/1092480047?check_suite_focus=true

Manual test run change only

**Automation added to e2e**:
No

No

No

No


haouc (approved):
Looks good to me. Thanks.

PR #1208, opened 2020-09-11
Seems the rule got merged twice

mogren:
**What type of PR is this?**
cleanup

None

When the vlan changes were merged, this rule got duplicated. Not causing any issues, but 

Just duplicate code

Unit tests only

**Automation added to e2e**:
None

No

No

No


jayanthvn (approved):
LGTM :)

PR #1209, opened 2020-09-12
Fix logging texts for ENI cleanup

mogren:
**What type of PR is this?**
cleanup

None

The log lines when we are doing ENI cleanup don't look right.

Check logs when 

**Automation added to e2e**:

No

No

None


haouc (approved):
looks good to me.

PR #560, opened 2019-08-02
Restructure existing code to work with latest version of operator-sdk

adammw:
**WORK IN PROGRESS**

*Issue #:* #522

*Description of changes:* Restructure existing code to work with latest version of operator-sdk. 

Following the recommendation of the [operator-sdk migration guide](https://github.com/operator-framework/operator-sdk/blob/master/doc/migration/v0.1.0-migration-guide.md), I created a new operator-sdk project and copied as much of the existing code as possible. Most changes are around the controller logic, as the operator-sdk API has changed from combined Handle() methods which take change events, to Reconcile() which take reconciliation requests of a single object instead.

*Open Questions*
- Will this code be accepted for a new minor release?
- Should the log code be left alone with the built-in seelog-based utilities or migrated to `sigs.k8s.io/controller-runtime/pkg/log`

*TODO*
- [x] Fix/update tests
- [ ] Set correct log-levels to reduce log spam
- [ ] Increase reconciliation period, 5 seconds seems awfully low

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1214, opened 2020-09-14
avoid 6 api rpm per pod by removing operator framework cache-reset

grosser:
we see requests to all api endpoints every minute coming from the k8sclient.GetKubeClient being hardcoded to reset every minute [here](https://github.com/operator-framework/operator-sdk/blob/v0.0.7/pkg/k8sclient/client.go#L66)

not calling operator sdk did not help (doing this)
```
	kubeConfig, err := rest.InClusterConfig()
	if err != nil {
		panic(err)
	}
	kubeClient := kubernetes.NewForConfigOrDie(kubeConfig)
```
so I had to disable the cache-reset in operator-sdk

no more api spam now 🎉 

Screen Shot 2020-09-14 at 5 29 26 PM

so I'd suggest either using that branch (kinda shady) or updating operator sdk (bunch of work I guess) or getting rid of operator-sdk (also lots of work) 😞

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -48,3 +48,5 @@ require (
 	k8s.io/cri-api v0.0.0-20191107035106-03d130a7dc28
 	k8s.io/kube-openapi v0.0.0-20190510232812-a01b7d5d6c22 // indirect
 )
+
+replace github.com/operator-framework/operator-sdk => github.com/grosser/operator-sdk v0.0.8-0.20200914235856-4b738c6ed745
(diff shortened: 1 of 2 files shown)


grosser:
the first change did not work, but the second change works, no more api requests

Screen Shot 2020-09-14 at 5 29 26 PM

so I'd suggest either using that branch (kinda shady) or updating operator sdk (bunch of work I guess) or getting rid of operator-sdk (also lots of work) 😞

mogren:
@grosser Thanks a lot for looking at this. We are working fixing the issues related to Custom Networking. There are multiple issues directly related to the old version of the operator framework that we have. (Related issue in https://github.com/aws/containers-roadmap/issues/867.)

Like you said, migrating from [0.0.7 to something newer isn't trivial](https://github.com/operator-framework/operator-sdk/blob/2ad2ee217424fcfde9bb4cfba00d845f988ec511/doc/migration/v0.1.0-migration-guide.md), but given that [v1.0.0](https://github.com/operator-framework/operator-sdk/releases/tag/v1.0.0)  was just released, it's time to do something about this.

grosser:
idk how stable newly baked 1.0 will be though :D
... I did not dig deep enough to understand why operator-sdk is needed, so my naive suggestion would be to try and replace it with native code bit-by-bit and get rid of that big dependency

mogren:
@grosser The operator-framework is used for the [Custom Networking to work with the ENIConfig CRD](https://www.eksworkshop.com/beginner/160_advanced-networking/secondary_cidr/prerequisites/). The last attempt to replace it was this PR https://github.com/aws/amazon-vpc-cni-k8s/pull/560, but as you can see, replacing it isn't trivial. Also, I agree that a `v1.0.0` is probably not something we want to depend on, but I suspect that there will be a release or two before custom networking has been rewritten. Alternatively, we would go with [v0.19.3](https://github.com/operator-framework/operator-sdk/releases/tag/v0.19.3) since that is an even more recent release.

jayanthvn:
With CNI 1.8, we have removed the dependency on operator-sdk - https://github.com/aws/amazon-vpc-cni-k8s/pull/1419. I think we don't need this PR anymore.

PR #1225, opened 2020-09-22
Refactor EC2 Metadata (IMDS) code into new typed accessor library

anguslees:
Refactor EC2 Metadata code into a typed wrapper.  The idea is to split out the low-level mechanism (imds.go) from clever caching or other logic (awsutils.go).

Reviewer: This is intended to be a no-op "obviously safe" refactor of metadata lookup code.  If there's something that looks surprising or unclear, I'm happy to simplify it further.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1226, opened 2020-09-22
Disable codecov on patch

mogren:
**What type of PR is this?**
cleanup

Even a [PR with 84% patch test coverage](https://github.com/aws/amazon-vpc-cni-k8s/pull/1225/files) is so full of CodeCov warnings that it is hard to read. 

**Why do we need it**:
Just renaming an existing function or variable makes CodeCov treat it as a patch missing coverage and fail the check. We want to have all checks green for PRs to be merged. Some other projects having the same issue: argoproj/argo-cd#1926,  mozilla-mobile/firefox-tv#778, kata-containers/tests#508.

Open a PR, look at the file changes.

None

**Automation added to e2e**:
None

No

No

No

### DIFF
--- a/.github/.codecov.yml
+++ b/.github/.codecov.yml
@@ -19,6 +19,11 @@ coverage:
         only_pulls: true
         flags:
           - "unittest"
+    # Disable patch since it is noisy and not correct
+    patch:
+      default:
+        enabled: no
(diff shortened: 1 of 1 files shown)


haouc (approved):
lgtm

Issue #1241, opened 2020-10-01
aws-vpc-cni-init 1.7.3 Init:CrashLoopBackOff due to sysctl: cannot stat /proc/sys/net/ipv4/tcp_early_demux: No such file or directory


mickael-ange:
**What happened**:

I'm trying to upgrade aws-vpc-cni to 1.7.3 on an AWS EKS cluster version 1.17 with CentOS 7 self-managed nodes. My goal is to  to use [security groups for pods](https://docs.aws.amazon.com/eks/latest/userguide/security-groups-for-pods.html).

However `aws-vpc-cni-init` v1.7.3 does `Init:CrashLoopBackOff` due to `sysctl: cannot stat /proc/sys/net/ipv4/tcp_early_demux: No such file or directory`.

```
kubectl logs -n kube-system aws-node-zr8hs aws-vpc-cni-init
+ PLUGIN_BINS='loopback portmap bandwidth aws-cni-support.sh'
+ for b in '$PLUGIN_BINS'
+ '[' '!' -f loopback ']'
+ for b in '$PLUGIN_BINS'
+ '[' '!' -f portmap ']'
+ for b in '$PLUGIN_BINS'
+ '[' '!' -f bandwidth ']'
+ for b in '$PLUGIN_BINS'
+ '[' '!' -f aws-cni-support.sh ']'
+ HOST_CNI_BIN_PATH=/host/opt/cni/bin
+ echo 'Copying CNI plugin binaries ... '
Copying CNI plugin binaries ... 
+ for b in '$PLUGIN_BINS'
+ install loopback /host/opt/cni/bin
+ for b in '$PLUGIN_BINS'
+ install portmap /host/opt/cni/bin
+ for b in '$PLUGIN_BINS'
+ install bandwidth /host/opt/cni/bin
+ for b in '$PLUGIN_BINS'
+ install aws-cni-support.sh /host/opt/cni/bin
Configure rp_filter loose... 
+ echo 'Configure rp_filter loose... '
++ curl -X PUT http://169.254.169.254/latest/api/token -H 'X-aws-ec2-metadata-token-ttl-seconds: 60'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    56  100    56    0     0  56000      0 --:--:-- --:--:-- --:--:-- 56000
[...]

PR #1242, opened 2020-10-01
Ignore error on enabling TCP early demux for old kernels

mogren:
**What type of PR is this?**
bug

#1241

For old kernels that do not have the `tcp_early_demux` kernel flag, the init container throws an error:
```
[centos@ip-192-168-9-194 ~]$ sysctl -w "net.ipv4.tcp_early_demux=1"
sysctl: cannot stat /proc/sys/net/ipv4/tcp_early_demux: No such file or directory
```
We need to ignore this warning:
```
[centos@ip-192-168-9-194 ~]$ sysctl -e -w "net.ipv4.tcp_early_demux=1"
```
From https://linux.die.net/man/8/sysctl: 
-e  Use this option to ignore errors about unknown keys.

The "early demux" feature was added in kernel 3.6, but the flag to disable it was not added until 4.6. This means that TCP health checks for pods using per-pod security groups will not work until at least kernel 4.6. UDP or exec checks should still work for pods with ENIs.

Start a CentOS image, run `sysctl -w "net.ipv4.tcp_early_demux=1"`

See above

**Automation added to e2e**:
None

No

No

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -431,10 +431,11 @@ Type: Boolean as a String
 
 Default: `false`
 
-If `ENABLE_POD_ENI` is set to `true`, in order for the kubelet on the node to talk to pods using the per pod security group feature,
-`DISABLE_TCP_EARLY_DEMUX` should be set to `true`. This will increase the local TCP connection latency slightly, that is why it is not
- on by default. Details on why this is needed can be found in this [#1212 comment](https://github.com/aws/amazon-vpc-cni-k8s/pull/1212#issuecomment-693540666).
-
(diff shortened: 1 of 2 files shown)


jayanthvn (approved):
https://linux.die.net/man/8/sysctl (-e)option

Looks good :)

jayanthvn:
I was discussing with Claes, should we instead throw a warning something like - "Current kernel version doesn't support  disabling tcp v4 early demux and the issue #1212 will exists".

SaranBalaji90 (approved):

PR #1247, opened 2020-10-02
Match primary ENI IP correctly

### LINKED ISSUE #1246: init.sh script fails when grep for  primary IP returns multiple matches
**What happened**:

The grep in the code https://github.com/aws/amazon-vpc-cni-k8s/blob/59cd4ae07063caf063da52b91c65b379fc99bd86/scripts/init.sh#L28 could potentially return more than one match, causing the script to fail.
Here is an example when it fails - 
```
eth0: 192.168.0.10
eth1: 192.168.0.100
```
Failure occurs during upgrade.
Upgrade to amazon-k8s-cni:v1.7.3 is succcessful

Use the IP addresses from the description.

**Environment**:
- Kubernetes version (use `kubectl version`): any
- CNI Version: on upgrade to amazon-k8s-cni:v1.7.3 under right conditions
- OS (e.g: `cat /etc/os-release`): any
- Kernel (e.g. `uname -a`): any

mogren:
**What type of PR is this?**
bug

Fixes #1246

For the example in the issue:
```
eth0: 192.168.0.10
eth1: 192.168.0.100
```
Trying to grep for `192.168.0.10` will match both. We need to add the trailing `/` to the `grep` argument. This won't happen on new nodes, only when updating running nodes.

Inside a pod with multiple ENIs attached:
```
bash-4.2# ip -4 -o a
1: lo    inet 127.0.0.1/8 scope host lo\       valid_lft forever preferred_lft forever
2: eth0    inet 192.168.116.5/19 brd 192.168.127.255 scope global dynamic eth0\       valid_lft 1924sec preferred_lft 1924sec
5: eth1    inet 192.168.118.65/19 brd 192.168.127.255 scope global eth1\       valid_lft forever preferred_lft forever
6: eth2    inet 192.168.100.237/19 brd 192.168.127.255 scope global eth2\       valid_lft forever preferred_lft forever
bash-4.2# HOST_IP=192.168.11...
bash-4.2# ip -4 -o a | grep "$HOST_IP" | awk '{print $2}'
eth0
eth1
bash-4.2# ip -4 -o a | grep "$HOST_IP/" | awk '{print $2}'
eth0
```

**Automation added to e2e**:
None

No, fix updates when multiple ENIs with only a trailing number in the IP is different.

No

No

### DIFF
--- a/scripts/init.sh
+++ b/scripts/init.sh
@@ -25,7 +25,7 @@ done
 echo "Configure rp_filter loose... "
 TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 60")
 HOST_IP=$(curl -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/local-ipv4)
-PRIMARY_IF=$(ip -4 -o a | grep "$HOST_IP" | awk '{print $2}')
+PRIMARY_IF=$(ip -4 -o a | grep "$HOST_IP/" | awk '{print $2}')
 sysctl -w "net.ipv4.conf.$PRIMARY_IF.rp_filter=2"
 cat "/proc/sys/net/ipv4/conf/$PRIMARY_IF/rp_filter"
 


kishorj (approved):

jayanthvn (approved):

scripts/init.sh
```diff
@@ -25,7 +25,7 @@ done
 echo "Configure rp_filter loose... "
 TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 60")
 HOST_IP=$(curl -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/local-ipv4)
-PRIMARY_IF=$(ip -4 -o a | grep "$HOST_IP" | awk '{print $2}')
+PRIMARY_IF=$(ip -4 -o a | grep "$HOST_IP/" | awk '{print $2}')
```
anguslees, line 28: I think there are still plenty of other cases where this regex match could go wrong, particularly because `.` is a meta character:

Eg: `grep 10.1.1.1/` will match `42: dummy0 inet 10.110.1.121/32` - which is clearly not the same address.

I suggest `grep -wF "$HOST_IP"`  ... or better yet, match of the layer2 address instead:
```
PRIMARY_MAC=$(curl -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/mac)
PRIMARY_IF=$(ip -o link show | grep -F "link/ether $PRIMARY_MAC" | awk -F'[ :]+' '{print $2}')
```

Issue #1124, opened 2020-08-07
Default security group attached when no others declared


edward2a:
Following documentation here:
- https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html

This leads to believe that the SG behaviour is consistent in the sense that, if no SG's are declared when adding subnets, the same SG's will be used as the primary interface.
The actual behaviour is that ENIs created in the added subnet will have attachment to the VPC's default SG instead.
I believe this should not be the case and rather, by default, use the same SG's as the primary interface.

```
apiVersion: crd.k8s.amazonaws.com/v1alpha1
kind: ENIConfig
metadata: 
  name: us-west-2a
spec: 
  subnet: subnet-011b111c1f11fdf11
```
The code above results in ENIs with VPC's default SG.
Applying this configuration should either 'use the primary iface SGs' or 'fail to apply'.

Thank you,
Ed

jayanthvn:
Hi @edward2a 

This is handled in #1259. Hence closing this issue.

Thank you!

PR #1259, opened 2020-10-12
Use primary ENI SGs if SG is null for Custom networking

jayanthvn:
**What type of PR is this?**
Bug

#1124 

If SG is not provided in the ENIConfig then use the primary ENI SGs.

Created ENI config -
```
cat <<EOF  | kubectl apply -f -
kind: ENIConfig
metadata:
 name: $AZ1
spec:
  subnet: $CUST_SNET1
EOF
```

Before fix SG list is empty-
```
{"level":"info","ts":"2020-10-12T21:11:32.781Z","caller":"awsutils/awsutils.go:610","msg":"Using a custom network config for the new ENI"}
{"level":"info","ts":"2020-10-12T21:11:32.781Z","caller":"awsutils/awsutils.go:610","msg":"Creating ENI with security groups: [] in subnet: subnet-0ac93d8034479e1cf"}
{"level":"info","ts":"2020-10-12T21:11:33.157Z","caller":"awsutils/awsutils.go:610","msg":"Created a new ENI: eni-014407d03b4c13b5f"}
```
Terminated the nodes and new nodes after fix -

```
{"level":"info","ts":"2020-10-12T20:55:44.350Z","caller":"ipamd/ipamd.go:638","msg":"ipamd: using custom network config: [], subnet-0ac93d8034479e1cf"}
{"level":"info","ts":"2020-10-12T20:55:44.350Z","caller":"awsutils/awsutils.go:610","msg":"Using a custom network config for the new ENI"}
{"level":"info","ts":"2020-10-12T20:55:44.350Z","caller":"awsutils/awsutils.go:610","msg":"Creating ENI with security groups: [sg-04b933767a4f1d5a0 sg-0c23b7680a45ebfa0] in subnet: subnet-0ac93d8034479e1cf"}
{"level":"info","ts":"2020-10-12T20:55:44.833Z","caller":"awsutils/awsutils.go:610","msg":"Created a new ENI: eni-04344bd43a8d9af7d"}
```

**Automation added to e2e**:

Upgrades and downgrades should not be impacted. One thing to call out, eni should be deleted and added back for the change to take into effect like terminating the node. Just upgrade will find the existing ENI and re-use it so SG won't be updated.

No

No

```release-note

```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -700,7 +700,11 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
 
 	if useCustomCfg {
 		log.Info("Using a custom network config for the new ENI")
-		input.Groups = sg
+		if len(sg) != 0 {
+			input.Groups = sg
+		} else {
+			log.Warnf("No custom networking security group found, will use the node's primary ENI's SG: %s", input.Groups)
+		}
 		input.SubnetId = aws.String(subnet)
 	} else {
 		log.Info("Using same config as the primary interface for the new ENI")


pkg/awsutils/awsutils.go
```diff
@@ -700,7 +700,9 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
 
 	if useCustomCfg {
 		log.Info("Using a custom network config for the new ENI")
-		input.Groups = sg
+		if len(sg) != 0 {
+			input.Groups = sg
+		}
```
mogren, line 705: I think we should log a warning (or maybe just Info?) here, something like:
```
} else {
	log.Warnf("No custom networking security group found, will use same as the node: %s", input.Groups)
}
```

jayanthvn: Yeah, I initially thought of adding but since there was a log below which lists the SGs so I didn't do it. Will add a warning, it will be more clear when logs are checked. Thank you.

mogren (approved):
It needs a rebase, but otherwise it looks good. :shipit:

pkg/awsutils/awsutils.go
```diff
@@ -700,7 +700,11 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
 
 	if useCustomCfg {
 		log.Info("Using a custom network config for the new ENI")
-		input.Groups = sg
+		if len(sg) != 0 {
+			input.Groups = sg
+		} else {
+			log.Warnf("No custom networking security group found, will use same as the node: %s", input.Groups)
```
nithu0115, line 706: @jayanthvn, could you please change this wording to: "No custom networking security group found, will use the node's primary ENI's SGs:"

nithu0115 (approved):
/lgtm. Requesting minor rewording.

Issue #1265, opened 2020-10-19
New pods failing to start with `FailedCreatePodSandBox` warning for CNI versions 1.7.x with Cilium


YesemKebede:
**What happened**:

New pods started failing to come up after upgrading to eks CNI v1.7.0 from v1.6.0.  I was able to upgrade to v1.6.3 without any issue. I started to see the errors when I upgraded to 1.7.0. I also tried to upgrade to other version ( v1.7.2 and v1.7.5) but I am seeing the same issue. 

Here is the error I am seeing.
```
 Warning  FailedCreatePodSandBox  28s                 kubelet            Failed to create pod sandbox: rpc error: code = Unknown desc = failed to set up sandbox container "7e3423d27fc6f36276de03aa7f41ef6b6f02121f800b65b64b8073c6a207b696" network for pod "spinnaker-get-resource-type-3fc73e4e3611d9f4-ps4b7": networkPlugin cni failed to set up pod "spinnaker-get-resource-type-3fc73e4e3611d9f4-ps4b7_default" network: invalidcharacter '{' after top-level value

``` 
Here is the [cni log](https://github.com/aws/amazon-vpc-cni-k8s/files/5403389/eks_i-06f4c3b1d6a6a3b1a_2020-10-19_1616-UTC_0.6.2.tar.gz)

**Anything else we need to know?**:
- We have Cilium running in chaining mode (v1.8.4) 

**Environment**:
- Kubernetes version :`v1.17.9-eks-4c6976`
- CNI Version: Tried different versions but seeing same issue for (1.7.0, 1.7.2, 1.7.5)
- Kernel: `5.4.58-27.104.amzn2.x86_64`

jayanthvn:
Hi @YesemKebede 

Can you please confirm if you have set `AWS_VPC_K8S_PLUGIN_LOG_FILE` to stdout?

I checked IPAMD logs and I see IP allocation seems fine on the first look. We will further investigate the issue.

Thanks.

YesemKebede:
@jayanthvn  `AWS_VPC_K8S_PLUGIN_LOG_FILE` is set to `/var/log/aws-routed-eni/plugin.log`

jayanthvn:
Thanks @YesemKebede . We will look into it asap.

jayanthvn:
Hi @YesemKebede 

Can also please confirm how you upgraded from 1.6.3 to 1.7.X?

Thank you!

[...]

PR #1275, opened 2020-11-02
Output to stderr when no log file path is passed

couralex6:
**What type of PR is this?**
Bug fix

#1265 

Customers using Cilium in chaining mode were unable to create new pods after upgrading to CNI v1.7.x. Because no log file path was passed, zaplogger.go would default to stdout. 
This change ensures that logs are being written to stderr when no log file path is passed.

This was tested on a cluster with [Cilium enabled in chaining mode ](https://docs.cilium.io/en/v1.9.0-rc2/gettingstarted/cni-chaining-aws-cni/). I reproduced the bug by installing CNI v1.7.5, which prevented new pods from being created. Pods were able to come up with this fix.

**Automation added to e2e**:
Added Unit Tests to test the following scenarios:
* Empty log file path > log to `zapcore.Lock(os.Stderr)`
* Actual file path > log to file path
* String set to 'stdout' > log to `zapcore.Lock(os.Stdout)`

No

No

No

### DIFF
--- a/pkg/utils/logger/logger_test.go
+++ b/pkg/utils/logger/logger_test.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"go.uber.org/zap/zapcore"
+	"gopkg.in/natefinch/lumberjack.v2"
 )
 
 func TestEnvLogFilePath(t *testing.T) {
@@ -57,4 +58,28 @@ func TestLogLevelReturnsDefaultLevelWhenEnvSetToInvalidValue(t *testing.T) {
 	inputLogLevel := GetLogLevel()
 	expectedLogLevel = zapcore.DebugLevel
 	assert.Equal(t, expectedLogLevel, getZapLevel(inputLogLevel))
+}
+
+func TestGetPluginLogFilePathEmpty(t *testing.T) {
+	expectedWriter := zapcore.Lock(os.Stderr)
+	inputPluginLogFile := ""
+	assert.Equal(t, expectedWriter, getPluginLogFilePath(inputPluginLogFile))
+}
+
+func TestGetPluginLogFilePathStdout(t *testing.T) {
+	expectedWriter := zapcore.Lock(os.Stdout)
+	inputPluginLogFile := "stdout"
+	assert.Equal(t, expectedWriter, getPluginLogFilePath(inputPluginLogFile))
+}
+
+func TestGetPluginLogFilePath(t *testing.T) {
+	inputPluginLogFile := "/var/log/aws-routed-eni/plugin.log"
+	expectedLumberJackLogger := &lumberjack.Logger{
+		Filename:   "/var/log/aws-routed-eni/plugin.log",
+		MaxSize:    100,
+		MaxBackups: 5,
+		MaxAge:     30,
+		Compress:   true,
+	}
+	assert.Equal(t, zapcore.AddSync(expectedLumberJackLogger), getPluginLogFilePath(inputPluginLogFile))
 }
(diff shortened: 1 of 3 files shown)


pkg/utils/logger/zaplogger.go
```diff
@@ -111,7 +111,9 @@ func (logConfig *Configuration) newZapLogger() *structuredLogger {
 
 	logFilePath := logConfig.LogLocation
 
-	if logFilePath != "" && strings.ToLower(logFilePath) != "stdout" {
+	if logFilePath != "" {
```
jayanthvn, line 114: This will just fix cases where log file path is empty but we still need to fix for cases with chaining like cilium and cni is `aws cni`, logFilePath should be set to `AWS_VPC_K8S_PLUGIN_LOG_FILE `

nithu0115 (approved):
Thanks for the fix @couralex6, LGTM 🍔

pkg/utils/logger/zaplogger.go
jayanthvn: @couralex6 - Please attach UT cases.

nithu0115 (changes requested):
@couralex6, could you please update documentation under `AWS_VPC_K8S_PLUGIN_LOG_FILE` env var, saying that if external plugins like Cilium and chained together, the conf file does not include pluginLogfile variable to pass in via stdin when CNI plugin is called by kubelet. This will result in no log files under `/var/log/aws-routed-eni/plugin.log`. However, result of CNI plugin execution can be seen in Kubelet logs as CNI plugin by default write output of `cmdAdd` to os.Stdout. (something along these lines for justification)

jayanthvn:
> @couralex6, could you please update documentation under `AWS_VPC_K8S_PLUGIN_LOG_FILE` env var, saying that if external plugins like Cilium and chained together, the conf file does not include pluginLogfile variable to pass in via stdin when CNI plugin is called by kubelet. This will result in no log files under `/var/log/aws-routed-eni/plugin.log`. However, result of CNI plugin execution can be seen in Kubelet logs as CNI plugin by default write output of `cmdAdd` to os.Stdout. (something along these lines for justification)

--> default to `os.Stderr`

pkg/utils/logger/logger_test.go
```diff
@@ -58,3 +59,27 @@ func TestLogLevelReturnsDefaultLevelWhenEnvSetToInvalidValue(t *testing.T) {
...
+	inputPluginLogFile := "/var/log/aws-routed-eni/plugin.log"
+	expectedLumberJackLogger := &lumberjack.Logger{
+		Filename:   "/var/log/aws-routed-eni/plugin.log",
+		MaxSize:    100,
+		MaxBackups: 5,
+		MaxAge:     30,
+		Compress:   true,
+	}
+	assert.Equal(t, zapcore.AddSync(expectedLumberJackLogger), getPluginLogFilePath(inputPluginLogFile))
+}
```
nithu0115, line 85: 👍

nithu0115 (approved):
Nice, @couralex6 /lgtm.

jayanthvn (approved):
LGTM :)

shaikatz:
Hi, when this PR is going to be released?

jayanthvn:
Hi @shaikatz, We are planning for the next release, I will provide you the dates in a week or so.

AmitBenAmi:
Hi @jayanthvn 
There have been already 2 releases since 1.7.6, but none of them fixed this issue.
Is this fix expected to be on `v1.7.9` or any other release expected to be released soon?

jayanthvn:
Hi @AmitBenAmi 

Sorry for the delay. Yes we are planning for a release In Jan/Feb. This will include this PR. Will update the exact date asap.

Thanks!

yaron-idan:
Hey @jayanthvn, is there any update on the release date for this PR? We are still blocked from upgrading our CNI.

fawadkhaliq:
@yaron-idan the fix will be included in the next patch release (1.7.9) and sorry, we don't share release dates publicly. Expect the patch release to be out in the coming days (very soon)

Issue #1272, opened 2020-10-28
Initial ENI creation/allocation over allocates IPs


jacksontj:
**What would you like to be added**:

While investigating an issue with MINIMUM_IP_TARGET (https://github.com/aws/amazon-vpc-cni-k8s/pull/1270#discussion_r512958166) I found that ENIs are allocated with the full number of IPs they can have regardless of other configuration. This end up causing an additional API call to AWS as we (1) allocate the eni then (2) immediately release a bunch of IPs.

So the ask here is to change the logic when allocating ENIs to request the number of IPs it wants to avoid the additional API call.

**Why is this needed**:

Reduce AWS API calls.

PR #1279, opened 2020-11-05
Change tryAssignIPs to assign up to configured WARM_IP_TARGET

jacksontj:
**What type of PR is this?**

bug / feature

#1272 

Previously tryAssignIPs would always allocate all available IPs on the
ENI regardless of the number of IPs the node would want. A great
showcase of this behavior is on node bootstrap. Assuming we have a node
with maxIPsPerENI=30 and WARM_IP_TARGET=5, the following happens:
- Node starts with single IP
- tryAssignIPs allocates an additional 29 IPs
- decreaseIPPool will release 29-5=24 IPs

After the change this would look like:
- Node starts with a single IP
- tryAssignIPs allocates an additional 5 IPs

So in this very common flow we cut out an entire AWS API call per node.
In addition this behavior exhibits itself any time we need to scale up
IPs as tryAssignIPs always attempts to over-allocate IPs -- which then
just need to be released almost immediately.

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -15,6 +15,7 @@ package ipamd
 
 import (
 	"fmt"
+	"math"
 	"net"
 	"os"
 	"strconv"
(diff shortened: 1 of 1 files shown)


jacksontj:
cc @mogren

jayanthvn:
Hi @jacksontj 

Please kindly fill the PR template and attach unit test cases - logs showing the functionality is working should be sufficient.

Appreciate your help on this. Thank you!

nithu0115 (commented):
(lgtm) fwiw, we need to test if setting WARM_IP_TARGET as 5 and if I have 30 pending pods to be scheduled on a newly created node, and see if p99 of pod start time increases in this case and/or see if there are other any negative impacts.

nithu0115 (commented):
(lgtm) fwiw, we need to test if setting WARM_IP_TARGET as 5 and if I have 30 pending pods to be scheduled on a newly created node, and see if p99 of pod start time increases in this case and/or see if there are other any negative impacts.

jayanthvn:
Also, can you please add mock up unit test cases something similar to ones here - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/awsutils/awsutils_test.go. Thank you!

jayanthvn:
As @nithu0115 mentioned, we will test this change to verify if p99 pod start up time is increasing.

M00nF1sh (approved):
/lgtm

nithu0115:
I cherry picked this PR on top of 1.7.10 release and tested it. The change is looks good to me. 

Tested it with below 2 scenarios and saw expected behavior. One thing to note was, I saw an increase in number of AssignPrivateIpAddresses API calls as we are not going optimistic behavior anymore where we previously assumed that there were pending Pods.

Scenario1: Started with 1 node in the cluster and CNI configuration MINIMUM_IP_TARGET=5 and WARM_IP_TARGET=2 and launched 100 Pods. 
Scenario2: WARM_IP_TARGET=2 and launched 100 Pods.

Issue #817, opened 2020-01-29
Configurable reconcile timeout for ipamd


mogren:
Currently the reconcile loop for the CNI runs too often in large clusters. We should make it possible to change these settings, either based on number of nodes in the cluster, or as an environment variable to override the default setting.

mogren:
Checking local metadata should be fine. The calls to EC2 have been reduced in #909

jayanthvn:
Waiting for #1280 PR review.

jayanthvn:
Closing the issue for now, will wait for PR to address the comments.

PR #1280, opened 2020-11-05
Add env vars for IP reconcile intervals

jacksontj:
**What type of PR is this?**

feature

#817 

Adds config options to reduce AWS API callc ount.

n/a

**Automation added to e2e**:

This shouldn't break upgrades/downgrades; hasn't been tested in that scenario though

No

```release-note
Adds environment variables for controlling IP reconciliation.
```

### DIFF
--- a/pkg/ipamd/introspect.go
+++ b/pkg/ipamd/introspect.go
@@ -197,3 +197,20 @@ func getEnvBoolWithDefault(envName string, def bool) bool {
 	}
 	return def
 }
+
+func getEnvDurationWithDefault(envName string, def time.Duration) time.Duration {
+	inputStr, found := os.LookupEnv(envName)
+
+	if !found {
+		return def
+	}
+
+	if input, err := time.ParseDuration(inputStr); err == nil {
+		if input < 0 {
+			return def
+		}
+		log.Debugf("Using %s %v", envName, input)
+		return input
+	}
+	return def
+}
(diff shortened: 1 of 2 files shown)


jayanthvn:
Hi @jacksontj 

Thanks for opening the PR, we will review it. Can you please kindly attach UT cases and fill the remaining questions in the PR template?

Appreciate your help on this.

jacksontj:
I think I've answered most of the questions here; I'm not sure what UT cases means though?

couralex6:
@jacksontj UT = unit tests :)

jayanthvn:
> I think I've answered most of the questions here; I'm not sure what UT cases means though?

Thank you for updating the template. Mainly few logs attached from IPAMD showing the new variables are honored during IP reconciliation, scenario when environment variable is not set and interval fall back to default interval, downgrade/upgrade scenario with just image change and having the env variables set. Sorry for the trouble but this will help reviewers and others using this feature.

pkg/ipamd/introspect.go
```diff
@@ -197,3 +197,20 @@ func getEnvBoolWithDefault(envName string, def bool) bool {
...
+
+func getEnvDurationWithDefault(envName string, def time.Duration) time.Duration {
+	inputStr, found := os.LookupEnv(envName)
+
+	if !found {
+		return def
+	}
+
+	if input, err := time.ParseDuration(inputStr); err == nil {
+		if input < 0 {
```
jayanthvn, line 209: I think we should cap the min to be `defaultNodeIPPoolReconcileInterval` for `NodeIPPoolReconcileInterval` and `defaultDecreaseIPPoolInterval` for `DecreaseIPPoolInterval`. Anything below the defaults will be way too aggressive and lot of ec2 calls.

jayanthvn (changes requested):

jayanthvn:
PR needs rebase and also we have to update the README for new variable.  Also we need to set floor for the variable as mentioned here - https://github.com/aws/amazon-vpc-cni-k8s/pull/1280#discussion_r558611232

PR #1284, opened 2020-11-07
add protocol field to metrics container port configuration

dallasmarlow:
**What type of PR is this?**

bug

when attempting to perform a server side apply for the cni daemonset (using terraform kubernetes alpha provider) I received an error stating that the `protocol` field was omitted.

```
Error: rpc error: associative list with keys has an element that omits key field "protocol"
```

This PR defines a protocol within the metrics container spec.

This change is no-op when successfully applied.

Yes, however the change is no-op when successfully applied.

No


nithu0115 (approved):
The protocol by default is TCP. Just curious, what version k8s are you on ?

That said, I do not see any harm in explicitly setting the protocol.

couralex6 (commented):
LGTM, thank you for your PR!

couralex6 (approved):

Issue #1249, opened 2020-10-06
Stop re-pulling images on every container start


anguslees:
https://github.com/aws/amazon-vpc-cni-k8s/blob/47f1bc4d818fac1ef1ad958a4bcba4954df09bae/config/master/manifests.jsonnet#L147

We set imagePullPolicy=Always in numerous places in our manifests, and have done so for a long time afaics.  This causes extra boot delay, load on ECR, and increases the chances of a runtime failure impacting existing nodes.

Can anyone think of a reason to keep doing this?  I think we should remove this line (and thus fall back to the usual k8s IfNotExists default)

mogren:
I totally agree. Having this set is useful for testing, but it should be removed in our sample configs and in [eks-charts/stable/aws-vpc-cni](https://github.com/aws/eks-charts/tree/master/stable/aws-vpc-cni)

nzoueidi:
> I totally agree. Having this set is useful for testing, but it should be removed in our sample configs and in [eks-charts/stable/aws-vpc-cni](https://github.com/aws/eks-charts/tree/master/stable/aws-vpc-cni)

I noticed that there is no issue in the eks-charts repository, so I have created it https://github.com/aws/eks-charts/issues/333

jayanthvn:
Pending changes of #1285 to be merged. Will close this once the PR is merged.

PR #1285, opened 2020-11-08
Stop re-pulling image on every container start

nzoueidi:
**What type of PR is this?**
cleanup
feature

**Which issue does this PR fix**: #1249 

Change imagePullPolicy to IfNotPresent, to reduce the boot delay and the load on ECR.


dallasmarlow:
I think this PR would make more sense if the docker image used had a specific version tag (IMHO).

anguslees:
> I think this PR only makes sense for manifests using a specific version tag instead of `latest` (IMHO).

Agreed.  I note this is also exactly the Kubernetes default behaviour, so I suggest we just _remove_ `imagePullPolicy` entirely, rather than try to set it to either value.

config/master/manifests.jsonnet
```diff
@@ -144,7 +144,7 @@ local awsnode = {
           containers_:: {
             awsnode: {
               image: "%s/amazon-k8s-cni:%s" % [$.ecrRepo, $.version],
-              imagePullPolicy: "Always",
+              imagePullPolicy: "IfNotPresent",
```
anguslees, line 147: Another option is to set this conditionally in jsonnet:
```jsonnet
imagePullPolicy: if $.version == "latest" then "Always" else "IfNotPresent",
```

Tbh, I've found that more trouble than it's worth in the past (eg: a naive kubectl patch might change the image, but forget to update the imagePullPolicy) so again I suggest just removing the field entirely and using the k8s dynamic default.

nzoueidi: yeah, you are totally right, the behavior of k8s is by default "IfNotPresent". So we proceed with removing the fields and just mention the condition in the jsonnet manifest. I will update the PR accordingly.

nzoueidi:
I modified the changes and squashed my commits. Let me know what do you think.
cc @dallasmarlow  @anguslees

dallasmarlow:
@nzoueidi it looks like your change may have introduced a spacing issue.

nzoueidi:
It looks like the spaces are shown only in the PR of Github

anguslees:
I think you have not regenerated the manifests following your latest jsonnet change.  Please re-run `make generate`.

Also, I think I may have confused you above - since you've followed _both_ of my alternate solutions :sweat_smile:.  To be clearer: I think we should remove the jsonnet if/else conditional, and not set `imagePullPolicy` _at all_.  So, can you:
0. (Rebase, since github says your branch has fallen behind)
1. Remove `imagePullPolicy: if $.version == "latest" then "Always" else "IfNotPresent"` line from jsonnet.
2. Re-run `make generate`.
3. Confirm the generated yaml now has _no_ `imagePullPolicy`.
4. Update this PR with all that.

Awesome, thanks in advance.  And sorry for the run-around.

jayanthvn:
Hi @nzoueidi , 

Can you please make the changes suggested by @anguslees. Then I can merge the PR.

nzoueidi:
sorry for the delay, will do today

jayanthvn:
Thanks @nzoueidi

anguslees (approved):

jayanthvn:
Hi @nzoueidi 

Can you please rebase your branch? Then I should be able to merge it.

nzoueidi:
Thank you @jayanthvn, done.

Issue #1256, opened 2020-10-09
IPAMD uses default HTTP Client with no timeout which can lead to stuck go routines


abhipth:
**What happened**:
If the EC2 API call fails for let's say ec2 service endpoint is not reachable then it's possible the request is stuck forever because there is no timeout in the http client used by IPAMD.

If we don't pass the client in [aws config](https://github.com/aws/aws-sdk-go/blob/master/aws/defaults/defaults.go#L59) it will use the default client
```
WithHTTPClient(http.DefaultClient).
```
The default client has [no timeout](https://github.com/golang/go/blob/master/src/net/http/client.go#L109)
```
var DefaultClient = &Client{}
```
With the default client the Timeout is 0 which means the client connection would never time out.

Since we are using a timeout of 0 if the ec2 service is not responding for some reason then the connection will never timeout and the go routine will remain stuck and no further reconciliation will be performed. 

This leads to new pods not getting IPv4 Addresses and the Pod will remain in ContainerCreating state till the aws-node is restarted to allow the connection to establish again.

**What you expected to happen**:
Expected behavior is to timeout the connection and keep on retrying.

**How to reproduce it (as minimally and precisely as possible)**:
The easiest way to reproduce is by creating a new faulty client in one of the ```DeallocIPAddresses``` method and use the client to make the EC2 API Call. After deleting pods as soon as the workflow is hit the go routine will hang forever given that we passed a non existent endpoint.
```
 // Create a client with wrong endpoint
 sess, err := session.NewSession(
     &aws.Config{Region: aws.String("us-west-2"),
         MaxRetries: aws.Int(15), Endpoint:aws.String("nonexistantsite.com")})
```

**Fix**
After performing the bench marking and finding out the appropriate value for timeout, introduce the timeout by using a non default HTTP client 
```
sess, err := session.NewSession(
	&aws.Config{Region: aws.String("us-west-2"),
		MaxRetries: aws.Int(15),
		HTTPClient: &http.Client{
			Timeout: time.Second * 5,
	}})
```
 
**How to see if you may have been impacted by the problem**
1. If you see the following errors ``` RequestError: send request failed\ncaused by: Post https://ec2.<local-region-name>.amazonaws.com/: dial tcp: i/o timeout"``` [Optional, but can indicate the problem]
2. If you still have remaining capacity on your ENI's and the go routine that adds the IP Address to the Data Store is stuck and not logging anything then it's possible you may be facing the issue.

[...]

PR #1286, opened 2020-11-11
Adds timeout to HTTP clients

couralex6:
**What type of PR is this?**
Bug fix.

#1256 

**Automation added to e2e**:

None

No
```release-note

```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -20,8 +20,10 @@ import (
 	"fmt"
 	"math/rand"
 	"net"
+	"net/http"
 	"os"
 	"regexp"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -61,6 +63,9 @@ const (
 	// Stagger cleanup start time to avoid calling EC2 too much. Time in seconds.
 	eniCleanupStartupDelayMax = 300
 	eniDeleteCooldownTime     = 5 * time.Minute
+
+	// Http client timeout env for sessions
+	httpTimeoutEnv = "HTTP_TIMEOUT"
 )
 
 var (
@@ -74,6 +79,8 @@ var (
 	ErrNoNetworkInterfaces = errors.New("No network interfaces found for ENI")
 	// Custom user agent
 	userAgent = request.WithAppendUserAgent("amazon-vpc-cni-k8s")
+	// HTTP timeout default value in seconds (5 seconds)
+	httpTimeoutValue = 10 * time.Second
 )
 
 var log = logger.Get()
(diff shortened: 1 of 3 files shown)


pkg/publisher/publisher.go
```diff
@@ -90,8 +99,25 @@ type cloudWatchPublisher struct {
 
 // New returns a new instance of `Publisher`
 func New(ctx context.Context) (Publisher, error) {
+
+	httpTimeoutEnvInput := os.Getenv(httpTimeoutEnv)
```
nithu0115, line 103: same as above comment

pkg/publisher/publisher.go
```diff
@@ -60,6 +66,9 @@ var (
 		"CLUSTER_ID",
 		"Name",
 	}
+
+	// HTTP timeout default value in seconds (5 seconds)
+	httpTimeoutValue = 5 * time.Second
```
nithu0115, line 71: how did we come up with default value as 5 seconds?

nithu0115 (changes requested):
@couralex6, left some comments inline.

pkg/ec2wrapper/ec2wrapper.go
```diff
@@ -30,7 +42,24 @@ type EC2Wrapper struct {
 
 //NewMetricsClient returns an instance of the EC2 wrapper
 func NewMetricsClient() (*EC2Wrapper, error) {
-	metricsSession := session.Must(session.NewSession())
+
+	httpTimeoutEnvInput := os.Getenv(httpTimeoutEnv)
```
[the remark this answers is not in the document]
couralex6: Mostly to make sure we don't have a hard coded time out in the code. The env var is optional. Only needed when the default timeout has to be overwritten.

pkg/ec2wrapper/ec2wrapper.go
```diff
@@ -30,7 +42,24 @@ type EC2Wrapper struct {
...
+	// if httpTimeout is not empty, we convert value to int and overwrite default httpTimeoutValue
+	if httpTimeoutEnvInput != "" {
+		if input, err := strconv.Atoi(httpTimeoutEnvInput); err == nil && input >= 1 {
+			log.Debugf("Using HTTP_TIMEOUT %v", input)
+			httpTimeoutValue = time.Duration(input) * time.Second
```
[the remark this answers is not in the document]
couralex6: It is used on line 46

pkg/publisher/publisher.go
couralex6: Responded in other similar comment on other file

nithu0115 (approved):
/lgtm

jayanthvn (approved):
Looks good to me too. Please add release notes in the PR summary.

pkg/awsutils/awsutils.go
```diff
@@ -74,6 +79,8 @@ var (
 	ErrNoNetworkInterfaces = errors.New("No network interfaces found for ENI")
 	// Custom user agent
 	userAgent = request.WithAppendUserAgent("amazon-vpc-cni-k8s")
+	// HTTP timeout default value in seconds (5 seconds)
```
jayanthvn, line 82: nit: Update the comment to 10 seconds

PR #1311, opened 2020-12-07
Fix regex match for getting primary interface

jayanthvn:
**What type of PR is this?**
bug

#1247 

Fixing the current regex match to pick primary interface from primary MAC which is more reliable.

```
[root@ip-192-168-85-147 ec2-user]# PRIMARY_MAC=$(curl -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/mac)
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    17  100    17    0     0  17000      0 --:--:-- --:--:-- --:--:-- 17000
[root@ip-192-168-85-147 ec2-user]# PRIMARY_IF=$(ip -o link show | grep -F "link/ether $PRIMARY_MAC" | awk -F'[ :]+' '{print $2}')
[root@ip-192-168-85-147 ec2-user]# echo $PRIMARY_IF
eth0
```

```
kubectl logs aws-node-88hvw -c aws-vpc-cni-init -n kube-system
+ PLUGIN_BINS='loopback portmap bandwidth aws-cni-support.sh'
+ for b in '$PLUGIN_BINS'
+ '[' '!' -f loopback ']'
+ for b in '$PLUGIN_BINS'
+ '[' '!' -f portmap ']'
+ for b in '$PLUGIN_BINS'
+ '[' '!' -f bandwidth ']'
+ for b in '$PLUGIN_BINS'
+ '[' '!' -f aws-cni-support.sh ']'
+ HOST_CNI_BIN_PATH=/host/opt/cni/bin
+ echo 'Copying CNI plugin binaries ... '
+ for b in '$PLUGIN_BINS'
+ install loopback /host/opt/cni/bin
Copying CNI plugin binaries ...
+ for b in '$PLUGIN_BINS'
+ install portmap /host/opt/cni/bin
+ for b in '$PLUGIN_BINS'
+ install bandwidth /host/opt/cni/bin
+ for b in '$PLUGIN_BINS'
+ install aws-cni-support.sh /host/opt/cni/bin
Configure rp_filter loose...
+ echo 'Configure rp_filter loose... '
++ get_metadata local-ipv4
+++ curl -X PUT http://169.254.169.254/latest/api/token -H 'X-aws-ec2-metadata-token-ttl-seconds: 60'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    56  100    56    0     0  56000      0 --:--:-- --:--:-- --:--:-- 56000
++ TOKEN=AQAEAHGXwoPHd-WatjsIZZX_51JLMhk73OKTbNWO-XDkiMwHgJ9UWw==
++ attempts=60
++ false
++ '[' 1 -gt 0 ']'
++ '[' 60 -eq 0 ']'
+++ curl -H 'X-aws-ec2-metadata-token: AQAEAHGXwoPHd-WatjsIZZX_51JLMhk73OKTbNWO-XDkiMwHgJ9UWw==' http://169.254.169.254/latest/meta-data/local-ipv4
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    14  100    14    0     0  14000      0 --:--:-- --:--:-- --:--:-- 14000
++ meta=192.168.22.192
++ '[' 0 -gt 0 ']'
++ '[' 0 -gt 0 ']'
++ echo 192.168.22.192
+ HOST_IP=192.168.22.192
++ get_metadata mac
+++ curl -X PUT http://169.254.169.254/latest/api/token -H 'X-aws-ec2-metadata-token-ttl-seconds: 60'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    56  100    56    0     0  56000      0 --:--:-- --:--:-- --:--:-- 56000
++ TOKEN=AQAEAHGXwoP4xDK_8AY9PNKkOPBEwr-WQmbR2pDhyWKCpbR1Lqvmuw==
++ attempts=60
++ false
++ '[' 1 -gt 0 ']'
++ '[' 60 -eq 0 ']'
+++ curl -H 'X-aws-ec2-metadata-token: AQAEAHGXwoP4xDK_8AY9PNKkOPBEwr-WQmbR2pDhyWKCpbR1Lqvmuw==' http://169.254.169.254/latest/meta-data/mac
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    17  100    17    0     0  17000      0 --:--:-- --:--:-- --:--:-- 17000
++ meta=0a:2d:8e:94:14:75
++ '[' 0 -gt 0 ']'
++ '[' 0 -gt 0 ']'
++ echo 0a:2d:8e:94:14:75
+ PRIMARY_MAC=0a:2d:8e:94:14:75
++ ip -o link show
++ grep -F 'link/ether 0a:2d:8e:94:14:75'
++ awk '-F[ :]+' '{print $2}'
+ PRIMARY_IF=eth0
+ sysctl -w net.ipv4.conf.eth0.rp_filter=2
net.ipv4.conf.eth0.rp_filter = 2
+ cat /proc/sys/net/ipv4/conf/eth0/rp_filter
2
+ '[' true == true ']'
+ sysctl -w net.ipv4.tcp_early_demux=0
net.ipv4.tcp_early_demux = 0
+ echo 'CNI init container done'
CNI init container done
```

**Automation added to e2e**:

No

No

No

No

```release-note

```

### DIFF
--- a/scripts/init.sh
+++ b/scripts/init.sh
@@ -2,6 +2,26 @@
 
 set -euxo pipefail
 
+get_metadata()
+{
+    TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 60")
+    attempts=60
+    false
+    while [ "${?}" -gt 0 ]; do
+        if [ "${attempts}" -eq 0 ]; then
+        echo "Failed to get metdata"
+        exit 1
+        fi
+        meta=$(curl -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/${1})
+        if [ "${?}" -gt 0 ]; then
+            let attempts--
+            sleep 0.5
+            false
+        fi
+    done
+    echo "$meta"
+}
+
 PLUGIN_BINS="loopback portmap bandwidth aws-cni-support.sh"
 
 for b in $PLUGIN_BINS; do
@@ -23,9 +43,10 @@ done
 
 # Configure rp_filter
 echo "Configure rp_filter loose... "
-TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 60")
-HOST_IP=$(curl -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/local-ipv4)
-PRIMARY_IF=$(ip -4 -o a | grep "$HOST_IP/" | awk '{print $2}')
+
+HOST_IP=$(get_metadata 'local-ipv4')
+PRIMARY_MAC=$(get_metadata 'mac')
+PRIMARY_IF=$(ip -o link show | grep -F "link/ether $PRIMARY_MAC" | awk -F'[ :]+' '{print $2}')
 sysctl -w "net.ipv4.conf.$PRIMARY_IF.rp_filter=2"
 cat "/proc/sys/net/ipv4/conf/$PRIMARY_IF/rp_filter"
 


scripts/init.sh
```diff
@@ -25,7 +25,8 @@ done
 echo "Configure rp_filter loose... "
 TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 60")
 HOST_IP=$(curl -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/local-ipv4)
-PRIMARY_IF=$(ip -4 -o a | grep "$HOST_IP/" | awk '{print $2}')
+PRIMARY_MAC=$(curl -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/mac)
```
achevuru, line 28: Is it worth taking the dependency on instance metadata? Didn't we in the past experience some race conditions around instance metadata access? So, in this case what happens if the initial curl to get the PRIMARY_MAC fails? If we want to go down this route, should we add retries for failure scenarios?

jayanthvn: I assume the past issues with metadata are not for primary ENI and it is already used for token and host_ip. @nithu0115 any comments on this?

nithu0115: I agree with @achevuru here. I have seen issues where IMDS can take up to 30s to provide the information of a new ENI after fetching the token. We should add retries to cover such unexpected failures.

jayanthvn: But say if metadata isn't updated then PRIMARY_IF will be null and following command will fail. Kubelet will try restart aws-node. Adding retry will be faster but do you think kubelet retry doesn't suffice this use case?

```
PRIMARY_IF=$(ip -o link show | grep -F "link/ether $PRIMARY_MAC" | awk -F'[ :]+' '{print $2}')
 sysctl -w "net.ipv4.conf.$PRIMARY_IF.rp_filter=2"
```

jayanthvn: @achevuru and @nithu0115 - I have added the retry for failure cases.

achevuru: Sure, but I believe it might be better if we handle that with in the init script. I mean, if one sees a bunch of restarts  that might invite questions/issues on what is causing the restarts and if it indicates a problem.

achevuru (approved):

scripts/init.sh
jayanthvn: Yeah makes sense.

scripts/init.sh
```diff
@@ -23,9 +42,11 @@ done
 
 # Configure rp_filter
 echo "Configure rp_filter loose... "
+
 TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 60")
```
nithu0115, line 46: Nit: Could you move this command into `get_metadata()` function as well ?

nithu0115 (approved):
👍

scripts/init.sh
jayanthvn: I had it initially inside get_metadata but that will lead to getting a token for each call. Hence moved it outside. Do you see any issue here?

nithu0115: Nope :D

PR #1329, opened 2020-12-14
Revert "Adds timeout to HTTP clients"

jayanthvn:
Reverts aws/amazon-vpc-cni-k8s#1286


jayanthvn:
I missed squash and merge. Hence reverting,

PR #1330, opened 2020-12-14
Adds timeout to HTTP clients

couralex6:
**What type of PR is this?**
Bug fix.

#1256 

**Automation added to e2e**:

None

No
```release-note

```


jayanthvn:
Thank you!

Issue #647, opened 2019-10-08
Configure aws sdk to use Region specific AWS STS service endpoints.


yeolahm:
## Issue Description
When worker nodes are launched within private subnets with no access to the internet.
VPC is configured with the VPC private endpoints for services EC2, ECR, S3 and STS.
Amazon VPC CNI Plugin for K8S is configured with the service account and mapped to IAM Role for restricted access.
In this case, while aws-node pod calling AWS EC2 API (DescribeNetworkInterfaces) fails since it can not reach sts.amazonaws.com global endpoint for AWS STS service to get the token to authenticate the API call.
```
[ERROR]	Failed to get ENI eni-xxxxxxxx information from EC2 control plane WebIdentityErr: failed to retrieve credentials
caused by: RequestError: send request failed
caused by: Post https://sts.amazonaws.com/: dial tcp 52.94.241.129:443: i/o timeout
2019-10-08T10:47:49.502Z [WARN]	Error trying to set up ENI eni-xxxxxxxx: failed to retrieve ENI eni-xxxxxxxx IP addresses: failed to find ENI addresses for ENI eni-xxxxxxxx: failed to describe network interface: WebIdentityErr: failed to retrieve credentials
caused by: RequestError: send request failed
caused by: Post https://sts.amazonaws.com/: dial tcp 52.94.241.129:443: i/o timeout
```
If AWS SDK can be configured to use the region specific endpoint for STS (example- sts.eu-west-1.amazonaws.com),  it can resolve the endpoint with Private IP's from the VPC Private Interface endpoint or to the public IP's when VPCE is not configured.

## Note
AWS STS Private Endpoint is not required when pod's are not configured with the fine grained permissions through K8S serviceaccount.

maxstepanov:
aws-sdk-go has the changes required to get rid of this annoying behaviour merged in v1.25.18: 
https://github.com/aws/aws-sdk-go/releases/tag/v1.25.18

New configurable option was introduced to regulate the STS endpoint resolution for legacy regions.
> AWS_STS_REGIONAL_ENDPOINTS=regional for environment, or sts_regional_endpoints=regional in shared config file.

Looks like just a version bump...

nithu0115:
@jaypipes, we need to re-open this issue. 

We need `AWS_STS_REGIONAL_ENDPOINTS=regional` environment variable to make it regional. Also, just setting it to regional might not work in region is Hong Kong and Baharin. 

[...]

PR #1332, opened 2020-12-14
Add STS endpoint resolver

nithu0115:
**What type of PR is this?** 
Enhancement and Cleanup

**Which issue does this PR fix**: #647 

Add STS endpoint resolver

**Automation added to e2e**:

```release-note

```


pkg/awsutils/awsutils.go
```diff
@@ -353,19 +326,8 @@ func New(useCustomNetworking bool) (*EC2InstanceMetadataCache, error) {
...
-			HTTPClient: &http.Client{
-				Timeout: httpTimeoutValue,
-			},
-		},
-	)
-	if err != nil {
-		log.Errorf("Failed to initialize AWS SDK session %v", err)
-		return nil, errors.Wrap(err, "instance metadata: failed to initialize AWS SDK session")
-	}
+	awsCfg := aws.NewConfig().WithRegion(region)
```
jayanthvn, line 329: Can we move the "region" when we initialize "sess" ? That could avoid a copy.

pkg/awsutils/awssession/session.go
```diff
@@ -0,0 +1,78 @@
...
+	log = logger.Get()
+	// HTTP timeout default value in seconds (10 seconds)
+	httpTimeoutValue = 10 * time.Second
+)
+
+func httpTimeout() time.Duration {
+	httpTimeoutEnvInput := os.Getenv(httpTimeoutEnv)
+	// if httpTimeout is not empty, we convert value to int and overwrite default httpTimeoutValue
+	if httpTimeoutEnvInput != "" {
+		if input, err := strconv.Atoi(httpTimeoutEnvInput); err == nil && input >= 1 {
```
jayanthvn, line 44: Should we maybe cap the input to be >= 10 seconds (default value)?

pkg/awsutils/awssession/session.go
```diff
@@ -0,0 +1,78 @@
...
+			log.Debugf("Using HTTP_TIMEOUT %v", input)
+			httpTimeoutValue = time.Duration(input) * time.Second
+		}
```
jayanthvn, line 47: Maybe we can add a warning that the HTTP timeout value set is invalid?

pkg/publisher/publisher.go
```diff
@@ -126,14 +98,22 @@ func New(ctx context.Context) (Publisher, error) {
...
 	if err != nil {
 		return nil, errors.Wrap(err, "publisher: Unable to obtain region")
 	}
-	cloudwatchClient := cloudwatch.New(awsSession, aws.NewConfig().WithMaxRetries(cloudwatchClientMaxRetries).WithRegion(region))
+
+	// Get AWS session
+	awsCfg := aws.Config{
+		Region: aws.String(region),
+	}
+	sess = sess.Copy(&awsCfg)
```
jayanthvn, line 113: Same comment as before, if we can pass the region to "sess", then we can avoid a copy.

jayanthvn (changes requested):
Some minor nits, overall looks good to me.

pkg/awsutils/awssession/session.go
achevuru: Just curious...Actually, why would anyone prefer to reduce the default timeout value? What will it address? I mean, if anything one would probably want to increase the default timeout value? Should we instead, set the default 10s as the min limit and set an arbitrary max limit value?

jayanthvn: Yeah, thats why I was recommending we have input >= 10 which will set the min to 10 seconds. We can also have max based on the suggested max value from EC2.

pkg/awsutils/awssession/session.go
couralex6: +1

pkg/awsutils/awssession/session.go
```diff
@@ -0,0 +1,78 @@
...
+		return httpTimeoutValue
+	} else {
+		return httpTimeoutValue
+	}
+}
+
+func New() *session.Session {
+	awsCfg := aws.Config{
+		MaxRetries:          aws.Int(10),
```
couralex6, line 56: Would be nice to have a `maxRetries` constant to avoid hard coding the value here.

pkg/awsutils/awsutils.go
```diff
@@ -353,19 +326,8 @@ func New(useCustomNetworking bool) (*EC2InstanceMetadataCache, error) {
 	cache.useCustomNetworking = useCustomNetworking
 	log.Infof("Custom networking %v", cache.useCustomNetworking)
 
-	sess, err := session.NewSession(
-		&aws.Config{
-			Region: aws.String(cache.region), 
-			MaxRetries: aws.Int(15),
```
couralex6, line 359: Max retry was 15 here and 10 above. Do we know why we used different values for those two sessions?

pkg/awsutils/awssession/session.go
couralex6: Yes, and also warn that we are using default instead.

couralex6 (commented):
Nice to have all the sessions in one file now. Looks good overall, but it would be nice to have a warning message for invalid custom HTTP timeout values.

pkg/awsutils/awssession/session.go
nithu0115: Done.

pkg/awsutils/awsutils.go
nithu0115: I thought of this. To move sess.copy to awssession and get region by querying ec2 metadata, get availability zone and then do some parsing did not seem efficient way, and more work.

pkg/awsutils/awsutils.go
nithu0115: I am not sure why this discrepancy. I went ahead and assumed that 10 retries should be a good number to meet most of the use-cases. Let me know if you want me to increase it.

nithu0115:
@couralex6 @achevuru @jayanthvn, thanks for the review. Addressed your comments :D

jayanthvn (approved):
LGTM :)

achevuru (approved):

pkg/awsutils/awsutils.go
anguslees: Can't afaics?  We need to find the region _somehow_.  Assuming we use IMDS to find the region, then we still need a 'regionless' session to initialise the IMDS client, so that we can _find_ the region, so that we can configure a 'regionful' session to initialise the 'normal' AWS client.

nithu0115: you are right :D @anguslees

Issue #1325, opened 2020-12-11
Paginate DecribeNetworkInterface call


jayanthvn:
**What would you like to be added**:
Paginate DescribeNetworkInterfaceCall

**Why is this needed**:
With larger cluster i.e, with higher node scale on nodeInit we call DescribeNetworkInterface to query all the ENIs in the account for cleanup which will cause higher API calls with larger node scale. Hence pagination will help.

https://github.com/aws/amazon-vpc-cni-k8s/blob/a68e8106ad7a4fa058df6e86edbe5d617a09346e/pkg/awsutils/awsutils.go#L1396

https://github.com/aws/amazon-vpc-cni-k8s/blob/a68e8106ad7a4fa058df6e86edbe5d617a09346e/pkg/awsutils/awsutils.go#L1399

fawadkhaliq:
@jayanthvn @haouc since the ipamd instance is bound to an ec2 instance and ec2 instance belongs to an availability zone. Should we also add `availability-zone` to `DescribeNetworkInterfacesInput` [1] filters to further filter the entries ipamd does not need to process? The more we can filter, the better (less number of API calls when # of pages > 1 )

[1] https://godoc.org/github.com/datacratic/aws-sdk-go/service/ec2#DescribeNetworkInterfacesInput

jayanthvn:
@fawadkhaliq This call https://github.com/aws/amazon-vpc-cni-k8s/blob/a68e8106ad7a4fa058df6e86edbe5d617a09346e/pkg/awsutils/awsutils.go#L1399 is made to clean up all leaked ENIs in the Cx account. So you are referring to filter this on availability-zone? So still some of the ENIs might not get cleaned right?

SaranBalaji90:
As Jayanth pointed out, this go routine is specifically for cleaning up leaked ENIs. we can’t filter out by AZ unfortunately but Jayanth if we have cluster name in the tag and if that’s known to CNI plug-in we can use that as well.

fawadkhaliq:
I see. +1, that would help. However, only consider cluster name tag if it exists for the supported CNI versions to make sure we don't skip interfaces from older CNI versions

jayanthvn:
Yeah @SaranBalaji90 today we are not aware of the cluster name. But is it fine to just clean up ENIs in the same cluster? Because today we clean up all leaked ENIs in the Cx account and that will be limited to the cluster now.

fawadkhaliq:
[...]

PR #1333, opened 2020-12-14
Replace DescribeNetworkInterfaces with paginated version

haouc:
**What type of PR is this?**
This PR tends to solve a latency issue when IPAMD need cleanup big number of ENIs.

Paginate DecribeNetworkInterface call #1325

Non-pagintaed EC2 could cause latency increase dramatically for some customer.

Load tests were done.
**Automation added to e2e**:

NO

NO

NO
```release-note

```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -1431,16 +1431,11 @@ func (cache *EC2InstanceMetadataCache) getFilteredListOfNetworkInterfaces() ([]*
 		MaxResults: aws.Int64(describeENIPageSize),
 	}
 
-	outputENIs, err := cache.getENIsFromPaginatedDescribeNetworkInterfaces(input)
-	if err != nil {
-		return nil, errors.Wrap(err, "awsutils: unable to obtain filtered list of network interfaces")
-	}
-
-	networkInterfaces := make([]*ec2.NetworkInterface, 0)
-	for _, networkInterface := range outputENIs {
+	var networkInterfaces []*ec2.NetworkInterface
+	filterFn := func(networkInterface *ec2.NetworkInterface) error {
 		// Verify the description starts with "aws-K8S-"
 		if !strings.HasPrefix(aws.StringValue(networkInterface.Description), eniDescriptionPrefix) {
-			continue
+			return nil
 		}
 		// Check that it's not a newly created ENI
 		tags := getTags(networkInterface, aws.StringValue(networkInterface.NetworkInterfaceId))
@@ -1453,7 +1448,7 @@ func (cache *EC2InstanceMetadataCache) getFilteredListOfNetworkInterfaces() ([]*
 			}
 			if time.Since(parsedTime) < eniDeleteCooldownTime {
 				log.Infof("Found an ENI created less than 5 minutes ago, so not cleaning it up")
-				continue
+				return nil
 			}
 			log.Debugf("%v", value)
 		} else {
@@ -1461,9 +1456,16 @@ func (cache *EC2InstanceMetadataCache) getFilteredListOfNetworkInterfaces() ([]*
 			 * process of being attached by CNI versions v1.5.x or earlier.
 			 */
 			cache.tagENIcreateTS(aws.StringValue(networkInterface.NetworkInterfaceId), maxENIBackoffDelay)
-			continue
+			return nil
 		}
 		networkInterfaces = append(networkInterfaces, networkInterface)
+		return nil
+	}
+
+	err := cache.getENIsFromPaginatedDescribeNetworkInterfaces(input, filterFn)
+
+	if err != nil {
+		return nil, errors.Wrap(err, "awsutils: unable to obtain filtered list of network interfaces")
 	}
 
 	if len(networkInterfaces) < 1 {
@@ -1522,20 +1524,24 @@ func (cache *EC2InstanceMetadataCache) IsUnmanagedENI(eniID string) bool {
 }
 
 func (cache *EC2InstanceMetadataCache) getENIsFromPaginatedDescribeNetworkInterfaces(
-	input *ec2.DescribeNetworkInterfacesInput) ([]*ec2.NetworkInterface, error) {
-	outputENIs := make([]*ec2.NetworkInterface, 0)
+	input *ec2.DescribeNetworkInterfacesInput, filterFn func(networkInterface *ec2.NetworkInterface) error) error {
 	pageNum := 0
-	log.Debugf("Paginating describe ENI has page size: %d", *input.MaxResults)
+	var innerErr error
 	pageFn := func(output *ec2.DescribeNetworkInterfacesOutput, lastPage bool) (nextPage bool) {
 		pageNum++
 		log.Debugf("EC2 DescribeNetworkInterfaces succeeded with %d results on page %d",
 			len(output.NetworkInterfaces), pageNum)
-
-		outputENIs = append(outputENIs, output.NetworkInterfaces...)
-		// Loop is guided by nextToken, the func expect a false to exit.
-		return output.NextToken != nil
+		for _, eni := range output.NetworkInterfaces {
+			if err := filterFn(eni); err != nil {
+				innerErr = err
+				return false
+			}
+		}
+		return true
 	}
 
-	err := cache.ec2SVC.DescribeNetworkInterfacesPagesWithContext(context.Background(), input, pageFn, userAgent)
-	return outputENIs, err
+	if err := cache.ec2SVC.DescribeNetworkInterfacesPagesWithContext(context.TODO(), input, pageFn, userAgent); err != nil {
+		return err
+	}
+	return innerErr
 }
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -66,6 +66,9 @@ const (
 
 	// Http client timeout env for sessions
 	httpTimeoutEnv = "HTTP_TIMEOUT"
+
+	// the default page size when paginating the DescribeNetworkInterfaces call
+	describeENIPageSize = 1000
 )
 
 var (
@@ -355,7 +358,7 @@ func New(useCustomNetworking bool) (*EC2InstanceMetadataCache, error) {
 
 	sess, err := session.NewSession(
 		&aws.Config{
-			Region: aws.String(cache.region), 
+			Region: aws.String(cache.region),
 			MaxRetries: aws.Int(15),
 			HTTPClient: &http.Client{
 				Timeout: httpTimeoutValue,
@@ -1425,14 +1428,16 @@ func (cache *EC2InstanceMetadataCache) getFilteredListOfNetworkInterfaces() ([]*
 
 	input := &ec2.DescribeNetworkInterfacesInput{
 		Filters: []*ec2.Filter{tagFilter, statusFilter},
+		MaxResults: aws.Int64(describeENIPageSize),
 	}
-	result, err := cache.ec2SVC.DescribeNetworkInterfacesWithContext(context.Background(), input, userAgent)
+
+	outputENIs, err := cache.getENIsFromPaginatedDescribeNetworkInterfaces(input)
 	if err != nil {
 		return nil, errors.Wrap(err, "awsutils: unable to obtain filtered list of network interfaces")
 	}
 
 	networkInterfaces := make([]*ec2.NetworkInterface, 0)
-	for _, networkInterface := range result.NetworkInterfaces {
+	for _, networkInterface := range outputENIs {
 		// Verify the description starts with "aws-K8S-"
 		if !strings.HasPrefix(aws.StringValue(networkInterface.Description), eniDescriptionPrefix) {
 			continue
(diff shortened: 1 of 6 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -1515,3 +1520,25 @@ func (cache *EC2InstanceMetadataCache) IsUnmanagedENI(eniID string) bool {
...
+
+func (cache *EC2InstanceMetadataCache) getENIsFromPaginatedDescribeNetworkInterfaces(
+	input *ec2.DescribeNetworkInterfacesInput) ([]*ec2.NetworkInterface, error) {
+	outputENIs := make([]*ec2.NetworkInterface, 0)
+	pageNum := 0
+	log.Debugf("Paginating describe ENI has page size: %d", *input.MaxResults)
+	pageFn := func(output *ec2.DescribeNetworkInterfacesOutput, lastPage bool) (nextPage bool) {
+		pageNum++
+		log.Infof("EC2 DescribeNetworkInterfaces succeeded with %d results on page %d",
+			len(output.NetworkInterfaces), pageNum)
```
SaranBalaji90, line 1532: We should not use Info level for these.

pkg/awsutils/awsutils.go
```diff
@@ -1515,3 +1520,25 @@ func (cache *EC2InstanceMetadataCache) IsUnmanagedENI(eniID string) bool {
...
+
+		outputENIs = append(outputENIs, output.NetworkInterfaces...)
+		// if no more ENI, we want to exit even it is not the last page to avoid many blank calls.
+		if len(output.NetworkInterfaces) == 0 {
```
SaranBalaji90, line 1536: You should always use token/continue fields in the output object. For https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#DescribeNetworkInterfacesOutput use 'NextToken'

pkg/awsutils/awsutils.go
```diff
@@ -1515,3 +1520,25 @@ func (cache *EC2InstanceMetadataCache) IsUnmanagedENI(eniID string) bool {
...
+			lastPage = true
+		}
+		return !lastPage
```
SaranBalaji90, line 1539: You can simplify this by just saying 

return output.NextToken != nil

pkg/awsutils/awsutils.go
haouc: Makes sense. Have synced with EC2. The same suggestion. Thanks @SaranBalaji90.

pkg/awsutils/awsutils.go
SaranBalaji90: If possible we should filter out ENIs which are not required like it should have description aws-k8s etc. @jayanthvn looking at https://github.com/aws/amazon-vpc-cni-k8s/blob/508c28dc04e11393c6cdb68bfba0e8f1da07b3a0/pkg/awsutils/awsutils.go#L1410 seems like ipamd will clean up leaked ENIs for other clusters in the same account as well.

haouc: The cleanup does look for all ENIs in the same account.

pkg/awsutils/awsutils.go
haouc: Updated to debug.

pkg/awsutils/awsutils.go
haouc: Updated to use nextToken != nill

pkg/awsutils/awsutils_test.go
```diff
@@ -857,19 +858,30 @@ func TestEC2InstanceMetadataCache_cleanUpLeakedENIsInternal(t *testing.T) {
...
-	}
+	interfaces := []*ec2.NetworkInterface{{
+		Description: &description,
+		TagSet: []*ec2.Tag{
+			{Key: aws.String(eniNodeTagKey), Value: aws.String("test-value")},
+		},
+	}}
 
-	mockEC2.EXPECT().DescribeNetworkInterfacesWithContext(gomock.Any(), gomock.Any(), gomock.Any()).Return(result, nil)
+	helpCallMockedPagedDescribeNetworkInterface(t, mockEC2, interfaces, nil, 1)
```
SaranBalaji90, line 868: I would have named it differently - /s/help/setupDescribeNwInterfacesPagesMock...()

haouc: Will be update soon.

pkg/awsutils/awsutils.go
```diff
@@ -1515,3 +1520,22 @@ func (cache *EC2InstanceMetadataCache) IsUnmanagedENI(eniID string) bool {
...
+	input *ec2.DescribeNetworkInterfacesInput) ([]*ec2.NetworkInterface, error) {
+	outputENIs := make([]*ec2.NetworkInterface, 0)
+	pageNum := 0
+	log.Debugf("Paginating describe ENI has page size: %d", *input.MaxResults)
+	pageFn := func(output *ec2.DescribeNetworkInterfacesOutput, lastPage bool) (nextPage bool) {
+		pageNum++
+		log.Debugf("EC2 DescribeNetworkInterfaces succeeded with %d results on page %d",
+			len(output.NetworkInterfaces), pageNum)
+
+		outputENIs = append(outputENIs, output.NetworkInterfaces...)
```
anguslees, line 1534: This avoids a huge datastructure on the EC2 server-side (yay), but still builds the same huge list in vpc-cni, just so we can _then_ iterate over the items in the list (boo).  Worse, it's now a function with an API that _can't_ be used in any other way, and it is no longer clear to a caller that this is the way it works internally.

Rather than this proposed helper function, can we call `DescribeNetworkInterfacesPagesWithContext` directly in the caller, and just do the `for _, networkInterface := range result.NetworkInterfaces { ... }` loop body inside the `pageFn` ?

SaranBalaji90: > Rather than this proposed helper function, can we call `DescribeNetworkInterfacesPagesWithContext` directly in the caller, and just do the `for _, networkInterface := range result.NetworkInterfaces { ... }` loop body inside the `pageFn` ?

I think recommendation from EC2 is to use MaxResult parameter for DescribeNetworkInterfacesPagesWithContext API?

pkg/awsutils/awsutils.go
```diff
@@ -1515,3 +1520,22 @@ func (cache *EC2InstanceMetadataCache) IsUnmanagedENI(eniID string) bool {
 	}
 	return false
 }
+
+func (cache *EC2InstanceMetadataCache) getENIsFromPaginatedDescribeNetworkInterfaces(
+	input *ec2.DescribeNetworkInterfacesInput) ([]*ec2.NetworkInterface, error) {
+	outputENIs := make([]*ec2.NetworkInterface, 0)
```
anguslees, line 1526: micro-nit: don't create a zero capacity array, just use the nil value.  That way it will avoid an allocation, particularly in the case where you don't append any entries at all (eg: an error return).

ie:
```golang
// Don't:
foos := make([]Foo, 0)
// Do:
var foos []Foo
```

See also https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices

pkg/awsutils/awsutils.go
```diff
@@ -1515,3 +1520,22 @@ func (cache *EC2InstanceMetadataCache) IsUnmanagedENI(eniID string) bool {
...
+		pageNum++
+		log.Debugf("EC2 DescribeNetworkInterfaces succeeded with %d results on page %d",
+			len(output.NetworkInterfaces), pageNum)
+
+		outputENIs = append(outputENIs, output.NetworkInterfaces...)
+		// Loop is guided by nextToken, the func expect a false to exit.
+		return output.NextToken != nil
+	}
+
+	err := cache.ec2SVC.DescribeNetworkInterfacesPagesWithContext(context.Background(), input, pageFn, userAgent)
```
anguslees, line 1539: This is not running in the background, so `context.Background()` is a misleading choice.  Either take a `ctx` arg from the caller (best), or use `context.TODO()` to indicate that it's not a meaningful context.

pkg/awsutils/awsutils.go
jayanthvn: @SaranBalaji90 Yes cleanup fetches all "available" ENIs with aws-K8S-* description in the cx account. This PR https://github.com/aws/amazon-vpc-cni-k8s/pull/1109 mitigates that issue for some extent. But yes cluster tag would surely restrict cleanup within the node's cluster. To use clusterName we need to have the CLUSTER_NAME env variable set or fetch from ec2 instance tag key -  "kubernetes.io/cluster/jay-v17-cluster = owned ".

pkg/awsutils/awsutils.go
anguslees: Sorry, my description was probably confusing.  I meant:

```golang
func (cache *EC2InstanceMetadataCache) getENIsFromPaginatedDescribeNetworkInterfaces(
￼	input *ec2.DescribeNetworkInterfacesInput, itemFn func(*ec2.NetworkInterface) bool) error {
    var innerErr error
    pageFn := func(output *ec2.DescribeNetworkInterfacesOutput, lastPage bool) bool {
        for _, eni := range output.NetworkInterfaces {
            if err := itemFn(eni); err != nil {
                innerErr = err
                return false
            }
        }
        return true
    }
    if err := cache.ec2SVC.DescribeNetworkInterfacesPagesWithContext(context.TODO(), input, pageFn, userAgent); err != nil {
        return err
    }
    return innerErr
}
```

... and then back in `getFilteredListOfNetworkInterfaces()`, do:
```golang
        // ...

        var networkInterfaces []*ec2.NetworkInterface
        visitorFn := func(networkInterface *ec2.NetworkInterface) error {
                // [Gus: This is the body of the original `for` loop, with only minor tweaks to control flow]

		// Verify the description starts with "aws-K8S-"
		if !strings.HasPrefix(aws.StringValue(networkInterface.Description), eniDescriptionPrefix) {
			return nil
		}
		// Check that it's not a newly created ENI
		tags := getTags(networkInterface, aws.StringValue(networkInterface.NetworkInterfaceId))
		if value, ok := tags[eniCreatedAtTagKey]; ok {
			parsedTime, err := time.Parse(time.RFC3339, value)
			if err != nil {
				log.Warnf("ParsedTime format %s is wrong so retagging with current TS", parsedTime)
				cache.tagENIcreateTS(aws.StringValue(networkInterface.NetworkInterfaceId), maxENIBackoffDelay)
			}
			if time.Since(parsedTime) < eniDeleteCooldownTime {
				log.Infof("Found an ENI created less than 5 minutes ago, so not cleaning it up")
				return nil
			}
			log.Debugf("%v", value)
		} else {
			/* Set a time if we didn't find one. This is to prevent accidentally deleting ENIs that are in the
			 * process of being attached by CNI versions v1.5.x or earlier.
			 */
			cache.tagENIcreateTS(aws.StringValue(networkInterface.NetworkInterfaceId), maxENIBackoffDelay)
		}

                networkInterfaces = append(networkInterfaces, networkInterface)
		return nil
        }

	input := &ec2.DescribeNetworkInterfacesInput{
￼		Filters: []*ec2.Filter{tagFilter, statusFilter},
￼		MaxResults: aws.Int64(describeENIPageSize),
￼	}
￼	err := getENIsFromPaginatedDescribeNetworkInterfaces(input, visitorFn)
        if err != nil {
￼		return nil, errors.Wrap(err, "awsutils: unable to obtain filtered list of network interfaces")
￼	}

        // .. continue with rest of function, and eventually return networkInterfaces, nil
￼```

```

pkg/awsutils/awsutils.go
anguslees: Fwiw, you can also just always `return true` if you want to process every page, and then stop.  There's no need to use `lastPage` ([`request.Pagination`](https://docs.aws.amazon.com/sdk-for-go/api/aws/request/#Pagination) uses the same `p.HasNextPage()` check for both the `lastPage` variable, and when to stop the `p.Next()` iterator)

anguslees: > Yes cleanup fetches all "available" ENIs with aws-K8S-* description in the cx account. 

(Aside: It's a larger change, but fwiw we could also run this loop _once_ per cluster, instead of on every node.  Just move the code out into a separate "cleanup" pod that we run as a Deployment replicas=1.  It doesn't make much sense to do a global cleanup _times_ the number of nodes :worried: )

pkg/awsutils/awsutils.go
haouc: This is so nice and minimizes the size of list holding results data locally. Will try this out. Thanks @anguslees

pkg/awsutils/awsutils.go
abhipth: An alternate could be to do it under leader election from one of the node and don't do it if it has already happened in the last x interval.

anguslees (approved):

jayanthvn (approved):
LGTM :)

pkg/awsutils/awsutils.go
anguslees: > An alternate could be to do it under leader election from one of the node and don't do it if it has already happened in the last x interval.

￼Agreed, using a Lease is the other obvious solution.  I think this means ~10k watches (ie: 1/node) from every aws-node pod on a shared Lease/ConfigMap, which is ok - and better than running the loop 10k times as we currently do.

Issue #1340, opened 2020-12-22
EC2 Metadata not in sync is causing aws-node be stuck in Crash Loopback state.


kakarotbyte:
**IPAMD is restarting with the below messages.**

    {"level":"debug","ts":"2020-11-21T12:18:10.565Z","caller":"awsutils/awsutils.go:388","msg":"Update ENI eni-xxxxxxxxxx"}
    {"level":"error","ts":"2020-11-21T12:18:10.773Z","caller":"aws-k8s-agent/main.go:28","msg":"Initialization failure: ipamd: can not initialize with AWS SDK interface: refreshSGIDs: unable to update the ENI's SG: InvalidNetworkInterfaceID.NotFound: The networkInterface ID 'eni-xxxxxxxxxxxx' does not exist\n\tstatus code: 400, request id: aaaaaa-bbbbbb-cccccc-ccccc-sssssss"}

From [awsutils l458]( https://github.com/aws/amazon-vpc-cni-k8s/blob/v1.7.5/pkg/awsutils/awsutils.go#L458) We understand **ModifyNetworkInterfaceAttribute** action is triggering the above error message. 

From the Cloud trail API calls I was able to confirm that the ENI was created and deleted by VPC cni however. They were not cleared from EC2 metadata. Since [metadataMACPath function](https://github.com/aws/amazon-vpc-cni-k8s/blob/eb63a013949c1ad5335ac9c229b2e136a1896797/pkg/awsutils/awsutils.go#L640) will use the below snippet to make the metadata call which will gather the eni-id We are seeing the above issue. 

	eniID, err = cache.ec2Metadata.GetMetadata( + eniMAC + metadataInterface)

I have confirmed by running the below curl call that the deleted ENI is still present on the metadata. 
 
    curl 169.254.169.254/2020-10-27/meta-data/network/interfaces/macs/<macaddress>/<interface-id>

### Summary
 IPAMD should be able to handle out-of-synced IMDS in above cases. As This will cause aws-node to be in crashloopback state and cause pods scheduling onto the node to not have any ip's causing them to be stuck in creating state.

similar issue form the past [#1177](https://github.com/aws/amazon-vpc-cni-k8s/pull/1177)

**Why is this needed:**
CNI should handle similar issue related to imds gracefully instead of getting the node and all pods stuck.

falgofrancis:
Is there any work around for this issue? Currently we drain the node.

jayanthvn:
[...]

PR #1341, opened 2020-12-22
Gracefully handle failed ENI SG update

jayanthvn:
**What type of PR is this?**
Bug

#1340 

In refreshSGIDs, if IMDS contains a stale ENI or if ENI SG update fails we used to return an error. Problem is here - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/awsutils/awsutils.go#L440. We have a go routine but also we make this call (L440). If this call returns error, then awsutils.new return error (https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/ipamd.go#L294) and eventually ipamd.New return error causing init failure (https://github.com/aws/amazon-vpc-cni-k8s/blob/master/cmd/aws-k8s-agent/main.go#L56 ). Hence aws-node keeps restarting. We don't need to make a separate call and also the error need not be returned since in the next 30 seconds retry will be done. Once IMDS data is synced from Ec2, stale ENIs will be removed so IPAMD will be gracefully handling this. Also returning error will fail updating all the ENIs since if one of the ENIs in the list is stale then following ENIs wont be updated. This PR will fix even that issue.

In nodeInit, if the attachedENIs have stale ENI then describeAllENIs will fail for that ENI, so added a counter to track the number of times IMDS is out of sync. Also this ENI won't be added to datastore.

Reconciler, will again get AttachedENIs, this time if there is a stale ENI, This condition [https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/ipamd.go#L963] will be true but describeAllENIs will again prevent adding the ENI to the metadataResult [https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/ipamd.go#L969 ] and [https://github.com/aws/amazon-vpc-cni-k8s/blob/f96e713debda4cf21dd9ea5f0a122031a9e82b91/pkg/awsutils/awsutils.go#L1011 ] but we need to account the error.

#1340 

```
{"level":"debug","ts":"2020-12-22T21:14:22.792Z","caller":"awsutils/awsutils.go:387","msg":"Update ENI eni-0bbfc9102ff0d750e"}
{"level":"debug","ts":"2020-12-22T21:14:23.184Z","caller":"awsutils/awsutils.go:387","msg":"Update ENI eni-0e31591083f43cd00"}
{"level":"debug","ts":"2020-12-22T21:14:23.335Z","caller":"awsutils/awsutils.go:387","msg":"refreshSGIDs: unable to update the ENI eni-0e31591083f43cd00 SG - InvalidNetworkInterfaceID.NotFound: The networkInterface ID 'eni-0e31591083f43cd00' does not exist\n\tstatus code: 400, request id: 0b4366ac-6e58-4253-908f-7d805a4255c7"}
{"level":"debug","ts":"2020-12-22T21:14:23.335Z","caller":"awsutils/awsutils.go:387","msg":"Update ENI eni-0e31591083f43cd89"}
```
**Automation added to e2e**:

No

No

No

No

```release-note

```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -474,6 +474,11 @@ func (cache *EC2InstanceMetadataCache) RefreshSGIDs(mac string) error {
 			_, err = cache.ec2SVC.ModifyNetworkInterfaceAttributeWithContext(context.Background(), attributeInput)
 			awsAPILatency.WithLabelValues("ModifyNetworkInterfaceAttribute", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
 			if err != nil {
+				if aerr, ok := err.(awserr.Error); ok {
+					if aerr.Code() == "InvalidNetworkInterfaceID.NotFound" {
+						awsAPIErrInc("IMDSMetaDataOutOfSync", err)
+					}
+				}
 				awsAPIErrInc("ModifyNetworkInterfaceAttribute", err)
 				//No need to return error here since retry will happen in 30seconds and also
 				//If update failed due to stale ENI then returning error will prevent updating SG 
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -454,7 +454,7 @@ func (cache *EC2InstanceMetadataCache) RefreshSGIDs(mac string) error {
 		for _, eni := range allENIs {
 			eniIDs = append(eniIDs, eni.ENIID)
 		}
-
+		
 		newENIs := StringSet{}
 		newENIs.Set(eniIDs)
 
@@ -475,7 +475,10 @@ func (cache *EC2InstanceMetadataCache) RefreshSGIDs(mac string) error {
 			awsAPILatency.WithLabelValues("ModifyNetworkInterfaceAttribute", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
 			if err != nil {
 				awsAPIErrInc("ModifyNetworkInterfaceAttribute", err)
-				return errors.Wrap(err, "refreshSGIDs: unable to update the ENI's SG")
+				//No need to return error here since retry will happen in 30seconds and also
+				//If update failed due to stale ENI then returning error will prevent updating SG 
+				//for following ENIs since the list is sorted 
+			    log.Debugf("refreshSGIDs: unable to update the ENI %s SG - %v", eniID, err)
 			}
 		}
 	}
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -1015,6 +1015,7 @@ func (cache *EC2InstanceMetadataCache) DescribeAllENIs() (DescribeAllENIsResult,
 			if aerr.Code() == "InvalidNetworkInterfaceID.NotFound" {
 				badENIID := badENIID(aerr.Message())
 				log.Debugf("Could not find interface: %s, ID: %s", aerr.Message(), badENIID)
+				awsAPIErrInc("IMDSMetaDataOutOfSync", err)
 				// Remove this ENI from the map
 				delete(eniMap, badENIID)
 				// Remove the failing ENI ID from the EC2 API request and try again


nithu0115 (approved):
/lgtm 👍

pkg/awsutils/awsutils.go
```diff
@@ -455,7 +443,10 @@ func (cache *EC2InstanceMetadataCache) refreshSGIDs(mac string) error {
 			awsAPILatency.WithLabelValues("ModifyNetworkInterfaceAttribute", fmt.Sprint(err != nil)).Observe(msSince(start))
 			if err != nil {
 				awsAPIErrInc("ModifyNetworkInterfaceAttribute", err)
-				return errors.Wrap(err, "refreshSGIDs: unable to update the ENI's SG")
+				//No need to return error here since retry will happen in 30seconds and also
+				//If update failed due to stale ENI then returning error will prevent updating SG
```
fawadkhaliq, line 447: There are two aspects to metadata not being in sync 1) transient 2) permanent (let's say some bug in the underlying metadata provider) 

This change handles # 1 well. We should have a way to bubble # 2 to the user so they can raise an issue on the dependency if # 2 ever happens and work towards remedy

jayanthvn: Yeah makes sense, I will make the changes and update the PR.

jayanthvn:
Hi @jinja2 

I will update the PR with @fawadkhaliq comments. Will try to push this in this week.

Thanks!

jayanthvn:
@achevuru - If the ENI is stale then we won't add the ENI to ENIPool [https://github.com/aws/amazon-vpc-cni-k8s/blob/9db2ae62ecd0cb56f7fc20b80427fa6ff4e17a42/pkg/awsutils/awsutils.go#L1007] so secondary IPs won't be attached in the create path. 

But the reconcile path seems to be adding it back to ENIPool - https://github.com/aws/amazon-vpc-cni-k8s/blob/9db2ae62ecd0cb56f7fc20b80427fa6ff4e17a42/pkg/ipamd/ipamd.go#L1007 , Adding a describe ENI to verify the ENIs will add EC2 calls for each reconcile loop so we will have to cache the stale ENI and check in reconcile path if it was marked stale before, then we will have to skip it here. But wondering if it is possible for the new ENI allocated using the same ID as the stale ENI, if so then in Alloc ENI we have to search the local cache of stale ENIs and remove it.

pkg/awsutils/awsutils.go
```diff
@@ -474,6 +474,11 @@ func (cache *EC2InstanceMetadataCache) RefreshSGIDs(mac string) error {
 			_, err = cache.ec2SVC.ModifyNetworkInterfaceAttributeWithContext(context.Background(), attributeInput)
 			awsAPILatency.WithLabelValues("ModifyNetworkInterfaceAttribute", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
 			if err != nil {
+				if aerr, ok := err.(awserr.Error); ok {
+					if aerr.Code() == "InvalidNetworkInterfaceID.NotFound" {
```
achevuru, line 478: Curious, if this check is enough for a leaked ENI case? (active but not attached to any instance). Interface ID is still valid but the ENI is probably no longer attached to the Instance (or any instance in the cluster). I mean before the IMDS is updated, it is possible we will actually go and update the SGs on this instance. Should we check if the ENI is attached to an instance (from describeENI o/p) before we call Modify on that ENI.

Maybe we can count on the leaked eni routine to delete that ENI eventually(?)...VPC CNI actually goes ahead and updates the SGs on this ENI even though it is no longer tied to this cluster but the leaked ENI routine will eventually delete it. Is that the expectation here?

jayanthvn: Yeah, we might have to check `attachment.instance-id - The ID of the instance to which the network interface is attached.` Similar check was done here - https://github.com/aws/amazon-vpc-cni-k8s/blob/9db2ae62ecd0cb56f7fc20b80427fa6ff4e17a42/pkg/awsutils/awsutils.go#L1007 . Yeah the design looks like the leaked ENI will eventually delete the leaked ENI.

Issue #1342, opened 2020-12-23
Remove Logging of CNI Plugin version 


jochen-st:
**What would you like to be added**:

Remove Logging of CNI Plugin version, call is here: https://github.com/aws/amazon-vpc-cni-k8s/blob/de7368ab6cb32fcc11f207c114abab3bf22891a6/cmd/routed-eni-cni-plugin/cni.go#L335 or reduce severity.

**Why is this needed**:

Using AWS Container Insights and fluent EKS logging, every node produces entries every 5 Minutes to LogGroup `/aws/containerinsights/cluster/host` and `/aws/containerinsights/cluster/dataplane` in the default configuration.

Sample event from dataplane LogGroup:
~~~
{
    "message": "{\"level\":\"info\",\"ts\":\"2020-12-23T10:06:56.551Z\",\"caller\":\"/usr/local/go/src/runtime/proc.go:203\",\"msg\":\"CNI Plugin version: v1.7.5 ...\"}",
    "hostname": "ip-10-0-2-242.eu-central-1.compute.internal",
    "systemd_unit": "kubelet.service"
}
~~~
This adds no value, increases costs and hinders finding more important log entries.

jayanthvn:
Hi @jochen-st 

I can reduce the severity. Thanks for letting us know.

jayanthvn:
Merged into master, so closing the issue.

ddemo-miriade:
Hi jayanthvn,
I see the change is still in master branch but is not yet released. I'm still having the issue even after the CNI update to version 1.7.10.
Do you have a release date for the fix?

Regards,
Daniele

jayanthvn:
Hi @ddemo-miriade ,

Yes will take it up for 1.8 release. I can provide you a timeline in couple weeks. Thank you.

PR #1343, opened 2020-12-28
Reduce log level for CNI version print

jayanthvn:
**What type of PR is this?**
bug

https://github.com/aws/amazon-vpc-cni-k8s/blob/de7368ab6cb32fcc11f207c114abab3bf22891a6/cmd/routed-eni-cni-plugin/cni.go#L335 is log Info and keeps printing in logfile.

Reduces the log level to debug

#1342 

Yes

**Automation added to e2e**:

No

No

No

No

```release-note

```


nithu0115 (approved):
/lgtm 👍

PR #1352, opened 2021-01-19
Checkpoint Migration 2 - Remove read from CRI

couralex6:
**What type of PR is this?**

Feature

This PR enables `Checkpoint Migration 2`, which disables reads from CRI. Instead, IPAMD will read and write to a file in JSON format to keep track of pods and their IPs on reload: `/var/run/aws-node/ipam.json`. 

Removing access to CRI from unprivileged pod like aws-node will improve the security posture for the customer and would allow Cx to add SELinux policy restrictions so unprivileged pods couldn't connect to host sockets. Aws-node pod is a kube-system pod and the node can be accessed only by cluster admin and the IPAM.json file created will have read-write access only to root user. Hence this would make the system more secure.

This PR also removes the dockershim.sock mounts from `manifest.jsonnet` and the aws-node manifests under `config/master/`.

This change need significant testing before merging. Here is the current lists of tests performed:

_Deleting file_
1. Add pods > OK
2. Delete pods > OK
3. Restart node > OK

_Manually modifying IPs in the file_
1. Add pods > OK
2. Delete pods > OK
3. Restart node > OK

_Changing JSON format of file_
1. Add pods > OK
2. Delete pods > OK
3. Restart node > OK

**Automation added to e2e**:

This PR will break upgrades from CNI v1.6.x to CNI v1.8.x. Customers will need to upgrade to CNI v1.7.x before upgrading to CNI v1.8.

This is due to the move from using the CRI as storage for allocated IPs to a file based approach. In CNI v1.6 we read from CRI to keep track of allocated IPs. In v1.8, we only read from the file. This means that IPAMD will not know which IPs have already been allocated to existing pods.

To enable this transition, we started writing allocated IPs to `/var/run/aws-node/ipam.json` in CNI v1.7, but keep reading from CRI. Hence, clusters running CNI v1.7 already have the allocated IPs stored in the file to allow IPAMD to operate without querying CRI in v1.8.

Yes, this PR modifies the CNI deamonset config files to remove the dockershim.sock mounts. The new config files will be release with CNI v1.8.

Action required: Customers running CNI v1.6 will need to upgrade to CNI v1.7 before upgrading to CNI v1.8.

```release-note

```

### DIFF
--- a/config/master/aws-k8s-cni-cn.yaml
+++ b/config/master/aws-k8s-cni-cn.yaml
@@ -187,8 +187,6 @@
           "name": "log-dir"
         - "mountPath": "/var/run/aws-node"
           "name": "run-dir"
-        - "mountPath": "/var/run/dockershim.sock"
-          "name": "dockershim"
         - "mountPath": "/run/xtables.lock"
           "name": "xtables-lock"
       "hostNetwork": true
@@ -216,9 +214,6 @@
       - "hostPath":
           "path": "/etc/cni/net.d"
         "name": "cni-net-dir"
-      - "hostPath":
-          "path": "/var/run/dockershim.sock"
-        "name": "dockershim"
       - "hostPath":
           "path": "/run/xtables.lock"
         "name": "xtables-lock"
--- a/config/master/aws-k8s-cni-us-gov-east-1.yaml
+++ b/config/master/aws-k8s-cni-us-gov-east-1.yaml
@@ -187,8 +187,6 @@
           "name": "log-dir"
         - "mountPath": "/var/run/aws-node"
           "name": "run-dir"
-        - "mountPath": "/var/run/dockershim.sock"
-          "name": "dockershim"
         - "mountPath": "/run/xtables.lock"
           "name": "xtables-lock"
       "hostNetwork": true
@@ -216,9 +214,6 @@
       - "hostPath":
           "path": "/etc/cni/net.d"
         "name": "cni-net-dir"
-      - "hostPath":
-          "path": "/var/run/dockershim.sock"
-        "name": "dockershim"
       - "hostPath":
           "path": "/run/xtables.lock"
         "name": "xtables-lock"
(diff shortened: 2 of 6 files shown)


jayanthvn:
Thanks Alex. Let's not merge this until testing is completed.

fawadkhaliq:
Looking at the PR description, it's hard to figure the context behind this change, how this impacts users, upgrades, downgrades, what testing was done, what documentation impact this has. Can we add the appropriate info in the description pls

anguslees:
When are we removing the dockershim.sock mount from the aws-node manifest?  I think we should do it in this PR, since it is coupled to this code change anyway.

jayanthvn:
Closing this PR since this will be handled as part of v1.11.0 release.

anguslees:
> Closing this PR since this will be handled as part of v1.11.0 release

I don't think this happened as part of v1.11 release?

As far as I can see we're still exposing dockershim.sock (host root equivalent!) to the aws-node container.

Please reopen this issue.  Given the history of repeated delays on this fix, I suggest not closing it until after the fix is _actually_ in a release. 😅

jayanthvn:
> > Closing this PR since this will be handled as part of v1.11.0 release
> 
> 
> 
> I don't think this happened as part of v1.11 release?
> 
> 
> 
> As far as I can see we're still exposing dockershim.sock (host root equivalent!) to the aws-node container.
> 
> 
> 
> Please reopen this issue/PR.  Given the history of repeated delays on this fix, I suggest not closing it until after the fix is _actually_ in a release. 😅 

Hi Gus. Yes this wasn't removed as part of 1.11. Yang was working on different PR to handle this. Will reopen the PR until the new PR is open.

TBBle:
Did the replacement PR ever surface? https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/master/aws-k8s-cni-cn.yaml still shows dockershim mounts.

jayanthvn:
@TBBle - Yes the sock mounts are not removed yet. We are working on a prototype before completely moving to state file.

jayanthvn:
Moved to checkpoint migration 2 and switched to use state file instead of CRI socket for IP allocation pool restore - https://github.com/aws/amazon-vpc-cni-k8s/pull/2110. This is as part of 1.12.0 release - https://github.com/aws/amazon-vpc-cni-k8s/releases/tag/v1.12.0. Hence closing the issue.

PR #1353, opened 2021-01-20
Change defaultLogLevel to Info

sidorovdmitry:
**What type of PR is this?**
cleanup
documentation

None

The log level "Debug" should not be used by default on any system except cases when it's really necessary.
Right now VPC CNI Plugin and the metrics helper are running with the log level "Debug" by default.

`{"level":"info","ts":"2021-01-20T15:29:24.525Z","caller":"runtime/proc.go:203","msg":"Starting CNIMetricsHelper. Sending metrics to CloudWatch: true, LogLevel Debug"}`

Changed the verbosity of logs to Info by default


config/v1.7/aws-k8s-cni-cn.yaml
```diff
@@ -130,7 +130,7 @@
         - "name": "AWS_VPC_K8S_CNI_EXTERNALSNAT"
           "value": "false"
         - "name": "AWS_VPC_K8S_CNI_LOGLEVEL"
-          "value": "DEBUG"
+          "value": "INFO"
```
jayanthvn, line 133: You might want to change here - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/master/manifests.jsonnet#L169 and regenerate the configs under master folder also change has to be done here https://github.com/aws/amazon-vpc-cni-k8s/blob/master/charts/aws-vpc-cni/values.yaml#L44.

jayanthvn (changes requested):

config/v1.7/aws-k8s-cni-cn.yaml
sidorovdmitry: Done, thank you.

jayanthvn: Thank you! We will review the PR.

jayanthvn: Generally we keep this as debug mainly to collect the logs when the issue happens and most of the time repros are harder especially on prod environments. Can you please let us know why you would prefer to keep it default since it is already customizable?

sidorovdmitry: The general use case looks like this: people open the official documentation (let's say [this page](https://docs.aws.amazon.com/eks/latest/userguide/cni-metrics-helper.html)), get the URL of the manifest, and deploy it to their k8s clusters. There is no specific information on how to configure the application, so people could assume that it's stable, secure, and reliable by default. Then people open logs and see numerous messages with the list of metrics, values, and some internal information about objects in Golang:
Screenshot 2021-01-20 at 23 51 57

I'm not that confident that such log messages are useful and help to find the root cause of a possible issue. We just get huge logs that are pretty hard to read and analyze. In addition to that, debug logs are not free in terms of system resources and performance, and in some cases can be not secure.
So my idea is to enable the `DEBUG` level only when it's necessary (let's say development environment) and use something more silent by default, especially on production clusters. Most of the users don't need this detailed information and can increase verbosity if they want to.

jayanthvn: Yes I agree we need to update the documentation on this, maybe have a section for optimizations. You can please open an issue for documentation and I can collate all the recommendations and put it in a doc. But the logs which you are sharing here is from "cni metrics helper" (https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/master/cni-metrics-helper.yaml) and the changes you have done is for log level in ipamd.log/plugin.log which generally captures issues like pod creates/deletes, IP exhaustion, ENI exhaustion, IPAMD failed to start, CNI version, number of IPs used and so on which is very useful to debug issues.

PR #1355, opened 2021-01-20
revert k8s version since it broke circle ci

jayanthvn:
**What type of PR is this?**
bug

#1347

upgraded k8s version but the tar is not existing. reverting for now to stabilize integration testing.

yes

**Automation added to e2e**:

yes

no

no
```release-note

```


couralex6 (approved):
LGTM

PR #1350, opened 2021-01-15
upgrade github.com/containernetworking/plugins to v0.9.0

fr0stbyte:
**What type of PR is this?**
feature

**Which issue does this PR fix**:

https://github.com/containernetworking/plugins/pull/553

**What does this PR do / Why do we need it**:

conntrack table is not flushed for UDP hostPorts when pod is deleted. This has been fixed in the newer version of containernetworking/plugins (PR with the fix is linked above)

**If an issue # is not available please add repro steps and logs from IPAMD/CNI showing the issue**:

**Testing done on this change**:

**Automation added to e2e**:

n/a

**Will this break upgrades or downgrades. Has updating a running cluster been tested?**:

Not tested beyond unittests

**Does this change require updates to the CNI daemonset config files to work?**:

**Does this PR introduce any user-facing change?**:
Not that I am aware of

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1362, opened 2021-01-27
update plugins to v0.9.0

fr0stbyte:
**What type of PR is this?**
bug

complementary PR for https://github.com/aws/amazon-vpc-cni-k8s/pull/1350

although the dependencies pull in the correct version for `containernetworkking/plugins` the actual binaries are not built, but fetch from the release page of the project. This bumps the project version to bring everything inline.

We have rolled the patched cni to an internal cluster and verified that we're not seeing the drop of UDP traffic upon restart of the destination pods.

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -199,7 +199,7 @@ generate-limits:    ## Generate limit file go code
 	go run scripts/gen_vpc_ip_limits.go
 
 # Fetch the CNI plugins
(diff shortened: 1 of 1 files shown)


fawadkhaliq (approved):
thanks!

jayanthvn (approved):

PR #1364, opened 2021-01-28
ASG returns null for perf test

jayanthvn:
**What type of PR is this?**
bug

ASG query on performance test returns null hence all cases fail.

Add "cluster-autoscaler":{"enable":true}}" while creating k8s tester and to get the ASG name use describe node group.

N/A

Have started perf test on my account
**Automation added to e2e**:

no

no

no

```release-note

```


couralex6 (approved):
LGTM

PR #1365, opened 2021-01-29
Output to stderr when no log file path is passed

couralex6:
Cherry picked PR #1275 :

**What type of PR is this?**
Bug fix

#1265 

Customers using Cilium in chaining mode were unable to create new pods after upgrading to CNI v1.7.x. Because no log file path was passed, zaplogger.go would default to stdout. 
This change ensures that logs are being written to stderr when no log file path is passed.

This was tested on a cluster with [Cilium enabled in chaining mode ](https://docs.cilium.io/en/v1.9.0-rc2/gettingstarted/cni-chaining-aws-cni/). I reproduced the bug by installing CNI v1.7.5, which prevented new pods from being created. Pods were able to come up with this fix.

**Automation added to e2e**:
Added Unit Tests to test the following scenarios:
* Empty log file path > log to `zapcore.Lock(os.Stderr)`
* Actual file path > log to file path
* String set to 'stdout' > log to `zapcore.Lock(os.Stdout)`

No

No

No

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,7 +318,9 @@ Default: `/var/log/aws-routed-eni/plugin.log`
 Valid Values: `stderr` or a file path
 
 Specifies where to write the logging output for `aws-cni` plugin. Either to `stderr` or to override the default file (i.e., `/var/log/aws-routed-eni/plugin.log`).
-`Stdout` cannot be supported for plugin log, please refer to #1248 for more details.
+`Stdout` cannot be supported for plugin log, please refer to [#1248](https://github.com/aws/amazon-vpc-cni-k8s/issues/1248) for more details.
+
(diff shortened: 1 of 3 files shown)


fawadkhaliq:
In the PR description, let's please link the PR being backported and also consider copy-pasta of the description from the original PR

_(same comment applies to all the other backport PRs authored for 1.7.9)_

jayanthvn:
I agree with Fawad, also please indicate if the PR is cherrypicked or modified. Maybe in the subject line we can add as "Cherry-pick #<issue number> for rel 1.7.9"

fawadkhaliq (approved):

jayanthvn (approved):

PR #1367, opened 2021-01-29
update github.com/containernetworking/plugins to v0.9.0

couralex6:
Cherry picked PR #1350 :

**What type of PR is this?**
feature

https://github.com/containernetworking/plugins/pull/553

conntrack table is not flushed for UDP hostPorts when pod is deleted. This has been fixed in the newer version of containernetworking/plugins (PR with the fix is linked above)

**Automation added to e2e**:

n/a

Not tested beyond unittests

Not that I am aware of


fawadkhaliq:
Not the same diff as https://github.com/aws/amazon-vpc-cni-k8s/pull/1350 - what else are we adding here?

jayanthvn (approved):

PR #1370, opened 2021-01-30
Adds http timeout to aws sessions

couralex6:
This pr manually imports the changes brought by PR #1331 

**What type of PR is this?**
Bug fix.

#1256 

**Automation added to e2e**:

None

No
```release-note

```

### DIFF
--- a/pkg/awsutils/awssession/session.go
+++ b/pkg/awsutils/awssession/session.go
@@ -1,0 +1,81 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package awssession
+
+import (
+	"fmt"
+	"net/http"
+	"os"
+
+	"strconv"
+	"time"
+
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/request"
+	"github.com/aws/aws-sdk-go/aws/session"
+)
+
+// Http client timeout env for sessions
+const (
+	httpTimeoutEnv = "HTTP_TIMEOUT"
+	maxRetries     = 15
+)
+
+var (
+	version string
+	log     = logger.Get()
+	// HTTP timeout default value in seconds (10 seconds)
+	httpTimeoutValue = 10 * time.Second
+)
+
+func getHTTPTimeout() time.Duration {
+	httpTimeoutEnvInput := os.Getenv(httpTimeoutEnv)
+	// if httpTimeout is not empty, we convert value to int and overwrite default httpTimeoutValue
+	if httpTimeoutEnvInput != "" {
+		input, err := strconv.Atoi(httpTimeoutEnvInput)
+		if err == nil && input >= 10 {
+			log.Debugf("Using HTTP_TIMEOUT %v", input)
+			httpTimeoutValue = time.Duration(input) * time.Second
+			return httpTimeoutValue
+		}
+	}
+	log.Info("HTTP_TIMEOUT env is not set or set to less than 10 seconds, defaulting to httpTimeout to 10sec")
+	return httpTimeoutValue
+}
+
+// New will return an session for service clients
+func New() *session.Session {
+	awsCfg := aws.Config{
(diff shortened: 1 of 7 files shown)


pkg/awsutils/awssession/session.go
```diff
@@ -0,0 +1,81 @@
...
+	// if httpTimeout is not empty, we convert value to int and overwrite default httpTimeoutValue
+	if httpTimeoutEnvInput != "" {
+		input, err := strconv.Atoi(httpTimeoutEnvInput)
+		if err == nil && input >= 10 {
+			log.Debugf("Using HTTP_TIMEOUT %v", input)
+			httpTimeoutValue = time.Duration(input) * time.Second
+			return httpTimeoutValue
+		}
+	}
+	log.Warn("HTTP_TIMEOUT env is not set or set to less than 10 seconds, defaulting to httpTimeout to 10sec")
```
fawadkhaliq, line 54: log.Warn -> log.Info 

Nothing to warn about as most users will just go with the default. Consider logging a warning if the env variable is set and the value falls out of bounds or is invalid otherwise (i.e not a valid integar)

pkg/awsutils/awssession/session.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/request"
+	"github.com/aws/aws-sdk-go/aws/session"
+)
+
+// Http client timeout env for sessions
+const (
+	httpTimeoutEnv = "HTTP_TIMEOUT"
+	maxRetries     = 10
```
fawadkhaliq, line 33: Let's keep it the same value as before

pkg/awsutils/awssession/session.go
```diff
@@ -0,0 +1,81 @@
...
+	return sess
+}
+
+// injectUserAgent will inject app specific user-agent into awsSDK
+func injectUserAgent(handlers *request.Handlers) {
+	handlers.Build.PushFrontNamed(request.NamedHandler{
+		Name: fmt.Sprintf("%s/user-agent", "amazon-vpc-cni-k8s"),
+		Fn: request.MakeAddToUserAgentHandler(
+			"amazon-vpc-cni-k8s",
+			"version/"+version),
```
fawadkhaliq, line 79: nit: like line#76, use fmt.Sprintf here as well?

pkg/ec2wrapper/ec2wrapper.go
```diff
@@ -30,15 +30,15 @@ type EC2Wrapper struct {
...
+	sess := awssession.New()
 	ec2MetadataClient := ec2metadatawrapper.New(nil)
 
 	instanceIdentityDocument, err := ec2MetadataClient.GetInstanceIdentityDocument()
 	if err != nil {
 		return &EC2Wrapper{}, err
 	}
 
-	ec2ServiceClient := ec2.New(metricsSession, aws.NewConfig().WithMaxRetries(maxRetries).WithRegion(instanceIdentityDocument.Region))
+	ec2ServiceClient := ec2.New(sess, aws.NewConfig().WithMaxRetries(maxRetries).WithRegion(instanceIdentityDocument.Region))
```
fawadkhaliq, line 41: > aws.NewConfig().WithMaxRetries(maxRetries).WithRegion(instanceIdentityDocument.Region)

Isn't this overriding the config supplied in the session?

fawadkhaliq (changes requested):

jayanthvn (approved):

pkg/publisher/publisher.go
```diff
@@ -107,7 +107,7 @@ func New(ctx context.Context) (Publisher, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "publisher: Unable to obtain region")
 	}
-	cloudwatchClient := cloudwatch.New(awsSession, aws.NewConfig().WithMaxRetries(cloudwatchClientMaxRetries).WithRegion(region))
+	cloudwatchClient := cloudwatch.New(sess, aws.NewConfig().WithMaxRetries(cloudwatchClientMaxRetries).WithRegion(region))
```
fawadkhaliq, line 110: Need to change it here as well

pkg/ec2wrapper/ec2wrapper.go
```diff
@@ -30,15 +30,15 @@ type EC2Wrapper struct {
...
+	sess := awssession.New()
 	ec2MetadataClient := ec2metadatawrapper.New(nil)
 
 	instanceIdentityDocument, err := ec2MetadataClient.GetInstanceIdentityDocument()
 	if err != nil {
 		return &EC2Wrapper{}, err
 	}
 
-	ec2ServiceClient := ec2.New(metricsSession, aws.NewConfig().WithMaxRetries(maxRetries).WithRegion(instanceIdentityDocument.Region))
+	ec2ServiceClient := ec2.New(sess)
```
fawadkhaliq, line 41: Region info has been removed in the last patch and is not passed from elsewhere. You will need to configure it

fawadkhaliq (changes requested):

Issue #1375, opened 2021-02-05
Clean up vlan ip rules before setting up pod networks


SaranBalaji90:
**What happened**:
When pods are deleted without TerminationGracePeriod, host veth devices for the pods using security group are not removed as Ipamd plugin will not be able to query APIServer for the deleted pod. Therefore when new pods are created, hostveth uses invalid vlan routing table.

**What you expected to happen**:
Before setting up new pods, clean up all existing vlan id from ip rule.

**How to reproduce it (as minimally and precisely as possible)**:
By force deleting the pod using security group, ipamd leaves dangling resources in ip rules.

Similar to this - 
https://github.com/aws/amazon-vpc-cni-k8s/blob/1c1d4b91e8d21d979cad494fe8228c464d1d4d3b/cmd/routed-eni-cni-plugin/driver/driver.go#L491 we should perform clean up here to delete the hostVeth from ip rule 1b. https://github.com/aws/amazon-vpc-cni-k8s/blob/1c1d4b91e8d21d979cad494fe8228c464d1d4d3b/cmd/routed-eni-cni-plugin/driver/driver.go#L316

Issue #1374, opened 2021-02-05
Existing pod network not cleanedup when using security groups with stateful sets


hintofbasil:
**What happened**:
When using security groups with stateful sets we noticed that pods often lost connectivity when restarted.
The security group they were bound to allowed all connections inbound and outbound on 0.0.0.0/0.

After some investigation we discovered the bug seems to affect pods re-created on the same node with the same name.

**Attach logs**
[eks_i-08aff468a2d6ce527_2021-02-05_1421-UTC_0.6.2.tar.gz](https://github.com/aws/amazon-vpc-cni-k8s/files/5933132/eks_i-08aff468a2d6ce527_2021-02-05_1421-UTC_0.6.2.tar.gz)

**What you expected to happen**:

The pod should launch normally

**How to reproduce it (as minimally and precisely as possible)**:

Create a securityGroupPolicy

```yaml
apiVersion: vpcresources.k8s.aws/v1beta1
kind: SecurityGroupPolicy
metadata:
  name: cni-test
  namespace: default
spec:
  podSelector:
    matchLabels:
      app: cni-test
  securityGroups:
    groupIds:
      - <id>
```

Create a pod which uses the security group policy

```yaml
kind: Pod
metadata:
  name: cni-test
  namespace: default
  labels:
    app: cni-test
spec:
  containers:
    - name: alpine
      image: alpine
      command:
      - sleep
      - "1000000000"
```

Kill the pod then recreate the pod ensuring it is scheduled to the same node.  Then attempt to make an outbound connection from the pod

```bash
apk add curl
curl 1.1.1.1
```

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):

```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"clean", BuildDate:"2020-11-12T01:09:16Z", GoVersion:"go1.15.4", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.9-eks-d1db3c", GitCommit:"d1db3c46e55f95d6a7d3e5578689371318f95ff9", GitTreeState:"clean", BuildDate:"2020-10-20T22:18:07Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

[...]

PR #1376, opened 2021-02-05
Fix deletion of hostVeth rule for pods using security group

SaranBalaji90:
**What type of PR is this?**
bug

https://github.com/aws/amazon-vpc-cni-k8s/issues/1375
https://github.com/aws/amazon-vpc-cni-k8s/issues/1374

This cleans up any dangling resources for old pods using same name. This might happen if pods are force deleted.

Tested on dev cluster and verified the dangling rules are deleted before adding new routes.
```
[ec2-user@ip-192-168-65-167 ~]$ ip rule
10:	from all iif vlan.eth.1 lookup 101 
10:	from all iif vlan.eth.2 lookup 102 
10:	from all iif vlanaad671050aa [detached] lookup 101 
20:	from all lookup local 
[ec2-user@ip-192-168-65-167 ~]$ ip rule
10:	from all iif vlan.eth.1 lookup 101 
10:	from all iif vlan.eth.2 lookup 102 
10:	from all iif vlanaad671050aa lookup 102 
20:	from all lookup local 
[ec2-user@ip-192-168-65-167 ~]$ 
```

**Automation added to e2e**:
NA

No

No

No

```release-note

```

### DIFF
--- a/cmd/routed-eni-cni-plugin/driver/driver.go
+++ b/cmd/routed-eni-cni-plugin/driver/driver.go
@@ -313,12 +313,26 @@ func (os *linuxNetwork) SetupPodENINetwork(hostVethName string, contVethName str
 	vlanTableID := vlanID + 100
 	vlanLink := buildVlanLink(vlanID, parentIfIndex, eniMAC)
 
-	// 1. clean up if vlan already exists (necessary when trunk ENI changes).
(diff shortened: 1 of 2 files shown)


jayanthvn (approved):
LGTM :)

Issue #1074, opened 2020-07-06
Upgrade eniconfigs.crd.k8s.amazonaws.com to apiextensions.k8s.io/v1 by k8s 1.22


TBBle:
Per https://github.com/kubernetes/kubernetes/issues/82022, the currently-used API version [`apiextensions.k8s.io/v1beta1`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#customresourcedefinition-v1beta1-apiextensions-k8s-io) will stop being served by default in k8s 1.22.

This is worth looking at sooner rather than later, as it's not a trivial change. [`apiextensions.k8s.io/v1`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#customresourcedefinition-v1-apiextensions-k8s-io) makes a [structural schema mandatory](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema), and [the current CRD definition](https://github.com/aws/amazon-vpc-cni-k8s/blob/afdb125d67eeed00a75433a14548a735430ab2d9/config/master/aws-k8s-cni.yaml#L36-L50) doesn't provide a schema at all.

I note that this might relate to #745, as the operator-sdk only gained support for generating v1 CRDs in [0.17.0](https://github.com/operator-framework/operator-sdk/releases/tag/v0.17.0), and they became the default in [0.18.0](https://github.com/operator-framework/operator-sdk/releases/tag/v0.18.0). That said, I'm not sure if that matters, or if the CRD can be manually updated despite the old operator-sdk.

adawalli:
Any traction on this? I ran into this today when I realized that the current definition isn't compatible with Terraform's K8s Alpha plugin due to the lack of schema

jayanthvn:
Hi @adawalli 

Yes we will look into it.

jayanthvn:
There is already a PR #1390 and regarding the schema @achevuru can we see if can take it up as part of #1419?

stevehipwell:
This should be easy enough to fix with an empty schema for ENIConfig v1alpha1 and then maybe an updated ENIConfig version with a typed schema in the future. I've add a https://github.com/aws/amazon-vpc-cni-k8s/pull/1390#issuecomment-876220253 to the PR with the simple addition required.

PR #1390, opened 2021-02-25
update templates to use v1 of apiextensions

marcincuber:
**What type of PR is this?**

Update

With Kubernetes 1.19, apiextensions.k8s.io/v1beta1 has been deprecated. Latest templates of vpc cni should be using latest api versions which aren't deprecated.

### DIFF
--- a/config/v1.7/aws-k8s-cni-cn.yaml
+++ b/config/v1.7/aws-k8s-cni-cn.yaml
@@ -51,7 +51,7 @@
   - "list"
   - "watch"
 ---
-"apiVersion": "apiextensions.k8s.io/v1beta1"
+"apiVersion": "apiextensions.k8s.io/v1"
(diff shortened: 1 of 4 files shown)


TBBle:
Closes: #1074.

I'm not sure if that'll work to link them, it might need to be in the description.

stevehipwell:
The CRDs should have the following item added to the version under `versions` which will make them valid `v1` CRDs without needing to define the schema.

```yaml
    schema:
      openAPIV3Schema:
        type: object
        x-kubernetes-preserve-unknown-fields: true
```

jayanthvn:
Thanks for the PR and changes is already in the master from another PR [https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/master/aws-k8s-cni.yaml#L68].

Fixes #1074

PR #1396, opened 2021-03-04
Multi card support - Prevent route override for primary ENI across multi-cards ENAs

jayanthvn:
**What type of PR is this?**
Feature

No

ENAs can be on multi-card (P4 family) and CNI currently doesn't support non-zero network card index. So on nodeInit, CNI thinks all the attached ENIs are on card 0, hence this PR will make CNI aware of that. But CNI doesnt allocate pods or secondary IPs on non-zero network cards. Support will be in future releases.

N/A

NodeInit
```
{"level":"info","ts":"2021-03-02T19:04:57.922Z","caller":"ipamd/ipamd.go:363","msg":"Got network cardindex 0 for ENI eni-0aa1c65af9aa0f115"}
{"level":"info","ts":"2021-03-02T19:04:57.922Z","caller":"ipamd/ipamd.go:363","msg":"Got network cardindex 1 for ENI eni-0535397aaad78b390"}
{"level":"info","ts":"2021-03-02T19:04:57.922Z","caller":"ipamd/ipamd.go:363","msg":"Got network cardindex 2 for ENI eni-02142cb550d03b63b"}
{"level":"info","ts":"2021-03-02T19:04:57.922Z","caller":"ipamd/ipamd.go:363","msg":"Got network cardindex 3 for ENI eni-04e0ddb1cc00fca31"}
{"level":"debug","ts":"2021-03-02T19:04:57.922Z","caller":"ipamd/ipamd.go:312","msg":"DescribeAllENIs success: ENIs: 4, tagged: 0"}
{"level":"debug","ts":"2021-03-02T19:04:57.922Z","caller":"ipamd/ipamd.go:370","msg":"Skipping ENI eni-04e0ddb1cc00fca31: since on non-zero network card"}
{"level":"debug","ts":"2021-03-02T19:04:57.922Z","caller":"ipamd/ipamd.go:370","msg":"Skipping ENI eni-0535397aaad78b390: since on non-zero network card"}
{"level":"debug","ts":"2021-03-02T19:04:57.922Z","caller":"ipamd/ipamd.go:370","msg":"Skipping ENI eni-02142cb550d03b63b: since on non-zero network card"}
{"level":"debug","ts":"2021-03-02T19:04:57.922Z","caller":"ipamd/ipamd.go:312","msg":"Discovered ENI eni-0aa1c65af9aa0f115, trying to set it up"}
{"level":"debug","ts":"2021-03-02T19:04:57.922Z","caller":"ipamd/ipamd.go:768","msg":"DataStore Add an ENI eni-0aa1c65af9aa0f115"}
```

Reconciler - 
```
{"level":"debug","ts":"2021-03-02T19:06:03.610Z","caller":"ipamd/ipamd.go:962","msg":"Skipping ENI eni-02142cb550d03b63b: since on non-zero network card"}
{"level":"debug","ts":"2021-03-02T19:06:03.610Z","caller":"ipamd/ipamd.go:962","msg":"Skipping ENI eni-04e0ddb1cc00fca31: since on non-zero network card"}
{"level":"debug","ts":"2021-03-02T19:06:03.610Z","caller":"ipamd/ipamd.go:962","msg":"Skipping ENI eni-0535397aaad78b390: since on non-zero network card"}
{"level":"debug","ts":"2021-03-02T19:06:03.610Z","caller":"ipamd/ipamd.go:504","msg":"Reconcile existing ENI eni-0aa1c65af9aa0f115 IP pool"}
{"level":"debug","ts":"2021-03-02T19:06:03.610Z","caller":"ipamd/ipamd.go:1007","msg":"Reconcile and skip primary IP 192.168.72.127 on ENI eni-0aa1c65af9aa0f115"}
{"level":"debug","ts":"2021-03-02T19:06:03.610Z","caller":"ipamd/ipamd.go:504","msg":"Successfully Reconciled ENI/IP pool"}
{"level":"debug","ts":"2021-03-02T19:06:03.610Z","caller":"ipamd/ipamd.go:504","msg":"IP Address Pool stats: total: 49, assigned: 0"}
{"level":"debug","ts":"2021-03-02T19:07:03.617Z","caller":"ipamd/ipamd.go:504","msg":"Reconciling ENI/IP pool info because time since last 1m0.079149379s <= 1m0s"}
```

RefreshSGIDs - 

```
{"level":"debug","ts":"2021-03-02T19:04:58.014Z","caller":"awsutils/awsutils.go:592","msg":"Found CIDR 192.168.64.0/19 for ENI 0a:f7:72:1b:8c:8b"}
{"level":"debug","ts":"2021-03-02T19:04:58.018Z","caller":"awsutils/awsutils.go:592","msg":"Found IP addresses [192.168.70.10] on ENI 0a:f7:72:1b:8c:8b"}
{"level":"debug","ts":"2021-03-02T19:04:58.018Z","caller":"ipamd/ipamd.go:319","msg":"Update ENI eni-0aa1c65af9aa0f115"}
{"level":"info","ts":"2021-03-02T19:04:58.443Z","caller":"ipamd/ipamd.go:325","msg":"Found 192.168.0.0/16, added to ipamd cache"}
{"level":"info","ts":"2021-03-02T19:04:58.444Z","caller":"aws-k8s-agent/main.go:69","msg":"Serving RPC Handler version v1.8-dirty on 127.0.0.1:50051"}
```

**Automation added to e2e**:

Yes

No

No

```release-note

```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -155,6 +155,16 @@ type APIs interface {
 
 	// WaitForENIAndIPsAttached waits until the ENI has been attached and the secondary IPs have been added
 	WaitForENIAndIPsAttached(eni string, wantedSecondaryIPs int) (ENIMetadata, error)
+
+	//SetCNIunmanaged ENI
+	SetCNIUnmanagedENIs(eniID []string) error
+    
+	//isCNIUnmanagedENI
+	IsCNIUnmanagedENI(eniID string) bool
+    
+	//RefreshSGIDs
+	RefreshSGIDs(mac string) error 
+    
 }
 
 // EC2InstanceMetadataCache caches instance metadata
@@ -171,6 +181,7 @@ type EC2InstanceMetadataCache struct {
 	region              string
 	unmanagedENIs       StringSet
 	useCustomNetworking bool
+	cniunmanagedENIs    StringSet
 
 	imds   TypedIMDS
 	ec2SVC ec2wrapper.EC2
@@ -219,6 +230,7 @@ type DescribeAllENIsResult struct {
 	TagMap      map[string]TagMap
 	TrunkENI    string
 	EFAENIs     map[string]bool
+	MultiCardENIIDs []string
 }
 
 // msSince returns milliseconds since start.
@@ -397,17 +409,7 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context) 
 		return err
 	}
 	log.Debugf("Found subnet-id: %s ", cache.subnetID)
-
-	// retrieve security groups
-	err = cache.refreshSGIDs(mac)
-	if err != nil {
-		return err
-	}
-
-	// Refresh security groups and VPC CIDR blocks in the background
-	// Ignoring errors since we will retry in 30s
-	go wait.Forever(func() { _ = cache.refreshSGIDs(mac) }, 30*time.Second)
-
+	
 	// We use the ctx here for testing, since we spawn go-routines above which will run forever.
 	select {
 	case <-ctx.Done():
@@ -417,8 +419,8 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context) 
 	return nil
 }
 
-// refreshSGIDs retrieves security groups
-func (cache *EC2InstanceMetadataCache) refreshSGIDs(mac string) error {
+// RefreshSGIDs retrieves security groups
+func (cache *EC2InstanceMetadataCache) RefreshSGIDs(mac string) error {
 	ctx := context.TODO()
 
 	sgIDs, err := cache.imds.GetSecurityGroupIDs(ctx, mac)
@@ -457,7 +459,8 @@ func (cache *EC2InstanceMetadataCache) refreshSGIDs(mac string) error {
 		newENIs := StringSet{}
 		newENIs.Set(eniIDs)
 
-		filteredENIs := newENIs.Difference(&cache.unmanagedENIs)
+		tempfilteredENIs := newENIs.Difference(&cache.cniunmanagedENIs)
+		filteredENIs := tempfilteredENIs.Difference(&cache.unmanagedENIs)
 
 		sgIDsPtrs := aws.StringSlice(sgIDs)
 		// This will update SG for managed ENIs created by EKS.
@@ -1034,13 +1037,19 @@ func (cache *EC2InstanceMetadataCache) DescribeAllENIs() (DescribeAllENIsResult,
 
 	// Collect ENI response into ENI metadata and tags.
 	var trunkENI string
+	var multiCardENIIDs []string
 	efaENIs := make(map[string]bool, 0)
 	tagMap := make(map[string]TagMap, len(ec2Response.NetworkInterfaces))
 	for _, ec2res := range ec2Response.NetworkInterfaces {
+		log.Infof("Got network cardindex %v for ENI %v", aws.Int64Value(ec2res.Attachment.NetworkCardIndex), aws.StringValue(ec2res.NetworkInterfaceId))
 		if ec2res.Attachment != nil && aws.Int64Value(ec2res.Attachment.DeviceIndex) == 0 && !aws.BoolValue(ec2res.Attachment.DeleteOnTermination) {
 			log.Warn("Primary ENI will not get deleted when node terminates because 'delete_on_termination' is set to false")
 		}
 		eniID := aws.StringValue(ec2res.NetworkInterfaceId)
+		if aws.Int64Value(ec2res.Attachment.NetworkCardIndex) > 0 {
+			multiCardENIIDs = append(multiCardENIIDs, eniID)
+		}
+
 		eniMetadata := eniMap[eniID]
 		interfaceType := aws.StringValue(ec2res.InterfaceType)
 
@@ -1065,6 +1074,7 @@ func (cache *EC2InstanceMetadataCache) DescribeAllENIs() (DescribeAllENIsResult,
 		TagMap:      tagMap,
 		TrunkENI:    trunkENI,
 		EFAENIs:     efaENIs,
+		MultiCardENIIDs: multiCardENIIDs,
 	}, nil
 }
 
@@ -1506,4 +1516,20 @@ func (cache *EC2InstanceMetadataCache) getENIsFromPaginatedDescribeNetworkInterf
 		return err
 	}
 	return innerErr
+}
+
+//SetCNIUnmanagedENIs Set unmanaged ENI set
+func (cache *EC2InstanceMetadataCache) SetCNIUnmanagedENIs(eniID []string) error {
+    if len(eniID) != 0 {
+		cache.cniunmanagedENIs.Set(eniID)
+ 	}
+	return nil
+}
+
+//IsCNIUnmanagedENI returns if the eni is unmanaged
+func (cache *EC2InstanceMetadataCache) IsCNIUnmanagedENI(eniID string) bool {
+	if len(eniID) != 0 {
+		return cache.cniunmanagedENIs.Has(eniID)
+	}
+	return false
 }
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -361,8 +361,7 @@ func (c *IPAMContext) nodeInit() error {
 	log.Debugf("DescribeAllENIs success: ENIs: %d, tagged: %d", len(metadataResult.ENIMetadata), len(metadataResult.TagMap))
 	c.awsClient.SetCNIUnmanagedENIs(metadataResult.MultiCardENIIDs)
 	c.setUnmanagedENIs(metadataResult.TagMap)
-	tempENIIsMetadata := c.filterCNIUnmanagedENIs(metadataResult.ENIMetadata)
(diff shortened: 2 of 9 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -155,6 +155,16 @@ type APIs interface {
 
 	// WaitForENIAndIPsAttached waits until the ENI has been attached and the secondary IPs have been added
 	WaitForENIAndIPsAttached(eni string, wantedSecondaryIPs int) (ENIMetadata, error)
+
+	//SetCNIunmanaged ENI
+	SetCNIUnmanagedENIs(eniID []string) error
```
achevuru, line 160: Another option to consider - Why can't we instead just enhance existing `SetUnmanagedENIs` and `IsUnmanagedENI` functions to also consider ENIs belonging to Non-Zero NW cards instead of managing them via separate functions? Might be easy to remove the additional conditions when we support them eventually - instead of the separate functions this CR utilizes?

pkg/awsutils/awsutils.go
```diff
@@ -219,6 +230,7 @@ type DescribeAllENIsResult struct {
 	TagMap      map[string]TagMap
 	TrunkENI    string
 	EFAENIs     map[string]bool
+	MultiCardENIIDs []string
```
achevuru, line 233: To continue on the above comment - Maybe we can just check if the ENI is part of `MultiCardENIIDs` as well in `IsUnmanagedENI`. Same goes to `SetUnmanagedENIs`, we can append `MultiCardENIIDs`. Any issues do you see with that approach?

pkg/ipamd/ipamd.go
```diff
@@ -311,6 +312,17 @@ func New(k8sapiClient kubernetes.Interface, eniConfig *eniconfig.ENIConfigContro
 	if err != nil {
 		return nil, err
 	}
+
+	mac := c.awsClient.GetPrimaryENImac()
+	// retrieve security groups
+	err = c.awsClient.RefreshSGIDs(mac)
```
achevuru, line 318: Is this required? I'm assuming we're not going to update SGs on ENIs of Non-Zero NW cards anyways, right? Is this code block addressing some other known issue?

pkg/awsutils/awsutils.go
jayanthvn: Yes I did think about that, but today any ENIs tagged with "no_manage" is stored in that cache so didn't want to deviate from that. Also in future if we have a use case where CNI needs to manage ENIs we can use the new cache or decommission once we have the support. Since we are planning for supporting this very soon, didn't want to modify existing routines.

pkg/awsutils/awsutils.go
jayanthvn: Yeah kept it separate for readability. If we have a single DB then we can use the same routines but yes I wanted to keep it separate so that it will be easy to remove in the future release.

pkg/ipamd/ipamd.go
jayanthvn: Yeah, so this was called in NodeInit and this function gets attached ENIs and updates the SGs. Which again thinks the ENIs are secondary ENIs. So moved it after describe call since by then we would have built local cache on CNIUnmanaged ENIs.

pkg/ipamd/ipamd.go
```diff
@@ -194,6 +194,7 @@ type IPAMContext struct {
 	maxIPsPerENI         int
 	maxENI               int
 	unmanagedENI         int
+	cniunmanagedENI      int
```
fawadkhaliq, line 197: What's the use for `cniunmanagedENI` ?

pkg/ipamd/ipamd.go
```diff
@@ -1321,3 +1344,21 @@ func (c *IPAMContext) SetNodeLabel(key, value string) error {
...
+		// If we have CNI unmanaged ENIs, filter them out
+		if c.awsClient.IsCNIUnmanagedENI(eni.ENIID) {
+			log.Debugf("Skipping ENI %s: since on non-zero network card", eni.ENIID)
+			numFiltered++
+			continue
+		}
+		ret = append(ret, eni)
+	}
+	c.cniunmanagedENI = numFiltered
+	c.updateIPStats(numFiltered)
```
fawadkhaliq, line 1362: `updateIPStats ` updates the `enisMax ` and `ipMax` based on the absolute max minus what is passed as in input to this function.

```
func (c *IPAMContext) updateIPStats(unmanaged int) {
	ipMax.Set(float64(c.maxIPsPerENI * (c.maxENI - unmanaged)))
	enisMax.Set(float64(c.maxENI - unmanaged))
}
```

So this code conflicts and overrides what is done in function `filterUnmanagedENIs`. Whoever runs last (`filterCNIUnmanagedENIs` or `filterUnmanagedENIs`) will override the unmanaged number. This will be result in inaccurate `ipMax` and `enisMax` for node with multiple cards.

You can either merge the unmanaged ENIs in the same function (better approach) or update `updateIPStats` to handle this scenario.

fawadkhaliq (changes requested):

pkg/ipamd/ipamd.go
```diff
@@ -1321,3 +1344,21 @@ func (c *IPAMContext) SetNodeLabel(key, value string) error {
 func (c *IPAMContext) GetPod(podName, namespace string) (*v1.Pod, error) {
 	return c.k8sClient.CoreV1().Pods(namespace).Get(podName, metav1.GetOptions{})
 }
+
+// filterCNIUnmanagedENIs filters out ENIs on non-zero network cards such as p4 family
+func (c *IPAMContext) filterCNIUnmanagedENIs(enis []awsutils.ENIMetadata) []awsutils.ENIMetadata {
```
fawadkhaliq, line 1349: Why not just update the existing `filterUnmanagedENIs ` to handle the network cards? it'll be much simpler. You could 


Something like
```
// filterUnmanagedENIs filters out ENIs marked with the "node.k8s.amazonaws.com/no_manage" tag
func (c *IPAMContext) filterUnmanagedENIs(enis []awsutils.ENIMetadata) []awsutils.ENIMetadata {
	numFiltered := 0
	ret := make([]awsutils.ENIMetadata, 0, len(enis))
	for _, eni := range enis {
		// If we have unmanaged ENIs, filter them out
		if c.awsClient.IsUnmanagedENI(eni.ENIID) {
			log.Debugf("Skipping ENI %s: tagged with %s", eni.ENIID, eniNoManageTagKey)
			numFiltered++
			continue
		} else if c.awsClient.IsMulticardENI(eni.ENIID) {
			log.Debugf("Skipping ENI %s: since on non-zero network card", eni.ENIID)
			numFiltered++
			continue
		}

		ret = append(ret, eni)
	}
	c.unmanagedENI = numFiltered
	c.updateIPStats(numFiltered)
	return ret
}
```

pkg/ipamd/ipamd.go
jayanthvn: Yeah, I missed this part. Sure I can merge in the same function.

pkg/ipamd/ipamd.go
jayanthvn: Yeah, I can do this. I think this will also handle use case of unmanaged ENI on non-zero network card. Since we just have to account it once.

pkg/ipamd/ipamd.go
jayanthvn: I need to update it here - https://github.com/aws/amazon-vpc-cni-k8s/blob/56851f0905dba4852eb895ec1c7bd5b1876a9c67/pkg/ipamd/ipamd.go#L637. Will update the PR.

pkg/ipamd/ipamd.go
fawadkhaliq: ++, exactly

pkg/ipamd/ipamd.go
jayanthvn: removed this since we the final value in unmanagedENI.

pkg/ipamd/ipamd.go
jayanthvn: Updated the PR with this.

pkg/ipamd/ipamd.go
jayanthvn: Updated the PR.

pkg/ipamd/ipamd.go
```diff
@@ -356,6 +368,12 @@ func (c *IPAMContext) nodeInit() error {
 		retry := 0
 		for {
 			retry++
+
+			if c.awsClient.IsCNIUnmanagedENI(eni.ENIID) {
+				log.Infof("Skipping ENI %s since it is not on network card 0", eni.ENIID)
+				continue
+			}
+
```
fawadkhaliq, line 376: Have we tested this workflow?

Does not look like the right place to have this check and the `continue` statement. Seems like for a given ENI that is not on network card 0, this loop will be keep spinning.

You would want to move this outside the retry loop

jayanthvn: Yeah you are right, it should have been a break, because it will run until -

```
if retry > maxRetryCheckENI {
				log.Warnf("Reached max retry: Unable to discover attached IPs for ENI from metadata service (attempted %d/%d): %v", retry, maxRetryCheckENI, err)
				ipamdErrInc("waitENIAttachedMaxRetryExceeded")
				break
			}
```

jayanthvn: Nvm, this needs to be fixed. Thank you. Good catch.

fawadkhaliq (approved):

PR #1401, opened 2021-03-09
Replaces aws-k8s-tester with eksctl

couralex6:
**What type of PR is this?**
bug

This PR removes aws-k8s-tester from run-integration-test.sh and replaces it with eksctl. 

This PR also adds two config files for eksctl: `performance_cluster.yaml` and `test_cluster.yaml`.

`test_cluster.yaml` config contains "CLUSTERNAME" as a value for the name field. This name gets overwritten by a randomly generated name to allow concurrent tests to run.

After the transition to eksctl, I noticed that 2 cases of the upstream conformance tests were failing:

* should be able to change the type from ExternalName to NodePort
* should be able to create a functioning NodePort service

I added them to the 'Skip' ginkgo flag for now to get the CI back up and running.

**Automation added to e2e**:

```release-note

```


jayanthvn:
Overall, it looks good but lets check why the 2 cases are failing - 

```
should be able to change the type from ExternalName to NodePort
should be able to create a functioning NodePort service
```

PR #1297, opened 2020-11-25
Switch calico to be deployed with the Tigera operator

tmjd:
This switches the Calico install to be done using the Tigera operator. This PR includes a manifest to install the operator which will install Calico v3.17.

**What type of PR is this?**
update config?

**Which issue does this PR fix**:

**What does this PR do / Why do we need it**:

**If an issue # is not available please add repro steps and logs from IPAMD/CNI showing the issue**:

**Testing done on this change**:
I have tested upgrading a cluster that had a previous version of Calico (with Amazon VPC CNI) installed and also installed on a cluster that only had the Amazon VPC CNI plugin installed (with no existing Calico).

**Automation added to e2e**:

**Will this break upgrades or downgrades. Has updating a running cluster been tested?**:
Because of the way the upgrade will happen with the operator there is a problem upgrading on a small cluster, 3 nodes or less. This is because for 3 nodes or less the operator tries to deploy a typha for each node and the current calico install uses at least one typha and multiple typhas cannot run on a single node.
Once a cluster is upgraded with these changes, it will not be simple to downgrade back to a versions of Calico that was installed without the operator.

**Does this change require updates to the CNI daemonset config files to work?**:

This is updating the daemonset config files for Calico. A 'kubectl patch' of the image tag does not work because the change switches Calico to be installed by an oeprator.

**Does this PR introduce any user-facing change?**:

Users of Calico can now use `kubectl get tigerastatus` to get an overview status of the Calico installation. They also would have the installations.operator.tigera.io default resource available for making configuration changes.

```release-note
Switch Calico to be installed and managed by the Tigera Operator.
```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1410, opened 2021-03-24
Split Calico manifest into two

caseydavenport:
**What type of PR is this?**

bug

As mentioned in this discussion: https://github.com/aws/amazon-vpc-cni-k8s/pull/1297#issuecomment-801472720

The current manifest includes both the Installation CRD as well as an
instance of an Insatllation CR. Since they are both in the same
manfiest, the API server doesn't have time to register the CRD before
the CR is sent, and so the CR is rejected.

They should be in two manfiests, so the user can apply the CRD first and
then apply the CR.

This change will include a corresponding documentation change. 

Splits calico.yaml into two so they can be properly ordered.

**Automation added to e2e**:

```release-note
calico.yaml is split into two manifests - calico-operator.yaml and calico-crs.yaml
```

### DIFF
--- a/config/master/calico-crs.yaml
+++ b/config/master/calico-crs.yaml
@@ -1,0 +1,10 @@
+# This section includes base Calico installation configuration.
+# For more information, see: https://docs.projectcalico.org/v3.17/reference/installation/api#operator.tigera.io/v1.Installation
+apiVersion: operator.tigera.io/v1
+kind: Installation
+metadata:
+  name: default
+spec:
+  # Configures Calico policy configured to work with AmazonVPC CNI networking.
+  cni:
+    type: AmazonVPC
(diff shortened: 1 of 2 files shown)


tmjd (approved):
LGTM
Thank you

jayanthvn:
Thanks @caseydavenport . Can you please update the branch so that I can merge.

jayanthvn (approved):
Can you please point me to a documentation so that I can update in our readme and release notes? LGTM, thanks :)

caseydavenport:
@jayanthvn this PR has the corresponding documentation updates: https://github.com/awsdocs/amazon-eks-user-guide/pull/341

jayanthvn:
@caseydavenport - awesome, thank you so much.

PR #1419, opened 2021-04-07
Address Excessive API Server calls from CNI Pods

achevuru:
**What type of PR is this?**
Enhancement

CNI daemonset Pods currently run a controller to handle ENIConfig CRDs along with Node and Pod objects. CNI code base depends on a rather old version of operator-sdk (v0.0.7) and this combined with dynamic api group discovery is contributing to a rather high rate of API Server calls from each CNI pod (approx 8500 calls per hr from each node). As we scale the cluster, this becomes a clear problem and needs to be dealt with. 

PR attempts to address the above issue via a 2-way approach depending on the resource it is dealing with.

1. For ENIConfig and Node resources, we will now use a K8S client with a cache tied to it - following the list+watch approach for both these resources. Both ENIConfig and Node resources are accessed during the ENI creation flow when custom networking is enabled and we don't want to introduce an API server call during this workflow.

2. For Pods, we will make an API server call on need basis.

With the change, API Server call count reduced to about 30-35 per hr from each node from the current 8500 (approx) per hr from each node.

Verified both regular and custom networking mode.

No. Yes, upgrade scenario has been tested.

No

No


cmd/aws-k8s-agent/main.go
```diff
@@ -39,18 +37,17 @@ func _main() int {
...
-		log.Errorf("Failed to create client: %v", err)
+	//Check API Server Connectivity
+	if k8sapi.CheckAPIServerConnectivity() != nil{
 		return 1
 	}
 
-	eniConfigController := eniconfig.NewENIConfigController()
-	if ipamd.UseCustomNetworkCfg() {
-		go eniConfigController.Start()
+	standaloneK8SClient, k8sClient, err := k8sapi.CreateKubeClients()
```
anguslees, line 45: Not a fan of this "return multiple magic clients" API :(
Perhaps break it into a regular "create a client" function (ie: the old `CreateKubeClient`) and then a _separate_ function that creates a caching client from a 'raw' client.  Then it will be clear to the caller (here) what the two are, and it will also be obvious for the cases where you only want one of those clients.

cmd/aws-k8s-agent/main.go
```diff
@@ -39,18 +37,17 @@ func _main() int {
...
+	if err != nil{
+		return 1
 	}
 
-	ipamContext, err := ipamd.New(kubeClient, eniConfigController)
+	ipamContext, err := ipamd.New(standaloneK8SClient, k8sClient)
```
anguslees, line 50: Why does ipamd.New need both cached and uncached client?  That seems weird, or a sign that we need cache invalidation somewhere.

cmd/cni-metrics-helper/metrics/cni_metrics_test.go
```diff
@@ -1,15 +1,19 @@
 package metrics
 
 import (
+	eniconfigscheme "github.com/aws/amazon-vpc-cni-k8s/pkg/apis/crd/v1alpha1"
```
anguslees, line 4: Same as comment elsewhere: separate imports into stdlib, 3p, 1p

cmd/cni-metrics-helper/metrics/cni_metrics_test.go
```diff
@@ -18,33 +22,37 @@ var logConfig = logger.Configuration{
...
 }
 
-var log = logger.New(&logConfig)
+var testLog = logger.New(&logConfig)
 
 type testMocks struct {
-	ctrl               *gomock.Controller
 	clientset          *k8sfake.Clientset
-	discoverController *k8sapi.Controller
+    podWatcher         *defaultPodWatcher
```
anguslees, line 29: Nit: Bad indent - Does this mean our gofmt checks aren't working? :confused:

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -15,26 +15,28 @@
 package metrics
 
 import (
+	"context"
 	"bytes"
 	"fmt"
+	"github.com/prometheus/common/log"
```
anguslees, line 21: Same again :)

cmd/cni-metrics-helper/metrics/metrics_test.go
```diff
@@ -14,6 +14,7 @@
 package metrics
 
 import (
+	"golang.org/x/net/context"
```
anguslees, line 17: Can we just use `"context"` now?

cmd/cni-metrics-helper/metrics/pod_watcher.go
```diff
@@ -0,0 +1,52 @@
+package metrics
+
+import (
+	"context"
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
```
anguslees, line 5: ditto imports

cmd/cni-metrics-helper/metrics/pod_watcher.go
```diff
@@ -0,0 +1,52 @@
...
+//Returns aws-node pod info. Below function assumes CNI pods follow aws-node* naming format
+//and so the function has to be updated if the CNI pod name format changes.
+func (d *defaultPodWatcher) GetCNIPods(ctx context.Context) ([]string, error){
+	var CNIPods []string
+	var podList corev1.PodList
+	listOptions := client.ListOptions{
+		Namespace:     metav1.NamespaceSystem,
+	}
+
+	err := d.k8sClient.List(ctx, &podList, &listOptions)
```
anguslees, line 39: Please use pagination on all List calls (unless you _know_ the answer is going to be ~1 item)

pkg/eniconfig/eniconfig.go
```diff
@@ -16,18 +16,14 @@ package eniconfig
 
 import (
 	"context"
+	"fmt"
+	corev1 "k8s.io/api/core/v1"
```
anguslees, line 20: same

pkg/eniconfig/eniconfig.go
```diff
@@ -75,127 +61,28 @@ type ENIConfigInfo struct {
...
-
-	myENIConfig, ok := eniCfg.eni[eniCfg.myENI]
-
-	if ok {
-		return &v1alpha1.ENIConfigSpec{
-			SecurityGroups: myENIConfig.SecurityGroups,
-			Subnet:         myENIConfig.Subnet,
-		}, nil
+	for _, eni := range eniConfigsList.Items {
+		if eniConfigName == eni.Name {
```
anguslees, line 79: If we only want a single named item, can we use Get() here rather than List() ?

pkg/eniconfig/eniconfig.go
```diff
@@ -227,3 +114,32 @@ func getEniConfigLabelDef() string {
...
+func GetNodeSpecificENIConfigName(ctx context.Context, k8sClient client.Client) (string, error) {
+    var eniConfigName string
+	nodeList := corev1.NodeList{}
+	err := k8sClient.List(ctx, &nodeList)
+	if err != nil {
+		fmt.Errorf("Error while Node List Get: %s", err)
+	}
+	log.Debugf("Node Count: ", len(nodeList.Items))
+	for _, node := range nodeList.Items {
+		if node.Name == os.Getenv("MY_NODE_NAME") {
```
anguslees, line 127: Why is this a List() - can we just call Get(MY_NODE_NAME) ?

pkg/eniconfig/eniconfig_test.go
```diff
@@ -14,205 +14,215 @@ package eniconfig
 
 import (
 	"context"
-	"fmt"
+	//"fmt"
```
anguslees, line 17: left over debugging?

pkg/ipamd/introspect.go
```diff
@@ -15,13 +15,15 @@ package ipamd
 
 import (
 	"encoding/json"
+	"golang.org/x/net/context"
```
anguslees, line 18: Just `"context"` ?

pkg/ipamd/introspect.go
```diff
@@ -142,7 +144,9 @@ func eniV1RequestHandler(ipam *IPAMContext) func(http.ResponseWriter, *http.Requ
 
 func eniConfigRequestHandler(ipam *IPAMContext) func(http.ResponseWriter, *http.Request) {
 	return func(w http.ResponseWriter, r *http.Request) {
-		responseJSON, err := json.Marshal(ipam.eniConfig.Getter())
+		ctx := context.Background()
+		myENIConfig, _ := eniconfig.GetNodeSpecificENIConfigName(ctx, ipam.k8sClient)
```
anguslees, line 148: I think you're ignoring an error here.  Don't do that :P

pkg/ipamd/introspect.go
```diff
@@ -142,7 +144,9 @@ func eniV1RequestHandler(ipam *IPAMContext) func(http.ResponseWriter, *http.Requ
 
 func eniConfigRequestHandler(ipam *IPAMContext) func(http.ResponseWriter, *http.Request) {
 	return func(w http.ResponseWriter, r *http.Request) {
-		responseJSON, err := json.Marshal(ipam.eniConfig.Getter())
+		ctx := context.Background()
```
anguslees, line 147: Use `ctx := r.Context()` here.  This is a great example of where the context cancellation is useful (if the client goes away, it will cancel any outstanding downstream API calls).

pkg/ipamd/ipamd.go
```diff
@@ -14,17 +14,18 @@
 package ipamd
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"os"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
anguslees, line 21: imports again

pkg/ipamd/ipamd.go
```diff
@@ -187,9 +191,9 @@ var (
 type IPAMContext struct {
 	awsClient            awsutils.APIs
 	dataStore            *datastore.DataStore
-	k8sClient            kubernetes.Interface
+	standalonek8sClient  client.Client
+	k8sClient            client.Client
```
anguslees, line 195: As mentioned above - it smells like a bug if ipamd really requires both a cached and uncached client.  Can we consolidate these, or make the caching behaviour clearer?

pkg/ipamd/ipamd.go
```diff
@@ -329,6 +332,7 @@ func (c *IPAMContext) nodeInit() error {
 	ipamdActionsInprogress.WithLabelValues("nodeInit").Add(float64(1))
 	defer ipamdActionsInprogress.WithLabelValues("nodeInit").Sub(float64(1))
 	var err error
+	ctx := context.Background()
```
anguslees, line 335: Make this a `context.TODO()` - This is not an async operation, and ctx should really be passed in from the caller.

pkg/ipamd/ipamd.go
```diff
@@ -407,17 +411,18 @@ func (c *IPAMContext) nodeInit() error {
 		vpcCIDRs = c.updateCIDRsRulesOnChange(vpcCIDRs)
 	}, 30*time.Second)
 
-	if c.useCustomNetworking && c.eniConfig.Getter().MyENI != "default" {
+	eniConfigName,_ := eniconfig.GetNodeSpecificENIConfigName(ctx, c.k8sClient)
```
anguslees, line 414: Don't ignore errors.  You are proceeding with an 'unknown' eniconfig name, which is very different behaviour to what you intended (eg: the below code will go and set node labels to empty string, etc)

pkg/ipamd/ipamd.go
```diff
@@ -1332,15 +1342,27 @@ func (c *IPAMContext) SetNodeLabel(key, value string) error {
...
+	podKey := types.NamespacedName{
+		Namespace: namespace,
+		Name:      podName,
+	}
+	err := c.standalonek8sClient.Get(ctx, podKey ,&pod)
+	if err != nil {
+		return nil, fmt.Errorf("Error while trying to retrieve Pod Info: %s", err)
+	}
+	return &pod, nil
 }
```
anguslees, line 1368: fyi: missing trailing newline in this file

pkg/ipamd/ipamd.go
```diff
@@ -1332,15 +1342,27 @@ func (c *IPAMContext) SetNodeLabel(key, value string) error {
...
+	log.Debugf("Updated node %s with label %q: %q", c.myNodeName, key, value)
+
 	return nil
 }
 
 // GetPod returns the pod matching the name and namespace
-func (c *IPAMContext) GetPod(podName, namespace string) (*v1.Pod, error) {
-	return c.k8sClient.CoreV1().Pods(namespace).Get(podName, metav1.GetOptions{})
+func (c *IPAMContext) GetPod(podName, namespace string) (*corev1.Pod, error) {
+	ctx := context.Background()
```
anguslees, line 1356: This is _definitely_ not an async background operation :stuck_out_tongue: 
Make this `context.TODO()` to indicate that we need to update the code to take ctx as an argument (at some point).

pkg/ipamd/ipamd.go
```diff
@@ -1332,15 +1342,27 @@ func (c *IPAMContext) SetNodeLabel(key, value string) error {
...
+	var pod corev1.Pod
+
+	podKey := types.NamespacedName{
+		Namespace: namespace,
+		Name:      podName,
+	}
+	err := c.standalonek8sClient.Get(ctx, podKey ,&pod)
```
anguslees, line 1363: Why doesn't this use the cached client?

pkg/ipamd/ipamd_test.go
```diff
@@ -16,6 +16,10 @@ package ipamd
 import (
 	"errors"
 	"fmt"
+	"golang.org/x/net/context"
```
anguslees, line 19: imports

pkg/k8sapi/k8sutils.go
```diff
@@ -0,0 +1,77 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var log = logger.Get()
+
+// CreateKubeClient creates a k8s client
+func CreateKubeClients() (client.Client, client.Client, error) {
+	restCfg := ctrl.GetConfigOrDie()
```
anguslees, line 20: I think this function should return an error rather than crash.

pkg/k8sapi/k8sutils.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	go func() {
+		cache.Start(stopChan)
+	}()
+	cache.WaitForCacheSync(stopChan)
+
+	standaloneK8SClient, err := client.New(restCfg, client.Options{Scheme: vpcCniScheme})
+	k8sClient := client.DelegatingClient{
```
anguslees, line 36: Afaics the only difference here is the read cache?  Perhaps we should call them 'caching' and 'raw' clients rather than 'standalone' and .. 'not-standalone' ?

pkg/k8sapi/k8sutils.go
```diff
@@ -0,0 +1,77 @@
...
+	// creates the in-cluster config
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		panic(err.Error())
+	}
+
+	// creates the clientset
+	clientSet, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		panic(err.Error())
```
anguslees, line 57: Why are we crashing here, rather than returning an error?
(It seems odd to define a function that returns `error`, but then take such drastic steps to always return `nil`...)

pkg/k8sapi/k8sutils.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	return clientSet, nil
+}
+
+func CheckAPIServerConnectivity() error {
+	restCfg := ctrl.GetConfigOrDie()
```
anguslees, line 63: Why crash here?  This makes this Check() function a particularly risky thing to call..

cmd/aws-k8s-agent/main.go
```diff
@@ -39,18 +37,17 @@ func _main() int {
 
 	log.Infof("Starting L-IPAMD %s  ...", version)
 
-	kubeClient, err := k8sapi.CreateKubeClient()
-	if err != nil {
-		log.Errorf("Failed to create client: %v", err)
+	//Check API Server Connectivity
+	if k8sapi.CheckAPIServerConnectivity() != nil{
```
anguslees, line 41: ... why do this, rather than just wait until the first actual apiserver call to effectively do it for us?

(In particular, just because the apiserver was available _here_ doesn't mean it's _always_ going to be available during every other part of execution - so we still have to handle intermittent connectivity issues elsewhere.)

cmd/aws-k8s-agent/main.go
```diff
@@ -15,12 +15,10 @@
 package main
 
 import (
-	"os"
-
-	"github.com/aws/amazon-vpc-cni-k8s/pkg/eniconfig"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/k8sapi"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
+	"os"
```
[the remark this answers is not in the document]
achevuru: Will sort it..

cmd/cni-metrics-helper/metrics/cni_metrics_test.go
achevuru: Indent is alright in the IDE but off in the diff..will rerun gofmt

pkg/k8sapi/k8sutils.go
achevuru: Yes, read cache. Agree on the naming, will adjust the names..

cmd/aws-k8s-agent/main.go
achevuru: Sure, a success here doesn't guarantee the call will go through a while later. This was mainly done to be in parity with the current CNI codeflow, where we check API server connectivity and essentially crash if the initial API Server connectivity check fails. Sort of helps to catch or highlight basic connectivity issues to API server as part of CNI bootstrap..

https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/k8sapi/discovery.go#L87

pkg/ipamd/ipamd.go
achevuru: So, there are 3 different resources ipamd is acting on - `ENIConfigs`, `Nodes` (during ENI create flow) and `Pods` (during SGPP flow - to get pod annotations). We wanted to avoid a cached client for `Pods`(essentially avoid a watch on pod resource from every node) and so we're using a Cached client exclusively for `ENIConfigs` and `Nodes` and depend on raw client to get `Pod` objects.

pkg/ipamd/ipamd.go
achevuru: We wanted to avoid a cached client for Pods(essentially avoid a watch on pod resource from every node) and so we're using a Cached client exclusively for `ENIConfigs` and `Nodes` and depend on raw client to get `Pod` objects.

pkg/k8sapi/k8sutils.go
achevuru: Ok, yeah we can let the main function exit and log an error here..

pkg/ipamd/ipamd.go
achevuru: Ok. Made it a `context.Background()` because it is in a `**Init` function (or) I can pass the context in through the main function.

pkg/ipamd/introspect.go
achevuru: Makes sense. Will update it.

pkg/ipamd/introspect.go
achevuru: Yeah, eniConfigName will be "" if the function returns an err and we respond back with an empty string indicating no match. Let me add a check though...

cmd/aws-k8s-agent/main.go
achevuru: Will split these in to separate functions..

pkg/eniconfig/eniconfig.go
achevuru: Yeah, repurposed legacy code. Will change it.

cmd/cni-metrics-helper/metrics/pod_watcher.go
achevuru: Sure. Since the call is limited to `kube-system` namespace, I used a list here.

pkg/apis/crd/v1alpha1/eniconfig_types.go
```diff
@@ -0,0 +1,60 @@
...
+
+// ENIConfigSpec defines the desired state of ENIConfig
+type ENIConfigSpec struct {
+	SecurityGroups []string `json:"securityGroups"`
+	Subnet         string   `json:"subnet"`
+}
+
+// ENIConfigStatus defines the observed state of ENIConfig
+type ENIConfigStatus struct {
+	// Fill me
```
jayanthvn, line 34: Nit: If ENIConfigStatus is not used, can we remove it?

cmd/cni-metrics-helper/metrics/cni_metrics_test.go
anguslees: > Indent is alright in the IDE but off in the diff

Ah, your editor is probably using spaces rather than tabs.  golang likes actual tab chars - the opposite of (eg) python style.

cmd/cni-metrics-helper/metrics/pod_watcher.go
achevuru: Ok, I should've clarified in my initial comment but this code is specific to CNI Metrics helper and is not part of `aws-node`. We will only have one CNI Metrics Server pod per cluster + the call being specific to `kube-system` namespace is the reason why a `list` was initially preferred. 

Yeah, I missed that on Limit usage. Considering, the limitation on the cached client end and the call being limited to a single pod + single NS, I'm thinking it should be alright to revert back to no limit usage.

pkg/apis/crd/v1alpha1/eniconfig_types.go
achevuru: Yeah, just the standard `Spec` and `Status` objects for CRDs/K8S resources. Although, we're not using it right now for ENIConfig CRD, it is good to have them in place.

cmd/cni-metrics-helper/metrics/pod_watcher.go
anguslees: > We will only have one CNI Metrics Server pod per cluster

Right, but the response for a 10k cluster is still going to cause ~~100MB to be allocated on apiserver (even for a single List(kube-system) request).  It would be great if we didn't do that.  .. But we might not have a choice until controller-runtime is updated.

achevuru: Scoped it down further using a label selector to filter out aws-node pods and this should help over what we had originally. Verified that it works as expected.

cmd/cni-metrics-helper/metrics/pod_watcher.go
```diff
@@ -33,9 +32,15 @@ func NewDefaultPodWatcher(k8sClient client.Client, log logger.Logger) *defaultPo
 func (d *defaultPodWatcher) GetCNIPods(ctx context.Context) ([]string, error) {
 	var CNIPods []string
 	var podList corev1.PodList
+	labelSelector, _ := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{
```
anguslees, line 35: You're ignoring errors :)
Since the input here is all static, I think the only way this fails is if we make a typo in the labelselector somewhere (ie: a programmer error).  So I think it's reasonable to `panic()` if this returns an error.

We should _not_ ignore the error and continue, because we don't know what state the labelSelector might be in.  Eg: it might be partially constructed and not match any pods (or might have a dangling null and we crash).

cmd/cni-metrics-helper/metrics/pod_watcher.go
anguslees: Big :+1: to the new label selector approach.  The apiserver thanks you.

cmd/cni-metrics-helper/metrics/pod_watcher.go
achevuru: Sure, Will add a check.

pkg/ipamd/ipamd.go
achevuru: Yeah, we did discuss about it but opted to use cached client for the resources we depend upon during the ENI creation flow. No additional latency during ENI Create and not having to worry about retries if (for whatever reason) the API Server call fails at the most inopportune moment :). Also, the existing custom networking code is watching on Node label/annotation changes real time and adjusting the ENIConfig name applicable to that node although I am not sure if we actually recommend this. I think, we can revisit if this becomes inefficient as we start supporting clusters with 10k+ nodes.

cmd/cni-metrics-helper/metrics/pod_watcher.go
anguslees: > If this is a new requirement it should be documented.

Yeah, fair call.  The previous code assumed/required that the pods were all called `aws-node-xxx`, so we're just switching one hard-coded assumption for another.  I agree we should mention that somewhere and/or make the labelselector configurable.

achevuru: @danbeaulieu Default manifest sets this particular label - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/v1.7/aws-k8s-cni.yaml#L74.

But agree on documenting this for CNI Metrics helper. If someone wants to use CNI metrics helper than this is required.

anguslees (approved):
lgtm!

cmd/cni-metrics-helper/metrics/pod_watcher.go
anguslees: Right.  It's ~easy to verify the code is correct now - but I'm thinking about what it looks like after patches by 20 different people who didn't realise these assumptions.

Fail loudly and obviously at compile time == best; fail loudly and obviously at every execution == next best; [.. other options]; fail by ignoring errors and silently doing the wrong thing sometimes at runtime == worst.

jayanthvn (changes requested):
Looks good, can you please fix the conflicts in ipamd.go.

achevuru:
> Looks good, can you please fix the conflicts in ipamd.go.

Done.

jayanthvn (approved):

PR #1434, opened 2021-04-21
[Preview] Prefix delegation feature development

jayanthvn:
**What type of PR is this?**
Feature

n/a

Current implementation allocates secondary IPs per ENI which will be used for pod IPs. With the new VPC feature each secondary IP will be replaced with /28 prefix in the ENI's subnet. IPs will be allocated from this subnet.

Yes and in progress

**Automation added to e2e**:

Yes

Yes

Yes

Yes

```release-note
Prefix delegation feature
```


pkg/ipamd/datastore/data_store.go
```diff
@@ -781,20 +1036,36 @@ func (ds *DataStore) UnassignPodIPv4Address(ipamKey IPAMKey) (ip string, deviceN
...
 	}
 	addr.UnassignedTime = time.Now()
+	if eni.IsPDenabled && addr.Prefix == "" {
+		ds.log.Infof("Prefix delegation is enabled and the IP is from secondary pool hence no need to update prefix pool")
+		ds.total--
+		mismatchedStore = true
+	} else if addr.Prefix != "" {
+		//Regular pod deletes for PD enabled and if pd is disabled but prefix is populated i.e, knob disable scenario
+		ds.log.Infof("Dump for prefix %v", addr.Prefix)
+		ds.log.Infof("DUMP - %v", eni.IPv4Prefixes[addr.Prefix])
```
anguslees, line 1056: Left over debugging.  Needs to be at least demoted to `Debugf()` (or removed)

pkg/ipamd/datastore/data_store.go
```diff
@@ -781,20 +1036,36 @@ func (ds *DataStore) UnassignPodIPv4Address(ipamKey IPAMKey) (ip string, deviceN
...
+		deleteIPv4AddrfromPrefix(eni.IPv4Prefixes[addr.Prefix], addr)
+		//Remove the IP from eni DB
+		delete(eni.IPv4Addresses, addr.Address)
+		if !eni.IsPDenabled {
+			ds.log.Infof("Prefix delegation is disabled but IP belongs to prefix pool")
+			mismatchedStore = true
```
anguslees, line 1062: I was initially confused what `mismatchedStore` meant.  Is `isSecondaryIP` equivalent and clearer?

pkg/ipamd/datastore/data_store.go
```diff
@@ -144,18 +146,24 @@ type ENI struct {
...
+	IPv4Prefixes map[string]*ENIPrefix
 }
 
 // AddressInfo contains information about an IP, Exported fields will be marshaled for introspection.
 type AddressInfo struct {
 	IPAMKey        IPAMKey
 	Address        string
 	UnassignedTime time.Time
+	Prefix         string
+	IPIndex        int
```
anguslees, line 166: I think we can derive `IPIndex` on the fly from `Address` and `Prefix` - even better if we store `Address` and `Prefix` as parsed `net.IP` (or `net.IPNet`) types.

pkg/ipamd/datastore/data_store.go
```diff
@@ -841,6 +1112,39 @@ func (ds *DataStore) FreeableIPs(eniID string) []string {
...
+	for _, prefix := range eni.IPv4Prefixes {
+		DBlen := (int)(prefix.AllocatedIPs.IPsPerPrefix / 8)
+
+		ds.log.Infof("In get IP from prefix - %s and octet len %d", prefix, DBlen)
+		var octet int
+		for octet = 0; octet < DBlen; octet++ {
+			if prefix.AllocatedIPs.UsedIPs[octet]&0xFF != 0 {
+				break
+			}
+		}
```
anguslees, line 1138: This should be a method on UsedIPs (or AllocatedIPs) - don't leak the (slightly complicated / error-prone) representation outside those classes.

pkg/ipamd/datastore/data_store.go
```diff
@@ -841,6 +1112,39 @@ func (ds *DataStore) FreeableIPs(eniID string) []string {
...
+	eni := ds.eniPool[eniID]
+	if eni == nil {
+		// Can't free any IPs from an ENI we don't know about...
+		return []string{}
+	}
+	ds.log.Infof("In freeable prefix for eni %s", eniID)
+
+	freeable := make([]string, 0, len(eni.IPv4Prefixes))
+	for _, prefix := range eni.IPv4Prefixes {
+		DBlen := (int)(prefix.AllocatedIPs.IPsPerPrefix / 8)
```
anguslees, line 1130: local variables should start with lowercase

pkg/ipamd/datastore/data_store.go
```diff
@@ -841,6 +1112,39 @@ func (ds *DataStore) FreeableIPs(eniID string) []string {
...
+
+		ds.log.Infof("In get IP from prefix - %s and octet len %d", prefix, DBlen)
```
anguslees, line 1132: This is not `Info` worthy - particularly inside a loop like this.  Demote to `Debug` or remove.

pkg/ipamd/datastore/prefix_store.go
```diff
@@ -0,0 +1,176 @@
...
+// UsedIPs will be a bitmap of size - IPsPerPrefix which will keep track of IPs used
+type PrefixIPsStore struct {
+	UsedIPs      []byte
+	CooldownIPs  []time.Time
+	IPsPerPrefix int
+}
+
+// ENIPrefix will have the prefixes allocated for an ENI and the list of allocated IPs under
+// the prefix
+type ENIPrefix struct {
```
anguslees, line 33: This should have a `String()` method, or (better?) have an easy way to get `Prefix`/`PrefixLen` as a `net.IPNet` (and then we could get a string from there).

Then we can remove several other places that are currently doing their own `prefix + "/" strconv.Itoa(PrefixLen)`

pkg/ipamd/datastore/data_store.go
```diff
@@ -883,8 +1187,56 @@ func (ds *DataStore) GetENIIPs(eniID string) ([]string, error) {
...
+	var ipPool = make([]string, 0, len(eni.IPv4Prefixes))
+	for prefix, prefixData := range eni.IPv4Prefixes {
+		ip := (prefix + "/" + strconv.Itoa(prefixData.PrefixLen))
+		ipPool = append(ipPool, ip)
+	}
+	return ipPool, nil
+}
+
+func (ds *DataStore) GetFreePrefixes() int {
+	freePrefixes := 0
```
anguslees, line 1222: I think we should hold the datastore lock here.  I agree we probably don't need it when walking the structures as we do (golang reference counting will save us from walking an object that other code frees concurrently) - but I don't trust ourselves to always get that correct.

pkg/ipamd/datastore/data_store.go
```diff
@@ -883,8 +1187,56 @@ func (ds *DataStore) GetENIIPs(eniID string) ([]string, error) {
...
+				freePrefixes++
+			}
+		}
+
+	}
+	return freePrefixes
+}
+
+func (ds *DataStore) CleanupDataStore(eniID string, ipv4 string, force bool, enableIpv4PrefixDelegation bool) error {
+	var err error
```
anguslees, line 1235: We need to hold the datastore lock during this function.

pkg/ipamd/datastore/data_store.go
```diff
@@ -883,8 +1187,56 @@ func (ds *DataStore) GetENIIPs(eniID string) ([]string, error) {
...
+	if !enableIpv4PrefixDelegation {
+		err = ds.DelIPv4AddressFromStore(eniID, ipv4, force)
```
anguslees, line 1237: I'm a bit confused here - isn't what we're deleting a property of the address itself (ie: is it a secondary IP or PD IP?) and not a property of what mode we're running in?

iow: what if the address was attached to this instance as a secondaryIP, but vpc-cni was started with PD enabled.   
 Now we're trying to free that unused secondary IP.  What happens in this function?

pkg/ipamd/ipamd.go
```diff
@@ -579,66 +609,70 @@ func (c *IPAMContext) tryUnassignIPsFromAll() {
...
 				}
-			}
 
-			// Deallocate IPs from the instance if they aren't used by pods.
-			if err := c.awsClient.DeallocIPAddresses(eniID, deletedIPs); err != nil {
-				log.Warnf("Failed to decrease IP pool by removing IPs %v from ENI %s: %s", deletedIPs, eniID, err)
-			} else {
-				log.Debugf("Successfully decreased IP pool by removing IPs %v from ENI %s", deletedIPs, eniID)
+				// Deallocate IPs from the instance if they aren't used by pods.
+				if err = c.awsClient.DeallocIPAddresses(eniID, deletedIPsOrPrefixes, false); err != nil {
```
anguslees, line 627: I might be confused by the diff, but I think this needs to be moved out a level.  If I'm reading it correctly, I _think_ you're deallocating `deletedIPsOrPrefixes` inside the same loop as where you're building that list.

pkg/ipamd/ipamd.go
```diff
@@ -806,10 +895,24 @@ func (c *IPAMContext) addENIaddressesToDataStore(ec2Addrs []*ec2.NetworkInterfac
 			ipamdErrInc("addENIaddressesToDataStoreAddENIIPv4AddressFailed")
 		}
 	}
-	total, assigned := c.dataStore.GetStats()
+	total, assigned, _ := c.dataStore.GetStats()
 	log.Debugf("IP Address Pool stats: total: %d, assigned: %d", total, assigned)
 }
 
+func (c *IPAMContext) addENIprefixesToDataStore(ec2Addrs []*ec2.Ipv4PrefixSpecification, eni string) {
+	log.Infof("In addENIPrefixedToDataStore for %s and no of prefixes %d", eni, len(ec2Addrs))
```
anguslees, line 903: Should be `Debug`

pkg/ipamd/ipamd.go
```diff
@@ -850,8 +953,29 @@ func getWarmENITarget() int {
...
+		log.Debugf("Using WARM_PREFIX_TARGET %v", input)
+		return input
+	}
+	return noWarmPrefixTarget
+}
+
+func logPoolStats(total int, used int, maxAddrsPerENI int, Ipv4PrefixDelegation bool) {
+	if !Ipv4PrefixDelegation {
+		log.Debugf("IP pool stats: total = %d, used = %d, c.maxIPsPerENI = %d", total, used, maxAddrsPerENI)
+	} else if Ipv4PrefixDelegation {
```
anguslees, line 976:
```suggestion
	} else {
```

.. also - I think these two branches are identical??

pkg/ipamd/ipamd.go
```diff
@@ -992,15 +1094,28 @@ func (c *IPAMContext) nodeIPPoolReconcile(interval time.Duration) {
...
-		eniIPPool, err := c.dataStore.GetENIIPs(attachedENI.ENIID)
-		if err == nil {
-			// If the attached ENI is in the data store
-			log.Debugf("Reconcile existing ENI %s IP pool", attachedENI.ENIID)
-			// Reconcile IP pool
-			c.eniIPPoolReconcile(eniIPPool, attachedENI, attachedENI.ENIID)
-			// Mark action, remove this ENI from currentENIs map
-			delete(currentENIs, attachedENI.ENIID)
-			continue
+		if !c.enableIpv4PrefixDelegation {
```
anguslees, line 1097: Don't we need to reconcile both SIP and PDs?  (What if the mode is toggled and we still had some pods running on the 'other' mode?)

pkg/ipamd/ipamd.go
```diff
@@ -1075,6 +1190,53 @@ func (c *IPAMContext) eniIPPoolReconcile(ipPool []string, attachedENI awsutils.E
 	}
 }
 
+func (c *IPAMContext) eniPrefixPoolReconcile(ipPool []string, attachedENI awsutils.ENIMetadata, eni string) {
+	attachedENIIPs := attachedENI.IPv4Prefixes
+	needEC2Reconcile := true
+	// Here we can't trust attachedENI since the IMDS metadata can be stale. We need to check with EC2 API.
+	log.Infof("Found prefix pool count %d for eni %s\n", len(ipPool), eni)
+	for _, dumpIP := range ipPool {
+		log.Infof("DUMP from eni pool %s", dumpIP)
```
anguslees, line 1199: `Debugf`  (and elsewhere in this function)

pkg/ipamd/ipamd.go
```diff
@@ -1204,6 +1442,10 @@ func enablePodENI() bool {
 	return getEnvBoolWithDefault(envEnablePodENI, false)
 }
 
+func enableIpv4PrefixDelegation() bool {
```
anguslees, line 1445: This sounds like it actually enables PD - perhaps call it "UseIPv4PrefixDelegation()" or some other different verb tense.

anguslees (commented):
Haven't finished reviewing ipamd.go yet, but pushing what I've got so far.

High level: We have a _lot_ of "if enablePD $this else $that".  This code structure is a lot of tech debt to carry forward, since we're going to have to think about both branches all the time.  I _expected_ to see only the place where we allocate new IPs (and thus prefer PD or SIP) to change based on whether PD is enabled.

jayanthvn:
> High level: We have a _lot_ of "if enablePD $this else $that". This code structure is a lot of tech debt to carry forward, since we're going to have to think about both branches all the time. I _expected_ to see only the place where we allocate new IPs (and thus prefer PD or SIP) to change based on whether PD is enabled.

Sure Gus, I will revisit the code remove unnecessary branching.

Makefile
```diff
@@ -285,6 +287,16 @@ generate-k8s-yaml:
 
 release: generate-k8s-yaml upload-resources-to-github
 
+setup-ec2-sdk-override:
+	@if [ "$(EC2_SDK_OVERRIDE)" = "y" ] ; then \
```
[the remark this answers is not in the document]
jayanthvn: I was trying to add similar to this - https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/appmesh_models_override/setup.sh and https://github.com/aws/aws-app-mesh-controller-for-k8s/blob/master/Makefile#L96. So that if any new feature needs to generate the model, they can run the script. Maybe I will remove from make file and keep this script in a helper directory and provide comments on how to use it?

misc/eni-max-pods.txt
```diff
@@ -11,7 +11,7 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 #
-# This file was generated at 2021-01-24T12:55:41+11:00
+# This file was generated at 2021-04-09T07:14:42Z
```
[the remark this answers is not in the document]
jayanthvn: I think this is unrelated. I had regenerated the vpc_ip_resource_limit.go to have "nitro" hypervisor type. That seems to have picked up these instance. Will re-check these instances.

pkg/awsutils/awsutils.go
```diff
@@ -164,26 +170,29 @@ type APIs interface {
 
 	//RefreshSGIDs
 	RefreshSGIDs(mac string) error
+
+	//GetInstanceHypervisorFamily returns the hypervisor family for the instance
+	GetInstanceHypervisorFamily() (string, error)
```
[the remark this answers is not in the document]
jayanthvn: I am using this in ipamd.go [https://github.com/jayanthvn/amazon-vpc-cni-k8s/blob/pd_dev/pkg/ipamd/ipamd.go#L319] to fail if PD is enable on non-nitro instances.

pkg/awsutils/awsutils.go
```diff
@@ -975,6 +1006,35 @@ func (cache *EC2InstanceMetadataCache) GetIPv4sFromEC2(eniID string) (addrList [
...
 
+// GetIPv4PrefixesFromEC2 calls EC2 and returns a list of all addresses on the ENI
+func (cache *EC2InstanceMetadataCache) GetIPv4PrefixesFromEC2(eniID string) (addrList []*ec2.Ipv4PrefixSpecification, err error) {
+	eniIds := make([]*string, 0)
+	eniIds = append(eniIds, aws.String(eniID))
+	input := &ec2.DescribeNetworkInterfacesInput{NetworkInterfaceIds: eniIds}
+
+	start := time.Now()
+	result, err := cache.ec2SVC.DescribeNetworkInterfacesWithContext(context.Background(), input)
+	awsAPILatency.WithLabelValues("DescribeNetworkInterfaces", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
```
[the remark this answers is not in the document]
jayanthvn: Agreed, will add a wrapper function and update all the callers.

pkg/awsutils/awsutils.go
```diff
@@ -1262,15 +1349,25 @@ func (cache *EC2InstanceMetadataCache) waitForENIAndIPsAttached(eni string, want
...
-				eniIPCount := len(returnedENI.IPv4Addresses)
-				if eniIPCount <= 1 {
-					log.Debugf("No secondary IPv4 addresses available yet on ENI %s", returnedENI.ENIID)
+				var eniIPCount int
+
+				if cache.ipv4PrefixDelegation {
+					eniIPCount = len(returnedENI.IPv4Prefixes)
+					log.Debugf("Found prefix count - %d", eniIPCount)
+				} else {
+					eniIPCount = len(returnedENI.IPv4Addresses)
```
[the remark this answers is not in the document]
jayanthvn: Right will make the change, while implementing I was thinking of adding separate debug lines but I think it is not needed here.

pkg/awsutils/awsutils_test.go
```diff
@@ -151,10 +202,28 @@ func TestInitWithEC2metadataErr(t *testing.T) {
...
 		metadataMACPath: primaryMAC + " " + eni2MAC,
-		metadataMACPath + eni2MAC + metadataDeviceNum:  eni2Device,
-		metadataMACPath + eni2MAC + metadataInterface:  eni2ID,
-		metadataMACPath + eni2MAC + metadataSubnetCIDR: subnetCIDR,
-		metadataMACPath + eni2MAC + metadataIPv4s:      eni2PrivateIP,
+		metadataMACPath + eni2MAC + metadataDeviceNum:    eni2Device,
+		metadataMACPath + eni2MAC + metadataInterface:    eni2ID,
+		metadataMACPath + eni2MAC + metadataSubnetCIDR:   subnetCIDR,
+		metadataMACPath + eni2MAC + metadataIPv4s:        eni2PrivateIP,
+		metadataMACPath + eni2MAC + metadataIPv4Prefixes: nil,
```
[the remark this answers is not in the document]
jayanthvn: You are right if the prefix is not allocated then the key won't be present, it is not used in this test. Shouldn't be here will remove it.

pkg/awsutils/imds.go
```diff
@@ -177,12 +177,31 @@ func (imds TypedIMDS) getCIDRs(ctx context.Context, key string) ([]net.IPNet, er
 	return cidrs, nil
 }
 
+func (imds TypedIMDS) getPrefixes(ctx context.Context, key string) ([]string, error) {
```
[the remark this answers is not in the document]
jayanthvn: Yes, it is identical. Reason I added was for readability but thinking about it again, it should be fine to replace with GetLists since the caller is GetLocalIPv4Prefixes.

pkg/awsutils/imds.go
```diff
@@ -177,12 +177,31 @@ func (imds TypedIMDS) getCIDRs(ctx context.Context, key string) ([]net.IPNet, er
...
 // GetLocalIPv4s returns the private IPv4 addresses associated with the interface.  First returned address is the primary address.
 func (imds TypedIMDS) GetLocalIPv4s(ctx context.Context, mac string) ([]net.IP, error) {
 	key := fmt.Sprintf("network/interfaces/macs/%s/local-ipv4s", mac)
 	return imds.getIPs(ctx, key)
 }
 
+// GetLocalIPv4Prefixes returns the private IPv4 addresses associated with the interface.  First returned address is the primary address.
+func (imds TypedIMDS) GetLocalIPv4Prefixes(ctx context.Context, mac string) ([]string, error) {
+	key := fmt.Sprintf("network/interfaces/macs/%s/ipv4-prefix", mac)
+	prefixes, err := imds.getPrefixes(ctx, key)
```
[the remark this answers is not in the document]
jayanthvn: Yeah, I can do that and later store in `Ipv4PrefixSpecification` as `Ipv4Prefix: aws.String(ipv4.String())`

pkg/ipamd/datastore/prefix_store.go
```diff
@@ -0,0 +1,176 @@
...
+	//"/32" IPv4 prefix
+	ipv4DefaultPrefixSize = 32
+)
+
+var log = logger.Get()
+
+// PrefixIPsStore will hold the size of the prefix Eg: 16 for /28 prefix
+// UsedIPs will be a bitmap of size - IPsPerPrefix which will keep track of IPs used
+type PrefixIPsStore struct {
+	UsedIPs      []byte
```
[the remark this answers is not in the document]
jayanthvn: Yes, I totally agree I wanted to make the change post preview, something like have an index DB which will be min (max IPs per ENI, max_pods), getNextIP will give the next free index which will be added to the prefix.

pkg/ipamd/datastore/prefix_store.go
```diff
@@ -0,0 +1,176 @@
...
+	CooldownIPs  []time.Time
```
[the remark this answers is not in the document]
jayanthvn: Yeah, I was trying to see if I can avoid the cool down array so each time a new IP is generated out of the prefix DB, I had to maintain a cool down cache of recently freed and check if the IP is present, if so move on to the next index.

pkg/ipamd/datastore/prefix_store.go
jayanthvn: Yes, I was actually working on that. Will update it.

pkg/ipamd/rpc_handler.go
```diff
@@ -195,7 +195,24 @@ func (s *server) DelNetwork(ctx context.Context, in *rpc.DelNetworkRequest) (*rp
 		IfName:      in.IfName,
 		NetworkName: in.NetworkName,
 	}
-	ip, deviceNumber, err := s.ipamContext.dataStore.UnassignPodIPv4Address(ipamKey)
+	eni, ip, deviceNumber, mismatchedStore, err := s.ipamContext.dataStore.UnassignPodIPv4Address(ipamKey)
+	if eni != nil && mismatchedStore {
+		if s.ipamContext.enableIpv4PrefixDelegation {
```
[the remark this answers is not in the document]
jayanthvn: With the current implementation when PD is enabled, I am just checking the prefix pool in the reconciler and based on warm prefix target, I am freeing the prefix if allocated more. So when a pod with secondary IP is released then I am forcefully deallocating the secondary IPs. Sure I can check to move the logic here - https://github.com/jayanthvn/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/ipamd.go#L481

pkg/ipamd/ipamd.go
jayanthvn: Ahh my bad, wanted to add prefix related debug.

pkg/ipamd/ipamd.go
jayanthvn: Yeah, I am handling that in the delete path. After the IP is deleted, I either delete the secondary IP or the prefix based on the mode. But as you suggested, I will fix up the reconciler and remove the cleanup in the delete path.

pkg/ipamd/datastore/data_store.go
```diff
@@ -285,7 +294,7 @@ type CheckpointEntry struct {
 // ReadBackingStore initialises the IP allocation state from the
 // configured backing store.  Should be called before using data
 // store.
-func (ds *DataStore) ReadBackingStore() error {
+func (ds *DataStore) ReadBackingStore(enableIpv4PrefixDelegation bool) error {
```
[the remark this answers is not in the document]
jayanthvn: Ahh missed during cleanup.

pkg/ipamd/datastore/data_store.go
```diff
@@ -144,18 +146,24 @@ type ENI struct {
...
 	IsEFA bool
+	// IsPDenabled indicates whether prefix delegation is enabled
+	IsPDenabled bool
 	// DeviceNumber is the device number of ENI (0 means the primary ENI)
 	DeviceNumber int
 	// IPv4Addresses shows whether each address is assigned, the key is IP address, which must
 	// be in dot-decimal notation with no leading zeros and no whitespace(eg: "10.1.0.253")
 	IPv4Addresses map[string]*AddressInfo
+	// Key is the prefix and value is prefix, used IPs under the prefix
+	IPv4Prefixes map[string]*ENIPrefix
```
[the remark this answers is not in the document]
jayanthvn: Makes sense :)

pkg/ipamd/datastore/data_store.go
```diff
@@ -343,22 +352,69 @@ func (ds *DataStore) ReadBackingStore() error {
...
+			ds.log.Infof("Got IP to recover %s\n", allocation.IPv4)
+			ipv4Prefix := getPrefixFromIPv4Addr(allocation.IPv4)
+			ds.log.Infof("Retrieved prefix %s", ipv4Prefix.String())
+
+			eni := eniPrefixes[ipv4Prefix.String()]
+			if eni == nil {
+				ds.log.Infof("datastore: Sandbox %s uses unknown IPv4 %s - presuming stale/dead", allocation.IPAMKey, allocation.IPv4)
+				continue
+			}
+			eniPrefixDB := ds.eniPool[eni.ID].IPv4Prefixes[ipv4Prefix.String()]
```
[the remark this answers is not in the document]
jayanthvn: Yeah, it will never be `nil`

pkg/ipamd/ipamd.go
```diff
@@ -300,9 +311,20 @@ func New(k8sapiClient kubernetes.Interface, eniConfig *eniconfig.ENIConfigContro
...
 	c.disableENIProvisioning = disablingENIProvisioning()
 	c.enablePodENI = enablePodENI()
+
+	hypervisorType, err := c.awsClient.GetInstanceHypervisorFamily()
+	if err != nil {
+		log.Error("Failed to get hypervisor type")
+		return nil, err
+	}
+	if hypervisorType != "nitro" && c.enableIpv4PrefixDelegation {
+		log.Error("Hypervisor is not nitro based and prefix delegation is not supported")
```
fawadkhaliq, line 325: Hypervisor is not nitro based and prefix delegation is not supported -> Prefix delegation is not supported non-nitro instances %s ..

pkg/ipamd/datastore/data_store.go
```diff
@@ -343,22 +352,69 @@ func (ds *DataStore) ReadBackingStore() error {
...
+
+			IPindex := getPrefixIndexfromIP(allocation.IPv4, ipv4Prefix)
+			if err := eniPrefixDB.AllocatedIPs.SetUnsetIPallocation(IPindex); err != nil {
+				ds.log.Infof("Invalid index, verify getPrefixIndex")
+				return err
+			}
+
+			err := ds.AddPrefixIPv4AddressToStore(eni.ID, allocation.IPv4)
+			if err != nil && err.Error() != IPAlreadyInStoreError {
+				ds.log.Infof("Did it come here in restore?")
```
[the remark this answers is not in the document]
jayanthvn: Yes, will remove it.

pkg/ipamd/datastore/data_store.go
```diff
@@ -343,22 +352,69 @@ func (ds *DataStore) ReadBackingStore() error {
...
+			eniPrefixDB.FreeIps--
+
+			IPindex := getPrefixIndexfromIP(allocation.IPv4, ipv4Prefix)
+			if err := eniPrefixDB.AllocatedIPs.SetUnsetIPallocation(IPindex); err != nil {
+				ds.log.Infof("Invalid index, verify getPrefixIndex")
+				return err
+			}
+
+			err := ds.AddPrefixIPv4AddressToStore(eni.ID, allocation.IPv4)
+			if err != nil && err.Error() != IPAlreadyInStoreError {
```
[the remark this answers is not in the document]
jayanthvn: Yeah we are recovering from the IPs, shouldn't have duplicates. If there are any we need to flag an error.

pkg/ipamd/ipamd.go
```diff
@@ -562,13 +591,14 @@ func (c *IPAMContext) tryFreeENI() {
 	}
 }
 
-// tryUnassignIPsFromAll determines if there are IPs to free when we have extra IPs beyond the target and warmIPTargetDefined
+// tryUnassignIPsorPrefixesFromAll determines if there are IPs to free when we have extra IPs beyond the target and warmIPTargetDefined
 // is enabled, deallocate extra IP addresses
-func (c *IPAMContext) tryUnassignIPsFromAll() {
-	if _, over, warmIPTargetDefined := c.ipTargetState(); warmIPTargetDefined && over > 0 {
+func (c *IPAMContext) tryUnassignIPsOrPrefixesFromAll() {
```
fawadkhaliq, line 596: Consider keeping them separate. These are two different (similar looking) workflows but something you can do without touching the existing functions. Instead of `tryUnassignIPsOrPrefixesFromAll`, you can have `tryUnassignIPsFromAll` and `tryUnassignPrefixesFromAll` and you won't have to pass the `c.enableIpv4PrefixDelegation` flag all the way through functions and you won't have overload variables based on the `c.enableIpv4PrefixDelegation` flag. It'll be simpler and easier to maintain. You could, of course merge, any common functionality into a helper if there's a need

pkg/ipamd/ipamd.go
```diff
@@ -579,66 +609,70 @@ func (c *IPAMContext) tryUnassignIPsFromAll() {
...
 	}
 }
 
-// findFreeableIPs finds and returns IPs that are not assigned to Pods but are attached
+// findFreeableIPsOrPrefixes finds and returns IPs that are not assigned to Pods but are attached
 // to ENIs on the node.
-func (c *IPAMContext) findFreeableIPs(eni string) ([]string, error) {
-	freeableIPs := c.dataStore.FreeableIPs(eni)
-
+func (c *IPAMContext) findFreeableIPsOrPrefixes(eni string) ([]string, error) {
```
fawadkhaliq, line 643: Same feedback as `tryUnassignIPsOrPrefixesFromAll `

pkg/ipamd/ipamd.go
```diff
@@ -726,15 +769,22 @@ func (c *IPAMContext) tryAllocateENI() error {
...
 }
 
-// For an ENI, try to fill in missing IPs on an existing ENI
-func (c *IPAMContext) tryAssignIPs() (increasedPool bool, err error) {
-	// If WARM_IP_TARGET is set, only proceed if we are short of target
-	short, _, warmIPTargetDefined := c.ipTargetState()
-	if warmIPTargetDefined && short == 0 {
+// For an ENI, try to fill in missing IPs on an existing ENI with PD disabled
+// try to fill in missing Prefixes on an existing ENI with PD enabled
+func (c *IPAMContext) tryAssignIPsOrPrefixes() (increasedPool bool, err error) {
```
fawadkhaliq, line 774: Same feedback as `tryUnassignIPsOrPrefixesFromAll`

pkg/ipamd/datastore/data_store.go
jayanthvn: Yeah, make sense to have it as `isSecondaryIP`

pkg/ipamd/datastore/data_store.go
jayanthvn: Yeah agreed.

pkg/ipamd/ipamd.go
```diff
@@ -1344,3 +1604,137 @@ func (c *IPAMContext) SetNodeLabel(key, value string) error {
...
+	} else if c.enableIpv4PrefixDelegation {
+		//Single PD - allocate one prefix per ENI and new add will be new ENI + prefix
+		//Multi - allocate one prefix per ENI and new add will be new prefix or new ENI + prefix
+		maxPrefixesPerENI, maxIpsPerPrefix, _ = datastore.GetPrefixDelegationDefaults()
+		maxIPsPerENI = maxPrefixesPerENI * maxIpsPerPrefix
+		log.Infof("max prefix %d max ips %d", maxPrefixesPerENI, maxIPsPerENI)
+		return maxIPsPerENI, maxPrefixesPerENI, nil
+	}
+	return 0, 0, nil
+}
```
fawadkhaliq, line 1703: Could change to: 

```
func (c *IPAMContext) GetIPv4Limit() (int, int, error) {
	var maxIPsPerENI, maxPrefixesPerENI, maxIpsPerPrefix int
	var err error
	if !c.enableIpv4PrefixDelegation {
		maxIPsPerENI, err = c.awsClient.GetENIIPv4Limit()
		maxPrefixesPerENI = 0
		if err != nil {
			return 0, 0, err
		}
	} else if c.enableIpv4PrefixDelegation {
		//Single PD - allocate one prefix per ENI and new add will be new ENI + prefix
		//Multi - allocate one prefix per ENI and new add will be new prefix or new ENI + prefix
		maxPrefixesPerENI, maxIpsPerPrefix, _ = datastore.GetPrefixDelegationDefaults()
		maxIPsPerENI = maxPrefixesPerENI * maxIpsPerPrefix
		log.Infof("max prefix %d max ips %d", maxPrefixesPerENI, maxIPsPerENI)
	}
	return maxIPsPerENI, maxPrefixesPerENI, nil
}
```

fawadkhaliq (changes requested):

pkg/ipamd/ipamd.go
jayanthvn: Yes, I had it as 2 separate functions but based on internal feedback I merged them. Since even I wanted to keep the existing functions untouched. Sure will separate it out.

pkg/ipamd/datastore/data_store.go
jayanthvn: So if vpc-cni started with PD enabled and is not assigned, I free it up at nodeInit - https://github.com/jayanthvn/amazon-vpc-cni-k8s/blob/pd_dev/pkg/ipamd/ipamd.go#L428. If the pod is assigned associated with secondary IP then on delete, I go free the IP.

jayanthvn: CleanupDataStore with PD enabled will just cleanup prefixes and associated IPs in the prefix.

pkg/ipamd/ipamd.go
jayanthvn: oh good catch, merge messed it up :(

pkg/ipamd/ipamd.go
jayanthvn: I had removed it, merge from internal repo to GitHub messed this up :(

config/master/manifests.jsonnet
```diff
@@ -174,6 +174,7 @@ local awsnode = {
                 DISABLE_INTROSPECTION: "false",
                 DISABLE_METRICS: "false",
                 ENABLE_POD_ENI: "false",
+                ENABLE_PREFIX_DELEGATION: "true",
```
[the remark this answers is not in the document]
jayanthvn: Default is `false`. This came up recently will update it.

pkg/awsutils/awsutils.go
```diff
@@ -1296,7 +1397,39 @@ func (cache *EC2InstanceMetadataCache) waitForENIAndIPsAttached(eni string, want
 }
 
 // DeallocIPAddresses allocates numIPs of IP address on an ENI
-func (cache *EC2InstanceMetadataCache) DeallocIPAddresses(eniID string, ips []string) error {
+func (cache *EC2InstanceMetadataCache) DeallocIPAddresses(eniID string, ips []string, igonoreCachedPDflag bool) error {
```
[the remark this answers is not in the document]
jayanthvn: This is mainly for scenarios where the Pod is associated with secondaryIP and the mode is PD enabled. When he pod is deleted, we need to unassign the IP instead of prefix. Hence added a override flag.

pkg/ipamd/datastore/data_store.go
anguslees: > if vpc-cni started with PD enabled and is not assigned, I free it up at nodeInit

I think only unassigned IPs get freed at startup.  Pods that are still running while ipamd is restarted will continue running (good), and I _think_ when they eventually exit, we will reach this function with their secondary IP?

Makefile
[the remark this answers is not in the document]
anguslees: > Maybe I will remove from make file and keep this script in a helper directory and provide comments on how to use it?

I like this plan.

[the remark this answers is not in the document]
jayanthvn: Thank you, will do that.

pkg/ipamd/datastore/data_store.go
jayanthvn: Yes the running pod will continue but when they exit, I am force cleaning it up here - https://github.com/jayanthvn/amazon-vpc-cni-k8s/blob/pd_dev/pkg/ipamd/rpc_handler.go#L206.

pkg/ipamd/datastore/prefix_store.go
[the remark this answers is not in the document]
jayanthvn: Modified the entire datastore implementation, will be updating the PR.

pkg/ipamd/rpc_handler.go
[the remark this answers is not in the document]
jayanthvn: Discussed with Gus, I will be taking this up after merging this PR. This needs some rework to merge the PD and Secondary IP databases and fix up the reconciler code.

pkg/ipamd/datastore/data_store.go
jayanthvn: Fixed this to be net.IP

pkg/ipamd/datastore/prefix_store.go
jayanthvn: Fixed this.

pkg/awsutils/awsutils.go
```diff
@@ -315,8 +331,27 @@ func (i instrumentedIMDS) GetMetadataWithContext(ctx context.Context, p string)
 	return result, nil
 }
 
+func (cache *EC2InstanceMetadataCache) GetNetworkInterfacesWithContext(ctx context.Context, eniIds []*string) (*ec2.DescribeNetworkInterfacesOutput, error) {
```
M00nF1sh, line 334: 1. start with lower case for private functions. seems `GetNetworkInterfacesWithContext` is not exposed.
2. it's better to use `eniIds []string`. avoid use pointers when unnecessary.

pkg/awsutils/awsutils.go
```diff
@@ -315,8 +331,27 @@ func (i instrumentedIMDS) GetMetadataWithContext(ctx context.Context, p string)
...
+	input := &ec2.DescribeNetworkInterfacesInput{NetworkInterfaceIds: eniIds}
+
+	start := time.Now()
```
M00nF1sh, line 337: not related to this PR(since all other functions are already collecting metrics this way).
but for aws api related metrics, it's better to collect it this way: https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/pkg/aws/metrics/collector.go
which is not intrusive to application code.

pkg/awsutils/awsutils.go
```diff
@@ -315,8 +331,27 @@ func (i instrumentedIMDS) GetMetadataWithContext(ctx context.Context, p string)
...
+	result, err := cache.ec2SVC.DescribeNetworkInterfacesWithContext(context.Background(), input)
```
M00nF1sh, line 338: seems this API is paginated, is it handled within ec2SVC?

pkg/awsutils/awsutils.go
```diff
@@ -315,8 +331,27 @@ func (i instrumentedIMDS) GetMetadataWithContext(ctx context.Context, p string)
...
+		awsAPIErrInc("DescribeNetworkInterfaces", err)
+		log.Errorf("Failed to get ENI %s information from EC2 control plane %v", eniIds, err)
+		return nil, errors.Wrap(err, "failed to describe network interface")
+	}
+	return result, nil
+}
+
 // New creates an EC2InstanceMetadataCache
-func New(useCustomNetworking bool) (*EC2InstanceMetadataCache, error) {
+func New(useCustomNetworking, ipv4PrefixDelegation bool) (*EC2InstanceMetadataCache, error) {
```
M00nF1sh, line 354: something like `useIPv4PrefixDelegation` or `enableIPv4PrefixDelegation`

pkg/awsutils/awsutils.go
```diff
@@ -340,6 +375,9 @@ func New(useCustomNetworking bool) (*EC2InstanceMetadataCache, error) {
 	cache.useCustomNetworking = useCustomNetworking
 	log.Infof("Custom networking %v", cache.useCustomNetworking)
 
+	cache.ipv4PrefixDelegation = ipv4PrefixDelegation
+	log.Infof("PD enabled %v", cache.ipv4PrefixDelegation)
```
M00nF1sh, line 379: this log is customer visible, better `s/PD/Prefix Delegation`

pkg/awsutils/awsutils.go
jayanthvn: Yeah makes sense, will add a todo and take up as part of clean up efforts.

pkg/awsutils/awsutils.go
```diff
@@ -549,13 +587,26 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
 			PrivateIpAddress: aws.String(ip4.String()),
 		}
 	}
+	ec2ipv4Prefixes := make([]*ec2.Ipv4PrefixSpecification, 0)
+	if (eniMAC == primaryMAC && !cache.useCustomNetworking) || (eniMAC != primaryMAC) {
```
M00nF1sh, line 591: i think we need some comments here for the condition here to help ppl understand.

pkg/awsutils/awsutils.go
jayanthvn: Pagination is done only for this func - https://github.com/aws/amazon-vpc-cni-k8s/blob/9ad4860bbe2f7fe8548670447ac69c8ee989d872/pkg/awsutils/awsutils.go#L1326 since we do describe on all ENIs, the calls here does describe on only certain ENIs.

pkg/awsutils/awsutils.go
```diff
@@ -549,13 +587,26 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
 			PrivateIpAddress: aws.String(ip4.String()),
 		}
 	}
+	ec2ipv4Prefixes := make([]*ec2.Ipv4PrefixSpecification, 0)
```
M00nF1sh, line 590: the golang idiomatic way of slice with 0 length is `var ec2ipv4Prefixes []*ec2.Ipv4PrefixSpecification`

pkg/awsutils/awsutils.go
jayanthvn: Yes will add a comment, so we had to include a flag for useCustomNetworking since we have functions like refresh SGs which shouldn't update the SG on secondary ENIs.

pkg/awsutils/awsutils.go
```diff
@@ -1218,9 +1278,21 @@ func (cache *EC2InstanceMetadataCache) AllocIPAddresses(eniID string, numIPs int
 	}
 
 	log.Infof("Trying to allocate %d IP addresses on ENI %s", needIPs, eniID)
-	input := &ec2.AssignPrivateIpAddressesInput{
-		NetworkInterfaceId:             aws.String(eniID),
-		SecondaryPrivateIpAddressCount: aws.Int64(int64(needIPs)),
+
+	input := &ec2.AssignPrivateIpAddressesInput{}
+
+	if cache.ipv4PrefixDelegation {
```
M00nF1sh, line 1284: let's create separate function for allocate IpAddress vs allocate IPPrefix.

pkg/awsutils/awsutils.go
```diff
@@ -1262,15 +1338,23 @@ func (cache *EC2InstanceMetadataCache) waitForENIAndIPsAttached(eni string, want
 		for _, returnedENI := range enis {
 			if eni == returnedENI.ENIID {
 				// Check how many Secondary IPs have been attached
-				eniIPCount := len(returnedENI.IPv4Addresses)
-				if eniIPCount <= 1 {
+				var eniIPCount int
+				if cache.ipv4PrefixDelegation {
```
M00nF1sh, line 1342: better to have separate function for ipv6 address vs prefix

pkg/awsutils/awsutils.go
```diff
@@ -1296,16 +1384,42 @@ func (cache *EC2InstanceMetadataCache) waitForENIAndIPsAttached(eni string, want
 }
 
 // DeallocIPAddresses allocates numIPs of IP address on an ENI
-func (cache *EC2InstanceMetadataCache) DeallocIPAddresses(eniID string, ips []string) error {
+func (cache *EC2InstanceMetadataCache) DeallocIPAddresses(eniID string, ips []string, ignoreCachedPDflag bool) error {
```
M00nF1sh, line 1387: this `ignoreCachedPDflag` is not clear from interface design. (i can understand it means ignore the ipv4PrefixDelegation setting).
We should change this into two separate function for ipv4 address and for prefix

pkg/ipamd/datastore/data_store.go
```diff
@@ -144,18 +145,23 @@ type ENI struct {
 	IsTrunk bool
 	// IsEFA indicates whether this ENI is tagged as an EFA
 	IsEFA bool
+	// IsPDenabled indicates whether prefix delegation is enabled
+	IsPDenabled bool
```
M00nF1sh, line 149: nit: s/IsPDenabled/IsPDEnabled

pkg/awsutils/awsutils.go
jayanthvn: The reason I didn't separate the functions was because we need to duplicate the functions expect for the input and if any fixes, we need to make sure it goes in 2 places. Hence I added the check at the last step to keep the changes minimal.

pkg/awsutils/awsutils.go
jayanthvn: Same as dealloc, I didn't separate it because of code duplication. Will sync up with you offline.

pkg/awsutils/awsutils.go
jayanthvn: Input to the DescribeNetworkInterface func is array of pointers hence I am passing it to avoid multiple conversions - `input := &ec2.DescribeNetworkInterfacesInput{NetworkInterfaceIds: eniIds}`. Since this is not exposed to any other APIs, is it fine to keep it as pointers?

```

    // One or more network interface IDs.
    //
    // Default: Describes all your network interfaces.
    NetworkInterfaceIds []*string `locationName:"NetworkInterfaceId" locationNameList:"item" type:"list"`
```

pkg/awsutils/awsutils.go
M00nF1sh: just for future reference. 
Ideally high-level logic like `useCustomNetworking` shouldn't interact with low-level component like this one.
I'd prefer a function like `getENIMetadata(eniMAC string, opts... eniMetadataOpts)`, which eniMetadataOpts have options like whether to include prefix. and we have a set of default attributes to fetch. and the high-level code can pass different opts based on it's business logic(e.g. useCustomNetworking)

pkg/awsutils/awsutils.go
M00nF1sh: if it accepts a list, we need to handle pagination correctly within that function. or change it to accept a single eni.

pkg/awsutils/awsutils.go
```diff
@@ -1296,16 +1388,42 @@ func (cache *EC2InstanceMetadataCache) waitForENIAndIPsAttached(eni string, want
...
+	ipsInput := aws.StringSlice(ips)
+
+	input := &ec2.UnassignPrivateIpAddressesInput{}
+	if !cache.useIPv4PrefixDelegation || ignoreCachedPDflag {
+		input = &ec2.UnassignPrivateIpAddressesInput{
+			NetworkInterfaceId: aws.String(eniID),
+			PrivateIpAddresses: ipsInput,
+		}
+	} else {
+		input = &ec2.UnassignPrivateIpAddressesInput{
```
M00nF1sh, line 1402: why we have both DeallocIPAddresses & DeallocPrefixAddresses and DeallocIPAddresses contains same logic as DeallocPrefixAddresses for prefix

M00nF1sh (approved):
discussed with Jayanth offline.
This is for preview only and will only merge into preview branch instead of master branch.

Approving this one so that we produce a preview version for testing.
However, this PR do need more work before merging into master.

jayanthvn:
> discussed with Jayanth offline.
> This is for preview only and will only merge into preview branch instead of master branch.
> 
> Approving this one so that we produce a preview version for testing.
> However, this PR do need more work before merging into master.

Thanks Yang. Will take up the restructuring part post preview.

pkg/awsutils/awsutils.go
jayanthvn: So if PD is enabled but we are freeing Secondary IP [this can happen during upgrades/knob disable to enable] then we set "ignoreCachedPDflag" so that we free the secondary IP.

pkg/ipamd/ipamd.go
jayanthvn: Discussed with Fawad offline, will see how I can restructure this. Mainly if we have 2 functions `tryUnassignIPsFromAll` and  `tryUnassignPrefixesFromAll`, both functions will be doing very similar tasks of getting the current warm setting, get free IPs/prefixes, clean up datastore and dealloc the prefix/IP so instead I pushed the logic of returning a IP or prefix to the final API - `findFreeableIPsOrPrefixes` which does - 

```
if !c.enableIpv4PrefixDelegation {
		freeableIPs = c.dataStore.FreeableIPs(eni)
	} else if c.enableIpv4PrefixDelegation {
		freeableIPs = c.dataStore.FreeablePrefixes(eni)
	}
```

pkg/ipamd/ipamd.go
jayanthvn: Same here, to avoid adding multiple conditions here - 

```
if c.isDatastorePoolTooLow() {
		c.increaseDatastorePool()
	} else if c.isDatastorePoolTooHigh() {
		c.decreaseDatastorePool(decreaseIPPoolInterval)
	}
```
I made a helper function which will tell based on the mode if the prefix pool is low/high or IP pool is low/high. Will take up the task of restructuring the code post this PR.

pkg/ipamd/ipamd.go
jayanthvn: Same as above, discussed about this. Will take it up in the following PR.

pkg/awsutils/awsutils.go
```diff
@@ -117,6 +117,9 @@ type APIs interface {
 	// GetIPv4sFromEC2 returns the IPv4 addresses for a given ENI
 	GetIPv4sFromEC2(eniID string) (addrList []*ec2.NetworkInterfacePrivateIpAddress, err error)
 
+	// GetIPv4PrefixesFromEC2 returns the IPv4 addresses for a given ENI
+	GetIPv4PrefixesFromEC2(eniID string) (addrList []*ec2.Ipv4PrefixSpecification, err error)
```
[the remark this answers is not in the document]
jayanthvn: Will take this up as part of reconciler changes to fix these APIs.

pkg/awsutils/awsutils.go
jayanthvn: Fixed the API.

fawadkhaliq (approved):
/lgtm for preview

(Jayanth will address code structure feedback as a fast follow up)

README.md
```diff
@@ -450,6 +450,30 @@ You can use the below command to enable `DISABLE_TCP_EARLY_DEMUX` to `true` -
...
+---
+
+`WARM_PREFIX_TARGET` 
+
+Type: Integer
+
+Default: None
+
+Specifies the number of free IPv4(/28) prefixes that the `ipamd` daemon should attempt to keep available for pod assignment on the node.
+This environment variable overrides `WARM_ENI_TARGET`, `WARM_IP_TARGET` and `MINIMUM_IP_TARGET` and works when `ENABLE_PREFIX_DELEGATION`
```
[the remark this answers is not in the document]
jayanthvn: Hi @asheldon,

Thanks for your feedback. With this feature(ENABLE_PREFIX_DELEGATION=true), we cannot support `WARM_ENI_TARGET, WARM_IP_TARGET, and MINIMUM_IP_TARGET` mainly because IPs will be reserved in chunks of 16 (/28 prefix). So say if you set WARM_IP_TARGET as 7, we need to still allocate 2 prefixes.

In the example which you have given for `WARM_PREFIX_TARGET == 1`, yes prefix D will be kept in the current implementation. I am planning to improve this by checking the total number of IPs and if it is sufficient for reaching the WARM_PREFIX_TARGET.

Yes it is logically equivalent to `WARM_IP_TARGET == 16` but we need to allocate in chunks of /28 prefixes with this feature hence we cannot have fine grained controls like existing WARM targets. Each secondary IP will be replaced by a /28 prefix hence we cannot support having both prefixes and secondary IPs.

README.md
```diff
@@ -450,6 +450,30 @@ You can use the below command to enable `DISABLE_TCP_EARLY_DEMUX` to `true` -
...
+---
+
+`ENABLE_PREFIX_DELEGATION` (Since v1.9)
+
+Type: Boolean as a String
+
+Default: `false`
+
+To enable IPv4 prefix delegation on nitro instances. Setting `ENABLE_PREFIX_DELEGATION` flag toggle to `true` will start allocating a /28 prefix 
+instead of a secondary IP in the ENIs subnet. The total number of prefixes and private IP addresses will be less than the
```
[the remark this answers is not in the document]
jayanthvn: Yes thats right.

README.md
```diff
@@ -450,6 +450,30 @@ You can use the below command to enable `DISABLE_TCP_EARLY_DEMUX` to `true` -
...
+limit on private IPs allowed by your instance. The current preview will support a single /28 prefix per ENI. Knob toggle while pods are running or if
+ENIs are attached is not supported. On toggling the knob, node should be recycled to set the new kubelet max pods value.
```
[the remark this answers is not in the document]
jayanthvn: Sure, will reword it. I mean here when the feature is enabled or disabled, max pods will change hence will need a kubelet restart or new node group with updated max pods value.

README.md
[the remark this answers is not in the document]
asheldon: Hi @jayanthvn. Thank you for the reply.  I think nodes with low density packing should require as few as one prefix, and nodes that have high-density packing should have as many prefixes as required, more or less. `WARM_PREFIX_TARGET` doesn't allow for both - the minimum will be two prefixes unless you give up the ability for a node to have more than one prefix worth of pods. Having each node use a minimum of 1+32 IPs makes total IP consumption on the network very fast.

The way I imagine this would work is different, and basically just treats WARM_IP_TARGET as a signal to IPAMD of how many IPs to keep in reserve, like it is now. WARM_IP_TARGET:

>Specifies the number of free IP addresses that the ipamd daemon should attempt to keep available for pod assignment on the node.

In combination with `ENABLE_PREFIX_DELEGATION=true`, I would interpret this to set a floor on how many IPs we want available. For example, if `WARM_IP_TARGET` were 6 and I were running 8 pods, I would consume only one prefix. This is because one prefix is 16 IPs, 8 are used and the number of warm IPs is also 8. 8 warm IPs meets the WARM_IP_TARGET of 6, so a second prefix is not required. Then 3 more pods start on this node, bringing us to 11 running pods and only 5 available IPs. Since the WARM_IP_TARGET is 6, we would then allocate a second prefix.  `MINIMUM_IP_TARGET` can essentially work the same. `MINIMUM_PREFIXES=CEILING(MINIMUM_IP_TARGET/16)`.  Similarly, nodes that reach 17 pods might consume only two prefixes, not three.

Nodes still consume a minimum of 1+16 IPs, but this gives cluster operators the ability to hint to the CNI how much buffer they really need around at a level that is lower than a single prefix, or more than a single prefix (but not an exact multiple of 16).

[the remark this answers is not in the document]
jayanthvn: Reg this - "the minimum will be two prefixes unless you give up the ability for a node to have more than one prefix worth of pods. Having each node use a minimum of 1+32 IPs makes total IP consumption on the network grow very fast." - The number of pods that can be scheduled on the nodes are not changed, do you mean something like, if we set WARM_PREFIX_TARGET = 1 and one node has 10 pods, then that node will have 1 new prefix + 6 IPs available and other node has just one pod, even then that node will have 1 new prefix + 15 IPs available, something like this? If the pod density per node is not high then WARM_PREFIX_TARGET can be set to 0. Allocating new prefixes are faster than allocating an ENI so mostly shouldn't impact much for the pod launch time.(I need to measure performance yet)

I agree this will be a good enhancement, but I don't think we can have it as WARM_IP_TARGET since it will change the definition from what we previously supported. As you mentioned WARM_IP_TARGET is the number of free IPs in datastore and if we support for PD then it will be similar to MINIUM_IP_TARGET because at the minimum a node can have either WARM_IP_TARGET number of IPs or (WARM_IP_TARGET + X) where X can go max up to 16, In the example you mentioned, if 11 are consumed then we will have in the warm pool [ 15 IPs + 6 (WARM_IP_TARGET) ].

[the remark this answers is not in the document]
asheldon: Hi @jayanthvn. Thanks again for your response and your time.

>  The number of pods that can be scheduled on the nodes are not changed, do you mean something like, if we set WARM_PREFIX_TARGET = 1 and one node has 10 pods, then that node will have 1 new prefix + 6 IPs available and other node has just one pod, even then that node will have 1 new prefix + 15 IPs available, something like this? If the pod density per node is not high then WARM_PREFIX_TARGET can be set to 0. Allocating new prefixes are faster than allocating an ENI so mostly shouldn't impact much for the pod launch time.

What I've seen in EKS is that when a pod is scheduled on a node that has 0 available IPs in IPAM, the pod gets into a bad state with no IP and has to be manually deleted. This can be triggered with scenarios like WARM_IP_TARGET=1 and scheduling just two pods at the same time.  Because of this behavior, I don't feel comfortable running nodes in configurations that might temporarily reach 0 available IPs. This includes WARM_IP_TARGET=0 or WARM_PREFIX_TARGET=0. 

I did not understand that the design here was to just-in-time allocate an additional prefix when all existing prefixes were exhausted and additional IPs were required, and I don't trust that behavior. I don't want to allocate the additional prefix at the last possible moment, but I also don't want to _always_ allocate an extra 16 IPs. WARM_IP_TARGET is an in-between. I want to allocate the additional prefix when I'm running low on available IPs, like 6 IPs, not 0 or 16.

>  I don't think we can have it as WARM_IP_TARGET since it will change the definition from what we previously supported. 

`WARM_IP_TARGET` has never been used before with `ENABLE_PREFIX_DELEGATION`, so I think it is okay for its definition to be adjusted to behave slightly differently in that context, as long as its behavior has the same essential reasoning to it. I want to tell the CNI how many IPs I want around just-in-case. If it can only deliver IPs in blocks of 16, well, I opted into that with ENABLE_PREFIX_DELEGATION, much the same way that the number of WARM_IPs can't fall below MINIMUM_IP_TARGET, which was introduced later.  

When WARM_IP_TARGET is set below the maximum number of IPs on a single ENI, it balances the need to start pods rapidly with IP consumption. We can start at least WARM_IP_TARGET pods at any time, but we also don't consume an entire ENI worth of IPs (up to 50!) to make this guarantee.

When we use WARM_IP_TARGET or MINIMUM_IP_TARGET today, the CNI determines how many ENIs it needs to meet that need and allocates / deallocates ENIs on-demand. When WARM_IP_TARGET and MINIMUM_IP_TARGET are used in combination with ENABLE_PREFIX_DELEGATION, I believe it should work the same: the CNI determines how many PDs are required to meet these targets and allocates that many PDs.  There will be at least WARM_IP_TARGET unused IPs in each PD, but the option may save entire prefixes worth of IP space.

> As you mentioned WARM_IP_TARGET is the number of free IPs in datastore and if we support for PD then it will be similar to MINIUM_IP_TARGET because at the minimum a node can have either WARM_IP_TARGET number of IPs or (WARM_IP_TARGET + X) where X can go max up to 16, In the example you mentioned, if 11 are consumed then we will have in the warm pool [ 15 IPs + 6 (WARM_IP_TARGET) ].

I'm not following you here.  If I enable prefix delegation and set MINIMUM_IP_TARGET=36 and WARM_IP_TARGET=4, then I I expect to always have at least three prefixes (to meet MINIMUM_IP_TARGET with 3*16=48 IPs).  If I run 45 pods on this node, I would fall to only 3 available IPs, and would expect to allocate a fourth prefix, bringing my usage to 45 pods / 64 IPs.  In the prior example, there are 11 pods running and a WARM_IP_TARGET of 6. Since a single PD contains only 16 IPs, there would be only 5 available IPs left and a second PD would be allocated. The minimum number of PDs consumed by the node would be equal to CEIL(WARM_IP_TARGET/16), so the minimum number of IPs consumed would be CEIL(WARM_IP_TARGET/16)*16.

| IPs IN USE | MINIMUM_IP_TARGET | WARM_IP_TARGET | WARM_PREFIX_TARGET  | PDs | UNUSED IPs |
| ---              | ---                                 | --                            | --                                       | --     | --    | 
| 33               | UNSET                         | UNSET                    | 1                                        | 4       | 31 |
| 33               | 36                                 | 6                             | UNSET                              | 3       | 15 |
| 47               | UNSET                         | UNSET                    | 1                                        | 4       | 17 |
| 47               | 36                                 | 6                             | UNSET                              | 4       | 17 |
| 49               | UNSET                         | UNSET                    | 1                                        | 4       | 31 |
| 49               | 36                                 | 6                             | UNSET                              | 4       | 15 |
| 21                | 20                                 | 10                           | UNSET                              | 2       | 12 |
| 8                 | UNSET                          | 6                             | UNSET                              | 1       | 8  |
| 16               | UNSET                          | 4                             | UNSET                              | 2       | 12  |
| 16               | UNSET                          | UNSET                   | 0                                        | 1       | 0  |

Unless WARM_PREFIX_TARGET is set to 0, it always consumes many more IPs than what I am suggesting, but when it it set to 0, it can reach 0 available IPs which is very low.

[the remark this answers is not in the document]
jayanthvn: Thanks a lot Aaron for the detailed explanation. As I mentioned this is a very nice enhancement and would like it take it up. I will open a follow up PR to add this support. This will be a good customer experience. Really appreciate your feedback and time for this :)

PR #1458, opened 2021-05-06
Binary for mtu and veth prefix check

cgchinmay:
**What type of PR is this?**
Added binary support for validating mtu and veth prefix for aws-node daemonset

Ginkgo test won't have to run exec commands, instead they can invoke this binary to validate mtu and veth prefix

To facilitate ginkgo testing

```
/tmp # ./aws-node-config --help
Usage of ./aws-node-config:
  -mtu-val int
    	Expected MTU values, specify when validating mtu (default 9001)
  -prefix string
    	veth prefix to be used, specify when validating veth (default "eni")
  -validate string
    	name of the test, valid values: mtu, veth
  -veth-count int
    	Number of expected veth links to have the new prefix. specify when validating veth (default 1)
/tmp # 
/tmp # 
/tmp # ./aws-node-config --validate mtu -mtu-val 1300
2021/05/06 20:33:36 MTU Check: Passed
/tmp # ./aws-node-config --validate mtu -mtu-val 9001
2021/05/06 20:33:43 MTU Check: Failed
/tmp # 
/tmp # 
/tmp # 
/tmp # 
/tmp # ./aws-node-config --validate veth -prefix veth -veth-count 2
2021/05/06 20:34:33 Veth Check: Failed
/tmp # ./aws-node-config --validate veth -prefix eni -veth-count 2
2021/05/06 20:34:41 Veth Check: Passed
```


test/agent/cmd/aws-node-config/main.go
```diff
@@ -0,0 +1,92 @@
...
+	VETH_PREFIX_CHECK = "veth"
+)
+
+func main() {
+	validate := ""
+	prefix := "eni"
+	vethCount := 1
+	mtuVal := 9001
+
+	flag.StringVar(&validate, "validate", "", "name of the test, valid values: mtu, veth")
```
abhipth, line 34: As we add more validations to this binary the flag combinations required for a given validation might get confusing.

test/agent/cmd/aws-node-config/main.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	actual := 0
+	for _, ni := range interfaces {
+		if strings.HasPrefix(ni.Name, prefix) {
+			actual++
+		}
+	}
+
+	if actual >= expected {
```
abhipth, line 63: I am wondering if this check could be made more strict by verifying the list of pods get the prefix. Btw, we can use host networking pod to test this already!

test/agent/cmd/aws-node-config/main.go
```diff
@@ -0,0 +1,92 @@
...
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	ok := false
+	for _, ni := range interfaces {
+		if strings.HasPrefix(ni.Name, "eth") {
+			if ni.MTU != expected {
+				log.Fatal("MTU Check: Failed")
```
abhipth, line 80: Can you also add which interface failed the check to make debugging easier.

test/agent/Dockerfile
```diff
@@ -26,6 +26,9 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build \
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build \
     -a -o metric-server cmd/metric-server/main.go
 
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build \
```
abhipth, line 29: Can you also add description of this new binary to the README

test/agent/cmd/aws-node-config/main.go
cgchinmay: Yes, so only the name of the test is mandatory flag, others are optional and get default value. To avoid confusion
1. We can either throw validation errors for any missing related flags.
OR
2. Have separate binaries for each of these small functionalities.
let me know if you have any ideas to avoid the flag confusion.

test/agent/cmd/aws-node-config/main.go
cgchinmay: Yes, will do that

test/agent/cmd/aws-node-config/main.go
abhipth: Validation errors could prevent incorrect input. However, my comment was regarding making it easier to decide what is the right flag combinations. Is it possible to evaluate if getting a JSON input can simplify this?

cgchinmay: We can pass json object as a serialized string. That way all related flags can be together

test/agent/cmd/aws-node-config/main.go
cgchinmay: I will take a look at the host networking binary

test/agent/cmd/aws-node-config/main.go
```diff
@@ -0,0 +1,156 @@
...
+	decoded, err := base64.StdEncoding.DecodeString(config)
+	if err != nil {
+		log.Fatalf("failed to descode base64 string %s: %v", config, err)
+	}
+
+	switch validate {
+	case MTU_CHECK:
+		validateMTUValue(decoded)
+	case VETH_PREFIX_CHECK:
+		validateVethPairCount(decoded)
```
abhipth, line 55: I think this can already be done using host networking binary. Did you get a chance to see if that can be used?

cgchinmay: Yes, I did, I thought we wanted to keep this binary separate. Do you want me to move this logic to host networking binary instead ?

test/framework/utils/utils.go
```diff
@@ -14,14 +14,30 @@
 package utils
 
 import (
+	appsV1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
 
+const (
```
abhipth, line 22: These are already defined in framework/utils/constants

abhipth (approved):
LGTM!

test/framework/utils/utils.go
cgchinmay: Oh okay, will make this minor change then

jayanthvn (approved):

PR #1460, opened 2021-05-11
use symmetric return path for non-VPC traffic

### LINKED ISSUE #1392: Traffic from outside VPC does not reach pod
**What happened**:

For some reason, UDP traffic from outside the VPC is not reaching the pod. Traffic flow logs show UDP traffic flowing through the load balancer and being sent to the correct internal address. UDP traffic has been confirmed (via `tcpdump`) to reach the primary IP of the ENI that hosts the pod's IP, but does not reach the pod. UDP traffic from inside the VPC does reach the pod as expected. TCP traffic from both inside and outside the VPC reaches the pod as expected.

AWS Support has confirmed that security groups are properly configured to allow access. 

**Attach logs**

Logs are attached to AWS Support case 7990499551 which I opened 16 days ago but remains unresolved. I feel like the ticket is not getting directed to the right team, which is why I am opening this issue. I think this is a CNI bug.

With a pod listening on a UDP port and a security group allowing access from `0.0.0.0/0`, I expect traffic targeted to that UDP port to be delivered, regardless of its source address, 

1. Create an EKS cluster with a node pool in a private subnet using default configuration.
1. Create a target Pod server on a secondary ENI on one of the nodes. The only way I know to do this is to first create enough pods to fill up the first ENI. 
2. Create a load balancer target group targeting the target pod on its pod IP address (not the ENI or Node IP address).  **Critical**: ensure "Preserve Client IP" is enabled.
3. Provision a Network load balancer to forward traffic to the target Pod. 

At this point, the configuration is complete and you should be able to communicate from the public internet to the Pod using the NLB's DNS name. However, the connection does not work. On the other hand, you can verify that the Pod is serving traffic properly by connecting to it from any host in the private subnet on the same IP as is specified in the target group.

<details><summary>~~Original version~~</summary>

I don't know. The entire configuration, from creating the VPC through creating the EKS cluster through deploying the Pod, Kubernetes service, and load balancer, was all provisioned with Terraform and Helm. The almost exactly the same configuration was applied in another account and works fine as expected. (The differences between the 2 configurations are things like ID names, DNS names, ARNs, etc. that have to be different between different installations, but basically the configurations are identical in what should be all relevant aspects.) 

As deployed, the only noticeable difference is that the working instance has the Pod deployed on the EC2 instance's primary ENI and the non-working instance has the pod deployed on a secondary ENI.
</details>

The default configuration leaves `aws-node` with `AWS_VPC_K8S_CNI_EXTERNALSNAT: false`. You can check with 
```
$  kubectl -n kube-system describe daemonset aws-node | grep EXTERNALSNAT
      AWS_VPC_K8S_CNI_EXTERNALSNAT:        false
```

Unfortunately, this is incompatible with allowing Pods to be reached on their private IPs when the source IP is outside the VPC. This issue can be fixed by setting `AWS_VPC_K8S_CNI_EXTERNALSNAT=true` but there is no AWS API for doing that, thus it is not practical to set in a large organization. 

1. The CNI should be smart enough to bypass SNAT when the connection is initiated from outside the node on the pod's private IP. 
1. There should be an API and all the corresponding ways (AWS console, `aws` CLI, Terraform, etc.) for setting `AWS_VPC_K8S_CNI_EXTERNALSNAT=true`

**Environment**:
- Kubernetes version (use `kubectl version`): v1.18.9-eks-d1db3c
- CNI Version: amazon-k8s-cni:v1.7.5-eksbuild.1
- Kernel (e.g. `uname -a`): `4.14.209-160.339.amzn2.x86_64 #1 SMP Wed Dec 16 22:44:04 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`

kishorj:
Configure permissive rp_filter on the secondary ENIs and use symmetric
return path for non-VPC traffic ingress from the seconary ENIs.

**What type of PR is this?**
bug

Fixes #1392 

When external SNAT is disabled, this PR does the following 
- configure permissive RP filter for the secondary ENIs as well
- setup iptables rules to mark ingress traffic from secondary ENIs
- setup ip rules to send the return traffic through the incoming interface

This fix is needed in order to support non-VPC traffic ingress from the secondary ENIs, for example in case of NLB with IP targets and preserve client IP enabled.

N/A

- ran the automated e2e tests
- Secondary connmark can be configured via env variable on the aws-node ds
  - ip route rules get reconciled with the new secondary connmark
  - iptables mangle chain will contain the rules with the old connmark as well, manual cleanup or node restart required
- With External SNAT disabled
  - pods are able to initiate connections to hosts outside the VPC, traffic is through the primary interface
  - in case of NLB-IP, verify traffic flows as expected for client IP preservation enabled/disabled, pods on primary/secondary interfaces
  - instance mode NLB works as expected
  - in cluster traffic works as expected
  - enabled security group per pod, verified traffic works as expected
- With external SNAT is enabled
  - pod specific ip route rule with the fwmark do not get configured. 
  - instance mode NLB works fine with IP preserve enabled/disabled, pods on primary/secondary interfaces
  - IP mode NLB works as expected with IP preserve enabled/disabled, pods on primary/secondary interfaces
  - pod can communicate to hosts within or outside VPC
  - enabled security group per pod, verified traffic flows 

**Automation added to e2e**:

upgrading to new version on a running cluster works fine

Added secondaryENIConnmark configuration in the file 10.aws.conflist for cni plugin to pickup the configured connmark.

```release-note
Fixes the issue with non-VPC traffic ingress from secondary ENIs when external SNAT feature is not used.
```


scripts/init.sh
```diff
@@ -42,13 +42,18 @@ for b in $PLUGIN_BINS; do
 done
 
 # Configure rp_filter
+echo "Configure default rp_filter loose... "
```
achevuru, line 45: Below set of changes will set `rp_filter` to `2` on all the interfaces and that might be a concern because we currently allow Cx to specify via tags the ENIs they don't want VPC CNI to manage. Same is true for P4d instances that support multiple network cards. Although, we only manage interfaces belonging to network card `0` this change will modify the settings on interfaces tied to non-zero network cards as well.

achevuru:
With this change, we will start marking the packets ingressing via secondary ENIs with `0x100` by default and that might be a breaking change for some customers if the connamrk value coincides with the any connmark values they are currently using. It is good that we do a provide an option for users to choose their own secondary connmark value but I feel it might be good if we let users explicitly enable this feature/option via env variable. By doing so, they know they are enabling the connmark on packets ingressing through secondary ENIs and if they want to change the default connmark value VPC CNI uses they can go ahead and change it. Right now, we do it by default and unless someone reads the release notes they might not be aware of this new change...

scripts/init.sh
kishorj: Are there any extra information that we can use to reliably ignore the interfaces, for example via IMDS? Since we are setting the default rp_filter, any new interfaces created after this point gets the permissive rp_filter configuration.

kishorj:
> With this change, we will start marking the packets ingressing via secondary ENIs with `0x100` by default and that might be a breaking change for some customers if the connamrk value coincides with any connmark values they are currently using. It is good that we do a provide an option for users to choose their own secondary connmark value but I feel it might be good if we let users explicitly enable this feature/option via env variable. By doing so, they know they are enabling the connmark on packets ingressing through secondary ENIs and if they want to change the default connmark value VPC CNI uses they can go ahead and change it. Right now, we do it by default and unless someone reads the release notes they might not be aware of this new change...

This is not a feature, but a bug fix. Something that should have worked but didn't. Having a separate flag to enable the fix specifically would be burdensome for a far more users than the ones that might have used the `0x100` connmark on their setup previously.

PR #1471, opened 2021-05-19
Code refactor - merge ENI's secondary and prefix store into single store of CIDRs 

jayanthvn:
**What type of PR is this?**
cleanup

Merges Secondary IP DB and Prefix DB

In the preview release we had 2 DBs, with this we will have a single DB ->
AvailableCIDRs->AssignedIPs->AddressInfo
Eg:
/32 -> /32 - {address info}
 or

/28 -> ip1/32 - {address info}
            ip2/32 - {address info}

Yes

**Automation added to e2e**:

No

No

No

No
```release-note

```


pkg/ipamd/datastore/data_store.go
```diff
@@ -663,51 +628,52 @@ func (ds *DataStore) AssignPodIPv4Address(ipamKey IPAMKey) (ipv4address string,
...
-			for _, addr := range eni.IPv4Addresses {
-				if !addr.Assigned() && !addr.inCoolingPeriod() {
-					ds.assignPodIPv4AddressUnsafe(ipamKey, eni, addr)
-					if err := ds.writeBackingStoreUnsafe(); err != nil {
-						ds.log.Warnf("Failed to update backing store: %v", err)
-						// Important! Unwind assignment
-						ds.unassignPodIPv4AddressUnsafe(addr)
-						return "", -1, err
+			for _, assignedaddr := range eni.AvailableIPv4Cidrs {
+				ds.log.Debugf("DUMP Assigned addr %v", assignedaddr)
```
achevuru, line 635: DUMP --> for debug purposes?

pkg/ipamd/datastore/data_store.go
```diff
@@ -1214,10 +1192,103 @@ func (ds *DataStore) GetFreePrefixes() int {
...
+		return cachedIP, nil
+	}
+
+	//If not in cooldown then generate next IP
+	ip, ipnet, err := net.ParseCIDR(availableCidr.Cidr.String())
+	if err != nil {
+		return "", err
+	}
+
+	for ip := ip.Mask(ipnet.Mask); ipnet.Contains(ip); getNextIPv4Addr(ip) {
```
achevuru, line 1259: We might end up looping through a particular prefix twice here in bunch of cases. Can't we merge both the loops and check `isAssgined` & `inCoolingPeriod` flags on an IP address as we're traversing the prefix? If there is a strong case, to maintain two for loops we can maybe optimize it by checking on `UsedIP` count after the first traversal before trying to walk through the prefix again. For example, if `UsedIP` count is 16 after the first traversal we can return right away.

pkg/ipamd/ipamd.go
```diff
@@ -806,17 +807,18 @@ func (c *IPAMContext) tryAssignIPs() (increasedPool bool, err error) {
...
 
-		c.addENIaddressesToDataStore(ec2Addrs, eni.ID)
+		c.addENIaddressesToDataStore(ec2Addrs, nil, eni.ID)
 		return true, nil
 	}
 	return false, nil
 }
 
 func (c *IPAMContext) tryAssignPrefixes() (increasedPool bool, err error) {
+	log.Debugf("In tryAssignPrefixes")
```
achevuru: Do we need this?

pkg/ipamd/ipamd.go
```diff
@@ -806,17 +807,18 @@ func (c *IPAMContext) tryAssignIPs() (increasedPool bool, err error) {
...
 	eni := c.dataStore.GetENINeedsIP(c.maxPrefixesPerENI, c.useCustomNetworking)
 	if eni != nil {
-		currentNumberOfAllocatedPrefixes := len(eni.IPv4Prefixes)
-
+		currentNumberOfAllocatedPrefixes := len(eni.AvailableIPv4Cidrs)
+		log.Debugf("Adding prefix to ENi %s ", eni.ID)
```
achevuru, line 819: ENi -> ENI

pkg/ipamd/ipamd.go
```diff
@@ -832,9 +834,11 @@ func (c *IPAMContext) tryAssignPrefixes() (increasedPool bool, err error) {
 			ipamdErrInc("increaseIPPoolGetENIprefixedFailed")
 			return true, errors.Wrap(err, "failed to get ENI Prefix addresses during IP allocation")
 		}
-		c.addENIprefixesToDataStore(ec2Prefixes, eni.ID)
+		c.addENIaddressesToDataStore(nil, ec2Prefixes, eni.ID)
+		//c.addENIprefixesToDataStore(ec2Prefixes, eni.ID)
```
achevuru: Can we remove the unused code here

pkg/ipamd/rpc_handler.go
```diff
@@ -198,16 +198,11 @@ func (s *server) DelNetwork(ctx context.Context, in *rpc.DelNetworkRequest) (*rp
 	eni, ip, deviceNumber, mismatchedStore, err := s.ipamContext.dataStore.UnassignPodIPv4Address(ipamKey)
 	if eni != nil && mismatchedStore {
 		if s.ipamContext.enableIpv4PrefixDelegation {
-			addr := eni.IPv4Addresses[ip]
-			if addr != nil && addr.Prefix == nil {
+			cidr := net.IPNet{IP: net.ParseIP(ip), Mask: net.IPv4Mask(255, 255, 255, 255)}
+			addr := eni.AvailableIPv4Cidrs[cidr.String()].IPv4Addresses[ip]
+			if addr != nil && addr.Prefix.IP == nil {
```
achevuru, line 203: We might not need these checks here as the above check guarantees Prefix.IP is nil for this case

pkg/ipamd/rpc_handler.go
```diff
@@ -198,16 +198,11 @@ func (s *server) DelNetwork(ctx context.Context, in *rpc.DelNetworkRequest) (*rp
 	eni, ip, deviceNumber, mismatchedStore, err := s.ipamContext.dataStore.UnassignPodIPv4Address(ipamKey)
```
achevuru, line 198: As discussed, we can do away with `mismatchedStore` and instead rely on `addr.Prefix.IP`value to determine if it's a secondary IP or Prefix.

pkg/ipamd/datastore/data_store.go
jayanthvn: Yeah, will clean it up in the later PR. Need it for few more iterations.

pkg/ipamd/datastore/data_store.go
```diff
@@ -1214,10 +1192,103 @@ func (ds *DataStore) GetFreePrefixes() int {
...
+  cached ip = 10.1.1.1/32
+  /28 ->  10.1.1.3/32[assigned]
+  return 10.1.1.1/32
+*/
+
+func getUnusedIP(availableCidr *AssignedIPv4Addresses) (string, error) {
+	//Check if there is any IP out of cooldown
+	var cachedIP string
+	for _, addr := range availableCidr.IPv4Addresses {
+		if !addr.Assigned() && !addr.inCoolingPeriod() {
```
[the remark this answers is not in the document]
jayanthvn: Yeah, this will never be called for non-PD case.

pkg/ipamd/datastore/data_store.go
jayanthvn: Yeah, the first loop is to avoid adding a cleanup thread to release the IPs. Yes, I can do that, before I used to check for FreeIPs, during this refactoring it got missed, will add it here - 
```
if availableCidr.IsPrefix &&   availableCidr.usedIPs != 16 {					
    strPrivateIPv4, err := ds.getFreeIPv4AddrfromCidr(availableCidr)
```

pkg/ipamd/rpc_handler.go
jayanthvn: Yeah, now it won't be needed. Will remove it.

pkg/ipamd/ipamd.go
jayanthvn: will remove it :)

pkg/ipamd/rpc_handler.go
achevuru: Yes..

pkg/ipamd/ipamd.go
```diff
@@ -1355,7 +1357,8 @@ func (c *IPAMContext) verifyAndAddPrefixesToDatastore(eni string, attachedENIIPs
 		}
 
 		// Try to add the IP
-		err := c.dataStore.AddIPv4PrefixToStore(eni, strPrivateIPv4)
+		_, ipv4CidrPtr, _ := net.ParseCIDR(strPrivateIPv4)
```
anguslees, line 1360: Probably want to check for err return here, otherwise we're going to get a crash when we dereference a nil `ipv4CidrPtr` on the next line.

pkg/ipamd/datastore/data_store.go
```diff
@@ -224,13 +224,13 @@ func (p *ENIPool) AssignedIPv4Addresses() int {
 }
 
 // FindAddressForSandbox returns ENI and AddressInfo or (nil, nil) if not found
-func (p *ENIPool) FindAddressForSandbox(ipamKey IPAMKey) (*ENI, *AddressInfo) {
+func (p *ENIPool) FindAddressForSandbox(ipamKey IPAMKey) (*ENI, *AddressInfo, *AssignedIPv4Addresses) {
```
anguslees, line 227: Nit: Can we put these return values in 'hierarchical' order? :)
`(*ENI, *AssignedIPv4Addresses, *AddressInfo)`

pkg/ipamd/datastore/data_store.go
```diff
@@ -151,23 +154,35 @@ type ENI struct {
...
+
+type AssignedIPv4Addresses struct {
+	//Cidr info /32 or /28 prefix
+	Cidr net.IPNet
+	//Key is the /32 IP either secondary IP or free /32 IP allocated from /28 prefix
+	IPv4Addresses map[string]*AddressInfo
+	//isPrefix
+	IsPrefix bool
+	//Cnt of number of UsedIPs
+	UsedIPs int
```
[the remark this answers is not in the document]
jayanthvn: Yeah I can do that, thats right then we don't worry about incorrect counters.

pkg/ipamd/datastore/data_store.go
```diff
@@ -151,23 +154,35 @@ type ENI struct {
 	DeviceNumber int
```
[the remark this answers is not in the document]
jayanthvn: The reason I added `isPDEnabled` was because to avoid passing PD enabled flags to functions such as  "AssignPodIPv4Address" but you are right, I can make it global at the datastore level.

pkg/ipamd/datastore/data_store.go
```diff
@@ -349,59 +366,51 @@ func (ds *DataStore) ReadBackingStore() error {
...
+		for cidr, _ := range eni.AvailableIPv4Cidrs {
+			eniCidrs[cidr] = eni
 		}
 	}
 
 	for _, allocation := range data.Allocations {
-		eni := eniIPs[allocation.IPv4]
+		ipv4Addr := net.IPNet{IP: net.ParseIP(allocation.IPv4), Mask: net.IPv4Mask(255, 255, 255, 255)}
+		eni := eniCidrs[ipv4Addr.String()]
 		if eni == nil {
```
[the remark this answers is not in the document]
jayanthvn: Makes sense, will do this.

pkg/ipamd/datastore/data_store.go
```diff
@@ -454,196 +465,153 @@ func (ds *DataStore) AddENI(eniID string, deviceNumber int, isPrimary, isTrunk,
...
+		ds.log.Infof("Added ENI(%s)'s IP %s to datastore", eniID, ipv4)
+	} else {
+		_, numIPsPerPrefix, _ := GetPrefixDelegationDefaults()
+		ds.total = ds.total + numIPsPerPrefix
+		ds.allocatedPrefix++
+		// Prometheus gauge
+		totalIPs.Set(float64(ds.total))
+		curENI.AvailableIPv4Cidrs[strIPv4Cidr] = &AssignedIPv4Addresses{Cidr: ipv4Cidr, IPv4Addresses: make(map[string]*AddressInfo), IsPrefix: isPrefix, UsedIPs: 0}
+		ds.log.Infof("Added ENI(%s)'s IPv4 Prefix %s to datastore", eniID, strIPv4Cidr)
+	}
```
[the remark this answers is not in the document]
jayanthvn: Nice :)

pkg/ipamd/datastore/data_store.go
```diff
@@ -454,196 +465,153 @@ func (ds *DataStore) AddENI(eniID string, deviceNumber int, isPrimary, isTrunk,
...
 	return nil
 }
 
 // AddPrefixIPv4AddressToStore add an IP of an ENI to data store
-func (ds *DataStore) AddPrefixIPv4AddressToStore(eniID string, ipv4 string) error {
+func (ds *DataStore) AddPrefixIPv4AddressToStore(eniID string, ipv4 string, cidr net.IPNet) error {
```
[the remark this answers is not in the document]
jayanthvn: Yeah, should be able to remove it.

pkg/ipamd/datastore/data_store.go
```diff
@@ -663,51 +631,52 @@ func (ds *DataStore) AssignPodIPv4Address(ipamKey IPAMKey) (ipv4address string,
...
-				if !addr.Assigned() && !addr.inCoolingPeriod() {
-					ds.assignPodIPv4AddressUnsafe(ipamKey, eni, addr)
-					if err := ds.writeBackingStoreUnsafe(); err != nil {
-						ds.log.Warnf("Failed to update backing store: %v", err)
-						// Important! Unwind assignment
-						ds.unassignPodIPv4AddressUnsafe(addr)
-						return "", -1, err
+			for _, assignedaddr := range eni.AvailableIPv4Cidrs {
+				ds.log.Debugf("DUMP Assigned addr %v", assignedaddr)
+				if !assignedaddr.IsPrefix {
```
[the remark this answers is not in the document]
jayanthvn: Yes will do that.

pkg/ipamd/datastore/data_store.go
```diff
@@ -929,16 +900,16 @@ func (ds *DataStore) RemoveUnusedENIFromStore(warmIPTarget, minimumIPTarget, war
 
 	removableENI := deletableENI.ID
 	if !ds.eniPool[removableENI].IsPDEnabled {
```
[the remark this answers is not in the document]
jayanthvn: I will try to unify, mainly the 2 cases are to fix the counts.

pkg/ipamd/datastore/data_store.go
```diff
@@ -980,15 +953,15 @@ func (ds *DataStore) RemoveENIFromDataStore(eniID string, force bool) error {
 	}
 
 	if !eni.IsPDEnabled {
```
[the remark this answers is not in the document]
jayanthvn: Yeah, will unify it once I make the pd enabled global at datastore level. This was added to fix up correct counts based on the feature.

pkg/ipamd/datastore/data_store.go
```diff
@@ -1114,13 +1093,12 @@ func (ds *DataStore) FreeablePrefixes(eniID string) []string {
 	}
 	ds.log.Debugf("In freeable prefix for eni %s", eniID)
 
-	freeable := make([]string, 0, len(eni.IPv4Prefixes))
-	for _, prefix := range eni.IPv4Prefixes {
-		if prefix.UsedIPs == 0 {
-			freeable = append(freeable, prefix.Prefix.String())
+	freeable := make([]string, 0, len(eni.AvailableIPv4Cidrs))
+	for _, assignedaddr := range eni.AvailableIPv4Cidrs {
+		if assignedaddr.IsPrefix && assignedaddr.UsedIPs == 0 {
```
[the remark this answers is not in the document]
jayanthvn: Yes, will add a function and replace UsedIPs, that will remove this check too.

pkg/ipamd/datastore/data_store.go
```diff
@@ -1155,28 +1136,26 @@ func (ds *DataStore) GetENIs() int {
 	return len(ds.eniPool)
 }
 
-// GetENIIPs returns the known (allocated & unallocated) ENI IPs.
-func (ds *DataStore) GetENIIPs(eniID string) ([]string, error) {
+// GetENICIDRs returns the known (allocated & unallocated) ENI IPs.
```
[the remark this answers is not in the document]
jayanthvn: Yes will do that.

pkg/ipamd/datastore/data_store.go
```diff
@@ -1214,10 +1195,103 @@ func (ds *DataStore) GetFreePrefixes() int {
 
 func (ds *DataStore) CleanupDataStore(eniID string, ipv4 string, force bool, enableIpv4PrefixDelegation bool) error {
```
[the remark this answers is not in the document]
jayanthvn: sure makes sense.

pkg/ipamd/datastore/data_store.go
```diff
@@ -1214,10 +1195,103 @@ func (ds *DataStore) GetFreePrefixes() int {
...
+
+	if cachedIP != "" {
+		return cachedIP, nil
+	}
+
+	//If not in cooldown then generate next IP
+	ip, ipnet, err := net.ParseCIDR(availableCidr.Cidr.String())
+	if err != nil {
+		return "", err
+	}
```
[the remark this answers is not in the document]
jayanthvn: Ahh right :(

pkg/ipamd/ipamd.go
```diff
@@ -876,41 +877,44 @@ func (c *IPAMContext) setupENI(eni string, eniMetadata awsutils.ENIMetadata, isT
...
+	c.addENIaddressesToDataStore(eniMetadata.IPv4Addresses, eniMetadata.IPv4Prefixes, eni)
 
 	return nil
 }
 
-// return primary ip address on the interface
-func (c *IPAMContext) addENIaddressesToDataStore(ec2Addrs []*ec2.NetworkInterfacePrivateIpAddress, eni string) {
-	for _, ec2Addr := range ec2Addrs {
-		if aws.BoolValue(ec2Addr.Primary) {
+func (c *IPAMContext) addENIaddressesToDataStore(ec2PrivateIpAddrs []*ec2.NetworkInterfacePrivateIpAddress, ec2PrefixAddrs []*ec2.Ipv4PrefixSpecification, eni string) {
```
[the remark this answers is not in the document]
jayanthvn: Yeah previously I had two functions `addENIprefixesToDataStore` and `addENIaddressesToDataStore` but due to last code review comments I merged it. Yeah will see if the caller can loop and call `AddIPv4CidrToStor`

pkg/ipamd/datastore/data_store.go
```diff
@@ -1214,10 +1195,103 @@ func (ds *DataStore) GetFreePrefixes() int {
...
+	for j := len(ip) - 1; j >= 0; j-- {
+		ip[j]++
+		if ip[j] > 0 {
+			break
+		}
+	}
+}
+
+//Given a IPv4 address, this will compute the associated prefix
+func getPrefixFromIPv4Addr(IPaddr string) net.IPNet {
```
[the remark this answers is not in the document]
jayanthvn: Yeah will re-evaluate.

pkg/ipamd/datastore/data_store.go
jayanthvn: Yeah makes sense :)

pkg/ipamd/ipamd.go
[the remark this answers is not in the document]
jayanthvn: I made this into 2 functions as before since we are also using it in setup ENI function.

jayanthvn:
@anguslees and @achevuru - I have updated the PR based on the review comments.

pkg/ipamd/datastore/data_store.go
```diff
@@ -145,38 +148,57 @@ type ENI struct {
...
-	IsPDEnabled bool
 	// DeviceNumber is the device number of ENI (0 means the primary ENI)
 	DeviceNumber int
 	// IPv4Addresses shows whether each address is assigned, the key is IP address, which must
 	// be in dot-decimal notation with no leading zeros and no whitespace(eg: "10.1.0.253")
-	IPv4Addresses map[string]*AddressInfo
-	// Key is the prefix and value is prefix, used IPs under the prefix
-	IPv4Prefixes map[string]*ENIPrefix
+	// Key is the IP address
+	AvailableIPv4Cidrs map[string]*AssignedIPv4Addresses
```
M00nF1sh, line 156: it's not clear to me what Key is reading this comment for a CIDR.

pkg/ipamd/datastore/data_store.go
```diff
@@ -145,38 +148,57 @@ type ENI struct {
...
 }
 
 // AddressInfo contains information about an IP, Exported fields will be marshaled for introspection.
 type AddressInfo struct {
 	IPAMKey        IPAMKey
 	Address        string
 	UnassignedTime time.Time
-	Prefix         net.IP
+	Prefix         net.IPNet
```
M00nF1sh, line 164: should this be called Cidr for consistency?

pkg/ipamd/datastore/data_store.go
jayanthvn: I can update the comment, Key will be "192.0.2.1/24"

pkg/ipamd/datastore/data_store.go
jayanthvn: Actually, I will delete this, we are not using it anymore.

pkg/ipamd/datastore/data_store.go
M00nF1sh: is this Key every used? if not it's better to be []*AssignedIPv4Addresses

pkg/ipamd/datastore/data_store.go
```diff
@@ -145,38 +148,56 @@ type ENI struct {
...
 	Address        string
 	UnassignedTime time.Time
-	Prefix         net.IP
 }
 
-func (e *ENI) findAddressForSandbox(ipamKey IPAMKey) *AddressInfo {
-	for _, addr := range e.IPv4Addresses {
-		if addr.IPAMKey == ipamKey {
-			return addr
+type AssignedIPv4Addresses struct {
```
M00nF1sh, line 166: what "Assigned" means here, Assigned to ENI or Assigned to Pod?
seems here it's means "Assigned" to ENI, but in the `AssignedIPv4Addresses` function, it means assigned to pod.

maybe pick a better name 😄

pkg/ipamd/datastore/data_store.go
```diff
@@ -207,13 +228,13 @@ func (p *ENIPool) AssignedIPv4Addresses() int {
 }
 
 // FindAddressForSandbox returns ENI and AddressInfo or (nil, nil) if not found
-func (p *ENIPool) FindAddressForSandbox(ipamKey IPAMKey) (*ENI, *AddressInfo) {
+func (p *ENIPool) FindAddressForSandbox(ipamKey IPAMKey) (*ENI, *AssignedIPv4Addresses, *AddressInfo) {
```
M00nF1sh, line 231: nit: map is passed by value, i don't think we need pointer receiver here.

pkg/ipamd/datastore/data_store.go
jayanthvn: Yes mainly in delete of the prefix while IPs are not used or during reducing the datastore, we can directly index rather than walking another loop.

pkg/ipamd/datastore/data_store.go
jayanthvn: Yeah, I can rename maybe to AllocatedIPv4Addresses?

pkg/ipamd/datastore/data_store.go
```diff
@@ -349,59 +372,36 @@ func (ds *DataStore) ReadBackingStore() error {
...
-				return err
-			}
-
-			err = ds.AddPrefixIPv4AddressToStore(eni.ID, allocation.IPv4)
-			if err != nil {
-				ds.log.Errorf("Failed to ADD PD IP %s on ENI %s", allocation.IPv4, eni.ID)
-				return err
+		ipv4Addr := net.ParseIP(allocation.IPv4)
+		found := false
+	eniloop:
```
M00nF1sh, line 386: it feels to me that we can make this more maintainable by abstract this into a function.

pkg/ipamd/datastore/data_store.go
```diff
@@ -349,59 +372,36 @@ func (ds *DataStore) ReadBackingStore() error {
 	ds.lock.Lock()
 	defer ds.lock.Unlock()
 
-	eniIPs := make(ENIPool)
-	eniPrefixes := make(ENIPool)
-
+	eniCidrs := make(ENIPool)
```
M00nF1sh, line 375: seems this eniCidrs is never used. If it's indeed like that, can we change AvailableIPv4Cidrs from map to slice.

jayanthvn: Yeah :)

pkg/ipamd/datastore/data_store.go
anguslees: I wondered this too, so I'm glad you raised it :)  They're not necessarily actually "assigned" or "allocated" :sweat_smile: and even in the `*ENI` struct we call them `AvailableIPv4Cidrs`

Maybe "AvailableIPv4Addresses" or "AttachedIPv4Addresses" - or skip the subjective term and just call it `CidrInfo` ?

The idea is that we have a 3-level nested structure that:
1. lists all the known ENIs (ENIPool - a collection of `*ENI`), that then includes a
2. list of all the current 'available' (or 'attached'?) known CIDRs (collection of ^this type), that tthen includes a
3. list of all the allocated (or in some version of cool-down) specific IPs.  This last one has to be sparse (we don't want to list _all_ the unallocated IPs) because an IPv6 CIDR is _huge_.

jayanthvn: During implementation I was thinking in terms of /32's assigned to pods within the Available Cidrs hence termed it as `AssignedIPv4Addresses` within `AvailableIPv4Cidrs`. I think we can make it `CidrInfo` and can reuse it for IPv6. 

Yes regarding [3], it is sparse and it includes just IPs assigned to pods and cool down IPs.

jayanthvn: Have updated the PR :)

pkg/ipamd/datastore/data_store.go
jayanthvn: There are quite a places even in the legacy code where we are passing the pointer (`AddressInfo*`). It would need quite some code changes so will discuss with you and take it up in a followup PR.

anguslees (approved):

pkg/ipamd/datastore/data_store.go
```diff
@@ -443,207 +445,113 @@ func (ds *DataStore) writeBackingStoreUnsafe() error {
...
 
-	if ipPrefix.UsedIPs != 0 {
-		if !force {
-			return errors.New(IPInUseError)
-		}
-		ds.log.Warnf("Force deleting assigned Prefix %s on eni %s", ipv4, eniID)
-		forceRemovedIPs.Inc()
-		//Cleanup datastore /32 ips
-		prefixDB := ipPrefix.AllocatedIPs
+	deletableCidr := curENI.AvailableIPv4Cidrs[strIPv4Cidr]
```
M00nF1sh, line 523: why not `deletableCidr, ok := curENI.AvailableIPv4Cidrs[strIPv4Cidr]`

pkg/ipamd/datastore/data_store.go
```diff
@@ -443,207 +445,113 @@ func (ds *DataStore) writeBackingStoreUnsafe() error {
...
 			// Continuing because 'force'
 		}
 	}
-
-	_, numIPsPerPrefix, _ := GetPrefixDelegationDefaults()
-	ds.total = ds.total - numIPsPerPrefix
-	ds.allocatedPrefix--
-	// Prometheus gauge
+	ones, bits := cidr.Mask.Size()
+	ds.total -= 1 << (bits - ones)
```
M00nF1sh, line 546: since this logic(`1 << (bits - ones)`) happens multiple places, it make sense to be a method on cidr.

pkg/ipamd/datastore/data_store.go
```diff
@@ -443,207 +445,113 @@ func (ds *DataStore) writeBackingStoreUnsafe() error {
...
+	if deletableCidr.IsPrefix {
+		ds.allocatedPrefix--
+	}
 	totalIPs.Set(float64(ds.total))
+	delete(deletableCidr.IPv4Addresses, ipv4)
```
M00nF1sh, line 551: why this is needed.

pkg/ipamd/datastore/data_store.go
```diff
@@ -655,67 +563,59 @@ func (ds *DataStore) AssignPodIPv4Address(ipamKey IPAMKey) (ipv4address string,
...
-					}
-
-					return addr.Address, eni.DeviceNumber, nil
+		for _, availableCidr := range eni.AvailableIPv4Cidrs {
+			var addr *AddressInfo
+			var strPrivateIPv4 string
+			var err error
+
+			if ds.isPDEnabled && availableCidr.IsPrefix {
+				strPrivateIPv4, err = ds.getFreeIPv4AddrfromCidr(availableCidr)
```
M00nF1sh, line 578: why for cidr, we use `getFreeIPv4AddrfromCidr(availableCidr)`
but for sip, we didn't use it. 
the place where we check for assigned/cooldown differs between this two code path (but ideally it can be treated same way)

pkg/ipamd/datastore/data_store.go
```diff
@@ -655,67 +563,59 @@ func (ds *DataStore) AssignPodIPv4Address(ipamKey IPAMKey) (ipv4address string,
...
-					}
-
-					//Adding here to eni DB, need to be removed while deleteing.
-					addr := eni.IPv4Addresses[strPrivateIPv4]
-					ds.assignPodIPv4AddressUnsafe(ipamKey, eni, addr)
-
-					addr.Prefix = prefix.Prefix.IP
+			addr = availableCidr.IPv4Addresses[strPrivateIPv4]
+			if addr == nil {
+				// addr is nil when we are using a new IP from prefix or SIP pool
```
M00nF1sh, line 597: this is hacky, as it couples the implementation detail of low level data structure in high level business code.


How about a method in CIDR like, so it hides the memory handling behind CIDR structure.
```
func (i *CIDR) tryAllocateFreeIPAddress() (*AddressInfo, bool) {
     if i.IsPrefix {
        return i.tryAllocateFreeIPAddressForPrefix()
     }
    tryAllocateFreeIPAddressForSIP()
}

func (i *CIDR) tryAllocateFreeIPAddressForPrefix() (*AddressInfo, bool) {
     ...
     do the allocation
}

func (i *CIDR) tryAllocateFreeIPAddressForSIP() (*AddressInfo, bool) {
    ....
   do the allocation
}
```

pkg/ipamd/datastore/data_store.go
```diff
@@ -655,67 +563,59 @@ func (ds *DataStore) AssignPodIPv4Address(ipamKey IPAMKey) (ipv4address string,
...
+			if ds.isPDEnabled && availableCidr.IsPrefix {
+				strPrivateIPv4, err = ds.getFreeIPv4AddrfromCidr(availableCidr)
+				if err != nil {
+					ds.log.Debugf("Unable to get IP address from prefix: %v", err)
+					//Check in next CIDR
+					continue
+				}
+				ds.log.Debugf("New IP from PD pool- %s", strPrivateIPv4)
+				if availableCidr.IPv4Addresses == nil {
+					availableCidr.IPv4Addresses = make(map[string]*AddressInfo)
```
M00nF1sh, line 586: nit: when will this IPv4Addresses ever be nil? seem it never happens from the code i reviewed.

I'm ok to have this safety check, but let's add some comment and debug statement here.
also, this is a sign that we are not encapsulation memory handling  logic good enough.

M00nF1sh (commented):
overall, we should improve the encapsulation here. 
ideally, the datastore shouldn't touch the internal members of CIDRs directly(which hides complicity for IPaddress/memory management behind cidr). also, we might need to consider when to use pointers vs when to use structures directly. (e.g. the addressInfo)

I'm open to merge this RP and improve in separate PRs given it's not release candidate and  there are a lot legacy code involved.

pkg/ipamd/datastore/data_store.go
jayanthvn: Discussed with Yang, there is lot of legacy code which needs this clean up, so will take up in future PRs.

pkg/ipamd/datastore/data_store.go
jayanthvn: Will address this with the next PR for warm ip target support.

pkg/ipamd/datastore/data_store.go
jayanthvn: Yeah, will fix this.

pkg/ipamd/datastore/data_store.go
jayanthvn: Will address this in the next PR.

pkg/ipamd/datastore/data_store.go
jayanthvn: Yes, this is not needed since we free up the CIDR block in the next line.

pkg/ipamd/datastore/data_store.go
jayanthvn: With PD, we try to get a free unused IP/32 from /28 hence I abstracted that in a function but whereas for SIP we already have a /32. As you suggested(next comment), in the following PR, I will create a function under CidrInfo to allocate a free IP.

pkg/ipamd/datastore/data_store.go
jayanthvn: Will push this logic into tryAllicateFreeIPAddress*

pkg/ipamd/datastore/data_store.go
jayanthvn: Yes, it is just for safety check, will add a debug

M00nF1sh (approved):
/lgtm

jayanthvn:
I have merged the PR to unblock the rest of the development. Please feel free drop in comments and I will take it up in the future PRs.

Issue #1423, opened 2021-04-09
Refresh CIDRs should follow up with adding iptables else traffic will be lost 


nithu0115:
**What happened**: We are adding more secondary CIDRs to our VPC. We see AWS VPC CNI plugin adds correct IP rules/routes but it is missing iptables due to which traffic is being dropped. 

**Attach logs**
````
source-ip-app-67bb5fbf99-qszv6   1/1     Running       0          12m     192.168.63.221    ip-192-168-48-37.us-west-2.compute.internal    <none>           <none> === eth2 

root@source-ip-app-67bb5fbf99-qszv6:/# curl 198.168.54.184
curl: (7) Failed to connect to 198.168.54.184 port 80: Connection timed out

nginx-6799fc88d8-8crh8           1/1     Running       0          5m3s    198.168.54.184    ip-198-168-8-134.us-west-2.compute.internal    <none>           <none> === eth2 

# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
3: eth0@if17: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9001 qdisc noqueue state UP group default
    link/ether 16:54:a0:0a:94:4f brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 198.168.54.184/32 scope global eth0
       valid_lft forever preferred_lft forever
````
#

**What you expected to happen**: Traffic from Pod `source-ip-app-67bb5fbf99-qszv6` to source Pod `nginx-6799fc88d8-8crh8` should be routable.  

**How to reproduce it (as minimally and precisely as possible)**: 

1) Create EKS cluster in VPC with CIDR `192.168.0.0/16` with 2 worker nodes in `192.168.0.0/16` CIDR
2) Create some nginx Pods on the 2 worker nodes created above. Make sure, there are Pods on secondary network interfaces.
3) After some time add a secondary CIDR `100.10.0.0/16` and launch worker in `100.10.0.0/16` CIDR.
4) Scale up above nginx deployment and ensure they land on above created worker node and on secondary ENIs of the worker nodes.
[...]

PR #1475, opened 2021-05-26
Use symmetric return path for non-VPC traffic - alternate solution

### LINKED ISSUE #1392: Traffic from outside VPC does not reach pod

kishorj:
**What type of PR is this?**
bug

Fixes #1392, #1423

When external SNAT is disabled, this PR does the following
- setup iptables rules to mark egress traffic from pods that are destined to external network
- external traffic will be sent through the primary interface

This is an alternate fix. It doesn't require permissive rp filter configuration on the secondary ENIs, or a new connmark. 

N/A

- When external SNAT is enabled, the iptables conmark rules added to the nat table get cleaned up
- With External SNAT disabled
   - connmark rules get added to the nat table
   - pods are able to initiate connections to hosts outside the VPC, traffic is through the primary interface
   - in case of NLB-IP, verify traffic flows as expected for client IP preservation enabled/disabled, pods on primary/secondary interfaces
   - UDP traffic works fine
   - instance mode NLB works as expected - verified externalTrafficPolicy Cluster, Local, IP preservation enabled/disabled, pods on primary/secondary interfaces
   - in-cluster traffic works as expected
   - enabled security group per pod, verified pods can communicate within the cluster, pods can receive client IP preserved traffic from the NLB

- With External SNAT enabled
    - connmark and SNAT rules get removed from the nat table
    - instance mode NLB works fine with IP preserve enabled/disabled, pods on primary/secondary interfaces
    - IP mode NLB works as expected with IP preserve enabled/disabled, pods on primary/secondary interfaces
    - pod can communicate to hosts within or outside VPC
    - enabled security group per pod, verified traffic flows

- New integration tests pass after some modification 

**Automation added to e2e**:

No impact on upgrade/downgrade. On, downgrade conn mark rules in the nat table remain until node is rebooted, but there is no impact on the functionality.

```release-note
Fixes the issue with non-VPC traffic ingress from secondary ENIs when external SNAT feature is not used.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,7 +376,7 @@ Type: String
 
 Default: `eni`
 
-Specifies the veth prefix used to generate the host-side veth device name for the CNI. The prefix can be at most 4 characters long.
+Specifies the veth prefix used to generate the host-side veth device name for the CNI. The prefix can be at most 4 characters long. The prefixes `eth` or `vlan` or `lo` are reserved and cannot be specified. We recommend using exclusive prefix for the host side veth device names.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -376,7 +376,7 @@ Type: String
 
 Default: `eni`
 
-Specifies the veth prefix used to generate the host-side veth device name for the CNI. The prefix can be at most 4 characters long. The prefixes `eth` or `vlan` or `lo` are reserved and cannot be specified. We recommend using exclusive prefix for the host side veth device names.
+Specifies the veth prefix used to generate the host-side veth device name for the CNI. The prefix can be at most 4 characters long. The prefixes `eth`, `vlan` and `lo` are reserved by the CNI plugin and cannot be specified. We recommend using prefix name not shared by any other network interfaces on the worker node instance.
 
 ---
 
--- a/pkg/networkutils/network.go
+++ b/pkg/networkutils/network.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/retry"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/pkg/errors"
 	"golang.org/x/sys/unix"
@@ -303,7 +304,26 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDRs []string, primaryMAC string, pr
 	if err != nil {
 		return errors.Wrap(err, "host network setup: failed to create iptables")
 	}
+	iptableRules, err := n.buildIptablesSNATRules(vpcCIDRs, primaryAddr, primaryIntf, ipt)
+	if err != nil {
+		return err
+	}
+	if err := n.updateIptablesRules(iptableRules, ipt); err != nil {
+		return err
+	}
 
+	iptableRules, err = n.buildIptablesConnmarkRules(vpcCIDRs, ipt)
+	if err != nil {
+		return err
+	}
+	if err := n.updateIptablesRules(iptableRules, ipt); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (n *linuxNetwork) buildIptablesSNATRules(vpcCIDRs []string, primaryAddr *net.IP, primaryIntf string, ipt iptablesIface) ([]iptablesRule, error) {
 	type snatCIDR struct {
 		cidr        string
 		isExclusion bool
@@ -318,12 +338,6 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDRs []string, primaryMAC string, pr
 		allCIDRs = append(allCIDRs, snatCIDR{cidr: cidr, isExclusion: true})
 	}
 
-	// if excludeSNATCIDRs or vpcCIDRs have changed they need to be cleared
-	snatStaleRulesToCheck, err := listCurrentSNATRules(ipt)
-	if err != nil {
-		return errors.Wrapf(err, "host network setup: failed to get SNAT chain rules to clear")
-	}
-
 	log.Debugf("Total CIDRs to program - %d", len(allCIDRs))
 	// build IPTABLES chain for SNAT of non-VPC outbound traffic and excluded CIDRs
 	var chains []string
@@ -332,7 +346,7 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDRs []string, primaryMAC string, pr
 		log.Debugf("Setup Host Network: iptables -N %s -t nat", chain)
 		if err := ipt.NewChain("nat", chain); err != nil && !containChainExistErr(err) {
 			log.Errorf("ipt.NewChain error for chain [%s]: %v", chain, err)
-			return errors.Wrapf(err, "host network setup: failed to add chain")
+			return []iptablesRule{}, errors.Wrapf(err, "host network setup: failed to add chain")
 		}
 		chains = append(chains, chain)
 	}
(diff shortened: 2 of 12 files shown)


pkg/networkutils/network_test.go
```diff
@@ -224,25 +224,15 @@ func TestUpdateRuleListBySrc(t *testing.T) {
 			[]string{"10.10.0.0/16", "10.11.0.0/16"},
 			[]string{"10.12.0.0/16", "10.13.0.0/16"},
 			[]netlink.Rule{origRule},
-			make([]netlink.Rule, 4),
-			make([]*net.IPNet, 4),
-			[]int{origRule.Table, origRule.Table, origRule.Table, origRule.Table},
+			make([]netlink.Rule, 1),
```
anguslees, line 227: Nit: Some of these test parameters look like they're not useful anymore - and we can probably simplify these structures and the below code.

anguslees (approved):

pkg/networkutils/network_test.go
kishorj: ack. I will do it in a follow up commit.

jayanthvn (approved):

pkg/networkutils/network.go
```diff
@@ -451,10 +452,98 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDRs []string, primaryMAC string, pr
...
+
+	var iptableRules []iptablesRule
+	log.Debugf("Setup Host Network: iptables -t nat -A PREROUTING -i eni+ -m comment --comment \"AWS, outbound connections\" -m state --state NEW -j AWS-CONNMARK-CHAIN-0")
+	iptableRules = append(iptableRules, iptablesRule{
+		name:        "connmark rule for non-VPC outbound traffic",
+		shouldExist: !n.useExternalSNAT,
+		table:       "nat",
+		chain:       "PREROUTING",
+		rule: []string{
+			"-i", "eni+", "-m", "comment", "--comment", "AWS, outbound connections",
```
kishorj, line 485: If AWS_VPC_K8S_CNI_VETHPREFIX is configured to something else than eni, this doesn't work. Need to make it based on the veth prefix configuration.

kishorj: fixed in new commit.

README.md
```diff
@@ -376,7 +376,7 @@ Type: String
 
 Default: `eni`
 
-Specifies the veth prefix used to generate the host-side veth device name for the CNI. The prefix can be at most 4 characters long.
+Specifies the veth prefix used to generate the host-side veth device name for the CNI. The prefix can be at most 4 characters long. The prefixes `eth` or `vlan` or `lo` are reserved and cannot be specified. We recommend using exclusive prefix for the host side veth device names.
```
achevuru, line 379: Are they really reserved though? We do create an eth0/1 in container ns. If the idea is to encourage users to avoid specifying some common linux interface prefixes, maybe we can have a generic statement here suggesting to avoid them(?). Because we probably can include more prefixes beyond the ones specified here (tun, tap, br etc)....

scripts/entrypoint.sh
```diff
@@ -42,6 +42,18 @@ validate_env_var()
...
         exit 1     
     fi 
+
+    if [[ ${#AWS_VPC_K8S_CNI_VETHPREFIX} -gt 4 ]]; then
+        log_in_json error "AWS_VPC_K8S_CNI_VETHPREFIX cannot be longer than 4 characters"
+        exit 1
+    fi
+
+    case ${AWS_VPC_K8S_CNI_VETHPREFIX} in
+        eth|vlan|lo)
```
achevuru, line 52: Same comment as above

README.md
kishorj: These are not reserved by linux, these are reserved only by the CNI plugin.

achevuru: Might help to call that out specifically. Also, we basically don’t want any other interfaces on the instance to share the same prefix as host veth interfaces and I think we should specify that in the above description so that customers know the prefixes to avoid in addition to the above three we are validating.

kishorj: updated readme.

scripts/entrypoint.sh
kishorj: updated readme

achevuru (approved):

jayanthvn:
Lets wait for the integration tests to finish.

PR #1483, opened 2021-06-04
Support mixed instances with PD

jayanthvn:
**What type of PR is this?**
Enhancement

We need to support mixed instance types and on non-nitro need to fallback to secondary mode.

Fallback to secondary on non-nitro

Yes
**Automation added to e2e**:

No

No

No

No

```release-note

```

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -322,7 +322,7 @@ func New(k8sapiClient kubernetes.Interface, eniConfig *eniconfig.ENIConfigContro
 		return nil, err
 	}
 	if hypervisorType != "nitro" && c.enableIpv4PrefixDelegation {
-		log.Warnf("Prefix delegation is not supported on non-nitro instance %s hence falling back to secondary mode", c.awsClient.GetInstanceType())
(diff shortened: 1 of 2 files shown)


pkg/ipamd/ipamd.go
```diff
@@ -322,8 +322,8 @@ func New(k8sapiClient kubernetes.Interface, eniConfig *eniconfig.ENIConfigContro
 		return nil, err
 	}
 	if hypervisorType != "nitro" && c.enableIpv4PrefixDelegation {
-		log.Error("Prefix delegation is not supported non-nitro instance - " + c.awsClient.GetInstanceType())
-		return nil, err
+		log.Warnf("Prefix delegation is not supported on non-nitro instance %s hence falling back to secondary mode", c.awsClient.GetInstanceType())
```
achevuru, line 325: nit: secondary mode -> `secondary IP address mode` (or) `default mode`. I'm thinking `default mode` might  be better because I don't think any of our docs talk about secondary mode.

jayanthvn: Yeah, I am fine with either of the suggestions. Will update the PR.

achevuru (approved):

PR #1499, opened 2021-06-09
Fix CNI crashing when there is no available IP addresses.

M00nF1sh:
Currently when CNI is invoked to AddNetwork while there is no available IP address in IPAMD, CNI crashes. 
It happens due to two bug below:
1. when IPAMD returns "no ip available error", the err variable get overwritten to nil when get VPCCIDRs. Thus we are returning a success response with empty IPv4Address. When CNI tries to setup veth-pair with empty IPv4Address, it fails and we invokes DelNetwork, and triggers the second bug below
2. when DelNetwork returns err(due to the pod sandbox is not found, which is expected), access r.Success will cause nil-pointer exception and crash CNI

**What type of PR is this?**

```
Tested with 1.19 cluster:
{"level":"warn","ts":"2021-06-09T20:54:14.695Z","caller":"rpc/rpc.pb.go:501","msg":"Send AddNetworkReply: unable to assign IPv4 address for pod, err: assignPodIPv4AddressUnsafe: no available IP addresses"}
```
**Automation added to e2e**:

```release-note

```

### DIFF
--- a/cmd/routed-eni-cni-plugin/cni.go
+++ b/cmd/routed-eni-cni-plugin/cni.go
@@ -207,9 +207,7 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 		if delErr != nil {
 			log.Errorf("Error received from DelNetwork grpc call for container %s: %v",
 				args.ContainerID, delErr)
-		}
-
-		if !r.Success {
+		} else if !r.Success {
 			log.Errorf("Failed to release IP of container %s: %v",
 				args.ContainerID, delErr)
 		}
--- a/cmd/routed-eni-cni-plugin/cni.go
+++ b/cmd/routed-eni-cni-plugin/cni.go
@@ -208,8 +208,7 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 			log.Errorf("Error received from DelNetwork grpc call for container %s: %v",
 				args.ContainerID, delErr)
 		} else if !r.Success {
-			log.Errorf("Failed to release IP of container %s: %v",
-				args.ContainerID, delErr)
+			log.Errorf("Failed to release IP of container %s", args.ContainerID)
 		}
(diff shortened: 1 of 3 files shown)


cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -207,9 +207,7 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 		if delErr != nil {
 			log.Errorf("Error received from DelNetwork grpc call for container %s: %v",
 				args.ContainerID, delErr)
-		}
-
-		if !r.Success {
+		} else if !r.Success {
```
achevuru, line 210: nit: We will end up here only if delErr is nil right? So, don't see any value printing delErr in the below error message.

achevuru: Also, why not extend this to scenarios where add fails as well.

https://github.com/aws/amazon-vpc-cni-k8s/blob/master/cmd/routed-eni-cni-plugin/cni.go#L164

M00nF1sh: @achevuru because when add fails, we always return within `err != nil `check.

M00nF1sh: > nit: We will end up here only if delErr is nil right? So, don't see any value printing delErr in the below error message.

make sense.

jayanthvn: I think later sometime we need to simplify this, like for instance make Add and Del functions return just one structure with the response and error, now there are two variables and anyone adding a piece of code should be aware of setting response when err is nil.

M00nF1sh: yeah, this whole piece needs to restructured and better handled.
like how to handle err vs r.success. (seems there is a duplicate here)

jayanthvn: yeah agreed.

jayanthvn (approved):
LGTM :)

Issue #1451, opened 2021-04-30
ENI allocates/frees in loop with custom networking and WARM_ENI_TARGET=0


jayanthvn:
**What happened**:

If custom networking is enabled and WARM_ENI_TARGET = 0, nodes which have no pods including coreDNS behind the secondary ENI will continuously free and add the secondary ENI.

**Attach logs**

```
cat /var/log/aws-routed-eni/ipamd.log | grep "Created a new"
{"level":"info","ts":"2021-04-30T03:48:00.292Z","caller":"awsutils/awsutils.go:674","msg":"Created a new ENI: eni-00403d554ade0d048"}
{"level":"info","ts":"2021-04-30T05:25:14.616Z","caller":"awsutils/awsutils.go:674","msg":"Created a new ENI: eni-07fc36da573f67a17"}
{"level":"info","ts":"2021-04-30T05:29:24.806Z","caller":"awsutils/awsutils.go:674","msg":"Created a new ENI: eni-0059aeefc485ad649"}
{"level":"info","ts":"2021-04-30T05:33:35.817Z","caller":"awsutils/awsutils.go:674","msg":"Created a new ENI: eni-00adbdaf675289a01"}
{"level":"info","ts":"2021-04-30T05:37:45.499Z","caller":"awsutils/awsutils.go:674","msg":"Created a new ENI: eni-006765cc28a46d9d2"}
{"level":"info","ts":"2021-04-30T05:41:55.402Z","caller":"awsutils/awsutils.go:674","msg":"Created a new ENI: eni-0178184612cc5184d"}
[ec2-user@ip-192-168-23-109 ~]$ cat /var/log/aws-routed-eni/ipamd.log | grep "Successfully deleted"
{"level":"info","ts":"2021-04-30T05:25:09.195Z","caller":"retry/retry.go:70","msg":"Successfully deleted ENI: eni-00403d554ade0d048"}
{"level":"info","ts":"2021-04-30T05:29:19.357Z","caller":"retry/retry.go:70","msg":"Successfully deleted ENI: eni-07fc36da573f67a17"}
{"level":"info","ts":"2021-04-30T05:33:30.377Z","caller":"retry/retry.go:70","msg":"Successfully deleted ENI: eni-0059aeefc485ad649"}
{"level":"info","ts":"2021-04-30T05:37:40.043Z","caller":"retry/retry.go:70","msg":"Successfully deleted ENI: eni-00adbdaf675289a01"}
[...]

PR #1515, opened 2021-06-17
Default WARM_PREFIX_TARGET to 1

jayanthvn:
**What type of PR is this?**
enhancement

WARM_PREFIX_TARGET is implemented similar to WARM_ENI_TARGET. Today we do support WARM_ENI_TARGET to 0 but the problem with this [1] with custom networking we allocate secondary ENI since primary ENI is not used and if there are no pods then IPAMD will determine this as an extra ENI since the target is 0 and free the ENI. Later IPAMD will see "available is 0 and WARM_ENI_TARGET is 0" so allocates a new ENI. This happens continuously by the reconciler adding to more EC2 calls on the account. This is tracked with this issue - #1451 [2] Without custom networking, scaling will be very slow - Initially on nodeInit and ENI will allocated Max secondary IPs and once those IPs are used, the next ENI will be allocated after the next reconciler starts and enters here - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/ipamd.go#L911 and then few more seconds for IPs to be available in IMDS so until then pods will be stuck in container creating. Similar fix done in this PR should be done even for WARM_ENI_TARGET but that will break upgrades if someone is already using this so it would need reach out and documentation update.

Similar issue will happen with WARM_PREFIX TARGET, prefixes will be attached and removed in a loop by the reconciler and when more prefixes are needed we need to wait for the reconciler.

<Sorry @anguslees copying your explanation below since it will help our documentation :D :D >
[1]"I am IP constrained" (allocate/free to maintain a small warm pool -> WARM_IP_TARGET=$small; WARM_ENI/PREFIX=$small or 0; possibly choose SIP-mode)
[2]"I am AWS-API-calls constrained" (allocate/free in large batches -> WARM_ENI/PREFIX=$small or $big)
[3]"I want to burst as quickly as I can" (maintain a large warm pool -> WARM_ENI/PREFIX=$big)

[1] With prefix delegation, we won't support WARM_{PREFIX,IP}_TARGET = 0 and MINIMUM_IP_TARGET=0.
[2] default WARM_PREFIX_TARGET to 1.

If WARM_{PREFIX,IP}_TARGET = 0/unset and MINIMUM_IP_TARGET=0/unset, then aws-node will fail to start with an error.

WARM_PREFIX_TARGET = 0 but WARM_IP_TARGET or MINIMUM_IP_TARGET is non zero will work since WARM_IP_TARGET /MINIMUM_IP_TARGET overrides WARM_PREFIX_TARGET

One combination which this PR won't handle is WARM_PREFIX_TARGET = non-zero but WARM_IP_TARGET or MINIMUM_IP_TARGET is zero, since this is supported with secondary mode and this will be evaluated along with WARM_ENI_TARGET.

n/a

```
kubectl logs aws-node-ppwcj -n kube-system
{"level":"info","ts":"2021-06-17T19:43:57.967Z","caller":"entrypoint.sh","msg":"Validating env variables ..."}
{"level":"error","ts":"2021-06-17T19:43:57.968Z","caller":"entrypoint.sh","msg":"Setting WARM_PREFIX_TARGET = 0 is not supported while WARM_IP_TARGET/MINIMUM_IP_TARGET is not set. Please configure either one of the WARM_{PREFIX/IP}_TARGET or MINIMUM_IP_TARGET env variables"}
```
**Automation added to e2e**:

No

No

No

```release-note
WARM_{PREFIX,IP}_TARGET = 0/unset and MINIMUM_IP_TARGET=0/unset is not supported with PD.
```

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -128,7 +128,7 @@ const (
 
 	//envWarmPrefixTarget is used to keep a /28 prefix in warm pool.
 	envWarmPrefixTarget = "WARM_PREFIX_TARGET"
-	noWarmPrefixTarget  = 0
+	defaultWarmPrefixTarget  = 0
 )
 
 var log = logger.Get()
@@ -961,17 +961,17 @@ func getWarmPrefixTarget() int {
 	inputStr, found := os.LookupEnv(envWarmPrefixTarget)
 
 	if !found {
-		return noWarmPrefixTarget
+		return defaultWarmPrefixTarget
 	}
 
 	if input, err := strconv.Atoi(inputStr); err == nil {
 		if input < 0 {
-			return noWarmPrefixTarget
+			return defaultWarmPrefixTarget
 		}
 		log.Debugf("Using WARM_PREFIX_TARGET %v", input)
 		return input
 	}
-	return noWarmPrefixTarget
+	return defaultWarmPrefixTarget
 }
 
 func logPoolStats(total int, used int, maxAddrsPerENI int, Ipv4PrefixDelegation bool) {
@@ -1039,19 +1039,8 @@ func (c *IPAMContext) shouldRemoveExtraPrefixes() int {
 	available := total - used
 
 	freePrefixes := c.dataStore.GetFreePrefixes()
-        _, numIPsPerPrefix, _ := datastore.GetPrefixDelegationDefaults()
-
 	over = max(freePrefixes-c.warmPrefixTarget, 0)
 
-	//if warm prefix target is 0, we should not make available to reach 0 since reconciler will enter into a churn
-	// of allocating and freeing prefixes similar to issue. This will add continous EC2 calls on nodes which have
-	// no prefixes used. Hence better to handle it. 
-	// [ENI allocates/frees in loop with custom networking and WARM_ENI_TARGET=0 #1451]
-	// But Will sync up internally with the team and see if need to even support WARM_ENI_TARGET = 0 and WARM_PREFIX_TARGET = 0
-	if over > 0 && ((available - (over * numIPsPerPrefix)) <= 0) && c.warmPrefixTarget == 0 {
-		log.Debugf("Warm prefix target is 0 and we will go below available hence capping to have atleast 1 free prefix")
-		over -= 1
-	}
 	logPoolStats(total, used, c.maxIPsPerENI, c.enableIpv4PrefixDelegation)
 	log.Debugf("shouldRemoveExtraPrefixes available %d over %d warm_prefix_target %d", available, over, c.warmPrefixTarget)
 	
@@ -1807,7 +1796,7 @@ func (c *IPAMContext) isDatastorePoolTooHigh() bool {
 }
 
 func (c *IPAMContext) warmPrefixTargetDefined() bool {
-	return c.warmPrefixTarget >= noWarmPrefixTarget && c.enableIpv4PrefixDelegation
+	return c.warmPrefixTarget >= defaultWarmPrefixTarget && c.enableIpv4PrefixDelegation
 }
 
 //DeallocCidrs frees IPs and Prefixes from EC2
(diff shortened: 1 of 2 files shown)


pkg/ipamd/ipamd.go
```diff
@@ -1807,7 +1796,7 @@ func (c *IPAMContext) isDatastorePoolTooHigh() bool {
 }
 
 func (c *IPAMContext) warmPrefixTargetDefined() bool {
-	return c.warmPrefixTarget >= noWarmPrefixTarget && c.enableIpv4PrefixDelegation
+	return c.warmPrefixTarget >= defaultWarmPrefixTarget && c.enableIpv4PrefixDelegation
```
M00nF1sh, line 1799: will we ever change `defaultWarmPrefixTarget` to a value than `0`?
If we change it to be same thing like 2, then this read as `c.warmPrefixTarget >= 2`. I feel this should be something like `c.warmPrefixTarget >= 0`

scripts/entrypoint.sh
```diff
@@ -35,13 +35,54 @@ log_in_json()
...
+
 validate_env_var()
 {
     log_in_json info "Validating env variables ..."
     if [[ "${AWS_VPC_K8S_PLUGIN_LOG_FILE,,}" == "stdout" ]]; then
         log_in_json error "AWS_VPC_K8S_PLUGIN_LOG_FILE cannot be set to stdout"
         exit 1     
     fi 
+
+    if is_prefix_delegation_enabled && unsupported_prefix_target_conf && warm_ip_target_not_configured && min_ip_target_not_configured ; then
```
M00nF1sh, line 82: can we merge this into a single function instead?
like validate_warm_prefix_conf

we added 4 un-necessary functions to support purpose, while these checks can be simply lined in a single function, and reads better.

jayanthvn: Yes initially I had it one line but mainly for readability I did this.

M00nF1sh (commented):
High level comment:
1. we should fix the reconcile logic. It's clearly a bug if it's allocate/deallocate ENI in loops under specific condition. From the code i read, it seems the purpose of `poolTooLow := available < c.maxIPsPerENI*c.warmENITarget || (c.warmENITarget == 0 && available == 0)` is `when warmENITarget is 0, we only allocate one more ENI when available is 0`. However, the ENIDeallocation logic is not honer above, which should be fixed.
2. Ideally the fix can be as simple as normalize warmENITarget and warmIPTarget at startup, so that if warmENITarget==0, warmIPTarget==0, we normalize it to be as "warmIPTarget==1", and remove this hacky logic of `c.warmENITarget == 0 && available == 0` above.
3. Ideally warmPrefixTarget should be handled similar to warmENITarget as above. Setting it to 0 should give customer the same effect of warmIPTarget==1.

M00nF1sh:
I'm ok to have this restriction at beginning and relax it later with a better fix.

pkg/ipamd/ipamd.go
jayanthvn: `defaultWarmPrefixTarget ` is 1 in this PR and also setting just WARM_PREFIX_TARGET = 0 won't be supported. If someone sets `WARM_IP_TARGET = num` and `WARM_PREFIX_TARGET = 0 or any num` then WARM_IP_TARGET will take precedence.

jayanthvn:
> High level comment:
> 
> 1. we should fix the reconcile logic. It's clearly a bug if it's allocate/deallocate ENI in loops under specific condition. From the code i read, it seems the purpose of `poolTooLow := available < c.maxIPsPerENI*c.warmENITarget || (c.warmENITarget == 0 && available == 0)` is `when warmENITarget is 0, we only allocate one more ENI when available is 0`. However, the ENIDeallocation logic is not honer above, which should be fixed.
> 2. Ideally the fix can be as simple as normalize warmENITarget and warmIPTarget at startup, so that if warmENITarget==0, warmIPTarget==0, we normalize it to be as "warmIPTarget==1", and remove this hacky logic of `c.warmENITarget == 0 && available == 0` above.
> 3. Ideally warmPrefixTarget should be handled similar to warmENITarget as above. Setting it to 0 should give customer the same effect of warmIPTarget==1.

Yes I agree, that's why would want to not support WARM_PREFIX_TARGET = 0 now. Later fix up WARM_{ENI,PREFIX}_TARGET which will need rewriting warm pool high and low logic.

M00nF1sh (approved):
/lgtm

PR #1516, opened 2021-06-18
Prefix Delegation: Merge from prefix delegation preview branch to master

jayanthvn:
**What type of PR is this?**
feature

This is an enhancement to achieve higher pod density per node.

Current implementation allocates secondary IPs per ENI which will be used for pod IPs. With the new VPC feature each secondary IP will be replaced with /28 prefix in the ENI's subnet. IPs will be allocated from this subnet.

N/A

Yes

**Automation added to e2e**:

Yes

New feature so downgrades won't be supported.

Yes

ENABLE_PREFIX_DELEGATION to enable the feature
WARM_PREFIX_TARGET to allocate /28 prefixes and store in warm pool.

Yes, scale per node will increase and max pods per node needs to be updated.

```release-note
Prefix delegation feature to support higher pod density per node.
```


jayanthvn:
locally UTs are running fine and even build is good - 

```
PASS
ok  	github.com/aws/amazon-vpc-cni-k8s/pkg/utils/retry	0.002s	coverage: 100.0% of statements
?   	github.com/aws/amazon-vpc-cni-k8s/pkg/utils/ttime	[no test files]
?   	github.com/aws/amazon-vpc-cni-k8s/pkg/utils/ttime/mocks	[no test files]

 % [amazon-vpc-cni-k8s]
dev-dsk-varavaj-2b-72f02457 % make build-linux
go build -mod=mod -buildmode=pie -ldflags '-s -w -X main.version=v1.6.4-rc1-171-gd0f990ea -X pkg/awsutils/awssession.version=v1.6.4-rc1-171-gd0f990ea' -o aws-k8s-agent     ./cmd/aws-k8s-agent
# github.com/aws/amazon-vpc-cni-k8s/cmd/aws-k8s-agent
loadinternal: cannot find runtime/cgo
go build -mod=mod -buildmode=pie -ldflags '-s -w -X main.version=v1.6.4-rc1-171-gd0f990ea -X pkg/awsutils/awssession.version=v1.6.4-rc1-171-gd0f990ea' -o aws-cni           ./cmd/routed-eni-cni-plugin
# github.com/aws/amazon-vpc-cni-k8s/cmd/routed-eni-cni-plugin
loadinternal: cannot find runtime/cgo
go build -mod=mod -buildmode=pie -ldflags '-s -w -X main.version=v1.6.4-rc1-171-gd0f990ea -X pkg/awsutils/awssession.version=v1.6.4-rc1-171-gd0f990ea' -o grpc-health-probe ./cmd/grpc-health-probe
# github.com/aws/amazon-vpc-cni-k8s/cmd/grpc-health-probe
loadinternal: cannot find runtime/cgo
```

pkg/awsutils/awsutils.go
```diff
@@ -1377,7 +1371,7 @@ func (cache *EC2InstanceMetadataCache) waitForENIAndIPsAttached(eni string, want
 				// At least some are attached
 				eniMetadata = returnedENI
 				// ipsToAllocate will be at most 1 less then the IP limit for the ENI because of the primary IP
-				if (eniIPCount > wantedSecondaryIPsOrPrefixes && !cache.useIPv4PrefixDelegation) || (eniIPCount >= wantedSecondaryIPsOrPrefixes && cache.useIPv4PrefixDelegation) {
+				if (eniIPCount > wantedSecondaryIPsOrPrefixes && !cache.enableIpv4PrefixDelegation) || (eniIPCount >= wantedSecondaryIPsOrPrefixes && cache.enableIpv4PrefixDelegation) {
```
achevuru, line 1374: So the diff between PD case and secondary IP case ( > and >=) is due to the presence of Primary IP in secondary IP mode? But, we don't use the Primary IP for Pods anyways, isn't it? Can you elaborate a bit on it here?

pkg/ipamd/datastore/data_store.go
```diff
@@ -1238,3 +1238,36 @@ func (ds *DataStore) FindFreeableCidrs(eniID string) []CidrInfo {
 func DivCeil(x, y int) int {
 	return (x + y - 1) / y
 }
+
+// CheckFreeableENIexists will return true if there is an ENI which is unused.
+// Could have just called getDeletbleENI, this is just to optimize a bit.
+func (ds *DataStore) CheckFreeableENIexists() bool {
```
achevuru, line 1244: Ok, don't we already have an equivalent function? (i.e.,) getDeletableENI? I see the only additional checks in that function are for WARM IP/Prefix targets and we need them anyways right? Can you expand a bit on the need for this?

pkg/ipamd/ipamd.go
```diff
@@ -1058,6 +1056,10 @@ func (c *IPAMContext) shouldRemoveExtraENIs() bool {
 	if shouldRemoveExtra {
 		logPoolStats(total, used, c.maxIPsPerENI, c.enableIpv4PrefixDelegation)
 		log.Debugf("It might be possible to remove extra ENIs because available (%d) >= (ENI/Prefix target + 1 (%d) + 1) * addrsPerENI (%d)", available, warmTarget, c.maxIPsPerENI)
+	} else if c.enableIpv4PrefixDelegation {
+		// When prefix target count is reduced, datastorehigh would have deleted extra prefixes hence available will be less than prefix target
```
achevuru, line 1060: It should only delete extra prefixes that we no longer need to meet the reduced prefix target(?)..Why will this delete more prefixes than what is required to maintain WARM_PREFIX_TARGET?

pkg/ipamd/rpc_handler.go
```diff
@@ -204,11 +204,11 @@ func (s *server) DelNetwork(ctx context.Context, in *rpc.DelNetworkRequest) (*rp
...
 	cidr := net.IPNet{IP: net.ParseIP(ip), Mask: net.IPv4Mask(255, 255, 255, 255)}
 	cidrStr := cidr.String()
-	if eni != nil && eni.AvailableIPv4Cidrs[cidrStr] != nil {
-		if s.ipamContext.enableIpv4PrefixDelegation && eni.AvailableIPv4Cidrs[cidrStr].IsPrefix == false {
+	if eni != nil {
+		if s.ipamContext.enableIpv4PrefixDelegation && eni.AvailableIPv4Cidrs[cidrStr] != nil && eni.AvailableIPv4Cidrs[cidrStr].IsPrefix == false {
 			log.Debugf("IP belongs to secondary pool with PD enabled so free IP from EC2")
 			s.ipamContext.tryUnassignIPFromENI(eni.ID)
-		} else if !s.ipamContext.enableIpv4PrefixDelegation && eni.AvailableIPv4Cidrs[cidrStr].IsPrefix == true {
+		} else if !s.ipamContext.enableIpv4PrefixDelegation && eni.AvailableIPv4Cidrs[cidrStr] == nil {
```
achevuru, line 211: Since this is a `DelNetwork` call, we know the IP/Prefix we just freed is attached to this particular ENI and so why should we check for internal data structure (i.e.,) `AvailableIPv4Cidrs[cidrStr]` to remove the IP/Prefix from the actual ENI? Shouldn't we always free/remove the IP/Prefix whenever there is a mismatch (sec ip in PD mode and prefix in sec ip mode)? Also, am assuming `eni.AvailableIPv4Cidrs[cidrStr] ` can never be nil for a just released IP? (mainly because we need it for cool down period check?)

pkg/awsutils/awsutils.go
jayanthvn: Yes the difference is having Primary IP in `returnedENI.IPv4Addresses` and wantedSecondaryIPsOrPrefixes will have maxIPsPerENI in secondary Mode which will not include primary IP. . But having said that because of the recent fix with warm IP targets, we need to check `eniIPCount >= wantedSecondaryIPsOrPrefixes` for both cases since eniIPCount has already removed primary IP and wantedSecondaryIPsOrPrefixes will either have max IPs or IPs based on warm targets.

pkg/ipamd/datastore/data_store.go
jayanthvn: Yes I agree but even before calling `tryFreeENI` which eventually calls `getDeletableENI` , I am just checking if the ENI has some pods. This is just a minor optimization because we call this function every reconcile time and `getDeletableENI` does lot more work to check if it is needed for warm targets.

pkg/ipamd/ipamd.go
jayanthvn: Oh I just need to reword the comment. So DatastoreHigh check would have deleted prefixes which are more than WARM_PREFIX_TARGET and these deleted prefixed might be spread across ENIs and those ENIs will be deletable now. So `shouldRemoveExtraENIs ` should return true now so that `tryFreeENI` frees it up.

pkg/ipamd/rpc_handler.go
jayanthvn: `cidrStr`  will be a /32 IP assigned to pod and that can be either from the secondary pool or prefix pool. With secondary IP mode `eni.AvailableIPv4Cidrs[cidrStr]` will not be `nil` but we need to free this IP only when we are in PD mode. Whereas `eni.AvailableIPv4Cidrs[cidrStr]` will be `nil` in PD mode since the key is `IP/28` and the prefix might be freeable in secondary mode. We can discuss this offline.

pkg/awsutils/awsutils.go
achevuru: Ok, then probably we can remove the PD check and combine them in to one check

pkg/ipamd/datastore/data_store.go
achevuru: Right but we can't anyways free the ENI without those checks. Are you saying this will be called much more frequently? Maybe we can replace the initial part of `getDeletableENI` with this function so that we can add any additional conditions in this function alone.

typo in comment: getDeletbleENI -> getDeletableENI

pkg/ipamd/rpc_handler.go
achevuru: Agree. I guess what I'm trying to say is that we should always free a secondary IP attached to an ENI in PD mode irrespective of whether `eni.AvailableIPv4Cidrs[cidrStr]` is nil or not and the same applies to a Prefix in secondary IP mode. Sure, let's take it offline.

pkg/ipamd/datastore/data_store.go
jayanthvn: Right this will be called much more frequently because of the reconciler. Yes that be done.

pkg/awsutils/awsutils.go
jayanthvn: Yeah will be updating the PR.

Makefile
```diff
@@ -144,7 +144,7 @@ docker-func-test: docker     ## Run the built CNI container image to use in func
 # Run unit tests
 unit-test: export AWS_VPC_K8S_CNI_LOG_FILE=stdout
 unit-test:    ## Run unit tests
-	go test -v -coverprofile=coverage.txt -covermode=atomic $(ALLPKGS)
+	 go test -v -mod=mod -coverprofile=coverage.txt -covermode=atomic ./pkg/...
```
M00nF1sh, line 147: nit: seems a unnecessary space added at beginning.

M00nF1sh, line 147: should -mod=mod replaced with `VENDOR_OVERRIDE_FLAG`?

jayanthvn: Yeah, will update it.

pkg/awsutils/awsutils.go
```diff
@@ -315,15 +332,34 @@ func (i instrumentedIMDS) GetMetadataWithContext(ctx context.Context, p string)
...
 
 	return result, nil
 }
 
+func (cache *EC2InstanceMetadataCache) getNetworkInterfacesWithContext(ctx context.Context, eniID string) (*ec2.DescribeNetworkInterfacesOutput, error) {
+	eniIds := []*string{aws.String(eniID)}
+	input := &ec2.DescribeNetworkInterfacesInput{NetworkInterfaceIds: eniIds}
+
+	start := time.Now()
+	result, err := cache.ec2SVC.DescribeNetworkInterfacesWithContext(context.Background(), input)
```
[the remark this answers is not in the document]
jayanthvn: Actually I didn't add this function, maybe because of PD branch and master branch merge. Let me remove this from PR.

pkg/awsutils/awsutils.go
```diff
@@ -315,15 +332,34 @@ func (i instrumentedIMDS) GetMetadataWithContext(ctx context.Context, p string)
...
+	awsAPILatency.WithLabelValues("DescribeNetworkInterfaces", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			if aerr.Code() == "InvalidNetworkInterfaceID.NotFound" {
+				return nil, ErrENINotFound
+			}
+		}
+		awsAPIErrInc("DescribeNetworkInterfaces", err)
+		log.Errorf("Failed to get ENI %s information from EC2 control plane %v", eniIds, err)
```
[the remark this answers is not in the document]
jayanthvn: Same as above.

pkg/ipamd/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
M00nF1sh, line 1: do we want to commit this one

pkg/ipamd/datastore/data_store.go
```diff
@@ -50,6 +51,9 @@ const (
 
 	// UnknownENIError is an error when caller tries to access an ENI which is unknown to datastore
 	UnknownENIError = "datastore: unknown ENI"
+
+	// IPv4 /32 prefix for host routes
+	ipv4DefaultPrefixSize = 32
```
M00nF1sh, line 56: nit: seems not used

pkg/ipamd/datastore/data_store.go
```diff
@@ -410,77 +454,99 @@ func (ds *DataStore) AddENI(eniID string, deviceNumber int, isPrimary, isTrunk,
...
+// AddIPv4AddressToStore add CIDR of an ENI to data store
+func (ds *DataStore) AddIPv4CidrToStore(eniID string, ipv4Cidr net.IPNet, isPrefix bool) error {
 	ds.lock.Lock()
 	defer ds.lock.Unlock()
 
+	strIPv4Cidr := ipv4Cidr.String()
+	ds.log.Infof("Adding %s to DS for %s", strIPv4Cidr, eniID)
 	curENI, ok := ds.eniPool[eniID]
 	if !ok {
+		ds.log.Infof("unkown ENI")
```
M00nF1sh, line 478: nit: unknown

pkg/ipamd/datastore/data_store.go
```diff
@@ -410,77 +454,99 @@ func (ds *DataStore) AddENI(eniID string, deviceNumber int, isPrimary, isTrunk,
...
-	_, ok = curENI.IPv4Addresses[ipv4]
+	_, ok = curENI.AvailableIPv4Cidrs[strIPv4Cidr]
 	if ok {
+		ds.log.Infof("IP already in DS")
 		return errors.New(IPAlreadyInStoreError)
 	}
 
-	ds.total++
-	// Prometheus gauge
+	curENI.AvailableIPv4Cidrs[strIPv4Cidr] = &CidrInfo{
```
M00nF1sh, line 488: nit: it's seems ideal to me to have a variable captures `&CidrInfo ...` first. so we can avoid the map lookup in`ds.total += curENI.AvailableIPv4Cidrs[strIPv4Cidr].Size()`

pkg/ipamd/datastore/data_store.go
```diff
@@ -546,9 +645,23 @@ func (ds *DataStore) unassignPodIPv4AddressUnsafe(addr *AddressInfo) {
...
+// GetStats returns total number of IP addresses, number of assigned IP addresses and total prefixes
+func (ds *DataStore) GetStats() (int, int, int) {
+	ds.lock.Lock()
+	defer ds.lock.Unlock()
+
+	totalIPs := 0
+	assignedIPs := 0
+	for _, eni := range ds.eniPool {
+		for _, cidr := range eni.AvailableIPv4Cidrs {
+			if (ds.isPDEnabled && cidr.IsPrefix) || (!ds.isPDEnabled && !cidr.IsPrefix) {
```
M00nF1sh, line 657: nit: this `(ds.isPDEnabled && cidr.IsPrefix) || (!ds.isPDEnabled && !cidr.IsPrefix)` happens multiple times.
It can be replaced with something like `IsCIDReligibleForAllocation`

pkg/ipamd/datastore/data_store.go
```diff
@@ -776,25 +948,29 @@ func (ds *DataStore) UnassignPodIPv4Address(ipamKey IPAMKey) (ip string, deviceN
...
 
 	ds.unassignPodIPv4AddressUnsafe(addr)
 	if err := ds.writeBackingStoreUnsafe(); err != nil {
 		// Unwind un-assignment
 		ds.assignPodIPv4AddressUnsafe(ipamKey, eni, addr)
-		return "", 0, err
+		return nil, "", 0, err
 	}
 	addr.UnassignedTime = time.Now()
+	if ds.isPDEnabled && availableCidr.IsPrefix == false {
```
M00nF1sh, line 966: nit: `availableCidr.IsPrefix == false` => `!availableCidr.IsPrefix`

pkg/ipamd/datastore/data_store.go
```diff
@@ -872,8 +1078,30 @@ func (ds *DataStore) GetENIs() int {
...
+			ipPool = append(ipPool, assignedAddr.Cidr.IP.String())
+		} else {
+			prefixPool = append(prefixPool, assignedAddr.Cidr.String())
+		}
+	}
+	return ipPool, prefixPool, nil
+}
+
+// GetENIPrefixes returns the known (allocated & unallocated) ENI Prefixed.
+func (ds *DataStore) GetENIPrefixes(eniID string) ([]string, error) {
```
M00nF1sh, line 1104: why we have this given we already have `GetENICIDRs`

pkg/ipamd/datastore/data_store.go
```diff
@@ -882,9 +1110,164 @@ func (ds *DataStore) GetENIIPs(eniID string) ([]string, error) {
...
 	}
 	return ipPool, nil
 }
+
+func (ds *DataStore) GetFreePrefixes() int {
+	ds.lock.Lock()
+	defer ds.lock.Unlock()
+
+	freePrefixes := 0
+	for _, other := range ds.eniPool {
```
M00nF1sh, line 1127: nit: other is confusing

pkg/ipamd/ipamd.go
```diff
@@ -583,83 +612,82 @@ func (c *IPAMContext) tryFreeENI() {
...
 // is enabled, deallocate extra IP addresses
-func (c *IPAMContext) tryUnassignIPsFromAll() {
-	if _, over, warmIPTargetDefined := c.ipTargetState(); warmIPTargetDefined && over > 0 {
+func (c *IPAMContext) tryUnassignCidrsFromAll() {
+
+	_, over, warmTargetDefined := c.datastoreTargetState()
+
+	//WARM IP targets not defined then check if WARM_PREFIX_TARGET is defined.
+	if !warmTargetDefined {
+		over = c.shouldRemoveExtraPrefixes()
```
M00nF1sh, line 623: nit: `shouldRemoveExtraPrefixes` -> `computeOverPrefixes`

pkg/ipamd/ipamd.go
```diff
@@ -783,7 +835,36 @@ func (c *IPAMContext) tryAssignIPs() (increasedPool bool, err error) {
...
-		c.addENIaddressesToDataStore(ec2Addrs, eni.ID)
+
+		c.addENIsecondaryIPsToDataStore(ec2Addrs, eni.ID)
+		return true, nil
+	}
+	return false, nil
+}
+
+func (c *IPAMContext) tryAssignPrefixes() (increasedPool bool, err error) {
+	toAllocate := c.getPrefixesNeeded()
```
M00nF1sh, line 846: nit: it feels to me we should have a common function that tells how many prefix are needed and how many prefixes are unneeded.

pkg/ipamd/ipamd.go
```diff
@@ -783,7 +835,36 @@ func (c *IPAMContext) tryAssignIPs() (increasedPool bool, err error) {
...
+	// Returns an ENI which has space for more prefixes to be attached, but this
+	// ENI might not suffice the WARM_IP_TARGET/WARM_PREFIX_TARGET
+	eni := c.dataStore.GetENINeedsIP(c.maxPrefixesPerENI, c.useCustomNetworking)
+	if eni != nil {
+		currentNumberOfAllocatedPrefixes := len(eni.AvailableIPv4Cidrs)
+		err = c.awsClient.AllocIPAddresses(eni.ID, min((c.maxPrefixesPerENI-currentNumberOfAllocatedPrefixes), toAllocate))
```
M00nF1sh, line 852: nit: we should hide two different API calls behind a single AllocIPAddresses

pkg/ipamd/ipamd.go
```diff
@@ -1186,6 +1374,79 @@ func (c *IPAMContext) verifyAndAddIPsToDatastore(eni string, attachedENIIPs []*e
 	return seenIPs
 }
 
+// verifyAndAddPrefixesToDatastore updates the datastore with the known Prefixes. Prefixes who are out of cooldown gets added
+// back to the datastore after being verified against EC2.
+func (c *IPAMContext) verifyAndAddPrefixesToDatastore(eni string, attachedENIIPs []*ec2.Ipv4PrefixSpecification, needEC2Reconcile bool) map[string]bool {
+	var ec2VerifiedAddresses []*ec2.Ipv4PrefixSpecification
+	seenIPs := make(map[string]bool)
+	for _, privateIPv4 := range attachedENIIPs {
```
M00nF1sh, line 1382: nit: it should be prefix/cidr for these variables..

pkg/ipamd/ipamd.go
```diff
@@ -1186,6 +1374,79 @@ func (c *IPAMContext) verifyAndAddIPsToDatastore(eni string, attachedENIIPs []*e
...
+			log.Debugf("found in cooldown")
+			if recentlyFreed {
+				log.Debugf("Reconcile skipping IP %s on ENI %s because it was recently unassigned from the ENI.", strPrivateIPv4, eni)
+				continue
+			} else {
+				if needEC2Reconcile {
+					// IMDS data might be stale
+					log.Debugf("This IP was recently freed, but is now out of cooldown. We need to verify with EC2 control plane.")
+					// Only call EC2 once for this ENI
+					if ec2VerifiedAddresses == nil {
```
M00nF1sh, line 1403: instead of check `ec2VerifiedAddresses == nil`, we should have a bool variable tracks whether GetIPv4PrefixesFromEC2 have been called or not.
otherwise, this logic's correctness relies on whether GetIPv4PrefixesFromEC2 returns empty slice or nil slice when no prefix found. we shouldn't have such dependencies in golang. (empty slice/nil slice should be treated semantically equal)

pkg/ipamd/ipamd.go
```diff
@@ -1186,6 +1374,79 @@ func (c *IPAMContext) verifyAndAddIPsToDatastore(eni string, attachedENIIPs []*e
...
+			continue
+		}
+		found, recentlyFreed := c.reconcileCooldownCache.RecentlyFreed(strPrivateIPv4)
+		if found {
+			log.Debugf("found in cooldown")
+			if recentlyFreed {
+				log.Debugf("Reconcile skipping IP %s on ENI %s because it was recently unassigned from the ENI.", strPrivateIPv4, eni)
+				continue
+			} else {
+				if needEC2Reconcile {
```
M00nF1sh, line 1399: nit: we should split the loop into two. 
1. first go though the attachedENIIPs to see filter out attachedENIIPs that is not RecentlyFreed
2. then if the filtered list contains items and needEC2Reconcile, we do the APICall
3. then we go through the filtered list and do the rest of the logic.

pkg/ipamd/ipamd.go
```diff
@@ -1186,6 +1374,79 @@ func (c *IPAMContext) verifyAndAddIPsToDatastore(eni string, attachedENIIPs []*e
...
+						continue
+					}
+				}
+				// The IP can be removed from the cooldown cache
+				// TODO: Here we could check if the IP is still used by a pod stuck in Terminating state. (Issue #1091)
+				c.reconcileCooldownCache.Remove(strPrivateIPv4)
+			}
+		}
+
+		err = c.dataStore.AddIPv4CidrToStore(eni, *ipv4CidrPtr, true)
```
M00nF1sh, line 1434: so no matter whether the CIDR is in really attached we added it to the store, is the logic correct?

M00nF1sh: i see, the continue breaks the for loop in 5 levels above..

pkg/ipamd/ipamd.go
```diff
@@ -1401,3 +1707,216 @@ func (c *IPAMContext) GetPod(podName, namespace string) (*corev1.Pod, error) {
...
+
+func (c *IPAMContext) tryUnassignPrefixesFromENIs() {
+	eniInfos := c.dataStore.GetENIInfos()
+	for eniID := range eniInfos.ENIs {
+		c.tryUnassignPrefixFromENI(eniID)
+	}
+}
+
+func (c *IPAMContext) tryUnassignPrefixFromENI(eniID string) {
+	FreeablePrefixes := c.dataStore.FreeablePrefixes(eniID)
```
M00nF1sh, line 1758: nit "freeablePrefixes"

pkg/awsutils/imds.go
```diff
@@ -35,23 +35,57 @@ type TypedIMDS struct {
...
-
-	return strings.Fields(data), nil
+	return strings.Fields(data), err
 }
 
 // GetAZ returns the Availability Zone in which the instance launched.
 func (imds TypedIMDS) GetAZ(ctx context.Context) (string, error) {
-	return imds.GetMetadataWithContext(ctx, "placement/availability-zone")
+	az, err := imds.GetMetadataWithContext(ctx, "placement/availability-zone")
+	if err != nil {
```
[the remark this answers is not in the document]
jayanthvn: Sounds good, I will add it for post GA work item.s

pkg/ipamd/datastore/data_store.go
```diff
@@ -582,9 +695,19 @@ func (ds *DataStore) isRequiredForWarmIPTarget(warmIPTarget int, eni *ENI) bool
...
-			otherWarmIPs += len(other.IPv4Addresses) - other.AssignedIPv4Addresses()
+			for _, otherPrefixes := range other.AvailableIPv4Cidrs {
+				if (ds.isPDEnabled && otherPrefixes.IsPrefix == true) || (!ds.isPDEnabled && otherPrefixes.IsPrefix == false) {
+					otherWarmIPs += otherPrefixes.Size() - otherPrefixes.AssignedIPv4AddressesInCidr()
+				}
+			}
 		}
 	}
+
+	if ds.isPDEnabled {
```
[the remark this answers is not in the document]
jayanthvn: As discussed on call, this section will normalize IPs in terms of prefixes.

pkg/awsutils/awsutils.go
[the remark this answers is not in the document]
jayanthvn: This is merge induced issue, for instance - 

```
// GetIPv4sFromEC2 calls EC2 and returns a list of all addresses on the ENI
func (cache *EC2InstanceMetadataCache) GetIPv4sFromEC2(eniID string) (addrList []*ec2.NetworkInterfacePrivateIpAddress, err error) {
	eniIds := make([]*string, 0)
	eniIds = append(eniIds, aws.String(eniID))
	input := &ec2.DescribeNetworkInterfacesInput{NetworkInterfaceIds: eniIds}
	start := time.Now()
	result, err := cache.ec2SVC.DescribeNetworkInterfacesWithContext(context.Background(), input)
	awsAPILatency.WithLabelValues("DescribeNetworkInterfaces", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
	if err != nil {
		if aerr, ok := err.(awserr.Error); ok {
			if aerr.Code() == "InvalidNetworkInterfaceID.NotFound" {
				return nil, ErrENINotFound
			}
		}
		awsAPIErrInc("DescribeNetworkInterfaces", err)
		log.Errorf("Failed to get ENI %s information from EC2 control plane %v", eniID, err)
		return nil, errors.Wrap(err, "failed to describe network interface")
	}
	// Shouldn't happen, but let's be safe
	if len(result.NetworkInterfaces) == 0 {
		return nil, ErrNoNetworkInterfaces
	}
	firstNI := result.NetworkInterfaces[0]
	return firstNI.PrivateIpAddresses, nil
}
```



```
PD branch has this -

// GetIPv4sFromEC2 calls EC2 and returns a list of all addresses on the ENI
func (cache *EC2InstanceMetadataCache) GetIPv4sFromEC2(eniID string) (addrList []*ec2.NetworkInterfacePrivateIpAddress, err error) {
	result, err := cache.getNetworkInterfacesWithContext(context.Background(), eniID)
	if err != nil {
		return nil, err
	}
	// Shouldn't happen, but let's be safe
	if len(result.NetworkInterfaces) == 0 {
		return nil, ErrNoNetworkInterfaces
	}
	firstNI := result.NetworkInterfaces[0]
	return firstNI.PrivateIpAddresses, nil
}
```
I will make it inline with master.

pkg/awsutils/awsutils.go
```diff
@@ -567,12 +607,29 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
 		}
 	}
 
+	var ec2ipv4Prefixes []*ec2.Ipv4PrefixSpecification
+	//Get prefix on primary ENI when custom networking is enabled is not needed. Everytime we
```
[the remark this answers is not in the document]
jayanthvn: Option 1, will never happen since with custom networking we recycle the nodes.

Option 2, I will update the comment.

pkg/ipamd/.vscode/settings.json
jayanthvn: yeah this is not needed. will remove it.

pkg/ipamd/datastore/data_store.go
jayanthvn: As discussed, will take this up post GA.

pkg/ipamd/datastore/data_store.go
jayanthvn: Yes this again has come from merge, I am not using this anymore.

pkg/ipamd/datastore/data_store.go
jayanthvn: This function is removed.

pkg/ipamd/ipamd.go
jayanthvn: Renamed it to 1computeExtraPrefixesOverWarmTarget`

pkg/ipamd/ipamd.go
jayanthvn: As discussed on call, will add it to post GA AI.

pkg/ipamd/ipamd.go
jayanthvn: Agreed, will fix it up post GA.

pkg/ipamd/ipamd.go
jayanthvn: Discussed with Yang. For now will keep it similar to secondary IPs and the right way will be to split the logic into two sections which will first fetch from EC2 once if needed and then compare with what we have from IMDS. Will update the comment.

pkg/ipamd/ipamd.go
jayanthvn: Thanks Yang, this makes sense. We should do this even for the secondary IP function.

pkg/ipamd/ipamd.go
jayanthvn: Yeah, the 2 loop comment should fix this too :D

achevuru (approved):

pkg/awsutils/awsutils.go
```diff
@@ -1232,9 +1315,21 @@ func (cache *EC2InstanceMetadataCache) AllocIPAddresses(eniID string, numIPs int
 	}
 
 	log.Infof("Trying to allocate %d IP addresses on ENI %s", needIPs, eniID)
-	input := &ec2.AssignPrivateIpAddressesInput{
-		NetworkInterfaceId:             aws.String(eniID),
-		SecondaryPrivateIpAddressCount: aws.Int64(int64(needIPs)),
+	log.Debugf("PD enabled - %d", cache.enableIpv4PrefixDelegation)
```
M00nF1sh, line 1318: will %d works?

jayanthvn: Right, I can make it %t

pkg/awsutils/awsutils.go
```diff
@@ -1309,13 +1421,13 @@ func (cache *EC2InstanceMetadataCache) waitForENIAndIPsAttached(eni string, want
...
+// DeallocIPAddresses frees IP address on an ENI
 func (cache *EC2InstanceMetadataCache) DeallocIPAddresses(eniID string, ips []string) error {
-	log.Infof("Trying to unassign the following IPs %s from ENI %s", ips, eniID)
-	var ipsInput []*string
-	for _, ip := range ips {
-		ipsInput = append(ipsInput, aws.String(ip))
+	if len(ips) == 0 {
+		return nil
 	}
+	log.Infof("Trying to unassign the following IPs %s from ENI %s", ips, eniID)
```
M00nF1sh, line 1429: maybe %v

pkg/awsutils/awsutils.go
```diff
@@ -1330,6 +1442,32 @@ func (cache *EC2InstanceMetadataCache) DeallocIPAddresses(eniID string, ips []st
...
+	prefixesInput := aws.StringSlice(prefixes)
+
+	input := &ec2.UnassignPrivateIpAddressesInput{
+		NetworkInterfaceId: aws.String(eniID),
+		Ipv4Prefixes:       prefixesInput,
+	}
+
+	start := time.Now()
+	_, err := cache.ec2SVC.UnassignPrivateIpAddressesWithContext(context.Background(), input)
+	awsAPILatency.WithLabelValues("UnassignPrivateIpAddresses", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
```
M00nF1sh, line 1464: UnassignPrivateIpAddresses-> DeallocPrefixAddresses

pkg/awsutils/awsutils.go
```diff
@@ -1330,6 +1442,32 @@ func (cache *EC2InstanceMetadataCache) DeallocIPAddresses(eniID string, ips []st
...
+	log.Debugf("Successfully freed IPs %v from ENI %s", ips, eniID)
+	return nil
+}
+
+// DeallocPrefixAddresses frees Prefixes on an ENI
+func (cache *EC2InstanceMetadataCache) DeallocPrefixAddresses(eniID string, prefixes []string) error {
+	if len(prefixes) == 0 {
+		return nil
+	}
+	log.Infof("Trying to unassign the following Prefixes %s from ENI %s", prefixes, eniID)
```
M00nF1sh, line 1454: maybe %v

pkg/awsutils/awsutils.go
```diff
@@ -1330,6 +1442,32 @@ func (cache *EC2InstanceMetadataCache) DeallocIPAddresses(eniID string, ips []st
...
+	input := &ec2.UnassignPrivateIpAddressesInput{
+		NetworkInterfaceId: aws.String(eniID),
+		Ipv4Prefixes:       prefixesInput,
+	}
+
+	start := time.Now()
+	_, err := cache.ec2SVC.UnassignPrivateIpAddressesWithContext(context.Background(), input)
+	awsAPILatency.WithLabelValues("UnassignPrivateIpAddresses", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
+	if err != nil {
+		awsAPIErrInc("UnassignPrivateIpAddresses", err)
```
M00nF1sh, line 1466: UnassignPrivateIpAddresses same here

pkg/awsutils/awsutils.go
```diff
@@ -1330,6 +1442,32 @@ func (cache *EC2InstanceMetadataCache) DeallocIPAddresses(eniID string, ips []st
...
+		log.Errorf("Failed to deallocate a Prefixes address %v", err)
+		return errors.Wrap(err, fmt.Sprintf("deallocate prefix: failed to deallocate Prefix addresses: %s", prefixes))
```
M00nF1sh, line 1468: %s->%v?

pkg/awsutils/awsutils.go
jayanthvn: Yeah post GA, I can split AssignPrivateIP and UnAssignPrivateIP into prefixes and IPs. Today check is done inside the function since on the input changes.

pkg/awsutils/awsutils.go
jayanthvn: Sure will handle it as fast follow up

M00nF1sh (approved):

PR #1522, opened 2021-06-29
Header formatting for env variable

jayanthvn:
**What type of PR is this?**
enhancement

N/A

Today it is hard to link an env variable. This will make it easy shareable.

n/a

https://github.com/jayanthvn/amazon-vpc-cni-k8s/tree/fix-readme#enable_prefix_delegation-since-v19

**Automation added to e2e**:

n/a

No

N/a

No
```release-note

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ To select an `ENIConfig` based upon availability zone set this to `failure-domai
 
 ---
 
-#### `AWS_VPC_ENI_MTU` (Since v1.6.0)
+#### `AWS_VPC_ENI_MTU` (v1.6.0+)
 
 Type: Integer as a String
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -472,7 +474,7 @@ limit on private IPs allowed by your instance. Setting or resetting of `ENABLE_P
 
 ---
 
-`WARM_PREFIX_TARGET` 
+#### `WARM_PREFIX_TARGET` (Since v1.9)
```
achevuru, line 477: nit: How about changing it to "v1.9+" instead?

jayanthvn: Yeah, I can do that.

achevuru (approved):

PR #1533, opened 2021-07-19
Updated test agent for snat test

cgchinmay:
**What type of PR is this?**
Using this PR for the test agent changes needed for SNAT tests: https://github.com/aws/amazon-vpc-cni-k8s/pull/1513

### DIFF
--- a/test/agent/cmd/snat-utils/main.go
+++ b/test/agent/cmd/snat-utils/main.go
@@ -1,0 +1,124 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"net/http"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/coreos/go-iptables/iptables"
+)
+
+func main() {
+	var testIPTableRules bool
+	var testExternalDomainConnectivity bool
+	var randomizedSNATValue string
+	var numOfCidrs int
+	var url string
+
+	flag.BoolVar(&testIPTableRules, "testIPTableRules", false, "bool flag when set to true tests validate if IPTable has required rules")
+	flag.StringVar(&randomizedSNATValue, "randomizedSNATValue", "prng", "value for AWS_VPC_K8S_CNI_RANDOMIZESNAT")
+	flag.IntVar(&numOfCidrs, "numOfCidrs", 1, "Number of CIDR blocks in customer VPC")
+	flag.BoolVar(&testExternalDomainConnectivity, "testExternalDomainConnectivity", false, "bool flag when set to true tests if the pod has internet access")
+	flag.StringVar(&url, "url", "https://aws.amazon.com/", "url to check for connectivity")
+
+	flag.Parse()
+
+	if testIPTableRules {
+		err := validateIPTableRules(randomizedSNATValue, numOfCidrs)
+		if err != nil {
+			log.Fatal(err)
+		}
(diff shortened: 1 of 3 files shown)


test/agent/cmd/snat-utils/main.go
```diff
@@ -0,0 +1,104 @@
...
+	if testExternalDomainConnectivity {
+		err := validateExternalDomainConnectivity(url)
+		if err != nil {
+			log.Fatal(err)
+		}
+		log.Println("External Domain Connectivity test passed")
+	}
+}
+
+func validateExternalDomainConnectivity(url string) error {
```
abhipth, line 46: Do we really need this, this can be simplified by doing a `curl --fail url` and check the exit code of the container?

cgchinmay: There were some limitations to using curl with fail flag. For instance not sure if it returns non zero exit code when the http status code is not 200. So used this approach

test/agent/cmd/snat-utils/main.go
```diff
@@ -0,0 +1,104 @@
...
+		expectedString = "random"
+	}
+
+	containsExpectedString := false
+	rule := ""
+
+	for i := 0; i <= numOfCidrs; i++ {
+		curr := fmt.Sprintf("AWS-SNAT-CHAIN-%d", i)
+		fmt.Printf("Checking: %s\n", curr)
+		chains, err := iptables.List("nat", curr)
```
achevuru, line 80: Naming the return value as`chains` for a `list` call on a (nat) table chain is confusing. We're going to get the rules under this chain from this call, we should name it as such. `rules` or something similar might be more appropriate.

test/agent/cmd/snat-utils/main.go
```diff
@@ -0,0 +1,104 @@
...
+
+		for _, chain := range chains {
+			if strings.Contains(chain, expectedString) {
+				rule = chain
+				containsExpectedString = true
+				break
+			}
+		}
+
+		if containsExpectedString {
```
achevuru, line 93: How will we ensure that we indeed have all the required chains? (i.e.,) how do you ensure that you have the `containsExpectedString` on the last chain?

Ideally, we have `numOfCidrs` and the actual values of those Cidrs. We can build the chains and then compare them against what is in the actual `nat` table. Doing so, will ensure we have what is required and nothing more (or less)...

test/agent/cmd/snat-utils/main.go
cgchinmay: Will update this

test/agent/cmd/snat-utils/main.go
cgchinmay: Okay, I will update this to build nat table chains instead and then make comparison

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## v1.9.0
```
achevuru: Why are these changes in this diff? They're already part of master...

Makefile
```diff
@@ -72,7 +72,7 @@ export GOPROXY = direct
 
 VENDOR_OVERRIDE_FLAG =
 # aws-sdk-go override in case we need to build against a custom version
-EC2_SDK_OVERRIDE ?= "y"
+EC2_SDK_OVERRIDE ?= "n"
```
achevuru: Same as above.

achevuru (approved):

Issue #1520, opened 2021-06-27
Add aws-node pod events when EC2/IMDS permissions/access are missing


jayanthvn:
**What would you like to be added**:
Add pod events for missing permissions for either EC2 or IMDS. This will help debugging.

Currently without required IAM permission CNI pod will be stuck with following log lines

sh-4.2$ kubectl logs aws-node-5td62 -n kube-system
{"level":"info","ts":"2021-06-27T17:59:21.804Z","caller":"entrypoint.sh","msg":"Validating env variables ..."}
{"level":"info","ts":"2021-06-27T17:59:21.805Z","caller":"entrypoint.sh","msg":"Install CNI binary.."}
{"level":"info","ts":"2021-06-27T17:59:21.816Z","caller":"entrypoint.sh","msg":"Starting IPAM daemon in the background ... "}
{"level":"info","ts":"2021-06-27T17:59:21.817Z","caller":"entrypoint.sh","msg":"Checking for IPAM connectivity ... "}

Kubelet will have following error message 

runtime network not ready: NetworkReady=false reason:NetworkPluginNotReady message:docker: network plugin is not ready: cni config uninitialized

**Why is this needed**:
Faster triaging

jayanthvn:
#1536 merged to master.

Issue #389, opened 2019-04-10
Document how to determine IAM role permissions issues


mentalisttraceur:
`aws-node` daemonset enters `CrashLoopBackoff` state when the EC2 instance lacks the necessary IAM role without giving any indication of what the problem is in the logs.

jaypipes:
@mentalisttraceur apologies for the long delay in responding here. How do you determine that the reason the daemonset goes into CrashLoopBackoff is due to improper IAM role permissions? I presume we can trap for the permissions failure and log a warning, etc, but I'd like to understand how you are determining this problem at the moment (given, as you say, there are no log messages indicating the IAM role permissions error).

mentalisttraceur:
@jaypipes When we just added an IAM role with the right permissions to the EC2 instance, and made no other changes, the daemonset stopped crashing.

jaypipes:
@mentalisttraceur can you provide a snippet of your logs before and after adding the IAM role so that we can determine what is not being logged or reported as an error? Thanks!

jaypipes:
The deliverable here is to test specifically what happens when you deploy the aws-node daemonset and have improper IAM role permissions. We need to document what this looks like from the k8s user perspective, how to dig into the pod logs to determine it is an IAM permission issue.

jayanthvn:
As part of #1520 we can update the documentation.

PR #1536, opened 2021-07-20
Add event recorder utils to raise aws-node pod events

sushrk:
**What type of PR is this?**
feature

https://github.com/aws/amazon-vpc-cni-k8s/issues/1520

Add event recorder utils to handle raising aws-node pod events for faster triaging. This PR adds missing permissions pod event when EC2 API call fails due to missing CNI policy from the IAM role. 

Repro steps: Detach AmazonEKS_CNI_Policy from IAM role and restart aws-node pods. The pods will not be ready due to missing permissions.

1. Nodegroup creation when CNI policy is missing 
2. Restart aws-node pods when CNI policy missing

Events can be retrieved with "kubectl describe pod aws-node-xxx" & "kubectl get events":
```
2m29s       Warning   MissingIAMPermissions   pod/aws-node-lps4s   Unauthorized operation: failed to call ec2:DescribeNetworkInterfaces due to missing permissions. Please refer https://github.com/aws/amazon-vpc-cni-k8s/blob/master/docs/iam-policy.md to attach relevant policy to IAM role
```
Ginkgo test run:
```
ipamd git:(addPodEvent) ✗ ginkgo -v -- \                                                       
 --aws-region=$AWS_REGION \
 --aws-vpc-id=$VPC_ID \
 --ng-name-label-val=$NG_NAME_LABEL_VAL
Running Suite: VPC IPAMD Test Suite - /Users/ravsushm/go/src/github.com/aws/amazon-vpc-cni-k8s/test/integration-new/ipamd
=========================================================================================================================
Random Seed: 1649808955

Will run 1 of 26 specs
[BeforeSuite] 
/Users/ravsushm/go/src/github.com/aws/amazon-vpc-cni-k8s/test/integration-new/ipamd/ipamd_suite_test.go:39
STEP: Delete coredns addon if it exists 04/12/22 17:16:22.748
[BeforeSuite] PASSED [4.469 seconds]
[BeforeSuite] 
/Users/ravsushm/go/src/github.com/aws/amazon-vpc-cni-k8s/test/integration-new/ipamd/ipamd_suite_test.go:39

  Begin Captured GinkgoWriter Output >>
    STEP: Delete coredns addon if it exists 04/12/22 17:16:22.748
  << End Captured GinkgoWriter Output
SSSSSSSSSSS
test aws-node pod event when iam role is missing VPC_CNI policy
  unauthorized event must be raised on aws-node pod
  /Users/ravsushm/go/src/github.com/aws/amazon-vpc-cni-k8s/test/integration-new/ipamd/ipamd_event_test.go:104
STEP: getting the iam role 04/12/22 17:16:23.151
STEP: getting the node instance role 04/12/22 17:16:23.689
STEP: detaching VPC_CNI policy and restart aws-node pods 04/12/22 17:16:24.07
STEP: checking aws-node pods not running 04/12/22 17:20:23.582
STEP: attaching VPC_CNI policy and restart aws-node pods 04/12/22 17:20:28.6
STEP: checking aws-node pods are running 04/12/22 17:25:04.428
• [SLOW TEST] [531.288 seconds]
test aws-node pod event
/Users/ravsushm/go/src/github.com/aws/amazon-vpc-cni-k8s/test/integration-new/ipamd/ipamd_event_test.go:32
  when iam role is missing VPC_CNI policy
  /Users/ravsushm/go/src/github.com/aws/amazon-vpc-cni-k8s/test/integration-new/ipamd/ipamd_event_test.go:35
    unauthorized event must be raised on aws-node pod
    /Users/ravsushm/go/src/github.com/aws/amazon-vpc-cni-k8s/test/integration-new/ipamd/ipamd_event_test.go:104

  Begin Captured GinkgoWriter Output >>
    STEP: getting the iam role 04/12/22 17:16:23.151
    STEP: getting the node instance role 04/12/22 17:16:23.689
    STEP: detaching VPC_CNI policy and restart aws-node pods 04/12/22 17:16:24.07
    STEP: checking aws-node pods not running 04/12/22 17:20:23.582
    STEP: attaching VPC_CNI policy and restart aws-node pods 04/12/22 17:20:28.6
    STEP: checking aws-node pods are running 04/12/22 17:25:04.428
  << End Captured GinkgoWriter Output
SSSSSSSSSSSSSS
[AfterSuite] 
/Users/ravsushm/go/src/github.com/aws/amazon-vpc-cni-k8s/test/integration-new/ipamd/ipamd_suite_test.go:61
[AfterSuite] PASSED [0.000 seconds]
[AfterSuite] 
/Users/ravsushm/go/src/github.com/aws/amazon-vpc-cni-k8s/test/integration-new/ipamd/ipamd_suite_test.go:61

Ran 1 of 26 Specs in 535.762 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 25 Skipped
PASS | FOCUSED
```
**Automation added to e2e**:
None

No. Upgrade tested.

No.

No.

### DIFF
--- a/charts/aws-vpc-cni/templates/clusterrole.yaml
+++ b/charts/aws-vpc-cni/templates/clusterrole.yaml
@@ -33,3 +33,7 @@ rules:
     resources:
       - '*'
     verbs: ["list", "watch"]
+  - apiGroups: ["", "events.k8s.io"]
+    resources:
+      - events
+    verbs: ["create", "patch", "list", "get"]
--- a/cmd/aws-k8s-agent/main.go
+++ b/cmd/aws-k8s-agent/main.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/k8sapi"
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/eventrecorder"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/version"
 )
@@ -61,6 +62,8 @@ func _main() int {
 		log.Errorf("Failed to create cached kube client: %s", err)
 		return 1
 	}
+
+	eventrecorder.InitEventRecorder(rawK8SClient)
 
 	ipamContext, err := ipamd.New(rawK8SClient, cacheK8SClient)
 	if err != nil {
--- a/config/master/aws-k8s-cni.yaml
+++ b/config/master/aws-k8s-cni.yaml
@@ -70,6 +70,10 @@ rules:
     resources:
       - '*'
     verbs: ["list", "watch"]
+  - apiGroups: ["", "events.k8s.io"]
+    resources:
+      - events
+    verbs: ["create", "patch", "list", "get"]
 ---
 # Source: aws-vpc-cni/templates/clusterrolebinding.yaml
 apiVersion: rbac.authorization.k8s.io/v1
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/awsutils/awssession"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/ec2wrapper"
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/eventrecorder"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/retry"
 	"github.com/aws/aws-sdk-go/aws"
@@ -36,6 +37,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
@@ -77,6 +79,9 @@ var (
 )
 
 var log = logger.Get()
+
+// event recorder to raise aws-node pod events eg. if EC2 API calls fail due to UnauthorizedOperation error
+var eventRecorder *eventrecorder.EventRecorder
 
 var (
 	awsAPILatency = prometheus.NewSummaryVec(
@@ -411,6 +416,8 @@ func New(useCustomNetworking, disableENIProvisioning, v4Enabled, v6Enabled bool)
 	if err != nil {
 		return nil, err
 	}
+	// event recorder to raise events for failed EC2 API calls
+	eventRecorder = eventrecorder.Get()
 
 	// Clean up leaked ENIs in the background
 	if !disableENIProvisioning {
@@ -555,6 +562,7 @@ func (cache *EC2InstanceMetadataCache) RefreshSGIDs(mac string) error {
 						awsAPIErrInc("IMDSMetaDataOutOfSync", err)
 					}
 				}
+				CheckAPIErrorAndBroadcastEvent(err, "ec2:ModifyNetworkInterfaceAttribute")
 				awsAPIErrInc("ModifyNetworkInterfaceAttribute", err)
 				//No need to return error here since retry will happen in 30seconds and also
 				//If update failed due to stale ENI then returning error will prevent updating SG
@@ -695,6 +703,7 @@ func (cache *EC2InstanceMetadataCache) awsGetFreeDeviceNumber() (int, error) {
 	result, err := cache.ec2SVC.DescribeInstancesWithContext(context.Background(), input)
 	awsAPILatency.WithLabelValues("DescribeInstances", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
 	if err != nil {
+		CheckAPIErrorAndBroadcastEvent(err, "ec2:DescribeInstances")
 		awsAPIErrInc("DescribeInstances", err)
 		log.Errorf("awsGetFreeDeviceNumber: Unable to retrieve instance data from EC2 control plane %v", err)
 		return 0, errors.Wrap(err,
@@ -758,6 +767,7 @@ func (cache *EC2InstanceMetadataCache) AllocENI(useCustomCfg bool, sg []*string,
 	_, err = cache.ec2SVC.ModifyNetworkInterfaceAttributeWithContext(context.Background(), attributeInput)
 	awsAPILatency.WithLabelValues("ModifyNetworkInterfaceAttribute", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
 	if err != nil {
+		CheckAPIErrorAndBroadcastEvent(err, "ec2:ModifyNetworkInterfaceAttribute")
 		awsAPIErrInc("ModifyNetworkInterfaceAttribute", err)
 		err := cache.FreeENI(eniID)
 		if err != nil {
@@ -787,6 +797,7 @@ func (cache *EC2InstanceMetadataCache) attachENI(eniID string) (string, error) {
 	attachOutput, err := cache.ec2SVC.AttachNetworkInterfaceWithContext(context.Background(), attachInput)
 	awsAPILatency.WithLabelValues("AttachNetworkInterface", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
 	if err != nil {
+		CheckAPIErrorAndBroadcastEvent(err, "ec2:AttachNetworkInterface")
 		awsAPIErrInc("AttachNetworkInterface", err)
 		log.Errorf("Failed to attach ENI %s: %v", eniID, err)
 		return "", errors.Wrap(err, "attachENI: failed to attach ENI")
@@ -835,6 +846,7 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
 	result, err := cache.ec2SVC.CreateNetworkInterfaceWithContext(context.Background(), input)
 	awsAPILatency.WithLabelValues("CreateNetworkInterface", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
 	if err != nil {
+		CheckAPIErrorAndBroadcastEvent(err, "ec2:CreateNetworkInterface")
 		awsAPIErrInc("CreateNetworkInterface", err)
 		log.Errorf("Failed to CreateNetworkInterface %v", err)
 		return "", errors.Wrap(err, "failed to create network interface")
@@ -884,6 +896,7 @@ func (cache *EC2InstanceMetadataCache) TagENI(eniID string, currentTags map[stri
 		_, err := cache.ec2SVC.CreateTagsWithContext(context.Background(), input)
 		awsAPILatency.WithLabelValues("CreateTags", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
 		if err != nil {
+			CheckAPIErrorAndBroadcastEvent(err, "ec2:CreateTags")
 			awsAPIErrInc("CreateTags", err)
 			log.Warnf("Failed to tag the newly created ENI %s:", eniID)
 			return err
@@ -942,6 +955,7 @@ func (cache *EC2InstanceMetadataCache) freeENI(eniName string, sleepDelayAfterDe
 		_, ec2Err := cache.ec2SVC.DetachNetworkInterfaceWithContext(context.Background(), detachInput)
 		awsAPILatency.WithLabelValues("DetachNetworkInterface", fmt.Sprint(ec2Err != nil), awsReqStatus(ec2Err)).Observe(msSince(start))
 		if ec2Err != nil {
+			CheckAPIErrorAndBroadcastEvent(err, "ec2:DetachNetworkInterface")
 			awsAPIErrInc("DetachNetworkInterface", ec2Err)
 			log.Errorf("Failed to detach ENI %s %v", eniName, ec2Err)
 			return errors.New("unable to detach ENI from EC2 instance, giving up")
@@ -982,6 +996,7 @@ func (cache *EC2InstanceMetadataCache) getENIAttachmentID(eniID string) (*string
 				return nil, ErrENINotFound
 			}
 		}
+		CheckAPIErrorAndBroadcastEvent(err, "ec2:DescribeNetworkInterfaces")
 		awsAPIErrInc("DescribeNetworkInterfaces", err)
 		log.Errorf("Failed to get ENI %s information from EC2 control plane %v", eniID, err)
 		return nil, errors.Wrap(err, "failed to describe network interface")
@@ -1019,6 +1034,7 @@ func (cache *EC2InstanceMetadataCache) deleteENI(eniName string, maxBackoffDelay
 					return nil
 				}
 			}
+			CheckAPIErrorAndBroadcastEvent(ec2Err, "ec2:DeleteNetworkInterface")
 			awsAPIErrInc("DeleteNetworkInterface", ec2Err)
 			log.Debugf("Not able to delete ENI: %v ", ec2Err)
 			return errors.Wrapf(ec2Err, "unable to delete ENI")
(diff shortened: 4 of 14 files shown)


sushrk:
@jayanthvn @abhipth please review.

pkg/ipamd/ipamd.go
```diff
@@ -385,8 +386,11 @@ func (c *IPAMContext) nodeInit() error {
 
 	metadataResult, err := c.awsClient.DescribeAllENIs()
 	if err != nil {
+		// Check error and broadcast event in case of missing EC2 permissions
+		c.CheckEC2Permissions(err)
```
abhipth, line 390: The functions returns an error, but we ignoring it here.

pkg/ipamd/ipamd.go
```diff
@@ -1920,3 +1924,30 @@ func (c *IPAMContext) getPrefixesNeeded() int {
...
+		var podList corev1.PodList
+		err := c.rawK8SClient.List(ctx, &podList, &listOptions)
+		if err != nil {
+			log.Errorf("Failed to get pods: %v", err)
+			return err
+		}
+
+		for _, pod := range podList.Items {
+			log.Debugf("Broadcast event on pod %s", pod.Name)
+			printErr := strings.Replace(nodeInitErr.Error(), "\n", "", 1)
```
abhipth, line 1947: Could you also print the event in the PR description. Does this print the encoded message in the event too?

IMO, something like this would be more descriptive and not mention the "node" role as the user could also be using IRSA.

*Reason:* MissingIAMPolicy
*Message:* "aws-node failed to start because it's unauthorized to perform EC2 API calls. Please verify AmazonEKS_CNI_Policy is attached to the supplied IAM Role" 

I think if we send out the entire error message in Pod event, it will not get de-duped and be shown as separate events. However, the warning above will get de-dupeed like this - 37m (x19 over 18h)

pkg/k8sapi/k8sutils.go
```diff
@@ -5,17 +5,39 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 var log = logger.Get()
 
+var eventRecorder record.EventRecorder
+
+func InitRecorder() {
+	restCfg, err := ctrl.GetConfig()
```
abhipth, line 25: Can we reuse GetKubeClientSet and remove these 8 lines?

pkg/ipamd/ipamd.go
sushrk: Removed the return and printed an error message instead, thanks!

pkg/k8sapi/k8sutils.go
```diff
@@ -5,17 +5,39 @@ import (
...
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 var log = logger.Get()
 
+var eventRecorder record.EventRecorder
```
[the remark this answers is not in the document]
sushrk: Makes sense, moved it to ipamd.
Thanks!

pkg/k8sapi/k8sutils.go
sushrk: GetKubeClient was using "InClusterConfig" that was causing an issue earlier; but moving the recorder to ipamd seems to work. 
Changed it, thanks!

jayanthvn:
Can you also update the documentation. Maybe we can include it here [https://github.com/aws/amazon-vpc-cni-k8s/blob/master/docs/troubleshooting.md]. 

Ref : https://github.com/aws/amazon-vpc-cni-k8s/issues/389

pkg/ipamd/ipamd.go
```diff
@@ -1920,3 +1949,30 @@ func (c *IPAMContext) getPrefixesNeeded() int {
 	log.Debugf("ToAllocate: %d", toAllocate)
 	return toAllocate
 }
+
+func (c *IPAMContext) BroadcastMissingPermissionsEvent(nodeInitErr error) {
+	printErr := strings.Replace(nodeInitErr.Error(), "\n", "", 1)
+	log.Errorf("Failed to create pod, please verify AmazonEKS_CNI_Policy is attached to supplied IAM Role: {%v}", printErr)
```
sftim, line 1955: - This error implies the cluster operator must use a specific IAM policy. I think any equivalent policy would be valid too.
- I think `UnauthorizedOperation` could be the result of an explicit deny, so that error could message could be shown even when the IAM role has that permission attached.
- do we _always_ know that the principal running this code is an IAM role?

I think I'd prefer to know which action was found to be forbidden.

jayanthvn:
@sushrk - Can you please fix the branch conflicts?

pkg/ipamd/ipamd.go
jayanthvn: @sftim - Yes it makes sense, we can log the action which was forbidden.

sushrk: Thank you. I've modified the alarm message. 
The error log shows the request-id for the unauthorized operation.

pkg/ipamd/ipamd.go
```diff
@@ -412,8 +422,30 @@ func New(rawK8SClient client.Client, cachedK8SClient client.Client) (*IPAMContex
...
+		log.Errorf("Failed to get client set", err)
+		return nil, err
+	}
+
+	eventBroadcaster := record.NewBroadcaster()
+	eventBroadcaster.StartRecordingToSink(&typedcorev1.EventSinkImpl{
+		Interface: clientSet.CoreV1().Events(""),
+	})
+	c.eventRecorder = eventBroadcaster.NewRecorder(clientgoscheme.Scheme, corev1.EventSource{
+		Component: "aws-node",
```
jayanthvn, line 436: This can be changed "aws-node" -> awsNode

pkg/ipamd/ipamd.go
```diff
@@ -412,8 +422,30 @@ func New(rawK8SClient client.Client, cachedK8SClient client.Client) (*IPAMContex
...
+		Host:      c.myNodeName,
+	})
+
 	err = c.nodeInit()
 	if err != nil {
+		log.Debugf("Failed to initialize node, err: %v", err)
+		var awsErr awserr.Error
+		if errors.As(err, &awsErr) {
```
jayanthvn, line 444: I think we should move this logic to another function. The function can take in "err". Initially do sanity checks and then we can have case statements for each error code we need to broadcast. In that way adding new events will be easy, just need to call the function and add a new case inside the function.

pkg/ipamd/ipamd.go
```diff
@@ -462,8 +494,9 @@ func (c *IPAMContext) nodeInit() error {
 
 	metadataResult, err := c.awsClient.DescribeAllENIs()
 	if err != nil {
-		return errors.New("ipamd init: failed to retrieve attached ENIs info")
+		return errors.Wrapf(err, "ipamd init: failed to retrieve attached ENIs info")
```
jayanthvn, line 497: This can be errors.Wrap since we just have a message.

pkg/ipamd/ipamd.go
```diff
@@ -2188,3 +2221,29 @@ func (c *IPAMContext) isConfigValid() bool {
 
 	return true
 }
+
+func (c *IPAMContext) BroadcastMissingPermissionsEvent(nodeInitErr error) {
```
jayanthvn, line 2225: As mentioned above, we need to have `BroadcastEvent` and inside that each case based on the error code should add the appropriate event. In that way we can reuse most of the code for newer events.

pkg/ipamd/ipamd.go
```diff
@@ -2188,3 +2221,29 @@ func (c *IPAMContext) isConfigValid() bool {
...
+	}
+
+	var podList corev1.PodList
+	err := c.rawK8SClient.List(ctx, &podList, &listOptions)
+	if err != nil {
+		log.Errorf("Failed to get pods, cannot broadcast missing permissions event: %v", err)
+		return
+	}
+
+	for _, pod := range podList.Items {
```
jayanthvn, line 2243: Will this not be just one pod? Since only one instance of aws-node will be running. Am I missing anything?

jayanthvn (changes requested):

pkg/ipamd/ipamd.go
sushrk: Yes, in this case, it will be just one pod for the specified label & fields. 
Using the List operation since we set list options. This is also more generic than getting one pod, as it can be used in case we want to raise events on other pods.

jayanthvn: Yeah but we won't be raising events for other pods. Lets just keep it to aws-node pod.

pkg/ipamd/ipamd.go
sushrk: Sounds good, will update the PR with these changes.

cmd/aws-k8s-agent/main.go
```diff
@@ -59,6 +60,8 @@ func _main() int {
 	ipamContext, err := ipamd.New(rawK8SClient, cacheK8SClient)
 	if err != nil {
 		log.Errorf("Initialization failure: %v", err)
+		// Allow time for event to be broadcast in case of missing permissions
```
achevuru, line 63: Curious, why do we need this? Do we see a pod tear down in this case? Isn't this just a container restart case? and so even if the event broadcast takes a while, it should eventually show up on Pod events. On top of it, we retry on ec2 call failures and that should give us ample time. I'm assuming this is here based on what you observed during testing but trying to understand more on why this is required.

pkg/ipamd/ipamd.go
```diff
@@ -2188,3 +2221,29 @@ func (c *IPAMContext) isConfigValid() bool {
 
 	return true
 }
+
+func (c *IPAMContext) BroadcastMissingPermissionsEvent(nodeInitErr error) {
+	log.Errorf("Failed to start aws-node due to missing policy, verify AmazonEKS_CNI_Policy(or equivalent) is attached to supplied IAM Role, err:%v", nodeInitErr)
```
achevuru, line 2226: `AmazonEKS_CNI_Policy` doesn't apply for v6 clusters though and in addition we now support cluster scoped policy. We should just point them to refer to our docs - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/docs/iam-policy.md

pkg/ipamd/ipamd.go
achevuru: I mean we can't either. We will have no idea why other pods failed.

achevuru: Ok, why do we even need this though? All we need is the Pod name where this blob of code is running. Why can't we just source it via env variable and be done with it. What am I missing? We do that for Node Name right now - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/master/aws-k8s-cni.yaml#L207 and we can do the same for Pod Name. Why do we need to make an API Server call here? We can do away with the code where we are creating kubeclient etc.

I'm suggesting something like below - 

```
        - name: MY_POD_NAME
          valueFrom:
            fieldRef:
              fieldPath: metadata.name
```

sushrk: Thanks! Previously we couldn't get the pod name without listing the pods, but adding the ENV will work.
Will update the PR.

cmd/aws-k8s-agent/main.go
sushrk: That's right, I observed that events take a while to show up. If the container restarts before that, the events were getting lost. 
I suspect this is because we start a new broadcaster on each restart, so this might still be required. Will look into this a bit more & update.

achevuru: If this is needed, let's move it to the `BroadcastEvent` function instead. We can land here for a variety of reasons and this is only required for a particular scenario this PR is addressing.

achevuru:
I've a generic comment - PR only captures the scenario when NodeInit fails (due to missing permissions for `DescribeNetworkInterfaces` call). However, CNI relies on a bunch of other EC2/VPC APIs(Attach/Detach/Modify ENIs, Assign/Unassign IPs etc) and it will be very useful to log an event whenever we receive a 403 for any of these API calls. Our Event message can specifically call out the API for which we ran in to Unauthorized error along with a generic statement pointing to relevant resources. With the current PR, we only log an event if the CNI pod is missing permissions for `DescribeNetworkInterfaces` call but I believe the end goal is to alert the customer via events if they are missing permissions for any EC2/VPC API that CNI relies on for it's operation.

jayanthvn:
Thats right @achevuru. End goal is to capture all important events. Hence making the function common will make it easy to extend.

cmd/aws-k8s-agent/main.go
sushrk: sleep statement is not required here as the BroadcastEvent call is moved to awsutils.

achevuru:
Thanks for making it generic. However, I only see the event being logged for `DescribeNetworkInterfaces' call. Now that we've generic infra in place - we should log all the `Unauthorized` errors for any of the EC2 calls VPC CNI relies on (i.e.,) Create/Delete/Attach/Detach/Modify ENI & Assign/Unassign IPs.

jayanthvn:
> Thanks for making it generic. However, I only see the event being logged for `DescribeNetworkInterfaces` call. Now that we've generic infra in place - we should log all the `Unauthorized` errors for any of the EC2 calls VPC CNI relies on (i.e.,) Create/Delete/Attach/Detach/Modify ENIs & Assign/Unassign IPs.

Agreed.

jayanthvn:
Can you please fix the conflicts?

config/master/aws-k8s-cni.yaml
```diff
@@ -70,6 +70,10 @@ rules:
     resources:
       - '*'
     verbs: ["list", "watch"]
+  - apiGroups: ["", "events.k8s.io"]
```
jayanthvn, line 73: This has to be updated in charts - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/charts/aws-vpc-cni/templates/clusterrole.yaml as well and then generate the yaml.

jayanthvn, line 73: Let's update the remaining yaml i.e, cn, gov regions. `make generate-cni-yaml` should generate all yamls in build folder.

pkg/awsutils/awsutils.go
```diff
@@ -755,6 +769,11 @@ func (cache *EC2InstanceMetadataCache) AllocENI(useCustomCfg bool, sg []*string,
 	_, err = cache.ec2SVC.ModifyNetworkInterfaceAttributeWithContext(context.Background(), attributeInput)
 	awsAPILatency.WithLabelValues("ModifyNetworkInterfaceAttribute", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
 	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
```
jayanthvn, line 772: We have these if statements in multiple places, I am thinking we can instead move all of this inside `BroadcastMissingPermissionsEvent`.

Inside `BroadcastMissingPermissionsEvent` we can check `ok := err.(awserr.Error)` and return if !ok.

sftim:
@sushrk if you know how to do a [rebase and squash](https://garrytrinder.github.io/2020/03/squashing-commits-using-interactive-rebase), I recommend that you do. That would make these changes easier to review and to accept.

pkg/awsutils/awsutils.go
```diff
@@ -36,6 +37,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
+	v1 "k8s.io/api/core/v1"
```
jayanthvn, line 40: Can we move this to events.k8s.io v1. Upstream seems to be moving away from corev1/Events to events.k8s.io/Events. ref:https://github.com/kubernetes/kubernetes/pull/102832

pkg/ipamd/ipamd.go
sushrk: We are not saving any API calls by adding ENV for POD_NAME since we still need to make an API call to get the pod object. 
This code is moved to event recorder utils, but the logic will remain same.

jayanthvn (approved):
lgtm

charts/aws-vpc-cni/templates/clusterrole.yaml
```diff
@@ -33,3 +33,7 @@ rules:
     resources:
       - '*'
     verbs: ["list", "watch"]
+  - apiGroups: ["", "events.k8s.io"]
+    resources:
+      - events
+    verbs: ["create", "patch", "list", "get"]
```
a7i, line 39: why does it need `patch`?

sushrk: This is required for event aggregation. For eg, without `patch` we wouldn't see the aggregated info [x12 over 21s] below:
`Warning  MissingIAMPermissions  20s (x12 over 21s)  aws-node           Unauthorized operation ...`

FYI we removed `get` recently as it was not required.

a7i: thanks for the explanation!

PR #1549, opened 2021-07-26
[wip] Shrink to a 'minimal' container image

anguslees:
Reduce the 'release' container image to just the binaries we need.
Smaller size, and also removes false cve-scanner alerts in binaries we
don't use anyway.

To avoid Dockerfile duplication, this change also collapses several images
to use separate targets from a single `Dockerfile.release`.

```release-note
Switch to 'minimal' container images.  Should have no visible effect, unless you derive from these images and expect additional commands to be available.
```

[WIP]  Still todo: change the build workflows to use docker buildx.


anguslees:
Urgh, and I should probably split this into two separate PRs: 1/ minimal image, 2/ build multiarch image via buildx

anguslees:
Abandoning, in favour of similar eks-distro based approaches.

Issue #1554, opened 2021-07-26
initcontainer should have requests/limits


grosser:
**What would you like to be added**:
https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/master/aws-k8s-cni.yaml#L199
needs requests/limits

this has worked well for us:
```yaml
        resources:
          requests:
            cpu: 20m
            memory: 32Mi
          limits:
            cpu: 200m
            memory: 128Mi
```

**Why is this needed**:
unlimited containers are can cause issues / are blocked by our cluster policies

jayanthvn:
Init container is a short lived pod hence I assume this was not added by default.

grosser:
anything against adding it ?
... I can make the PR

jayanthvn:
Happy to take PR :)

Also we think setting 200m/128Mi for limits might be high for a init container. Maybe 50m/64Mi is sufficient?

grosser:
I was unable to get metrics out due to it running so quickly
tried 50m/64Mi and it worked
https://github.com/aws/amazon-vpc-cni-k8s/pull/1559

jayanthvn:
Cool thanks for trying it out :)

jayanthvn:
PR is merged.

PR #1559, opened 2021-07-27
set requests/limits for initcontainer

grosser:
**What type of PR is this?**

feature

https://github.com/aws/amazon-vpc-cni-k8s/issues/1554

adds requests/limits for initcontainer so it does not get rejected when cluster requires it and serves as example for users that want to set it, also prevents cni from consuming too much cpu/memory

deployed to our clusters and did not see oomkills or slow boot

No

yes

```release-note
set default init container requests and limits
```


jayanthvn:
I have re-triggered the integration tests.

jayanthvn:
Limits look good. But for the release, we have to verify testing and also update this with managed add ons.

jayanthvn (approved):
Thanks for taking it up :)

PR #1563, opened 2021-07-30
Added Multus artifacts to config folder

cgchinmay:
NA

Provides multus config files for Multus Image installation for different regions

Verified Image Installation from build-beta account

No

### DIFF
--- a/config/multus/README.md
+++ b/config/multus/README.md
@@ -1,0 +1,21 @@
+## Setup
+The following Multus installation will set up aws-vpc-cni as the default plugin.
+
+The Multus daemonset assumes that 10-aws.conflist (aws-vpc-cni) config file is present in the /etc/cni/net.d/ directory on your worker node to generate the multus config from it. 
+
+Download the latest version of the [yaml](../multus/) and apply it the cluster.
+```
+kubectl apply -f aws-k8s-multus.yaml
+```
+
+Check logs for Multus pod
+```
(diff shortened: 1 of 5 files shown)


jayanthvn:
It would be good to add something like this - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/master/manifests.jsonnet and using "go generate" we can rebuild the manifest. Also any plans to add charts (https://github.com/aws/amazon-vpc-cni-k8s/tree/master/charts) folder?

cgchinmay:
> It would be good to add something like this - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/master/manifests.jsonnet and using "go generate" we can rebuild the manifest. Also any plans to add charts (https://github.com/aws/amazon-vpc-cni-k8s/tree/master/charts) folder?

Will do that. The charts are not yet planned for this release. Will have to decide the timeline on that

fawadkhaliq (approved):

cgchinmay:
> It would be good to add something like this - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/master/manifests.jsonnet and using "go generate" we can rebuild the manifest. Also any plans to add charts (https://github.com/aws/amazon-vpc-cni-k8s/tree/master/charts) folder?

I will create another PR to add templating file jsonnet. Since we need these artifacts by Monday, merging this PR

PR #1574, opened 2021-08-16
Upgrade controller-runtime dependency in test dir

sushrk:
**What type of PR is this?**
Cleanup/bug

N/A

Recent helm upgrade in test dir caused k8s dependencies to be upgraded. Upgrading controller-runtime to match these upgrades.

N/A

Verified CNI integration test cases run.

**Automation added to e2e**:
N/A

No.

No

No.


jayanthvn:
There is another PR which fixes the commits - #1582

PR #1580, opened 2021-08-19
Knob to enable bandwidth plugin

### LINKED ISSUE #1437: Container runtime reads partially written CNI plugin config and fails
**What happened**
Container runtime can read a partially written CNI config and fail. This is because ipamd installation/restart writes the CNI plugin configuration directly to [CNI config directory](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/entrypoint.sh#L109) (e.g. `/host/etc/cni/net.d`). The container runtime net config sync loop (e.g. in [containerd](https://github.com/containerd/containerd/blob/0867c99b1a69cda79ae7354cbe54b05ffb6093ac/pkg/cri/server/cni_conf_syncer.go#L73)) gets triggered on fs change events and if the contents of CNI config are not fully written, container runtime will fail with an error like `unexpected end of JSON input`. 

We should write to tmp and move the _fully written_ file to CNI config directory

**Attach logs**
```
msg="failed to reload cni configuration after receiving fs
change event(\"/etc/cni/net.d/10-aws.conflist\": WRITE)" error="cni config load failed: failed to load CNI config list file /etc/cni/net.d/10-aws.conflist: error parsing configuration li
st: unexpected end of JSON input: invalid cni config: failed to load cni config"
```

**What you expected to happen**
We should write the conf to a temp file and move the already written file to CNI config directory so the container runtime always sees a fully written file and not incomplete/invalid JSON forcing an unnecessary failure on the container runtime. 

It's a race. Restart ipamd and container runtime at the same time.

**Anything else we need to know?**
Nope

**Environment**
- Kubernetes version (use `kubectl version`): N/A
- CNI Version: All VPC CNI versions
- OS (e.g: `cat /etc/os-release`): N/A
- Kernel (e.g. `uname -a`): N/A

jayanthvn:
**What type of PR is this?**
Feature

Update 10-aws.conflist file when bandwidth plugin is needed

When knob is enabled , we add bandwidth plugin to 10-aws.conflist file.

Also fixes #1437 

n/a

```
[root@ip-192-168-32-207 ec2-user]# cat /etc/cni//net.d/10-aws.conflist 
{
  "cniVersion": "0.3.1",
  "name": "aws-cni",
  "plugins": [
    {
      "name": "aws-cni",
      "type": "aws-cni",
      "vethPrefix": "eni",
      "mtu": "1500",
      "pluginLogFile": "/var/log/aws-routed-eni/plugin.log",
      "pluginLogLevel": "Debug"
    },
    {
      "type": "portmap",
      "capabilities": {
        "portMappings": true
      },
      "snat": true
    },
    {
      "type": "bandwidth",
      "capabilities": {
        "bandwidth": true
      }
    }
  ]
}
```

Disable ENABLE_BANDWIDTH_PLUGIN

```
[root@ip-192-168-32-207 ec2-user]# cat /etc/cni//net.d/10-aws.conflist 
{
  "cniVersion": "0.3.1",
  "name": "aws-cni",
  "plugins": [
    {
      "name": "aws-cni",
      "type": "aws-cni",
      "vethPrefix": "eni",
      "mtu": "1500",
      "pluginLogFile": "/var/log/aws-routed-eni/plugin.log",
      "pluginLogLevel": "Debug"
    },
    {
      "type": "portmap",
      "capabilities": {"portMappings": true},
      "snat": true
    }
  ]
}
```
**Automation added to e2e**:

No

Yes, if user needs bandwidth plugin

```release-note
Bandwidth plugin support
```


scripts/entrypoint.sh
```diff
@@ -150,7 +152,14 @@ sed \
   -e s~__MTU__~"${AWS_VPC_ENI_MTU}"~g \
   -e s~__PLUGINLOGFILE__~"${AWS_VPC_K8S_PLUGIN_LOG_FILE}"~g \
   -e s~__PLUGINLOGLEVEL__~"${AWS_VPC_K8S_PLUGIN_LOG_LEVEL}"~g \
-  10-aws.conflist > "$HOST_CNI_CONFDIR_PATH/10-aws.conflist"
+  10-aws.conflist > "$TMP_AWS_CONFLIST_FILE"
+
+if [[ "$ENABLE_BANDWIDTH_PLUGIN" == "true" ]]; then
```
M00nF1sh, line 157: what's the drawback of enable bandwidth plugin by default?

M00nF1sh (commented):
looks good to me, except would like to understand why we don't enable it by default like the portmap. 
e.g. will it introduce performance issues if ppl don't use the bandwidth functionality?

M00nF1sh (approved):
/lgtm

scripts/entrypoint.sh
jayanthvn: I haven't measured the performance but we will be unnecessarily making the calls to bandwidth plugin. Hence enabled it with a knob.

PR #1582, opened 2021-08-19
Upgrading controller-runtime in test directory

sushrk:
What type of PR is this?
Cleanup/bug

Which issue does this PR fix:
N/A

What does this PR do / Why do we need it:
Recent helm upgrade in test dir caused k8s dependencies to be upgraded. Upgrading controller-runtime to match these upgrades.

If an issue # is not available please add repro steps and logs from IPAMD/CNI showing the issue:
N/A

Testing done on this change:
Verified CNI integration test cases run.

Automation added to e2e:
N/A

Will this break upgrades or downgrades. Has updating a running cluster been tested?:
No.

No

No.


test/integration-new/cni/pod_traffic_test_PD_toggle.go
```diff
@@ -42,10 +42,6 @@ var _ = Describe("Test pod networking with prefix delegation enabled <-> disable
 	JustBeforeEach(func() {
 		// TODO Gingko doesnt support beforeAll so while adding upgrades/downgrades will move this to a suite
 		if firstRun {
-			By("creating test namespace")
```
achevuru, line 45: Any reason we are removing this? Is the namespace carried over from the prior test suite?

abhipth: The suite had the namespace creation/deletion being done in Before/AftureSuite blocks. This change introduced by PD tests creates/deletes the same namespace in JustBeforeEach/JustAfterEach blocks. 

If the PD test runs first and delete the namespace, then the other tests would fail which previously relied on namespace to be present during test.  

I think we can either remove this, or change the PD test namespace to fix the issue.

jayanthvn (approved):
lgtm

PR #1595, opened 2021-08-27
Consider attached ENI to the node as unmanaged if its IP is not overlapping with the VPC CIDR block of the primary ENI

Rangudu:
**What type of PR is this?**
feature

Allows CNI to consider ENIs that belongs to a different VPC block than the primary ENI of the node as unmanaged

AWS KDA have an use-case where cross account/vpc eni attachment to the nodes in the EKS worker node group can be done to provide access to resource that might exists in a different VPC for security.
Hence in this case this ENI needs to be considered as unmanaged by the CNI so that any network setting that might have been created for this setup work are not entangled by the CNI actions.
So this change will check if the IP address of the ENI attached to the node overlaps with the VPC CIDR block of the primary ENI of the node if not then it will consider that ENI as unmanaged.  

n/a

Manually tested the changes in our beta stack with an VPC application setup

Following are the logs of the Ipamd where it is considering the X-ENI as unmanaged after the attachment is completed successfully

```
{"level":"info","ts":"2021-08-27T10:06:42.640Z","caller":"ipamd/ipamd.go:1124","msg":"Checking if ENI eni-03a56e99be96a1914 (devNum: 1) should be unmanaged"}
{"level":"info","ts":"2021-08-27T10:06:42.640Z","caller":"ipamd/ipamd.go:1124","msg":"Filtered all unmanaged ENIs: 0"}
{"level":"info","ts":"2021-08-27T10:06:42.640Z","caller":"ipamd/ipamd.go:1124","msg":"Checking if ENI eni-04f15b9ecbcbe34ef (devNum: 2) should be unmanaged"}
{"level":"info","ts":"2021-08-27T10:06:42.640Z","caller":"ipamd/ipamd.go:1124","msg":"Treating ENI eni-04f15b9ecbcbe34ef as unmanaged because it's VPC CIDR (172.31.0.0/16) doesn't match the VPC CIDR of the primary ENI (142.151.0.0/16)"}
{"level":"info","ts":"2021-08-27T10:06:42.640Z","caller":"ipamd/ipamd.go:1124","msg":"Checking if ENI eni-0a2c12d93743c3c01 (devNum: 0) should be unmanaged"}
{"level":"info","ts":"2021-08-27T10:06:42.640Z","caller":"ipamd/ipamd.go:1124","msg":"Filtered all unmanaged ENIs: 1"}
```

**Automation added to e2e**:
no

no

no

no

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -223,6 +223,9 @@ type ENIMetadata struct {
 	// SubnetIPv4CIDR is the ipv4 cider of network interface
 	SubnetIPv4CIDR string
 
+	// PrimaryVPCIPv4CIDRBlock is the ipv4 CIDR of the primary network interface's VPC
+	PrimaryVPCIPv4CIDRBlock string
+
 	// The ip addresses allocated for the network interface
 	IPv4Addresses []*ec2.NetworkInterfacePrivateIpAddress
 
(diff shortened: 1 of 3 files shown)


jayanthvn:
@Rangudu - I think we don't need this PR anymore?

Rangudu:
Yes we don't need

jayanthvn:
Thanks @Rangudu

PR #1596, opened 2021-08-27
Respect existing eniConfig label if set on node. 

### LINKED ISSUE #1593: CNI should respect eniConfig label if it already exists.
**What happened**:

We currently have a set of nodes that use custom networking. As part of a new application being deployed, we want to give the pods on these nodes an ENIConfig that is different from the ENIConfig for other nodes in the cluster (which is set based on AZ). To try and achieve this, we provided userdata to this set of nodes to have the kubelet set an initial eniConfig label. 

When the nodes start up, they are re-labeled by the CNI with the value of the label specified in `ENI_CONFIG_LABEL_DEF` (which ends up mapping to an AZ)

**Attach logs**

I'd expect the CNI to respect the eniConfig label if it already set on the node. 
Set the eniConfig label in the kubelet configuration. Start and join the node to a K8S cluster with the CNI installed. Notice that the label is modified.

I noticed that a workaround for this is possible by specifying the annotation, but from a user perspective, that's much harder to control/set compared to labels - labels can be set via kubelet options, whereas the annotation would need to be set by some controller or other component. 

**Environment**:
- Kubernetes version (use `kubectl version`): 1.21
- CNI Version: 1.9.0
- OS (e.g: `cat /etc/os-release`): BottleRocket 1.2.0
- Kernel (e.g. `uname -a`):

backjo:
Closes #1593 
Signed-off-by: Jonah Back <jonah@jonahback.com>

**What type of PR is this?**

Bug
#1593 

This PR makes the CNI respect existing eniConfig labels if they exist. This is useful for users who may want to setup custom ENIConfigurations, but want to manage it through node labels (which notably can be passed to kubelet at startup) instead of through the override annotation.

Manual testing on an EKS cluster to ensure that the CNI was properly choosing the right ENIConfig. 

**Automation added to e2e**:

It could be considered a breaking change if a user is currently setting these labels, since it would start using their specified ENIConfig instead of the one that the CNI chooses.

No

No
```release-note

```

### DIFF
--- a/pkg/eniconfig/eniconfig.go
+++ b/pkg/eniconfig/eniconfig.go
@@ -34,6 +34,10 @@ const (
 	eniConfigDefault              = "default"
 	eniConfigLabel                = "vpc.amazonaws.com/eniConfig"
 
+	// when this is defined, it is to be treated as the source of truth for the eniconfig.
+	// it is meant to be used for out-of-band mananagement of the eniConfig - i.e. on the kubelet or elsewhere
+	externalEniConfigLabel = "vpc.amazonaws.com/externalEniConfig"
+
 	// when "ENI_CONFIG_LABEL_DEF is defined, ENIConfigController will use that label key to
 	// search if is setting value for eniConfigLabelDef
 	// Example:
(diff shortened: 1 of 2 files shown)


backjo:
@jayanthvn ran the make format - needs an approval to allow running workflows.

backjo:
hey @jayanthvn  - anything needed here?

jayanthvn (approved):
LGTM. Thanks for the PR and UTs.

backjo:
Cool! Is it fair to expect this to land in the 1.9.2 release?

backjo:
@jayanthvn any ideas if/when this will land in master?

PR #1604, opened 2021-09-08
Unset GOOS/GOARCH for generate target

anguslees:
**What type of PR is this?**

bug

`make generate` runs `go run ...` (and other tools).  These tools need
to run on the native build host and not the target OS/arch.

Running `make generate` on a mac currently gives:
```console
go run github.com/golang/mock/mockgen -destination mocks/awsutils_mocks.go -copyright_file ../../scripts/copyright.txt . APIs
fork/exec /var/folders/xp/g_qrz0_d25v2jzrdn08153qxpsx_z3/T/go-build305079594/b001/exe/mockgen: exec format error
pkg/awsutils/generate_mocks.go:16: running "go": exit status 1
```


anguslees:
I only have easy access to linux/amd64 - so it would be useful if someone with a darwin/mac (or other) could confirm the above actually works as intended...

achevuru (approved):

PR #1607, opened 2021-09-10
Support for no_manage=false

jayanthvn:
**What type of PR is this?**
Enhancement

Certain use cases need to use only specific ENIs for pod IPs similar to allow and deny list. Currently we support no_manage = true tag but certain use cases need no_manage=false.

`no_manage=false` will take precedence and will be considered for pod IP allocation and If `MANAGE_UNTAGGED_ENI` = `false` is set then any untagged ENIs will be marked as unmanaged by IPAMD and will not be considered for pod IP allocation.

N/A

yes. 
**Automation added to e2e**:

Will update in a follow up PR.

No

No

No

```release-note

```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -182,6 +182,9 @@ type APIs interface {
 
 	//Update cached prefix delegation flag
 	InitCachedPrefixDelegation(bool)
+
+	// GetInstanceID returns the instance ID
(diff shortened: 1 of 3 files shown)


M00nF1sh (approved):
/lgtm
let's add some UT for this function.

jayanthvn:
Yes, I am working on the UTs PR. Will send a follow up tonight.

PR #1616, opened 2021-09-13
Fallback for get hypervisor type and eni ipv4 limits

jayanthvn:
**What type of PR is this?**
Bug

For new instances get hypervisor type and get eni ipv4 limit will fail

Fallback for get instance type

```
{"level":"debug","ts":"2021-09-13T21:19:38.935Z","caller":"awsutils/awsutils.go:1317","msg":"Instance type limits are missing from vpc_ip_limits.go hence making an EC2 call to fetch the limits"}
{"level":"debug","ts":"2021-09-13T21:19:39.902Z","caller":"ipamd/ipamd.go:365","msg":"Instance hypervisor family nitro"}
{"level":"info","ts":"2021-09-13T21:19:39.902Z","caller":"ipamd/ipamd.go:374","msg":"Prefix Delegation enabled false"}
{"level":"debug","ts":"2021-09-13T21:19:39.902Z","caller":"ipamd/ipamd.go:379","msg":"Start node init"}
```
**Automation added to e2e**:

No
No

No

```release-note

```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -1249,12 +1249,46 @@ func (cache *EC2InstanceMetadataCache) AllocIPAddress(eniID string) error {
 	return nil
 }
 
+func (cache *EC2InstanceMetadataCache) fetchLimitsFromEC2() (InstanceTypeLimits, error) {
+	log.Debugf("Instance type limits are missing from vpc_ip_limits.go hence making an EC2 call to fetch the limits")
+	var eniLimits InstanceTypeLimits
+	describeInstanceTypesInput := &ec2.DescribeInstanceTypesInput{InstanceTypes: []*string{aws.String(cache.instanceType)}}
+	output, err := cache.ec2SVC.DescribeInstanceTypesWithContext(context.Background(), describeInstanceTypesInput)
+	if err != nil || len(output.InstanceTypes) != 1 {
+		log.Errorf("", err)
+		return InstanceTypeLimits{}, errors.New(fmt.Sprintf("Failed calling DescribeInstanceTypes for `%s`: %v", cache.instanceType, err))
+	}
+	info := output.InstanceTypes[0]
+	// Ignore any missing values
+	instanceType := aws.StringValue(info.InstanceType)
+	eniLimit := int(aws.Int64Value(info.NetworkInfo.MaximumNetworkInterfaces))
+	ipv4Limit := int(aws.Int64Value(info.NetworkInfo.Ipv4AddressesPerInterface))
+	hypervisorType := aws.StringValue(info.Hypervisor)
+	//Not checking for empty hypervisorType since have seen certain instances not getting this filled.
+	if instanceType != "" && eniLimit > 0 && ipv4Limit > 0 {
+		eniLimits = InstanceTypeLimits{
+			ENILimit:       eniLimit,
+			IPv4Limit:      ipv4Limit,
+			HypervisorType: hypervisorType,
+		}
+		InstanceNetworkingLimits[instanceType] = eniLimits
+	} else {
+		return InstanceTypeLimits{}, errors.New(fmt.Sprintf("%s: %s", UnknownInstanceType, cache.instanceType))
+	}
+	return eniLimits, nil
+}
+
 // GetENIIPv4Limit return IP address limit per ENI based on EC2 instance type
 func (cache *EC2InstanceMetadataCache) GetENIIPv4Limit() (int, error) {
 	eniLimits, ok := InstanceNetworkingLimits[cache.instanceType]
 	if !ok {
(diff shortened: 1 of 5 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -1293,8 +1312,13 @@ func (cache *EC2InstanceMetadataCache) GetENILimit() (int, error) {
 func (cache *EC2InstanceMetadataCache) GetInstanceHypervisorFamily() (string, error) {
 	eniLimits, ok := InstanceNetworkingLimits[cache.instanceType]
 	if !ok {
-		log.Errorf("Failed to get hypervisor info due to unknown instance type %s", cache.instanceType)
-		return "", errors.New(UnknownInstanceType)
+		// Fetch from EC2 API
+		var err error
+		eniLimits, err = cache.fetchLimitsFromEC2()
```
achevuru, line 1317: So, do we need to call the `fetchLimitsFromEC2` function independently from each of these GET functions? If it is missing in the packaged file, can we call it from one cental loc and gather all the data. I mean, it is not possible we miss one of these parameters for an instance present in the file right?

jayanthvn: Just want to keep the functions independent and have fall back mechanism if we ever use it separately. Also to keep the changes minimal for a patch release.

M00nF1sh: same here, we should avoid call the same API that returns static information multiple times.
maybe a simple cache with just map[instanceType]ENILimits?

jayanthvn: Effectively we do always return from the cache. GetInstanceHypervisorFamily() is first called and we fill the cache and later function will always return from the cache. But yeah we can move this functionality to common init and always return from cache for code readability.

pkg/awsutils/awsutils.go
```diff
@@ -1249,55 +1252,58 @@ func (cache *EC2InstanceMetadataCache) AllocIPAddress(eniID string) error {
...
+	if ok {
+		return nil
+	}
+
+	log.Debugf("Instance type limits are missing from vpc_ip_limits.go hence making an EC2 call to fetch the limits")
+	var eniLimits InstanceTypeLimits
+	describeInstanceTypesInput := &ec2.DescribeInstanceTypesInput{InstanceTypes: []*string{aws.String(cache.instanceType)}}
+	output, err := cache.ec2SVC.DescribeInstanceTypesWithContext(context.Background(), describeInstanceTypesInput)
+	if err != nil || len(output.InstanceTypes) != 1 {
+		log.Errorf("", err)
```
M00nF1sh, line 1266: why is the format empty.

pkg/awsutils/awsutils.go
```diff
@@ -1249,55 +1252,58 @@ func (cache *EC2InstanceMetadataCache) AllocIPAddress(eniID string) error {
...
+		return errors.New(fmt.Sprintf("Failed calling DescribeInstanceTypes for `%s`: %v", cache.instanceType, err))
+	}
+	info := output.InstanceTypes[0]
+	// Ignore any missing values
+	instanceType := aws.StringValue(info.InstanceType)
+	eniLimit := int(aws.Int64Value(info.NetworkInfo.MaximumNetworkInterfaces))
+	ipv4Limit := int(aws.Int64Value(info.NetworkInfo.Ipv4AddressesPerInterface))
+	hypervisorType := aws.StringValue(info.Hypervisor)
+	//Not checking for empty hypervisorType since have seen certain instances not getting this filled.
```
M00nF1sh, line 1275: is our code handle empty hypervisorType properly during usage?

jayanthvn: Yes, there are certain instances where we are not getting hypervisorType - https://github.com/aws/amazon-vpc-cni-k8s/blob/8cb5dc2815f05818303b45ff8a55a68df890e6b7/pkg/awsutils/vpc_ip_resource_limit.go#L26. So I am checking not != nitro - https://github.com/aws/amazon-vpc-cni-k8s/blob/8cb5dc2815f05818303b45ff8a55a68df890e6b7/pkg/ipamd/ipamd.go#L370

pkg/awsutils/awsutils.go
```diff
@@ -185,6 +185,9 @@ type APIs interface {
 
 	// GetInstanceID returns the instance ID
 	GetInstanceID() string
+
+	// Verify if the InstanceNetworkingLimits has the ENI limits else make EC2 call to fill cache.
+	VerifyElseFetchLimitsFromEC2() error
```
M00nF1sh, line 190: this name feels a bit odd..
i'm not a naming expert either, but how about `FetchInstanceTypeLimits` or `EnsureInstanceTypeLimits` 😄

jayanthvn: Yeah sure :)

M00nF1sh (approved):
/lgtm

PR #1622, opened 2021-09-15
Fix ipamd test

vikasmb:
- Removed duplicate function definition in the package
- Removed invalid test case of warm_target_prefix = 0 with PD enabled

**What type of PR is this?**
Bug

Removed duplicate function definition in the package

Removed duplicate function definition in the package

Executed integ test in ipamd package

**Automation added to e2e**:

No

No
```release-note

```


jayanthvn (approved):
lgtm

PR #1628, opened 2021-09-20
Update CNI Manifests

achevuru:
**What type of PR is this?**
feature

Adds support for new env variables that allow an user to choose the IP address mode for VPC CNI.

Adds support for new env variables that allow an user to choose the IP address mode for VPC CNI.


jayanthvn (changes requested):
We will have to add it here too - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/master/manifests.jsonnet#L179

jayanthvn (approved):
lgtm

PR #1631, opened 2021-09-22
Handle delays tied to V6 interfaces

achevuru:
**What type of PR is this?**
bug

V6 addresses assigned to an interface might take a while before they transition from `tentative` state to `stable` state as all addresses need to go through Duplicate Address Detection (DAD). PR introduces a check to make sure the address is in `stable` state before CNI returns.

Verified that there is no packet loss observed  right after pod boot-up due to the issue documented above.

### DIFF
--- a/cmd/routed-eni-cni-plugin/driver/driver.go
+++ b/cmd/routed-eni-cni-plugin/driver/driver.go
@@ -19,6 +19,7 @@ import (
 	"net"
 	"os"
 	"syscall"
+	"time"
 
 	"github.com/pkg/errors"
 	"golang.org/x/sys/unix"
@@ -43,6 +44,8 @@ const (
 	fromContainerRulePriority = 1536
 	// Main routing table number
 	mainRouteTable = unix.RT_TABLE_MAIN
+
+	WAIT_INTERVAL = 50 * time.Millisecond
 )
 
 // NetworkAPIs defines network API calls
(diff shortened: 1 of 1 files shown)


cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -43,6 +44,8 @@ const (
 	fromContainerRulePriority = 1536
 	// Main routing table number
 	mainRouteTable = unix.RT_TABLE_MAIN
+
+	WAIT_INTERVAL = 50 * time.Millisecond
```
sramabad1, line 48: Good find ! Not sure if we plan to support anything beyond AL2 for initial cut.  But it might be worth checking if this delay holds good in other distributions as well. 

Side note : Going forward we will have multiple eth attachments on single pod with 5G to separate out different flows.  Having this delay as configurable option would help until we characterize the actual number for different use cases.

achevuru: So, the total wait time is actually `10s`. WAIT_INTERVAL` is essentially how long we wait before checking the status again. I'm assuming `10s` might be long enough and the function in `ip` package that most of the CNI plugins rely on is capping it @`10s` as well. I see that it usually takes between 1-2s in my testing but if we do run in to a specific requirement/use-case, I guess we can definitely consider making the upper bound configurable.

anguslees: Fwiw, I don't think it needs to be configurable - we should be able to just 'wait long enough' for every use case, and there's no benefit from timing out and aborting aggressively.

Re other distros: It would be odd to pick something wildly different from the current Linux kernel default values, and I expect the delay will always be around the few-seconds mark.  One alternative here is that we either use 'optimistic DAD' which allows userspace to use the address for _some_ purposes while it is still tentative.  A better alternative is to just disable DAD altogether on veth interfaces, because we control both ends anyway so there are no surprises here.  Meh, at best it gains 1-2s, and we can come back to this later.  Even if we disable DAD on veth, we're still going to want this function at some point for "real" network interfaces (eg: trunk, EFA, ENI+ipvlan).

We could also remove the above timer by using netlink events rather than polling (see [AddrSubscribe](https://github.com/vishvananda/netlink/blob/53455394aa2b7ada7a308d0328747acdd05e91d5/addr_linux.go#L296-L300)).  Again, meh, we can come back to this if this 50ms poll ever becomes an issue.

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -216,6 +225,48 @@ func (createVethContext *createVethPairContext) run(hostNS ns.NetNS) error {
...
+// If any addresses are still tentative after timeout seconds, then error.
+func WaitForAddressesToBeStable(ifName string, timeout int) error {
+	link, err := netlink.LinkByName(ifName)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve link: %v", err)
+	}
+
+	deadline := time.Now().Add(time.Duration(timeout) * time.Second)
+	for {
+		addrs, err := netlink.AddrList(link, netlink.FAMILY_ALL)
```
anguslees, line 240: I think we only want this for v6, so we may as well use that to our advantage here.
```suggestion
		addrs, err := netlink.AddrList(link, netlink.FAMILY_V6)
```

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -216,6 +225,48 @@ func (createVethContext *createVethPairContext) run(hostNS ns.NetNS) error {
...
+		if err != nil {
+			return fmt.Errorf("could not list addresses: %v", err)
+		}
+
+		if len(addrs) == 0 {
+			return nil
+		}
+
```
anguslees, line 248: I think the loop below already does the right thing for empty-array

```suggestion
```

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -208,6 +211,12 @@ func (createVethContext *createVethPairContext) run(hostNS ns.NetNS) error {
 		return errors.Wrap(err, "setup NS network: failed to add static ARP")
 	}
 
+	if createVethContext.v6Addr != nil && createVethContext.v6Addr.IP.To16() != nil {
+		if err := WaitForAddressesToBeStable(createVethContext.contVethName, 10); err != nil {
```
anguslees, line 215: 1. It's weird that you have WAIT_INTERVAL as a constant above, and then hardcode '10' here, even though this '10' is much more important :stuck_out_tongue: 
2. Don't just use `10` here, instead this function should take an arg of `time.Duration`.  Getting confused about time units (seconds? milliseconds? hours?) is a really common bug, so make that impossible:

```
const DADTimeout = 10 * time.Second

// ... here
		if err := WaitForAddressesToBeStable(createVethContext.contVethName, DADTimeout); err != nil {
		
// and below:
func WaitForAddressesToBeStable(ifName string, timeout time.Duration) error {
        deadline := time.Now().Add(timeout)
        // ...
}
```

anguslees:
(nice, code style comments only)

cmd/routed-eni-cni-plugin/driver/driver.go
achevuru: Yeah, fair enough. Will make the change.

anguslees (approved):

PR #1641, opened 2021-09-28
Knob to disable/enable leaked eni cleanup

jayanthvn:
**What type of PR is this?**
Enhancement

N/A

This adds 2 additional tags for ENIs allocated by aws-node -

```
    1. kubernetes.io/cluster/<cluster-name>: owned
    2. eks:eni:owner: amazon-vpc-cni
```
Also added a knob to disable leaked ENI collection.

n/a

Yes

UTs
```
PASS
ok  	github.com/aws/amazon-vpc-cni-k8s/pkg/utils/retry	0.003s	coverage: 100.0% of statements
?   	github.com/aws/amazon-vpc-cni-k8s/pkg/utils/ttime	[no test files]
?   	github.com/aws/amazon-vpc-cni-k8s/pkg/utils/ttime/mocks	[no test files]
?   	github.com/aws/amazon-vpc-cni-k8s/pkg/version	[no test files]
```
**Automation added to e2e**:

Will add as a follow up PR

No

No

2 additional tags will be created on ENIs allocated by aws-node
```release-note

```


pkg/awsutils/awsutils.go
```diff
@@ -411,7 +416,7 @@ func New(useCustomNetworking, disableENIProvisioning, v4Enabled, v6Enabled bool)
 	}
 
 	// Clean up leaked ENIs in the background
-	if !disableENIProvisioning {
+	if !disableENIProvisioning && !disableLeakedENICollection {
```
achevuru, line 419: Thinking about it..Why don't we disable this for v6 completely since v6 support in PD mode will not be creating any additional ENIs? We only have/need Primary ENI in v6 PD mode..

jayanthvn: Yeah we can. Only catch is if they moved from v4 to v6 cluster and there are leaked ENIs. Maybe should be fine?

achevuru: So, v6 cluster will have to be a brand new cluster and more importantly the proposed IPv6 IAM policy for an IPv6 cluster will not have permissions to `delete` an ENI, so it is not going to be of any use.

achevuru: Sure, we might but we can probably remove the v6 check at that point along with v6 specific checks we placed around custom nw code. Problem I see with this right now is the `ipamd` logs might be filled with 403s(`Unauthorized`) every hr when the `aws-node` pod attempts to delete a leaked ENI.

jayanthvn: Yeah makes sense to disable it for now until we have actual support for custom networking

pkg/awsutils/awsutils.go
```diff
@@ -54,6 +54,11 @@ const (
 	eniClusterTagKey        = "cluster.k8s.amazonaws.com/name"
 	additionalEniTagsEnvVar = "ADDITIONAL_ENI_TAGS"
 	reservedTagKeyPrefix    = "k8s.amazonaws.com"
+	clusterNameTagKeyFormat = "kubernetes.io/cluster/%s"
```
achevuru, line 57: We already add a `cluster name` tag - https://github.com/aws/amazon-vpc-cni-k8s/blob/88c12231b93a04cbe43eef59f5adea0b3d38ade4/pkg/awsutils/awsutils.go#L837. We use this for our Cluster specific IAM policy right now. MAO will also populate this env variable by default going forward. Do we still need another tag for cluster name?

jayanthvn:
/not stale. This will be needed once we move cleanup logic to VPC RC.

jayanthvn:
@M00nF1sh - As part of this PR, I will remove the tagging and have just the disable logic. Once we have the final design we can add the tagging..

jdn5126:
Discarding in favor of https://github.com/aws/amazon-vpc-cni-k8s/pull/2370

PR #1643, opened 2021-09-29
Move IPv6 settings to init container

achevuru:
**What type of PR is this?**
bug/cleanup

`aws-node` pod is running in to permission issues with certain AMIs while trying to configure IPv6 settings with `NET_ADMIN` permissions. We had to enable `privileged` mode to get around that. However, we don't want to enable `privileged` mode for `aws-node`.`aws-vpc-cni-init` container currently operates in privileged mode, so moving the corresponding IPv6 config settings over to init container.

Also, modified a metric name to make it generic to both IPv4 and IPv6 CIDRs.

Validated that `aws-node` no longer runs in to permission issues during boot-up and also validated basic traffic flows in IPv6 mode.

Yes, `aws-vpc-cni-init` container will now have a new env variable exposed - `ENABLE_IPv6`. Variable has to be set to `true` in IPv6 mode.


jayanthvn (approved):
lgtm

PR #1651, opened 2021-10-01
Make error count granular

### LINKED ISSUE #1636: InstanceMetadata awsApiErr elevated in cni 1.9
**Environment**:
- Kubernetes version (use `kubectl version`): 1.19
- CNI Version: 1.9
- OS (e.g: `cat /etc/os-release`): 
NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
HOME_URL="https://amazonlinux.com/"
- Kernel (e.g. `uname -a`):
Linux ip-10-0-208-198.ec2.internal 5.4.110-54.189.amzn2.x86_64 #1 SMP Mon Apr 26 21:25:38 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

Since upgrading to CNI version 1.9, we are seeing increased awsApiErr metrics:
awscni_aws_api_error_count{api="GetMetadata",error="EC2MetadataError"} 1290

It seems like this commit https://github.com/aws/amazon-vpc-cni-k8s/pull/1544/files pulled in this https://github.com/aws/aws-sdk-go/pull/3962 which is exposing 404s without retries. 

Interestingly, I would expect to see errors in the ipamd log about this, but grepping for error in these logs doesn't yield any results. Our aws-node pods also seem to be functioning properly. Is there any case where awscni_aws_api_error_count would be incremented despite being retried somewhere in ipamd?

Any help/support would be appreciated! Happy to provide any additional diagnostics that would be helpful!

jayanthvn:
**What type of PR is this?**
Bug

Fixes #1636

Currently we `awsAPIErrInc("GetMetadata", err)` in a common function which fetches from IMDS `GetMetadataWithContext`. We query for prefixes and secondary IPs but with secondary IP mode get prefixes will return 404 which is expected but the common functions logs the events as an error. Instead moving the logic to make the error count more granular. With this PR we will know the exact field in IMDS which is causing the error rather than just getMetaData which isnt sufficient for debugging.

Yes

No errors incrementing.

Screen Shot 2021-10-01 at 3 27 35 PM

**Automation added to e2e**:

n/a

no

no

no
```release-note

```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -363,7 +363,6 @@ func (i instrumentedIMDS) GetMetadataWithContext(ctx context.Context, p string) 
 	awsAPILatency.WithLabelValues("GetMetadata", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(duration)
 
 	if err != nil {
-		awsAPIErrInc("GetMetadata", err)
 		return "", newIMDSRequestError(p, err)
 	}
 
@@ -429,6 +428,7 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context) 
 	// retrieve availability-zone
 	cache.availabilityZone, err = cache.imds.GetAZ(ctx)
 	if err != nil {
+		awsAPIErrInc("GetAZ", err)
 		return err
 	}
 	log.Debugf("Found availability zone: %s ", cache.availabilityZone)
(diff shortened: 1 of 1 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -563,6 +570,7 @@ func (cache *EC2InstanceMetadataCache) GetAttachedENIs() (eniList []ENIMetadata,
 	// retrieve number of interfaces
 	macs, err := cache.imds.GetMACs(ctx)
 	if err != nil {
+		awsAPIErrInc("GetMACS", err)
```
achevuru, line 573: nit: `GetMACS` -> `GetMACs`

jayanthvn: Yeah will update

achevuru (approved):

PR #1675, opened 2021-10-08
Add readiness/liveness timeout and cleanup configs

jayanthvn:
**What type of PR is this?**
feature + cleanup

1. Random livenessProbe failures for aws-node with K8s 1.21. This is because of the bug fix in k8s 1.20: Fixing Kubelet Exec Probe Timeouts and KEP-1972: kubelet exec probe timeouts. aws-node has default timeout of 1s. Sometimes this is too short causing it to fail and pods to restart.
2. This timeout may not be suitable for all use cases so this can be configured using helm by setting `livenessProbeTimeoutSeconds` and `readinessProbeTimeoutSeconds`.
3. The configs in config/master was generated by manifests.jsonnet. Going forward the jsonnet will be deprecated and the configs will be generated by helm. The new config files in this PR are generated by running `make generate-cni-yamls`.

Increase the liveness and readiness timeouts
Cleanup the way manifests are generated to keep the manifests in sync with helm generated manifests.

#1425 

Yes

**Automation added to e2e**:

N/A

No

No

No

```release-note

```

### DIFF
--- a/charts/aws-vpc-cni/values.yaml
+++ b/charts/aws-vpc-cni/values.yaml
@@ -51,6 +51,8 @@ env:
   WARM_ENI_TARGET: "1"
   WARM_PREFIX_TARGET: "1"
   DISABLE_NETWORK_RESOURCE_PROVISIONING: "false"
+  ENABLE_IPv4: "true"
+  ENABLE_IPv6: "false"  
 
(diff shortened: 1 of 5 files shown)


emillg:
@jayanthvn How to make this working with Terraform "aws_eks_addon" resource?

Issue #1700, opened 2021-10-20
Count addresses in cooldown against used ones in datastoreTargetState()


veshij:
datastoreTargetState() calculates `short` value of ip addresses not taking into account addresses in cooldown.
On nodes with a high churn rate it leads to failures during address allocation (note 2 addresses are both "available" and in "cooldown").
```
{"level":"debug","ts":"2021-10-20T18:32:16.219Z","caller":"ipamd/ipamd.go:1214","msg":"Current warm IP stats: target: 2, total: 18, assigned: 16, available: 2, cooldown: 2, short: 0, over 0"}
{"level":"info","ts":"2021-10-20T18:32:18.997Z","caller":"rpc/rpc.pb.go:713","msg":"Received AddNetwork for NS /var/run/netns/cni-8497daaf-cab2-7929-ff1c-53f97a295633, Sandbox 2324678184bfbb2d12122195a190c25744cb9cdc195ae0363a63c5f3c98d21f9, ifname eth0"}
{"level":"debug","ts":"2021-10-20T18:32:18.997Z","caller":"rpc/rpc.pb.go:713","msg":"AddNetworkRequest: K8S_POD_NAME:\"1db8f1952a05cac8-wpnnj\"  K8S_POD_NAMESPACE:\"blackbox\"  K8S_POD_INFRA_CONTAINER_ID:\"2324678184bfbb2d12122195a190c25744cb9cdc195ae0363a63c5f3c98d21f9\"  ContainerID:\"2324678184bfbb2d121221
{"level":"debug","ts":"2021-10-20T18:32:18.997Z","caller":"datastore/data_store.go:680","msg":"AssignIPv4Address: IP address pool stats: total: 18, assigned 16"}
{"level":"debug","ts":"2021-10-20T18:32:18.997Z","caller":"datastore/data_store.go:757","msg":"Get free IP from prefix failed no free IP available in the prefix - 10.244.56.173/ffffffff"}
{"level":"debug","ts":"2021-10-20T18:32:18.997Z","caller":"datastore/data_store.go:680","msg":"Unable to get IP address from CIDR: no free IP available in the prefix - 10.244.56.173/ffffffff"}
...
{"level":"debug","ts":"2021-10-20T18:32:18.997Z","caller":"datastore/data_store.go:680","msg":"AssignPodIPv4Address: ENI eni-0f2c931402cffbd00 does not have available addresses"}
[...]

PR #1703, opened 2021-10-21
Delayed unused IP address cleanup.

veshij:
**What type of PR is this?**: feature
**Which issue does this PR fix**: https://github.com/aws/amazon-vpc-cni-k8s/issues/1700
**What does this PR do / Why do we need it**: TODO
**Testing done on this change**: Y
**Automation added to e2e**: No
**Will this break upgrades or downgrades. Has updating a running cluster been tested?**: No/No
**Does this PR introduce any user-facing change?**: N
```release-note
Free IP addresses are released with a 10-minute delay to avoid unnecessary churn on loaded nodes.
```

### DIFF
--- a/pkg/ipamd/datastore/data_store.go
+++ b/pkg/ipamd/datastore/data_store.go
@@ -34,6 +34,9 @@ const (
 	// in addressCoolingPeriod
 	addressCoolingPeriod = 30 * time.Second
 
+	// addressDelayedReleasePeriod is used to release unused IP after a period of inactivity.
+	addressDelayedReleasePeriod = 10 * time.Minute
+
 	// DuplicatedENIError is an error when caller tries to add an duplicate ENI to data store
 	DuplicatedENIError = "data store: duplicate ENI"
 
@@ -244,10 +247,10 @@ func (cidr *CidrInfo) AssignedIPAddressesInCidr() int {
 	}
 	return count
 }
-
 type CidrStats struct {
 	AssignedIPs int
 	CooldownIPs int
+	DelayedReleaseIPs int
 }
 
 // Gets number of assigned IPs and the IPs in cooldown from a given CIDR
@@ -258,6 +261,8 @@ func (cidr *CidrInfo) GetIPStatsFromCidr() CidrStats {
 			stats.AssignedIPs++
 		} else if addr.inCoolingPeriod() {
 			stats.CooldownIPs++
+		} else if addr.inDelayedReleasePeriod() {
+			stats.DelayedReleaseIPs++
 		}
 	}
 	return stats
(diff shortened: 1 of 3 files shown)


cgchinmay:
@veshij Could you update your branch. Just started reviewing it. Thanks

pkg/ipamd/datastore/data_store.go
```diff
@@ -34,6 +34,9 @@ const (
 	// in addressCoolingPeriod
 	addressCoolingPeriod = 30 * time.Second
 
+	// addressDelayedReleasePeriod is used to release unused IP after a period of inactivity.
+	addressDelayedReleasePeriod = 10 * time.Minute
```
jayanthvn, line 38: Still reviewing it, but on a high level this should be configurable.

jayanthvn:
Thanks for the changes. We will provide review comments and given this is in the critical IP allocation path, we would also like to test it on our cluster before merging. Currently we are working on a release and holidays are coming up so will keep you updated once we start with the testing for this PR.

pkg/ipamd/datastore/data_store.go
veshij: Sure, I can make it configurable.
afair there was a discussion in related issue that we don't want to introduce new env variables. Is it ok to expose it via env var with some safe default value?

achevuru:
I have a general comment/observation. Why can't we achieve this with WARM and MIN IP targets? Understand we want to avoid unnecessary assign/detach IP calls in a high churn env but a user can greatly minimize that by configuring WARM and MIN IP values based on their expected churn/use case, isn't it? Now, holding a free IP for 10 mins by default might be too conservative as we're taking these IPs from VPC IP pool and they are not just unique at cluster level.

veshij:
> I have a general comment/observation. Why can't we achieve this with WARM and MIN IP targets? Understand we want to avoid unnecessary assign/detach IP calls in a high churn env but a user can greatly minimize that by configuring WARM and MIN IP values based on their expected churn/use case, isn't it? Now, holding a free IP for 10 mins by default might be too conservative as we're taking these IPs from VPC IP pool and they are not just unique at cluster level.

* cooldown addresses need to be accounted as "busy" since those can't be assigned to pods by request. If we are making this change CNI will start allocating additional addresses for short period of time while recently released addresses are in cooldown unnecessary increasing  address churn and api requests.
* WARM/MIN ip target logic works perfectly fine in more or less stable environments when number of pods don't change over time and  targets are chosen correctly. Otherwise there are two options: either nodes are "over-provisioned" and many addresses are not used in the cluster or we have to wait each time pod is launched to allocate an address.

10 minutes limit can be safely decreased, it's just a timeout which we found to work great in our environment.
I can make it low by default (2*cooldown?) and add an environment variable to configure it.

veshij:
@jayanthvn @achevuru what do you think?

jayanthvn:
@veshij - We both are on PTO until next week end. We will get back to you on the week of 17th. Sorry for the delay.

Reg the environment variable, yes it would be good to have. If environment variable is "not configured" or "0" then we can fall back to existing design where WARM targets decide the number of IPs in the pool else if set then released IPs will stay in pool for "cooldown period + delayed release period".

veshij:
Sorry for the quite long delay.

> Reg the environment variable, yes it would be good to have. If environment variable is "not configured" or "0" then we can fall back to existing design where WARM targets decide the number of IPs in the pool else if set then released IPs will stay in pool for "cooldown period + delayed release period".
@jayanthvn I introduced the env variable, updated code to use it.
I defaulted it to 60 seconds (so address will spend 30s in "cooldown" and another 30s in delayed release state). In our environment even conservative 30s gives us quite good address reuse.
Please let me know if you'd prefer to default to 0.
I will update corresponding docs if this approach makes sense to you.
Thank you!

jayanthvn:
@veshij - Will keep it open.

jayanthvn:
/not stale. We need to continue discussion on this.

PR #1714, opened 2021-10-28
Renaming aws-calico charts to aws-tigera-operator

haouc:
**What type of PR is this?**
We should rename this Calico charts in VPC CNI repo to `aws-tigera-operator `which will be synced periodically with eks-charts new charts folder `aws-tigera-operator`.

Making Calico charts naming aligned with upstream and eks-charts

We are actually providing tigera operator now.  

**Automation added to e2e**:

**Does this PR introduce any user-facing change?**: no

```release-note

```

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -38,7 +38,7 @@ Test case added to lib/integration.sh
 If no, create an issue with enhancement/testing label
 -->
 
-** Will this PR introduce any new dependencies?**:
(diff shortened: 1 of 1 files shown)


jayanthvn (changes requested):
Readme needs to be updated - https://github.com/aws/amazon-vpc-cni-k8s/tree/master/charts/aws-calico 

Also here - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/generate-cni-yaml.sh#L122-L138

Issue #1287, opened 2020-11-11
Support Pod IRSA for CNI metrics helper


nithu0115:
**What would you like to be added**:
Support Pod IRSA feature for CNI metrics helper. 

Today, we rely heavily on getting region and EC2 instance ID information from EC2 metadata. However, when using Pod IRSA along with dropping iptables to reach EC2 IMDS (169.254.169.254), we will not be able to get this information and the Pod will fail to establish/create a session. 

* For region information: Add new environment variable to list. Something like AWS_REGION. However, starting 1.18+ Pod IRSA inject AWS_REGION variable into the Pod environment which will solve this issue.
* The major problem is with EKS cluster name as cluster names are found by querying EC2 instance ID tags which requires us to get EC2 instance ID from EC2 meta-data. 

CC: @jayanthvn

avoidik:
potential option here might be mapping of needed info to node or pod labels and then consume either using downward-api or direct query to Kubernetes API

groodt:
I've been tracking 3 important AWS addons: `aws-ebs-csi-driver`, `aws-efs-csi-driver` and `amazon-vpc-cni-k8s`.

`aws-ebs-csi-driver` is now operating without hostNetwork, so I imagine for the cni-metrics-helper, similar could be done https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/821#issuecomment-923413504

achevuru:
@groodt `cni-metrics-helper` runs in `kube-system` namespace but doesn't use `hostNetwork`. Are you referring to something else? 

https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/v1.9/cni-metrics-helper.yaml#L86

groodt:
We are currently running: https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/v1.7.2/config/v1.7/cni-metrics-helper.yaml

We have had to patch the Deployment through to `hostNetwork` for it to work. This is on a cluster following [AWS EKS Security Best Practice](https://docs.aws.amazon.com/eks/latest/userguide/best-practices-security.html). We have blocked IMDSv1 entirely. IMDSv2 is only permitted for containers that use `hostNetwork`, which should be limited as far as possible.

[...]

PR #1715, opened 2021-10-28
Fetch Region and CLUSTER_ID information from cni-metrics-helper env

cgchinmay:
**What type of PR is this?**
feature

Allows Customers to use cni-metrics-helper by completely blocking access to IMDS

Removes hard dependency of cni-metrics-helper on IMDS

#1287 

**Automation added to e2e**:
Added Unit test
Manually tested on sample cluster

Tested by disabling [pod-imds ](https://docs.aws.amazon.com/eks/latest/userguide/best-practices-security.html#restrict-ec2-credential-access)
```
"MetadataOptions": {
                        "State": "applied",
                        "HttpTokens": "required",
                        "HttpPutResponseHopLimit": 1,
                        "HttpEndpoint": "enabled"
                    },
```
No

Yes, if IMDS access needs to be blocked

Yes


cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -317,7 +317,9 @@ func produceCloudWatchMetrics(t metricsTarget, families map[string]*dto.MetricFa
 	for key, family := range families {
 		convertMetrics := convertDef[key]
 		mType := family.GetType()
+		log.Infof("mType: %v", mType)
```
achevuru: Can we replace `mType` with something that explains itself?

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -326,6 +328,7 @@ func produceCloudWatchMetrics(t metricsTarget, families map[string]*dto.MetricFa
 						Unit:       aws.String(cloudwatch.StandardUnitCount),
 						Value:      aws.Float64(action.data.curSingleDataPoint),
 					}
+					log.Infof("Publishing...")
```
achevuru: Do we need this additional log?

pkg/publisher/publisher.go
```diff
@@ -85,21 +85,31 @@ type cloudWatchPublisher struct {
...
+			clusterID = getClusterID(ec2Client)
+		}
+
+		// Get ec2metadata client
+		ec2MetadataClient := ec2metadatawrapper.New(sess)
 
-	region, err := ec2MetadataClient.Region()
-	if err != nil {
-		return nil, errors.Wrap(err, "publisher: Unable to obtain region")
+		region, err = ec2MetadataClient.Region()
```
achevuru, line 109: Why do we want to do this if we already have `region` available?

cmd/cni-metrics-helper/metrics/metrics.go
cgchinmay: I removed it, by mistake it got pushed

pkg/publisher/publisher.go
cgchinmay: It might not be available if we are not using IRSA. I will update this line to fetch only if we dont have region

achevuru: Yes, similar to the `ClusterID` check you have above.

jayanthvn (changes requested):
Can you also add a section here - https://github.com/aws/amazon-vpc-cni-k8s/tree/master/cmd/cni-metrics-helper. We can give steps on how to enable IRSA for cni-metrics helper and include this - 

```
Case 1: Cx not using IRSA, we need to get region and clusterID using IMDS (existing approach)
 	// Case 2: Cx using IRSA but not specified clusterID, we can still get this info if IMDS is not blocked
 	// Case 3: Cx blocked IMDS access and not using IRSA (which means region == "") OR
 	// not specified clusterID then its a Cx error
```

jayanthvn:
Also as part of this, lets update https://docs.aws.amazon.com/eks/latest/userguide/cni-metrics-helper.html for IRSA support.

config/master/manifests.jsonnet
```diff
@@ -297,40 +297,11 @@ local metricsHelper = {
       {
         apiGroups: [""],
         resources: [
-          "nodes",
```
jayanthvn, line 300: This needs to be updated too - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/charts/cni-metrics-helper/templates/clusterrole.yaml

pkg/publisher/publisher.go
```diff
@@ -85,21 +85,33 @@ type cloudWatchPublisher struct {
...
-	region, err := ec2MetadataClient.Region()
-	if err != nil {
-		return nil, errors.Wrap(err, "publisher: Unable to obtain region")
+		// If Customers have explicitly specified clusterID then skip auto generating it
+		if clusterID == "" {
+			clusterID = getClusterID(ec2Client)
+		}
+
+		// Get ec2metadata client
+		ec2MetadataClient := ec2metadatawrapper.New(sess)
```
achevuru, line 107: Why do this outside of the below `if` loop? We don't need to do this if `region` is already available isn't it

pkg/publisher/publisher.go
```diff
@@ -93,24 +93,25 @@ func New(ctx context.Context, region string, clusterID string) (Publisher, error
 	// Case 3: Cx blocked IMDS access and not using IRSA (which means region == "") OR
 	// not specified clusterID then its a Cx error
 	if region == "" || clusterID == "" {
```
achevuru, line 95: Minor nit around code structure. 

Code is structured as below right now.
`
if region == "" || clusterID == ""
{
   if clusterID == ""{
   }
   
   if region == "" {
   }
}
`
Why do we need the upper level `if region == "" || clusterID == ""`? Instead below will suffice right? Prior to this diff update, we had few additional things we were doing outside `if region == ""` and `if clusterID == ""`blocks but with the below changes, everything is confined to both the individual `if` blocks..

`
 if clusterID == ""{
   }
   
   if region == "" {
   }
`
From what I can see, looks like something we don't really need the upper check.

cgchinmay: Yes, we dont need it. Have Updated the logic

groodt:
Nice! Any ETA on `v1.10.2` being GA?

cgchinmay:
> Nice! Any ETA on `v1.10.2` being GA?

Will soon have one, but will be around mid February. If you want to use this change then we do have private image release. Change the image tag to v1.10.2-rc1 and apply this manifest.

cgchinmay:
> > Change the image tag to v1.10.2-rc1 and apply this manifest.
> 
> I've tried upgrading `amazon-vpc-cni-k8s` to v1.10.2-rc1 and unfortunately aws-node fails to start.
> 
> ```
> {"level":"info","ts":"2022-01-10T11:17:44.242Z","caller":"entrypoint.sh","msg":"Validating env variables ..."}
> {"level":"info","ts":"2022-01-10T11:17:44.339Z","caller":"entrypoint.sh","msg":"Install CNI binaries.."}
> {"level":"info","ts":"2022-01-10T11:17:44.644Z","caller":"entrypoint.sh","msg":"Starting IPAM daemon in the background ... "}
> {"level":"info","ts":"2022-01-10T11:17:44.739Z","caller":"entrypoint.sh","msg":"Checking for IPAM connectivity ... "}
> {"level":"info","ts":"2022-01-10T11:17:47.041Z","caller":"entrypoint.sh","msg":"Retrying waiting for IPAM-D"}
> panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x39 pc=0x55ed00eac7c8]
> 
> goroutine 552 [running]:
> github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd.(*IPAMContext).StartNodeIPPoolManager(0x0)
> 	/go/src/github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd/ipamd.go:633 +0x28
> created by main._main
> 	/go/src/github.com/aws/amazon-vpc-cni-k8s/cmd/aws-k8s-agent/main.go:64 +0x32c
> {"level":"info","ts":"2022-01-10T11:17:49.540Z","caller":"entrypoint.sh","msg":"Retrying waiting for IPAM-D"}
> ```

Add these env vars to your aws-node ds
- name: ENABLE_IPv4
 value: "true"
 - name: ENABLE_IPv6
 value: "false"

Or instead apply the manifest for [v1.10.1](https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/v1.10.1/config/master/aws-k8s-cni.yaml) first and then change the image tag.

groodt:
> Or instead apply the manifest for [v1.10.1](https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/v1.10.1/config/master/aws-k8s-cni.yaml) first and then change the image tag.

Thanks! It worked.

jayanthvn:
@cgchinmay - Can you update your branch? We can finish the review this week.

cmd/cni-metrics-helper/README.md
```diff
@@ -13,6 +13,94 @@ The following diagram shows how `cni-metrics-helper` works in a cluster:
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cni-metrics-helper
+subjects:
+  - kind: ServiceAccount
+    name: <IRSA name>
+    namespace: kube-system
+```
+
+3. Specify this IRSA in the cni-metrics-helper deployment spec alongwith CLUSTER_ID as the metric dimension
```
jayanthvn, line 68: Nit: Can you expand on what can be set for AWS_CLUSTER_ID? Maybe you can add a section in readme like this https://github.com/aws/amazon-vpc-cni-k8s#cluster_name

jayanthvn (approved):
Just one minor nit, rest lgtm.

cmd/cni-metrics-helper/README.md
cgchinmay: Updated. have a look :)

jayanthvn: Looks good, thanks.

https://github.com/cgchinmay/amazon-vpc-cni-k8s/tree/fix_imds_access_cw/cmd/cni-metrics-helper

PR #1726, opened 2021-11-05
entry point script migration to golang

jayanthvn:
**What type of PR is this?**
Enhancement

**Which issue does this PR fix**:
N/A

**What does this PR do / Why do we need it**:
Rewrite the entry point shell scripts in golang

This is first part of the change to update to minimal distro image.

**If an issue # is not available please add repro steps and logs from IPAMD/CNI showing the issue**:
N/A

**Testing done on this change**:

Yes

Test Cases

Startup Aws-node 
```
dev-dsk-varavaj-2b-fdf1da64 % kubectl logs aws-node-55brn -n kube-system
time="2021-11-02T22:12:16Z" level=info msg="Install CNI binaries.."
time="2021-11-02T22:12:16Z" level=info msg="Starting IPAM daemon ... "
time="2021-11-02T22:12:16Z" level=info msg="Checking for IPAM connectivity ... "
time="2021-11-02T22:12:17Z" level=info msg="Copying config file ... "
time="2021-11-02T22:12:17Z" level=info msg="Successfully copied CNI plugin binary and config file."
```

Startup Init-container
```
dev-dsk-varavaj-2b-fdf1da64 % kubectl logs aws-node-8qk9p -n kube-system -c aws-vpc-cni-init
Installed /host/opt/cni/bin/loopback
time="2021-11-02T22:24:23Z" level=info msg="Copying CNI plugin binaries ..."
Installed /host/opt/cni/bin/portmap
Installed /host/opt/cni/bin/bandwidth
Installed /host/opt/cni/bin/aws-cni-support.sh
time="2021-11-02T22:24:23Z" level=info msg="Copied all CNI plugin binaries to /host/opt/cni/bin\n"
time="2021-11-02T22:24:23Z" level=info msg="Found hostIP 192.168.12.6 and primaryMAC 0a:5e:a9:0e:e5:f3"
time="2021-11-02T22:24:23Z" level=info msg="Found primaryIF eth0"
time="2021-11-02T22:24:23Z" level=info msg="Updated entry for 2"
time="2021-11-02T22:24:23Z" level=info msg="Updated entry for 1"
time="2021-11-02T22:24:23Z" level=info msg="CNI init container done"
```

PD is Enable but WARM targets = 0
```
dev-dsk-varavaj-2b-fdf1da64 % kubectl logs aws-node-8nx4j -n kube-system
[...]

PR #1727, opened 2021-11-05
Use IMDSv2 token when fetching node ip in entrypoint

chlunde:
**What type of PR is this?**
bug

Use IMDSv2/tokens.

This is required in hardened setups (as recommended by the AWS
documentation)

Without the change, the CNI driver did not become ready on the master
branch.

This has been tested on nodes with the following launch template
options:

  MetadataOptions.HttpTokens required
  MetadataOptions.HttpPutResponseHopLimit 1

**Automation added to e2e**:

N/A, I can create an issue for this (switch e2e nodes to IMDSv2)

No, it worked with 1.9.3

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -125,12 +125,14 @@ wait_for_ipam() {
 get_node_primary_v4_address() {
     while :
     do
-        NODE_IP=$(curl http://169.254.169.254/latest/meta-data/local-ipv4)
+        token=$(curl -Ss -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 60")
+        NODE_IP=$(curl -H "X-aws-ec2-metadata-token: $token" -Ss http://169.254.169.254/latest/meta-data/local-ipv4)
         if [[ "${NODE_IP}" != "" ]]; then
             return 0
         fi
         # We sleep for 1 second between each retry
         sleep 1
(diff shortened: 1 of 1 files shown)


jayanthvn:
#1726 will replace this script and also fix this issue.

chlunde:
@jayanthvn cool, for 1.10.0 or .1? I can close this issue then. This works with 1.9.3 at the moment.

jayanthvn:
This will make it to 1.10.1, but do you need it based of 1.9 branch?

chlunde:
@jayanthvn no, I don't need it, but I think it works on 1.9 already? I had not noticed any issue with this on 1.9.3 but when trying out the bottlerocket+SG+stateful set issue I noticed the issue.

Looking at git blame, it seems like this was introduced with #1587

So 1.9.3 works for IMDSv2 customers, release 1.10.0 will not unless fixed.

For us this is not an issue because we're waiting for 1.10.1 anyway, but you might want some kind of fix for 1.10.0 to avoid regressions for customers using IMDSv2 only.

If you want another kind of fix I can close this.

johngmyers:
I confirm this fix works cherrypicked to the release-1.10 branch with kOps IPv4.

jayanthvn:
@chlunde - Yes will look into it. Thanks for bringing it up. 

@johngmyers - Thanks for verifying.

chlunde:
@jayanthvn I see 1.10 was already released so I'll close this

PR #1752, opened 2021-11-15
add support for running canary script in different regions

abhipth:
**What type of PR is this?**
Add region support for the canary script.

**Automation added to e2e**:

### DIFF
--- a/scripts/run-canary-test.sh
+++ b/scripts/run-canary-test.sh
@@ -39,7 +39,7 @@ function wait_for_addon_status() {
   local expected_status=$1
 
   if [ "$expected_status" =  "DELETED" ]; then
-    while $(aws eks describe-addon $ENDPOINT_FLAG --cluster-name "$CLUSTER_NAME" --addon-name $VPC_CNI_ADDON_NAME); do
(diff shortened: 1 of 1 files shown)


vikasmb (approved):
lgtm

orsenthil (approved):
Is there a possiblity to refactor these scripts (in future) so that repetition with this https://github.com/abhipth/amazon-vpc-resource-controller-k8s/blob/d2876abf229695a0be7cd10194f7f3d9f6b23efa/scripts/test/run-canary-test.sh could be avoided?

abhipth:
Yeah that's a good idea. The add-on installation code is duplicate in both places, I am not sure of a way to import scripts across different repos easily. We can possibly do this operation in Go and have a central framework used by all networking repositories. This will also get rid of duplication we have in our testing framework also.

orsenthil:
> I am not sure of a way to import scripts across different repos easily.

Oh. I failed to note the repo difference. I thought, it was only a directory difference.

Usually, the way is keep a separate utils folder for common tasks across repos.

a. Either use git submodule
b. Or hard-code dependency, and clone it before repo specific thing is called.

The current repetition is fine too! ( I didn't notice that they were two difference repos).

PR #1759, opened 2021-11-19
Automation script for running IT 

Shreya027:
**What type of PR is this?**
feature

Automates running of integration tests in the integration-new folder.

- Iterates and runs tests in new integration folder having sub-folders
- Takes an optional flag -f in order to provide FOCUS as a parameter to run selective tests
- Uses a set of ENV variables and derives other variables needed to run tests
- Created a new folder test inside script, all refactored test scripts will be added here in future.

Running script with focus flag as "CANARY"

```

% ./scripts/test/run-integration-tests.sh -f "CANARY"
/Volumes/workplace/go/src/github.com/amazon-vpc-cni-k8s/test/integration-new /Volumes/workplace/go/src/github.com/amazon-vpc-cni-k8s
Running integration test with the following configuration:
KUBE CONFIG: /Users/shrenaik/.kube/eksctl/clusters/test-cluster-ec2
CLUSTER_NAME: test-cluster-ec2
AWS_REGION: us-west-2
OS_OVERRIDE: darwin
FOCUS: CANARY
Running Suite: CNI Pod Networking Suite
=======================================
Random Seed: 1637357115
Will run 6 of 15 specs

STEP: creating test namespace
STEP: getting the node with the node label key eks.amazonaws.com/nodegroup and value 
```
Running script with no focus flag passed

```

% ./scripts/test/run-integration-tests.sh            
/Volumes/workplace/go/src/github.com/amazon-vpc-cni-k8s/test/integration-new /Volumes/workplace/go/src/github.com/amazon-vpc-cni-k8s
Running integration test with the following configuration:
KUBE CONFIG: /Users/shrenaik/.kube/eksctl/clusters/test-cluster-ec2
CLUSTER_NAME: test-cluster-ec2
AWS_REGION: us-west-2
OS_OVERRIDE: darwin
Running Suite: CNI Pod Networking Suite
=======================================
Random Seed: 1637357132
Will run 15 of 15 specs

STEP: creating test namespace
STEP: getting the node with the node label key eks.amazonaws.com/nodegroup and value 
```

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/.github/workflows/cron-test.yml
+++ b/.github/workflows/cron-test.yml
@@ -42,3 +42,7 @@ jobs:
           S3_BUCKET_NAME: ${{ secrets.S3_BUCKET_NAME }}
         run: |
           ./scripts/run-integration-tests.sh
+
+      - name: Run new integration tests
+        run: |
+          ./scripts/test/run-integration-tests.sh 
--- a/.github/workflows/cron-test.yml
+++ b/.github/workflows/cron-test.yml
@@ -42,7 +42,3 @@ jobs:
           S3_BUCKET_NAME: ${{ secrets.S3_BUCKET_NAME }}
         run: |
           ./scripts/run-integration-tests.sh
-
-      - name: Run new integration tests
-        run: |
-          ./scripts/test/run-integration-tests.sh 
--- a/scripts/test/run-integration-tests.sh
+++ b/scripts/test/run-integration-tests.sh
@@ -1,14 +1,18 @@
 #!/bin/bash
 
-set -e
-OS_OVERRIDE=$(go env GOOS)
+set -eo pipefail
+
 pushd ./test/integration-new
 
 echo "Running integration test with the following configuration:
-KUBE CONFIG: $KUBECONFIG
+KUBECONFIG: $KUBECONFIG
 CLUSTER_NAME: $CLUSTER_NAME
-AWS_REGION: $AWS_REGION
-OS_OVERRIDE: $OS_OVERRIDE"
+AWS_REGION: $AWS_REGION"
+
+
+if [[ -n "${ENDPOINT}" ]]; then
+  ENDPOINT_FLAG="--endpoint $ENDPOINT"
+fi
 
 while getopts "f:" arg; do
   case $arg in
@@ -16,20 +20,14 @@ while getopts "f:" arg; do
     echo "FOCUS: $FOCUS";
   esac
 done
-    
 
-if [[ -z "${OS_OVERRIDE}" ]]; then
-    OS_OVERRIDE=linux
-fi
-
-CLUSTER_INFO=$(aws eks describe-cluster --name $CLUSTER_NAME --region $AWS_REGION)
+CLUSTER_INFO=$(aws eks describe-cluster --name $CLUSTER_NAME --region $AWS_REGION $ENDPOINT_FLAG)
 
 VPC_ID=$(echo $CLUSTER_INFO | jq -r '.cluster.resourcesVpcConfig.vpcId')
 SERVICE_ROLE_ARN=$(echo $CLUSTER_INFO | jq -r '.cluster.roleArn')
 ROLE_NAME=${SERVICE_ROLE_ARN##*/}
+TEST_FAILED=false
  
-
-
 START=$SECONDS
 
 for dir in */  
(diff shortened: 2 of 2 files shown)


scripts/test/run-integration-tests.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+set -e
+OS_OVERRIDE=$(go env GOOS)
+pushd ./test/integration-new
```
jayanthvn, line 5: Once we have all the PRs merged, lets replace integration-new -> integration

jayanthvn (commented):
Lets add it in the existing nightly workflow - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/.github/workflows/cron-test.yml along with the existing integration test run.

Once we have successful runs then we can add it here - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/.github/workflows/integration-tests.yml which will be tested for each PR.

scripts/test/run-integration-tests.sh
```diff
@@ -0,0 +1,52 @@
...
+
+echo "Running integration test with the following configuration:
+KUBE CONFIG: $KUBECONFIG
+CLUSTER_NAME: $CLUSTER_NAME
+AWS_REGION: $AWS_REGION
+OS_OVERRIDE: $OS_OVERRIDE"
```
abhipth, line 11: We can remove this variable, as it's not required.

scripts/test/run-integration-tests.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+set -e
+OS_OVERRIDE=$(go env GOOS)
+pushd ./test/integration-new
+
+echo "Running integration test with the following configuration:
```
abhipth, line 7: Could we also expose an ENDPOINT env variable - https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/scripts/test/run-canary-test.sh#L21-L23

scripts/test/run-integration-tests.sh
```diff
@@ -0,0 +1,52 @@
...
+KUBE CONFIG: $KUBECONFIG
+CLUSTER_NAME: $CLUSTER_NAME
+AWS_REGION: $AWS_REGION
+OS_OVERRIDE: $OS_OVERRIDE"
+
+while getopts "f:" arg; do
```
abhipth, line 13: nit: we can get this variable from ENV variable as well like the other ones.

scripts/test/run-integration-tests.sh
```diff
@@ -0,0 +1,52 @@
...
+
+
+START=$SECONDS
+
+for dir in */  
+do
+
+    cd "${dir%*/}"  
+    if [ -n "$FOCUS" ]; then
+        ginkgo --focus="$FOCUS" -v -r -- --cluster-kubeconfig=$KUBECONFIG --cluster-name=$CLUSTER_NAME --aws-region=$AWS_REGION --aws-vpc-id=$VPC_ID
```
abhipth, line 40: with `set -e` if the firs test fails then all the other tests will be skipped as the script will exit immediately. It would be nice if we can allow all suites to run and later return with non zero exit code if any one suite failed. 

Did something similar here - https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/master/scripts/test/test-with-eksctl.sh#L136-L142

abhipth (changes requested):

scripts/test/run-integration-tests.sh
Shreya027: Hi, so I planned on passing it as a flag instead of ENV variable as I thought it would be more dynamic in comparison to other ENV variables that would more likely remain constant for multiple runs of the script. The focus parameter would most likely change for multiple runs, making passing the flag a simpler alternative. Let me know if you think otherwise or think that we should keep everything consistent as ENV.

.github/workflows/cron-test.yml
```diff
@@ -42,3 +42,7 @@ jobs:
           S3_BUCKET_NAME: ${{ secrets.S3_BUCKET_NAME }}
         run: |
           ./scripts/run-integration-tests.sh
+
+      - name: Run new integration tests
+        run: |
+          ./scripts/test/run-integration-tests.sh
```
abhipth, line 48: Will the script work without any environment variables?
Also, I prefer doing this as part of a different PR so it's easy to revert if there are issues.

abhipth (changes requested):

abhipth (approved):

.github/workflows/cron-test.yml
jayanthvn: @Shreya027 - You can test this on your local fork. Lets add the new script/test cases for nightly runs since it has been long pending.

PR #1764, opened 2021-11-23
Workflow to build arm and x86 images

jayanthvn:
**What type of PR is this?**
Workflow

Build multi arch images on PR push

Run `make docker all` on X86 and ARM runners

Yes
https://github.com/jayanthvn/amazon-vpc-cni-k8s/actions/runs/1492994412

**Automation added to e2e**:

Yes

No

No

No

```release-note

```

### DIFF
--- a/.github/workflows/build-multi-arch.yml
+++ b/.github/workflows/build-multi-arch.yml
@@ -1,0 +1,42 @@
+name: Build x64 and ARM64 images 
+
+on:
+  pull_request:
+    branches:
+      - 'master'
+      - 'release*'  
+
+jobs:
+  build_x64:
+    name: Build x64
+    runs-on: [self-hosted, linux, x64]
+    steps:
+
+    - name: Set up Go 1.14
+      uses: actions/setup-go@v2
+      with:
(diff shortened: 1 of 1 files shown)


srini-ram (commented):
/lgtm

.github/workflows/build-multi-arch.yml
```diff
@@ -0,0 +1,42 @@
...
+jobs:
+  build_x64:
+    name: Build x64
+    runs-on: [self-hosted, linux, x64]
+    steps:
+
+    - name: Set up Go 1.14
+      uses: actions/setup-go@v2
+      with:
+        go-version: '1.14.3'
```
orsenthil, line 18: In our make file , we build with go 1.16 

https://github.com/aws/amazon-vpc-cni-k8s/blob/master/Makefile#L64 

Does the worker version go-1.14 have any significance or relevance? Because we do `make docker ...` the targets below.

jayanthvn: I don't think there is any relevance. It is mainly to pull the code into go directory.

PR #1650, opened 2021-10-01
Add error messages in aws-k8s-agent initialization

ryan-joynbio:
**What type of PR is this?**

Feature: Usability

**Which issue does this PR fix**:

N/A

**What does this PR do / Why do we need it**:

Right now if there is an error from any one of these methods, the error never gets presented to the user and debugging is impossible.

**If an issue # is not available please add repro steps and logs from IPAMD/CNI showing the issue**:

N/A

**Testing done on this change**:

N/A

**Automation added to e2e**:

N/A

**Will this break upgrades or downgrades. Has updating a running cluster been tested?**:

No. No.

**Does this change require updates to the CNI daemonset config files to work?**:

No.

**Does this PR introduce any user-facing change?**:

No.

```release-note
Adds error messages during aws-k8s-agent initilization
```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1766, opened 2021-11-24
Do not silently fail in k8s agent initialization

jeromerobert:
**What type of PR is this?**

bug/cleanup

It logs aws-k8s-agent initialisation error. Currently, aws-k8s-agent may fail without any error message in the log file, for example if the API server cannot be found.

*Will this break upgrades or downgrades. Has updating a running cluster been tested?**:

No.

No.

Not a change that deserve inclusion in the release notes IMHO.

### DIFF
--- a/cmd/aws-k8s-agent/main.go
+++ b/cmd/aws-k8s-agent/main.go
@@ -39,17 +39,20 @@ func _main() int {
 	version.RegisterMetric()
 
 	//Check API Server Connectivity
-	if k8sapi.CheckAPIServerConnectivity() != nil {
+	if err := k8sapi.CheckAPIServerConnectivity(); err != nil {
+		log.Errorf("Failed to connect to the API server: %v", err)
 		return 1
 	}
 
 	rawK8SClient, err := k8sapi.CreateKubeClient()
 	if err != nil {
+		log.Errorf("Failed to create the Kubernetes client: %v", err)
 		return 1
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/aws-k8s-agent/main.go
```diff
@@ -39,17 +39,21 @@ func _main() int {
 	version.RegisterMetric()
 
 	//Check API Server Connectivity
-	if k8sapi.CheckAPIServerConnectivity() != nil {
+	err := k8sapi.CheckAPIServerConnectivity()
```
orsenthil, line 42: Doing a direct comparison to `nil` since there is no other return value seems better.

So, this line can reduced to simply adding a `log.Errorf`

jeromerobert: Do you mean this ?

```patch
--- a/cmd/aws-k8s-agent/main.go
+++ b/cmd/aws-k8s-agent/main.go
@@ -40,6 +40,7 @@ func _main() int {
 
        //Check API Server Connectivity
        if k8sapi.CheckAPIServerConnectivity() != nil {
+               log.Errorf("Failed to connect to the API server")
                return 1
        }
```
It would prevent to display the returned error while it contains valuable information.

ChrisMcKee:
Isnt this the same as #1650

jeromerobert:
Oh yes it is, I missed it.

Shreya027:
Closing PR as identical PR made earlier exists as mentioned above.

Issue #1626, opened 2021-09-16
Inadvertent newline in eniconfig.crd.k8s.amazonaws.com resource spec results in misleading error 


nbrational:
**What would you like to be added**:
Better error handling for input data errors in ENIConfig CRD

**Why is this needed**:

When using custom CNI, an accidental null string in securityGroups list results in a failed CreateNetworkInterface with a vague error `Client.InvalidParameterValue: user <AWS_ACCOUNT_ID> does not own a resource`.

Steps to replicate:-
- follow steps on [this](https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html) link
- Use following for ENIConfig (note the null string in securityGroups list)
```
apiVersion: crd.k8s.amazonaws.com/v1alpha1
kind: ENIConfig
metadata: 
  name: ap-southeast-2a
spec: 
  securityGroups: 
    - sg-0c4e4101aaa30c2ff
    - ""
  subnet: subnet-039aec6e2329771e8
```
- IPAMD logs show following error

```
{"level":"error","ts":"2021-03-15T09:58:21.256Z","caller":"ipamd/ipamd.go:627","msg":"Failed to increase pool size due to not able to allocate ENI AllocENI: failed to create ENI: failed to create network interface: InvalidParameterValue: user &lt;ACCOUNT_ID&gt; does not own a resource\n\tstatus code: 400, request id: 345bb7c3-14a3-4f7b-a515-b2b3fc58daf7"}
```

- CloudTrail has a failed CreateNetworkInterface:-
```
"errorCode": "Client.InvalidParameterValue",
    "errorMessage": "user 307862320347 does not own a resource",
```

- Errors are vague and it wasn't easy to identify the input error

jayanthvn:
Thanks for opening the issue. Happy to take PRs :)

PR #1772, opened 2021-12-01
Validate Security Groups before attempt to attach to ENI Config

orsenthil:
**What type of PR is this?**

Bug, Cleanup.

This tries to fix https://github.com/aws/amazon-vpc-cni-k8s/issues/1626

This adds a method to call describe all the security groups, and if an error is returned in the describe, it will fail the operation (return the err) to the high level call and skip Eni creation.

1. Validate SG from enicfg during the node initialization phase,  when custom networking is enabled and cache the validation.

2. When node pool update is called, the uncached SG will be validated and cache will be kept to update.

```
make unit-test

...

PASS
```

**Automation added to e2e**:

```release-note
No
```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -200,6 +200,9 @@ type APIs interface {
 
 	// FetchInstanceTypeLimits Verify if the InstanceNetworkingLimits has the ENI limits else make EC2 call to fill cache.
 	FetchInstanceTypeLimits() error
+
+	//ValidateSecurityGroups Validates a list of security group ids
+	ValidateSecurityGroups(securityGroupIds []string) error
 }
 
 // EC2InstanceMetadataCache caches instance metadata
@@ -490,7 +493,7 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context) 
 	return nil
 }
 
-// RefreshSGIDs retrieves security groups
+// RefreshSGIDs retrieves security groups attached to an interface
 func (cache *EC2InstanceMetadataCache) RefreshSGIDs(mac string) error {
 	ctx := context.TODO()
 
@@ -888,6 +891,18 @@ func (cache *EC2InstanceMetadataCache) TagENI(eniID string, currentTags map[stri
 		log.Debugf("Successfully tagged ENI: %s", eniID)
 		return nil
 	})
+}
+
+func (cache *EC2InstanceMetadataCache) ValidateSecurityGroups(securityGroups []string) error {
+	_, err := cache.ec2SVC.DescribeSecurityGroups(&ec2.DescribeSecurityGroupsInput{
+		GroupIds: aws.StringSlice(securityGroups),
+	})
+
+	if err != nil {
+		return errors.Wrap(err, "Invalid Security Group Id")
+	}
+
+	return nil
 }
 
 // containsPrivateIPAddressLimitExceededError returns whether exceeds ENI's IP address limit
--- a/pkg/awsutils/mocks/awsutils_mocks.go
+++ b/pkg/awsutils/mocks/awsutils_mocks.go
@@ -489,6 +489,20 @@ func (mr *MockAPIsMockRecorder) TagENI(arg0, arg1 interface{}) *gomock.Call {
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "TagENI", reflect.TypeOf((*MockAPIs)(nil).TagENI), arg0, arg1)
 }
 
+// ValidateSecurityGroups mocks base method
+func (m *MockAPIs) ValidateSecurityGroups(arg0 []string) error {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "ValidateSecurityGroups", arg0)
+	ret0, _ := ret[0].(error)
+	return ret0
+}
+
+// ValidateSecurityGroups indicates an expected call of ValidateSecurityGroups
+func (mr *MockAPIsMockRecorder) ValidateSecurityGroups(arg0 interface{}) *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "ValidateSecurityGroups", reflect.TypeOf((*MockAPIs)(nil).ValidateSecurityGroups), arg0)
+}
+
 // WaitForENIAndIPsAttached mocks base method
 func (m *MockAPIs) WaitForENIAndIPsAttached(arg0 string, arg1 int) (awsutils.ENIMetadata, error) {
 	m.ctrl.T.Helper()
--- a/pkg/ec2wrapper/client.go
+++ b/pkg/ec2wrapper/client.go
@@ -36,6 +36,7 @@ type EC2 interface {
 	ModifyNetworkInterfaceAttributeWithContext(ctx aws.Context, input *ec2svc.ModifyNetworkInterfaceAttributeInput, opts ...request.Option) (*ec2svc.ModifyNetworkInterfaceAttributeOutput, error)
 	CreateTagsWithContext(ctx aws.Context, input *ec2svc.CreateTagsInput, opts ...request.Option) (*ec2svc.CreateTagsOutput, error)
 	DescribeNetworkInterfacesPagesWithContext(ctx aws.Context, input *ec2svc.DescribeNetworkInterfacesInput, fn func(*ec2svc.DescribeNetworkInterfacesOutput, bool) bool, opts ...request.Option) error
+	DescribeSecurityGroups(input *ec2svc.DescribeSecurityGroupsInput) (*ec2svc.DescribeSecurityGroupsOutput, error)
 }
 
 // New creates a new EC2 wrapper
--- a/pkg/ec2wrapper/mocks/ec2wrapper_mocks.go
+++ b/pkg/ec2wrapper/mocks/ec2wrapper_mocks.go
@@ -249,6 +249,21 @@ func (mr *MockEC2MockRecorder) DescribeNetworkInterfacesWithContext(arg0, arg1 i
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "DescribeNetworkInterfacesWithContext", reflect.TypeOf((*MockEC2)(nil).DescribeNetworkInterfacesWithContext), varargs...)
 }
 
+// DescribeSecurityGroups mocks base method
+func (m *MockEC2) DescribeSecurityGroups(arg0 *ec2.DescribeSecurityGroupsInput) (*ec2.DescribeSecurityGroupsOutput, error) {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "DescribeSecurityGroups", arg0)
+	ret0, _ := ret[0].(*ec2.DescribeSecurityGroupsOutput)
+	ret1, _ := ret[1].(error)
+	return ret0, ret1
+}
+
+// DescribeSecurityGroups indicates an expected call of DescribeSecurityGroups
+func (mr *MockEC2MockRecorder) DescribeSecurityGroups(arg0 interface{}) *gomock.Call {
+	mr.mock.ctrl.T.Helper()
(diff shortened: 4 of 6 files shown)


jayanthvn:
Need to run `make format`

orsenthil:
@jayanthvn  - yes, will do that.  I will also refactor a bit, and see where to add tests for this. thank you.

orsenthil:
Hi @jayanthvn - this is ready for review.

pkg/awsutils/awsutils.go
```diff
@@ -727,6 +730,19 @@ func (cache *EC2InstanceMetadataCache) awsGetFreeDeviceNumber() (int, error) {
 // AllocENI creates an ENI and attaches it to the instance
 // returns: newly created ENI ID
 func (cache *EC2InstanceMetadataCache) AllocENI(useCustomCfg bool, sg []*string, subnet string) (string, error) {
+	var sgIds []string
+
+	for _, sgId := range sg {
```
achevuru, line 735: Why do we need to do this? I see we're converting them back to []*string in `ValidateSecurityGroups`?

pkg/awsutils/awsutils.go
```diff
@@ -727,6 +730,19 @@ func (cache *EC2InstanceMetadataCache) awsGetFreeDeviceNumber() (int, error) {
...
+		sgIds = append(sgIds, *sgId)
+	}
+
+	err := cache.ValidateSecurityGroups(sgIds)
+
+	if err != nil {
+		log.Errorf("AllocENI: Security Groups are Invalid: %v", sgIds)
```
achevuru, line 742: Can we wrap the error here?

pkg/awsutils/awsutils.go
```diff
@@ -890,6 +906,27 @@ func (cache *EC2InstanceMetadataCache) TagENI(eniID string, currentTags map[stri
...
+
+	_, err := cache.ec2SVC.DescribeSecurityGroups(&ec2.DescribeSecurityGroupsInput{
+		GroupIds: aws.StringSlice(securityGroupIds),
+	})
+
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case "InvalidGroupId.Malformed":
+				fallthrough
```
achevuru, line 919: Curious..Why `fallthrough` here? We're looking for error codes here and that should be an `either/or` scenario rather than an `and`, right? Is this even possible?

pkg/awsutils/awsutils.go
orsenthil: My mistake. Thanks for spotting. Somewhere between refactoring, I had this unnecessary double conversion. I will remove it and it will make it much better.

pkg/awsutils/awsutils.go
orsenthil: Sure.

pkg/awsutils/awsutils.go
orsenthil: I think, I followed through this example documentation on DescribeSecurityGroups - https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/ec2-example-security-groups.html

The reasoning here seems - we care about error with "Invalid Security Group Id" and show the details regarding `Malformed` or `NotFound` as wrapped error.  If the failure was due to something other than Malformed or NotFound, we still show them as wrapped error in Line 924.

Using that documentation example seems fine  to me as the context is present the wrapped error.  If you have an alternate suggestion, please share, I can try it.

orsenthil (commented):
Thanks for the review, Apurup.

pkg/awsutils/awsutils.go
achevuru: So, we're currently wrapping `err` even inside the switch (and not aerr.Message()) . What do you see when you test with an SG Id that has valid format vs an SG Id that is just malformed?

orsenthil: Here is the test output for the scenarios.

```
# Malformed Security Group Error Message

$ go run ec2_describe_security_groups.go foobar
Invalid Security Group: InvalidGroupId.Malformed: Invalid id: "foobar" (expecting "sg-...")
	status code: 400, request id: d8c66d06-ccbc-443a-93fc-734a809d73fb
exit status 1

$ go run ec2_describe_security_groups.go sg-foobar
Invalid Security Group: InvalidGroupId.Malformed: Invalid id: "sg-foobar"
	status code: 400, request id: d367a2fe-f742-40d0-9885-6d07d38089f7
exit status 1

# Invalid Security Group Error Message

go run ec2_describe_security_groups.go sg-0b89b5a84e6b93f51
Invalid Security Group: InvalidGroup.NotFound: The security group 'sg-0b89b5a84e6b93f51' does not exist
	status code: 400, request id: 76efcb2b-cec0-45da-bc07-8fc9002b10d7
exit status 1

# Valid Security Group

$ go run ec2_describe_security_groups.go sg-0b89...
Security Group:
{
  Description: "Allow SSH access",
  GroupId: "sg-0b89...",
  GroupName: "eksctl-senthil-17Nov-nodegroup-standard-workers-remoteAccess",
  [],
  VpcId: "vpc-xxx"
}
```

And it was against this code.

```
 func main() {
     if len(os.Args) < 2 {
         fmt.Println("Security Group ID required\nUsage: %s group_id ...", filepath.Base(os.Args[0]))
         os.Exit(1)
     }
     groupIds := os.Args[1:]

     sess, err := session.NewSession(&aws.Config{
         Region: aws.String("us-west-2")},
     )

     svc := ec2.New(sess)
     result, err := svc.DescribeSecurityGroups(&ec2.DescribeSecurityGroupsInput{
         GroupIds: aws.StringSlice(groupIds),
     })
     if err != nil {
         if aerr, ok := err.(awserr.Error); ok {
             switch aerr.Code() {
             case "InvalidGroupId.Malformed":
                 fallthrough
             case "InvalidGroup.NotFound":
                 e := errors.Wrap(err, "Invalid Security Group")
                 fmt.Println(e)
                 os.Exit(1)
             }
         }
         e := errors.Wrap(err, "Invalid Security Group")
         fmt.Println(e)
         os.Exit(1)
     }

     fmt.Println("Security Group:")
     for _, group := range result.SecurityGroups {
         fmt.Println(group)
     }
 }
```

orsenthil:
Hi @achevuru ; I have addressed the review comments. Please take a look.

pkg/awsutils/awsutils.go
orsenthil: @achevuru  - Here is thought, I can remove the entire switch case itself. I think, it doesn't add value other than while reading the program. I had used it from the documentation example and I think, documentation example motivation was to illustrate the different kinds of errors.   

Having it simply as 

```
     if err != nil {
         e := errors.Wrap(err, "Invalid Security Group")
         fmt.Println(e)
         os.Exit(1)
     }

``` 

will be better. What do you think?

achevuru: Above test o/p makes sense. We're not going to have both the error codes for the same call and yes I think the sample program you pointed to is just showing how to validate all the possible error codes related to Invalid SG Id. I think the above proposed modification will suffice for what we're trying to do here.

orsenthil: I have addressed this change. thank you.

achevuru:
I have a generic comment on when(and where) we can invoke the `ValidateSecurityGroups` from. Right now, this diff is calling it prior to an ENI create. We derive SGs from ENIConfig CRD spec and I'm thinking we can instead do that whenever we derive the ENIConfig CRD for that node. By doing so, we will only do it once per node (or whenever ENIConfig CRD gets updated) as opposed to calling it prior to every ENI create. 

https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/eniconfig/eniconfig.go#L86
https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/eniconfig/eniconfig.go#L120

Will help reduce the no.of describeSG API calls to close to `1` per node.

orsenthil:
> I'm thinking we can instead do that whenever we derive the ENIConfig CRD for that node.

Good idea. I will change it.

jayanthvn:
> > I'm thinking we can instead do that whenever we derive the ENIConfig CRD for that node.
> 
> Good idea. I will change it.

If you are adding it here https://github.com/aws/amazon-vpc-cni-k8s/blob/192f8472bd9524f739ef2b9d65d91a7ce3b04d09/pkg/ipamd/ipamd.go#L830 then it will be called for each ENI create right? We actually cannot reduce the describeSG call for 1 per node, it will always be `max(ENI) - 1`. Am I missing anything?

orsenthil:
I will verify the call graph, @jayanthvn . I placed this based on the assumption to call `before` AllocENI; but wasn't sure if we can optimize it further, until I read Apurup's comment; and wanted to try that.

jayanthvn:
> I will verify the call graph, @jayanthvn . I placed this based on the assumption to call `before` AllocENI; but wasn't sure if we can optimize it further, until I read Apurup's comment; and wanted to try that.

Sounds good. If we hadn't allowed ENIconfig update on runtime then we could have just validated the config on runtime but since we allow editing then we need to check before AllocENI. This will always be max(ENI)-1 calls but if we really want to optimize then one way I can think of is - 
1. validate ENIconfig and cache during nodeInit. -> 1 describeSG call
2. During allocENI, compare the ENiconfig retrieved with cached data and if there is a mismatch i.e, update scenario then revalidate which will lead to one more extra call. -> 1 extra describeSG call for each new update.

achevuru:
It will be much more than `maxENI-1` count as we free and allocate ENIs as pods are created/deleted on the node. Essentially we redo the whole thing every time we create an ENI. I'm suggesting that we validate it the first time we grab the ENIConfig CRD resource spec (once per node) in the above code path (eniconfig.go)  We used to have a controller for eniconfig but since we moved to an On-demand get, we need to cache the SGs tied to the ENIconfig of the node as suggested above and only make a describeSG call if we detect any changes to what we cached.

orsenthil:
Hi @achevuru , and @jayanthvn 
I spent sometime looking at where to best introduce this again.

Calling the validation logic in `(c *IPAMContext) nodeInit()` pkg/ipamd/ipamd.go seems like the place to validate this, *if* we want to call it only once at the Node Initialization. 

Any other places like in enconfig.go does not seem suitable for the following reasons.
- They are all invoked by `	go ipamContext.StartNodeIPPoolManager()` and will called at runtime beyond node initialization.
- Invoking it in `MyENIConfig` doesn't seem suitable as it is still invoked by `StartNodeIPPoolManager` (which has a for-ever loop).

So, no obvious advantage, plus it will involve a _significant refactor_ the codebase because we don't send context in to eniconfig methods. eniconfig.go seems more like config file reader. So, ruling out seems better.

The other place to add this is this location as mentioned above is
https://github.com/aws/amazon-vpc-cni-k8s/blob/eb06491215598f6ef22d8a8e0b85e62825747757/pkg/ipamd/ipamd.go#L534 

This will validate at Node Intialization level, but this will add a limitation. 
These Security Group *wont be* invoked when we update IP pool or monitor IP pool (updateIPPoolIfRequired) - So, this  will be a limitation and potential for tricky problems.  _Please let me know if this correct_

Given these, calling these at `AllocENI` as the current PR seems fine to me. It is invoked in all cases at the right level. It adds a cost, but I am unsure if that can be alleviated by moving the validation logic elsewhere.

jayanthvn:
Hi @orsenthil 

We can add the ENIConfig validator at nodeInit and also we need to cache the validated ENIConfig. 

So next time in tryAllocateENI after this line - https://github.com/aws/amazon-vpc-cni-k8s/blob/192f8472bd9524f739ef2b9d65d91a7ce3b04d09/pkg/ipamd/ipamd.go#L830 we compare the cachedENIConfig with ENIConfig retrieved and if there is a diff then again call ENIConfig validator else we don't need to validate the Config.

achevuru:
@orsenthil I mean we don't need to strictly do it from within `nodeInit` function if that is what you're referring to. By NodeInit, I'm referring to IPAMD init flow in general. Right now, we will call DescribeSG function for every ENI allocation and it is excessive in my opinion. Instead if we can validate them the first time we retrieve the ENIConfig for the node and cache the results, we can then only make additional calls if SG list changes in the ENIConfig CRD resource.

orsenthil:
We are closing this issue as we don't want to add additional API calls during startup.

If we revisit this, we can refer to the code in this PR.

PR #1787, opened 2021-12-09
Workflow file updates

vikasmb:
**What type of PR is this?**
cleanup

- Updated workflow files to use go 1.16 for consistency.
- Changed integ test script to use git commit id and deleted unused code
- Cleans stale docker images on runner hosts

- Updated workflow files to use go 1.16 for consistency.
 - Changed integ test script to use git commit id and deleted unused code
- Cleans stale docker images on runner hosts

**Automation added to e2e**:

No

No

No
```release-note

```

### DIFF
--- a/.github/workflows/build-multi-arch.yml
+++ b/.github/workflows/build-multi-arch.yml
@@ -29,7 +29,7 @@ jobs:
       
   build_arm:
     name: Build ARM64
-    runs-on: [self-hosted, linux, ARM64]
+    runs-on: [self-hosted, linux, arm64]
     steps:
 
     - name: Set up Go 1.16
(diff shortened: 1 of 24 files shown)


vikasmb:
make check-format is failing in unit test workflow. Looking into reason for the same.

.github/workflows/build-multi-arch.yml
```diff
@@ -1,4 +1,4 @@
-name: Build x64 and ARM64 images 
+name: Build x86 and ARM64 images
```
haouc, line 1: Why do we build x86?

vikasmb: I believe existing workflow is set to build for x86 mainly for consistency purpose. In this case, build on x86 is also done in unit-test workflow.

haouc (approved):
lgtm, thanks

Issue #1644, opened 2021-09-29
Container ID inconsistency results in deleted vlan device


JacobHenner:
**What happened**:

On several occasions, we've encountered an issue where pods scheduled on a node start with broken network connectivity and never recover. The containers within the pod will attempt to start repeatedly, but since the applications within depend on network connectivity, the containers exit in failure each time and the pod remains in CrashLoopBackoff.

Each time I've observed this personally, I've noticed it seems to be associated with many pods being scheduled on the node simultaneously. All of the pods scheduled around the same time have the same issue. However, the issue is not permanent - pods scheduled sometime after the issue has occurred start with functional network connectivity, and continue to work as expected.

After some investigation, I believe I've identified the immediate cause of the broken network connectivity. However, I'm unsure of the root cause, and whether the vpc-cni should be preventing the condition from occurring.

Comparing a healthy pod and an unhealthy pod from the same node, I noticed that the unhealthy pod does not have a vlan device (e.g. `vlan.eth.XY@eth1`) on the host, and that the associated routes that would otherwise be in place are missing. There is a veth device for the broken pod (e.g. `vlan7cf3df5c6e2@if3`).

Exploring this further, I found the following in the CNI logs (for a broken pod):

```jsonl
{"level":"info","ts":"2021-09-24T22:52:24.230-0400","caller":"routed-eni-cni-plugin/cni.go:117","msg":"Received CNI add request: ContainerID(30c860ea0116032d24b70be358ce519ca613b2a449a2ffaafd31992c3245359c) Netns(/proc/20683/ns/net) IfName(eth0) Args(IgnoreUnknown=1;K8S_POD_NAMESPACE=redacted;K8S_POD_NAME=redacted-bf4687b75-bsmsm;K8S_POD_INFRA_CONTAINER_ID=30c860ea0116032d24b70be358ce519ca613b2a449a2ffaafd31992c3245359c) Path(/opt/cni/bin) argsStdinData({\"cniVersion\":\"0.3.1\",\"mtu\":\"9001\",\"name\":\"aws-cni\",\"pluginLogFile\":\"/var/log/aws-routed-eni/plugin.log\",\"pluginLogLevel\":\"DEBUG\",\"type\":\"aws-cni\",\"vethPrefix\":\"eni\"})"}

[...]

PR #1792, opened 2021-12-13
Changes for rc image v1.10.2-rc1 [VlanID and cni-metrics-helper]

cgchinmay:
**What type of PR is this?**
feature

#1644 
#1715 

Changes for rc image v1.10.2-rc1

Tested the changes by running e2e and integration tests

**Automation added to e2e**:

No


cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -241,12 +249,15 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 
 	hostInterface := &current.Interface{Name: hostVethName}
 	containerInterface := &current.Interface{Name: args.IfName, Sandbox: args.Netns}
+	vlanInterface := &current.Interface{Name: vlanInterfaceName, Mac: fmt.Sprint(r.PodVlanId)}
+	log.Infof("Using vlanInterface: %v", vlanInterface)
```
srini-ram, line 254: For the final commit, let's explore if the actual vlan interface name can be used  rather than place holder name (const string). This may or may not affect CNI chaining use cases if we have duplicate intf name across SGPP containers.

srini-ram (approved):
/lgtm

cmd/routed-eni-cni-plugin/cni.go
cgchinmay: Yes, I have made a note of TODO changes

Issue #694, opened 2019-10-29
Automate CNI plugin upgrade testing


jaypipes:
Before releases, we manually perform a test of the upgrade process for the CNI plugin:

1. create 3 c3.xlarge node cluster
1. scale pod replica from 0 to 130
1. verify kubectl get deployment
1. upgrade cni to new version
1. verify kubectl get deployment
1. scale replica from 130 to 0
1. verify kubectl get deployment
1. scale replica from 0 to 130 again and verify

We should automate this and gate all PRs on a successful upgrade test

- [ ] Convert above steps into a Bash script
- [x] Add job to CircleCI pipeline, gating for all PRs

mogren:
The current test script doesn't scale up and down.

jayanthvn:
1. Scale up to 130 pods (Ref performance script)
2. Run traffic test from the new integration scripts.
3. Upon success, scale down to 0
4. Upgrade to latest CNI
5. Repeat 1-3 steps.

jayanthvn:
@cgchinmay - Can we include this issue as part of this PR - https://github.com/aws/amazon-vpc-cni-k8s/pull/1863

PR #1795, opened 2021-12-16
CNI IT to test for version upgrade/downgrade 

Shreya027:
**What type of PR is this?**
feature

Adds integration tests for testing upgrade/downgrade using user provided initial and final CNI version. Pod traffic, host networking and service connectivity test are run, first on applying initial addon version and again on applying final addon version to test for successful upgrade/downgrade.

- Takes parameterized input with $CNI_INITIAL_VERSION and $CNI_FINAL_VERSION  provided as follows:
```
ginkgo -v --failOnPending -- \                                                    
 --aws-region=$AWS_REGION \
 --aws-vpc-id=$VPC_ID \
 --ng-name-label-val=$NG_NAME_LABEL_VAL \
 --initial-version=$CNI_INITIAL_VERSION \
 --final-version=$CNI_FINAL_VERSION
```
- Added additional addon wrapper functions in `eks.go` under framework folders that can be re-used for addon related utilities
- Added code for applying/deleting/describing addons in ginkgo tests 
- Added code for running  pod traffic, service connectivity and host networking tests in upgrade testing

#694 

Ran all the 3 test files that gave all tests passed in at least one of the runs:

```     
/amazon-vpc-cni-k8s/test/integration-new/cni-version-testing/cni_version_service_connectivity_test.go:191
STEP: deleting test namespace

Ran 10 of 10 Specs in 1563.561 seconds
SUCCESS! -- 10 Passed | 0 Failed | 0 Pending | 0 Skipped
```
```
amazon-vpc-cni-k8s/test/integration-new/cni-version-testing/cni_version_host_networking_test.go:168
STEP: deleting test namespace

Ran 8 of 8 Specs in 1011.273 seconds
SUCCESS! -- 8 Passed | 0 Failed | 0 Pending | 0 Skipped
```
If PD and other values not correctly configured as master config manifest, the tests might show failures.
**Automation added to e2e**:

No, adds testing code

```release-note

```


cgchinmay:
Do we need a separate folder ? wondering this why this cannot be part of the existing cni folder.

test/integration-new/cni-version-testing/cni_version_host_networking_test.go
```diff
@@ -0,0 +1,316 @@
...
+				Should(BeNumerically(">", 0))
+			Expect(len(interfaceTypeToPodList.PodsOnPrimaryENI)).
+				Should(BeNumerically(">", 0))
+
+			By("generating the pod networking validation input to be passed to tester")
+			input, err := GetPodNetworkingValidationInput(interfaceTypeToPodList).Serialize()
+			Expect(err).NotTo(HaveOccurred())
+
+			By("validating host networking setup is setup correctly")
+			ValidateHostNetworking(NetworkingSetupSucceeds, input)
```
cgchinmay, line 120: This looks like duplicate logic. Is there a way by which we can avoid replicating this same logic in 2 places ?

cgchinmay (commented):
I think we should restructure test files so that we minimize the code duplication.

abhipth:
@Shreya027 thanks for the PR, before I had a couple of questions before I dive deep into the PR
- Is this test intended to test the active production system i.e the published add-ons? We have something similar [here](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/run-canary-test.sh#L99-L119) . But it's good to write it in Ginkgo.
- How does this scale, let's say we want to plug-in and plug-out new tests in upgrade/downgrade scenarios.
- How can we utilize this when doing a new release (an addon that is not yet published).

Shreya027:
Hi Abhinav, I have hopefully addressed your questions below:

- Yes it can be used for published add-ons. Ideally in an add-on release pipeline in the future. For instance, a pipeline that updates the default VPC CNI version for eks clusters
- Yes, it should be easy to scale as long as the existing tests are modular. It should a simple addition to existing test calls as below:

```
Context("when testing host networking on xxxx version", func() {
		HostNetworkingTest()
		PodTrafficTest()
		ServiceConnectivityTest()
		...
		....
	})
```

- I don't think we can utilize this for an addon that is not yet published. Because the code uses Addon Functions that seem to use published add-ons.

cgchinmay:
```
The below test is currently flaky, however it has had 8/8 tests passed in one of the runs
```
Could you elaborate on this ? Why is it flaky ?

test/framework/resources/aws/services/eks.go
```diff
@@ -23,7 +23,10 @@ import (
 type EKS interface {
 	DescribeCluster(clusterName string) (*eks.DescribeClusterOutput, error)
 	CreateAddon(addon string, clusterName string) (*eks.CreateAddonOutput, error)
+	CreateAddonWithVersion(addon string, clusterName string, addonVersion string) (*eks.CreateAddonOutput, error)
```
cgchinmay, line 26: If only change is the version, then I think we should create an AddonInput struct and pass it to the method, instead of creating a new method only for version. That way it would be easy to extend to other methods.

test/integration-new/cni/cni_version_test.go
```diff
@@ -0,0 +1,95 @@
...
+	initialCNIVersion   string
+	finalCNIVersion     string
+)
+
+var _ = Describe("test cluster upgrade/downgrade", func() {
+
+	It("should apply initial addon version successfully", func() {
+		By("getting initial cni version")
+		initialCNIVersion = f.Options.InitialCNIVersion
+		ApplyAddOn(initialCNIVersion)
```
cgchinmay, line 25: We should validate initialCNIVersion and finalCNIVersion that they are not empty.

test/framework/resources/aws/services/eks.go
Shreya027: I created a separate func because we have to add an additional parameter ```aws.String("OVERWRITE")```, for ```ResolveConflicts``` in case a version is specified in the eks.go file. If we have one function, we will have to perform a check there and modify the struct to set```ResolveConflicts``` parameter.

Shreya027:
> ```
> The below test is currently flaky, however it has had 8/8 tests passed in one of the runs
> ```
> 
> Could you elaborate on this ? Why is it flaky ?

By flaky, I mean some of the test errors seem to be transient. They run successfully on test re-runs. However, my code hasn't changed any of the test logic, it just runs the tests by making calls to them on upgrade/downgrade.

test/framework/resources/aws/services/eks.go
```diff
@@ -22,14 +22,24 @@ import (
...
 	DeleteAddon(addon string, clusterName string) (*eks.DeleteAddonOutput, error)
+	DescribeAddonVersions(addon string, k8sVersion string) (*eks.DescribeAddonVersionsOutput, error)
+	DescribeAddon(addon string, clusterName string) (*eks.DescribeAddonOutput, error)
 }
 
 type defaultEKS struct {
 	eksiface.EKSAPI
 }
 
+type CreateAddOnParams struct {
```
cgchinmay, line 35: nit: may be rename to AddOnInput so that it can be extended for Describe, Delete as well if needed.  Also add a comment when why are we having this struct.

test/framework/resources/aws/services/eks.go
```diff
@@ -22,14 +22,24 @@ import (
...
+	_ struct{} `type:"structure"`
+
+	AddonName    string `locationName:"addonName" type:"string" required:"true"`
+	ClusterName  string `location:"uri" locationName:"name" min:"1" type:"string" required:"true"`
+	AddonVersion string `locationName:"addonVersion" type:"string"`
```
cgchinmay, line 40: Add example of Version format in comments. like v1.7.5 or v1.7.5-eksbuild.1

test/integration-new/cni/cni_version_test.go
```diff
@@ -0,0 +1,106 @@
...
+	})
+
+})
+
+func ApplyAddOn(versionName string) {
+
+	By("getting the current addon")
+	describeAddonOutput, err = f.CloudServices.EKS().DescribeAddon("vpc-cni", f.Options.ClusterName)
+	if err == nil {
+		fmt.Printf("By applying addon %s\n", versionName)
```
cgchinmay, line 66: Change it to By instead . remove Print calls

test/integration-new/cni/cni_version_test.go
```diff
@@ -0,0 +1,104 @@
...
+	if err == nil {
+		By("checking if the current addon is same as addon to be applied")
+		if *describeAddonOutput.Addon.AddonVersion != versionName {
+
+			By("deleting the current vpc cni addon ")
+			_, err = f.CloudServices.EKS().DeleteAddon("vpc-cni", f.Options.ClusterName)
+			Expect(err).ToNot(HaveOccurred())
+
+			_, err = f.CloudServices.EKS().DescribeAddon("vpc-cni", f.Options.ClusterName)
+			for err == nil {
```
cgchinmay, line 74: Didnt understand this part. Why are we having a sleep and describing it again ?

test/integration-new/cni/cni_version_test.go
```diff
@@ -0,0 +1,104 @@
...
+				time.Sleep(5 * time.Second)
+				_, err = f.CloudServices.EKS().DescribeAddon("vpc-cni", f.Options.ClusterName)
+
```
cgchinmay, line 77: Didnt understand this part. Why are we having a sleep and describing it again ?

test/integration-new/cni/cni_version_test.go
cgchinmay, line 74: This could lead to an infinite loop.

test/integration-new/cni/cni_version_test.go
```diff
@@ -0,0 +1,104 @@
...
+	} else {
+		By(fmt.Sprintf("By applying addon %s\n", versionName))
+		_, err = f.CloudServices.EKS().CreateAddon(services.AddOnInput{AddonName: "vpc-cni", ClusterName: f.Options.ClusterName, AddonVersion: versionName})
+		Expect(err).ToNot(HaveOccurred())
+	}
+
+	var status = ""
+
+	By("waiting for  addon to be ACTIVE")
+	for status != "ACTIVE" {
```
cgchinmay, line 94: I think this can be extracted to utility function similar to daemonset.go. So this logic can be extended.

test/integration-new/cni/cni_version_test.go
```diff
@@ -0,0 +1,104 @@
...
+		describeAddonOutput, err = f.CloudServices.EKS().DescribeAddon("vpc-cni", f.Options.ClusterName)
+		Expect(err).ToNot(HaveOccurred())
+		status = *describeAddonOutput.Addon.Status
+		time.Sleep(5 * time.Second)
+	}
+
+	//Set the WARM_ENI_TARGET to 0 to prevent all pods being scheduled on secondary ENI
+	k8sUtils.AddEnvVarToDaemonSetAndWaitTillUpdated(f, "aws-node", "kube-system",
```
cgchinmay, line 102: What are the default settings when you apply an addon ?Just want to know why need these env settings ?

test/integration-new/cni/cni_version_test.go
```diff
@@ -0,0 +1,104 @@
...
+		Expect(err).ToNot(HaveOccurred())
+		finalCNIVersion = f.Options.FinalCNIVersion
+		ApplyAddOn(finalCNIVersion)
+
+	})
+
+	Context("when testing host networking on final version", func() {
+		HostNetworkingTest()
+		PodTrafficTest()
+		ServiceConnectivityTest()
```
cgchinmay, line 56: I think these 3 functions are 3 different kinds of tests, so should be separated out in 3 Contexts ? would like to know your thoughts

cgchinmay (commented):
Just 3 comments, rest looks good. Thanks 👍

cgchinmay:
Closing this PR, refer this instead: https://github.com/aws/amazon-vpc-cni-k8s/pull/1861

Issue #1801, opened 2022-01-03
Init container emits non-error messages to stderr


skpy:
**What happened**:
The amazon-vpc-cni-init container invokes [set -x](https://github.com/aws/amazon-vpc-cni-k8s/blob/ec409b0cd124923dc4f369483b4416499ab7bc3b/scripts/init.sh#L3) in `init.sh` which emits output to stderr.  Our log collection and monitoring system (DataDog) sees these messages as errors (because they're coming from stderr, not stdout) and trips our alerting threshold for number of errors per unit of time whenever the init container runs.

I have implemented a log processor in DataDog to forcibly set all `amazon-vpc-cni-init` logs as INFO, which avoids the superfluous alerts at the expense of masking any real errors. I could also modify our existing alerts to exclude amazon-vpc-cni-init, then make a new alert specifically for amazon-vpc-cni-init, if necessary.

It might also be worth adding `-s` to the curl commands ([1](https://github.com/aws/amazon-vpc-cni-k8s/blob/ec409b0cd124923dc4f369483b4416499ab7bc3b/scripts/init.sh#L7) and [2](https://github.com/aws/amazon-vpc-cni-k8s/blob/ec409b0cd124923dc4f369483b4416499ab7bc3b/scripts/init.sh#L15)) to suppress the status output from curl, which also emits to stderr.

**Environment**:
EKS 1.21

- Kubernetes version (use `kubectl version`): `Server Version: version.Info{Major:"1", Minor:"21+", GitVersion:"v1.21.2-eks-06eac09", GitCommit:"5f6d83fe4cb7febb5f4f4e39b3b2b64ebbbe3e97", GitTreeState:"clean", BuildDate:"2021-09-13T14:20:15Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"linux/amd64"}`
- CNI Version `v1.10.1-eksbuild.1`

jaypipes:
@skpy What's up, Skippy? :)

jaypipes:
I agree that it's generally poor Bash hygiene to have any non-error output be directed to `stderr`. We could either do `BASH_XTRACEFD=1` at the start of the init script or we could simply remove the use of `set -x`  entirely (as it's mostly a debug/troubleshooting need and not something that should be in production scripts). @jayanthvn what do you think?

skpy:
[...]

PR #1802, opened 2022-01-03
remove set -x from bash, add -Ss to curl

skpy:
**What type of PR is this?**

cleanup

#1801 

While not catastrophic, it's sub-optimal for the `init.sh` script to emit diagnostic data to stderr.  This PR removes the `-x` from the Bash `set` command.  `-x` is great for debugging and troubleshooting, but probably not necessary in the normal course of operations of this script.

I also added `-Ss` to the two `curl` calls in `init.sh`.  This suppresses the status output going to stderr, but will still emit errors in the event of actual errors.  This is consistent with other `curl` calls within this repo.

N/A

**Automation added to e2e**:
N/A

It ought not.  I have not tested this yet, as I don't currently have sufficient scaffolding to do so.

No

Yes, insofar as messages that had been emitted previously will no longer be emitted.  If someone was expecting / relying on those messages for some purpose, those expectations won't be satisfied.

```release-note

```

### DIFF
--- a/scripts/init.sh
+++ b/scripts/init.sh
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 
-set -euxo pipefail
+set -euo pipefail
 
 get_metadata()
 {
-    TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 60")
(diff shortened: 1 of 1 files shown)


jaypipes (approved):
:+1: thanks Skippy!

orsenthil (approved):
I like the `curl -Ss` addition.

I sometimes find the `set -x` useful while looking at the logs (to see the command that was executed after the various substitutions). I am good with this change.

PR #1813, opened 2022-01-21
Add "ip rule add" command to cni-proposal.md

a2ush:
**What type of PR is this?**
documentation

Update cni-proposal.md document.

Currently, [cni-proposal.md](https://github.com/aws/amazon-vpc-cni-k8s/blob/v1.10.1/docs/cni-proposal.md) indicates the example of the routing rules.
```
# ip rule list
0:	from all lookup local 
512:	from all to 10.0.97.30 lookup main <---------- to Pod's traffic
1025:	not from all to 10.0.0.0/16 lookup main 
1536:	from 10.0.97.30 lookup eni-1 <-------------- from Pod's traffic
```
However, there is no "ip rule add" command in the section of ["CNI Plugin Sequence"](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/docs/cni-proposal.md#cni-plugin-sequence).

I think this command (ip rule add) should be described in this section for introducing the work of amazon-vpc-cni-k8s in more detail.

[Note]
This cni plugin adds routing rules for launched pods.
The 1536 priority rule is not always added. So I think it's OK to describe just the rule of "to Pod's traffic" (512 priority rule) in this example.

https://github.com/aws/amazon-vpc-cni-k8s/blob/104354c45bb210c444bb3317e4199b696cfbb2f9/cmd/routed-eni-cni-plugin/driver/driver.go#L501-L527

N/A

N/A

**Automation added to e2e**:

N/A

N/A

N/A

```release-note
```

### DIFF
--- a/docs/cni-proposal.md
+++ b/docs/cni-proposal.md
@@ -149,11 +149,19 @@ Here are the wiring steps to enable pod to pod communication:
 	ip netns exec ns1 arp -i veth-1c -s 169.254.1.1 <veth-1's mac> /* add static ARP entry for default gateway */
 	```
 
-* On host side, add host route so that incoming Pod's traffic can be routed to Pod.
+* On host side, add host route and routing rule so that incoming Pod's traffic can be routed to Pod.
 
 	```
 	/* Pod's IP address is 20.0.49.215 */
 	ip route add 20.0.49.215/32 dev veth-1 /* add host route */
+	ip rule add from all to 20.0.49.215/32 table main prio 512 /* add routing rule */
+	```
+	Note:
+
+	When Pod's ip address is associated with the secondary ENI, the from-pod rule is also added.
+	For example, if the ip address "20.0.49.215" is associated with eth1, add the following rule.
+	```
+	ip rule add from 20.0.49.215/32 table 2 prio 1536  /* add from-pod rule */
 	```
 
 #### Life of a Pod to Pod Ping Packet


docs/cni-proposal.md
```diff
@@ -149,11 +149,12 @@ Here are the wiring steps to enable pod to pod communication:
 	ip netns exec ns1 arp -i veth-1c -s 169.254.1.1 <veth-1's mac> /* add static ARP entry for default gateway */
 	```
 
-* On host side, add host route so that incoming Pod's traffic can be routed to Pod.
+* On host side, add host route and routing rule so that incoming Pod's traffic can be routed to Pod.
 
 	```
 	/* Pod's IP address is 20.0.49.215 */
 	ip route add 20.0.49.215/32 dev veth-1 /* add host route */
+	ip rule add from all to 20.0.49.215/32 table main prio 512 /* add routing rule */
```
jayanthvn, line 157: 1356 rule is added for from-pod rule, only need it when it is not primary ENI - https://github.com/aws/amazon-vpc-cni-k8s/blob/5bb9811e5b21701a2d119dd58ad49b83017decda/cmd/routed-eni-cni-plugin/driver/driver.go#L325. Maybe you can include both mentioning the difference as primary and secondary ENI.

a2ush: Sorry for delay.
I'll try to include the both messages :)

a2ush: @jayanthvn
I tried it, but how is it?
And also, I modified the example comment along the current behavior. e236a333c313ce4f7396c44426fbcbfa78f4b246

a2ush:
@jayanthvn 
I resolved the conflict and pushed it again because [driver.go](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/cmd/routed-eni-cni-plugin/driver/driver.go) was updated.

Could you please confirm this PR again? Thank you.

jayanthvn (approved):
lgtm

PR #1816, opened 2022-01-24
Updated go version

cgchinmay:
**What type of PR is this?**
feature

NA

Updates go version

No

```release-note

```


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/aws/amazon-vpc-cni-k8s
 
-go 1.14
+go 1.17
```
jayanthvn, line 3: But still the base image is based of 1.16 - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/Makefile#L61

cgchinmay: will update that too

vikasmb: Can you also please upgrade the go version used in workflow files - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/.github/workflows/build-multi-arch.yml#L18 along with this change ? Thanks

docs/troubleshooting.md
```diff
@@ -202,3 +202,120 @@ If you're using v1.10.0, `aws-node` daemonset pod requires IMDSv1 access to obta
...
+check-format: format
+```
+
+4. Run make check-format in your local repo and we see the same error  
+
+```
+make: *** [Makefile:263: format] Error 1
+```
+
+5. Instead of running the make target, lets just run the standalone command. This command uses the FORMAT_FLAGS if available, if not then it uses the '-w' flag instead. So replace the value of the FORMAT_FLAGS and run the equivalent command.
```
jayanthvn, line 250: Do we need all these steps? `make format` should fix the formatting right?

docs/troubleshooting.md
```diff
@@ -202,3 +202,120 @@ If you're using v1.10.0, `aws-node` daemonset pod requires IMDSv1 access to obta
...
+
+## Debugging Unit Test Failures on PR submission
+
+### Run unit tests / Unit-tests (pull_request)
+The above workflow gets triggered for every PR. It setups up a self hosted github runner with ubuntu:latest image and runs the unit tests. 
+To debug failures, click on details and check the phase in which the workflow failed. Expand the failed phase and check for the failed command. Try one of the ways to debug further
+
+- Run the failed command on your machine  
+OR  
+- Provision an EC2 instance with ubuntu:latest image (as used by self hosted runner)
```
vikasmb, line 214: Provision an EC2 instance with ubuntu:latest image (as it is used by github hosted runner used for build and unit test execution.)

jayanthvn:
ARM build is failing - 

```
Step 13/17 : FROM public.ecr.aws/amazonlinux/amazonlinux:1
78
1: Pulling from amazonlinux/amazonlinux
79
no matching manifest for linux/arm64/v8 in the manifest list entries
```

cgchinmay:
> ARM build is failing -
> 
> ```
> Step 13/17 : FROM public.ecr.aws/amazonlinux/amazonlinux:1
> 78
> 1: Pulling from amazonlinux/amazonlinux
> 79
> no matching manifest for linux/arm64/v8 in the manifest list entries
> ```

Oh, taking a look

Issue #1775, opened 2021-12-06
ipamd loop creating and freeing ENI when WARM_ENI_TARGET=0 and no IP is being used on the node


dtnyn:
**What happened**:
When we set `WARM_ENI_TARGET=0 ` for our CNI configuration, we notice that when there is no pod on the node. The ipamd will detect that it met its requirement to request for a new ENI

```
{"level":"debug","ts":"2021-12-02T00:01:40.937Z","caller":"ipamd/ipamd.go:563","msg":"IP pool is too low: available (0) < ENI target (0) * addrsPerENI (14)"}
{"level":"debug","ts":"2021-12-02T00:01:40.937Z","caller":"ipamd/ipamd.go:564","msg":"Starting to increase pool size"}
```

However it will then delete this ENI as it is not in used due to no pod is consuming IP.

This created an issue where the ipamd will continuously create new ENI but then procceed to delete it afterwards

```
➜ cat ipamd.log | ag "(Created a new ENI)|(Successfully freed ENI)" | head -n 10
{"level":"info","ts":"2021-12-02T00:00:11.209Z","caller":"awsutils/awsutils.go:667","msg":"Created a new ENI: eni-05a14ca3e85f7bab2"}
{"level":"info","ts":"2021-12-02T00:01:23.778Z","caller":"awsutils/awsutils.go:850","msg":"Successfully freed ENI: eni-05a14ca3e85f7bab2"}
{"level":"info","ts":"2021-12-02T00:01:41.339Z","caller":"awsutils/awsutils.go:667","msg":"Created a new ENI: eni-0cd69037ff4b1992d"}
{"level":"info","ts":"2021-12-02T00:02:53.433Z","caller":"awsutils/awsutils.go:850","msg":"Successfully freed ENI: eni-0cd69037ff4b1992d"}
{"level":"info","ts":"2021-12-02T00:02:58.844Z","caller":"awsutils/awsutils.go:667","msg":"Created a new ENI: eni-0a9f1da35fb408dbc"}
{"level":"info","ts":"2021-12-02T00:04:14.765Z","caller":"awsutils/awsutils.go:850","msg":"Successfully freed ENI: eni-0a9f1da35fb408dbc"}
{"level":"info","ts":"2021-12-02T00:04:20.190Z","caller":"awsutils/awsutils.go:667","msg":"Created a new ENI: eni-05257e8d3edf05e46"}
[...]

PR #1820, opened 2022-01-25
Ad-hoc ENI when IP address is required and WARM_ENI_TARGET is 0

haouc:
**What type of PR is this?**
bug

#1775 
#1451 

When the env WARM_ENI_TARGET is set as 0 and none of existing ENIs has available IP address, ipamd will try to add and remove a new ENI infinitely since it always thinks IP availability is low (adding) but no warm ENI is set (deleting). This behavior could dramatically increase call volume to EC2 API. In some case (large clusters), the account could be throttled.

Non custom networking cluster:

1. assigned all IPs of primary ENI or the secondary ENI (t3 and c5 instance types) and checked no looping behavior(add/delete) occurring
2. soaking tests for 5 hours. Every 5 minutes, adding a request for IP into a fully assigned ENI, or removing an IP from an ENI in which only one IP was in use. Result showed new ENI can always been added and IP was successfully assigned from the ENI.
3. automation tests for WARM ENI in integration tests.

Custom networking enabled cluster:

1. no ENI was added if WARM_ENI_TARGET is 0 and no pod was created in the node.
2. new ENI and IP was added and assigned when creating a new pod.
3. ipamd won't add new ENI even if existing ENIs are full assigned for IP addresses.

Also confirmed when WARM_ENI_TARGET is not 0, all behavior stay same as before.

**Automation added to e2e**:

Unite tests were added. We should also add test cases in integration tests.
No

No

No
```release-note

```

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -2042,10 +2042,14 @@ func (c *IPAMContext) isDatastorePoolTooLow() bool {
 		totalIPs = maxIpsPerPrefix
 	}
 
-	poolTooLow := available < totalIPs*warmTarget || (warmTarget == 0 && available == 0)
+	poolTooLow := c.dataStore.GetENIs() == 0 || available < totalIPs*warmTarget
 	if poolTooLow {
 		log.Debugf("IP pool is too low: available (%d) < ENI target (%d) * addrsPerENI (%d)", available, warmTarget, totalIPs)
 		c.logPoolStats(stats)
+	} else if warmTarget == 0 && c.dataStore.IsIPStarvedInDS() {
+		poolTooLow = available <= 0
+		log.Debugf("Ipamd knows WARM_ENI_TARGET is 0 and DS may need more IP. Returning poolToLow (%v)(available is %d) and set ds IP starvation to false.", poolTooLow, available)
+		c.dataStore.SetIPStarvedInDS(false)
 	}
 	return poolTooLow
 
--- a/pkg/ipamd/datastore/data_store.go
+++ b/pkg/ipamd/datastore/data_store.go
@@ -307,6 +307,7 @@ type DataStore struct {
 	backingStore             Checkpointer
 	cri                      cri.APIs
 	isPDEnabled              bool
+	ipStarved                bool
 }
 
 // ENIInfos contains ENI IP information
@@ -803,6 +804,7 @@ func (ds *DataStore) AssignPodIPv4Address(ipamKey IPAMKey) (ipv4address string, 
 	}
 
 	ds.log.Errorf("DataStore has no available IP/Prefix addresses")
+	ds.ipStarved = true
 	return "", -1, errors.New("assignPodIPv4AddressUnsafe: no available IP/Prefix addresses")
 }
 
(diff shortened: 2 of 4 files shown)


pkg/ipamd/ipamd.go
```diff
@@ -2042,10 +2042,14 @@ func (c *IPAMContext) isDatastorePoolTooLow() bool {
 		totalIPs = maxIpsPerPrefix
 	}
 
-	poolTooLow := available < totalIPs*warmTarget || (warmTarget == 0 && available == 0)
+	poolTooLow := c.dataStore.GetENIs() == 0 || available < totalIPs*warmTarget
```
achevuru, line 2045: Can we expand on the logic we're attempting to use here via a detailed comment? Might help someone understand the rationale behind what we're checking here.

In particular, I'm curious if `c.dataStore.GetENIs()` ever be `0` and why are we checking this condition here? We should at least have Primary ENI in our datastore isn't it?

pkg/ipamd/ipamd.go
```diff
@@ -2042,10 +2042,14 @@ func (c *IPAMContext) isDatastorePoolTooLow() bool {
...
 	if poolTooLow {
 		log.Debugf("IP pool is too low: available (%d) < ENI target (%d) * addrsPerENI (%d)", available, warmTarget, totalIPs)
 		c.logPoolStats(stats)
+	} else if warmTarget == 0 && c.dataStore.IsIPStarvedInDS() {
+		poolTooLow = available <= 0
```
achevuru, line 2050: We appear to be setting `poolTooLow` in the `else if` check of `if poolTooLow`? Why can't we enhance the original statement in #2045? Also, we're setting ds.ipStarved back to false before the reconciler adds additional IPs/ENIs. Shouldn't we do it after the ENI or IP add succeeds and datastore has a non zero count of available addresses?

pkg/ipamd/ipamd.go
haouc: Yes, we should have a comment. GetENIs() should be always > 0. But we do have unit test check for 0 case. I wonder if we have some use case that makes 0 check valid. @jayanthvn do you know some context on this? Thanks.

pkg/ipamd/ipamd.go
haouc: poolTooLow doesn't check the special case of warm eni is zero. We want to avoid reconciler adding new ENI to DS if WARM_ENI is 0 and no extra IP is required. When poolTwooLOw is set in `else if`, it just overrides the variable's value for return. 
I prefer to reset the DS ipStarved back to `false` as soon as we can to avoid possible race from multiple requests. If IP allocation fails, the kubelet will try again and DS will set the property to true which will be recognized by ipam reconciler again.
Please let me know if this could cause failure or significant latency. I am also ok to keep the `true` state longer since the worst case would be ipam attaching extra ENI and will remove it eventually.

pkg/ipamd/ipamd.go
jayanthvn: Yes we should always have primary ENI.

pkg/ipamd/ipamd.go
jayanthvn: I agree with Apurup to merge the poolTooLow into one logic. Also instead of a flag, can't we just walk thru the ENIs and see if IPs are starved?

haouc: Watching the flag can avoid walking thru the ENIs every reconciling cycle. Moreover the checking is not necessary in most of use cases. When we use flag to deliver the signal, only the DS encounter IP starving, ipamd takes action.

pkg/ipamd/ipamd.go
haouc: If you are fine with these, I will update to remove unit tests for `== 0` and this check after the PR is approved and merged.

pkg/ipamd/ipamd.go
jayanthvn: But the ENI's max will be just few so shouldn't be a big walk. Maybe lets sync up offline?

haouc: Yeah, we can discuss offline. I think that the key is we need to do ad-hoc on the signal of IP starving instead of watching on ENIs since watching on ENIs without a signal can lead to the original ENIs churning problem.

dennisme: @jayanthvn @haouc Thanks for doing this PR! Where there any take aways from the offline discussion? Can I help on this issue in anyway?

Issue #1689, opened 2021-10-14
Add ARM based nodes for e2e and UT tests


jayanthvn:
**What would you like to be added**:
ARM based work nodes and runners

**Why is this needed**:
To test both AMD and ARM based images

vikasmb:
Changes have been completed and merged.
Multi-arch integ test sample execution is in https://github.com/aws/amazon-vpc-cni-k8s/runs/5056823481?check_suite_focus=true

PR #1824, opened 2022-01-26
Build multi-arch image in integration test.

vikasmb:
**What type of PR is this?**
Feature

Builds multi-arch (for x86 and arm64) images during integration test.

- Builds multi-arch (for x86 and arm64) images during integration test.
- Uses docker buildx to build and push images for x86 and ARM as part of integ test workflow.
- Uses buildkit defined [variables](https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope) to detect the target architecture variable set in dockerfile 
- For regular docker builds, arch variable is set in Makefile and thus, GOARCH build arg isn't needed anymore in docker files.

#1689 

Integ test for this PR passed in https://github.com/aws/amazon-vpc-cni-k8s/actions/runs/1758227018
`make all` and `make docker-unit-tests` passed with this change. 

**Automation added to e2e**:

No

No
```release-note

```


scripts/run-integration-tests.sh
```diff
@@ -131,13 +131,11 @@ if [[ $(docker images -q "$IMAGE_NAME:$TEST_IMAGE_VERSION" 2> /dev/null) ]]; the
 else
     echo "CNI image $IMAGE_NAME:$TEST_IMAGE_VERSION does not exist in repository."
     START=$SECONDS
-    make docker IMAGE="$IMAGE_NAME" VERSION="$TEST_IMAGE_VERSION"
-    docker push "$IMAGE_NAME:$TEST_IMAGE_VERSION"
+    make multi-arch-cni-build-push IMAGE="$IMAGE_NAME" VERSION="$TEST_IMAGE_VERSION"
```
orsenthil, line 134: Where is `TEST_IMAGE_VERSION` used in docker buildx command line?

vikasmb: It is used to set Version variable in Makefile: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/Makefile#L23

Makefile
```diff
@@ -60,7 +60,11 @@ IMAGE_ARCH_SUFFIX = $(addprefix -,$(filter $(ARCH),arm64))
 # GOLANG_IMAGE is the building golang container image used.
 GOLANG_IMAGE = public.ecr.aws/docker/library/golang:1.16-stretch
 # For the requested build, these are the set of Go specific build environment variables.
-export GOARCH ?= $(ARCH)
+ifdef TARGETARCH
```
anguslees, line 63: This works, but adds complexity and extra codepaths.  Can we just pass `make ARCH=$TARGETARCH` from the Dockerfile?  (or `make GOARCH=$TARGETARCH` ?  There are already too many $arch variables in this Makefile :confused: )

anguslees: Should we remove line 54?
https://github.com/aws/amazon-vpc-cni-k8s/blob/050d9be772edc1c8fa6dc2548e0fc98aa128c952/Makefile#L54

And just replace all this with `export GOARCH = $(ARCH)` - with the idea that $(ARCH) is usually unset.

I think if we _don't_ set GOARCH then it defaults to the current architecture, and all the UNAME_ARCH/ARCH stuff above is just a complicated way of doing the same thing.

... Alternatively, rip out everything to do with $(ARCH), $(GOARCH), and $(GOOS) in this file.  Then expect the caller (ie: Dockerfile) to set GOOS/GOARCH before invoking this Makefile, if cross-compiling.

Basically, we should set the target GOOS/GOARCH in one layer - either the Makefile or the Dockerfile.  And not some fight between both.

scripts/dockerfiles/Dockerfile.init
```diff
@@ -2,7 +2,7 @@ ARG golang_image
 
 FROM $golang_image as builder
 WORKDIR /go/src/github.com/aws/amazon-vpc-cni-k8s
-ARG GOARCH
+ARG TARGETARCH
```
anguslees, line 5: From my Makefile comment above, I think we can simplify things if we add `ENV ARCH=$TARGETARCH` here too (and clean up the Makefile a bit)

scripts/dockerfiles/Dockerfile.init
```diff
@@ -2,7 +2,7 @@ ARG golang_image
 
 FROM $golang_image as builder
```
anguslees, line 3: I think the build will go faster if we set `FROM --platform=$BUILDPLATFORM as builder` here.

(That means we'll cross-compile on the native architecture, rather compile in a qemu-emulated version of the target architecture)

Ditto for other Dockerfiles.

Makefile
anguslees: Also we can remove all the IMAGE_ARCH_SUFFIX stuff now that we have true multi-arch images:
https://github.com/aws/amazon-vpc-cni-k8s/blob/050d9be772edc1c8fa6dc2548e0fc98aa128c952/Makefile#L59

vikasmb: @anguslees Thanks for the feedback, I will work on removing the extra variables for `IMAGE_ARCH_SUFFIX`.

Regarding the switch statement in Makefile, it's currently needed as docker platform arguments such as `TARGETARCH` are set only when buildkit is used (Reference: From https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope, `This feature is only available when using the BuildKit backend`). 

Further, docker buildkit is not yet the default in linux (https://github.com/moby/moby/issues/40379). As such, to make the build process work in scenarios where buildkit isn't used, setting the arch variable in makefile is needed.

scripts/dockerfiles/Dockerfile.init
vikasmb: Good catch, yes, will change this part. 
https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/ has this recommendation too for faster multi-platform builds.

scripts/dockerfiles/Dockerfile.init
vikasmb: As noted in the other comment, TARGETARCH is available only when buildkit is enabled, which isn't the default yet in linux. Thus, this platform arg is set by docker only in case of multi-platform builds (where buildx and buildkit are used).

Makefile
vikasmb: `I think if we don't set GOARCH then it defaults to the current architecture, and all the UNAME_ARCH/ARCH stuff above is just a complicated way of doing the same thing.` 
That's correct and the build itself would succeed even if GOARCH isn't set. However, it would not be able to identify the correct version of networking plugin to download here - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/Makefile#L211

In case of buildkit driven builds, this would need to be derived from the TARGETARCH platform argument and in case of non buildkit builds, it would have to be derived from `uname` output.

scripts/dockerfiles/Dockerfile.init
vikasmb: Usage of BUILDPLATFORM in docker file for instruction `FROM --platform=$BUILDPLATFORM $golang_image as builder` has the same issue as TARGETARCH when buildkit isn't used, as it is a platform argument available with [buildkit](https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope) only.

```
Step 2/17 : FROM --platform=$BUILDPLATFORM $golang_image as builder
failed to parse platform : "" is an invalid component of "": platform specifier component must match "^[A-Za-z0-9_-]+$": invalid argument
```

We could mandate usage of buildkit by [setting](https://docs.docker.com/develop/develop-images/build_enhancements/#to-enable-buildkit-builds) `DOCKER_BUILDKIT=1` in Makefile prior to docker build. 
In that way, we would get the platform arguments and simplify the Makefile and obtain faster build speeds.
Thoughts on mandating the buildkit setting ?

Makefile
vikasmb: With the modified approach of setting buildkit at the outset in makefile with `export DOCKER_BUILDKIT=1` https://github.com/aws/amazon-vpc-cni-k8s/pull/1824/files, I have removed the extra executions for `UNAME_ARCH` and `ARCH`.

Now, value for `GOARCH` variable is directly derived from the docker platform arg value of `TARGETARCH`

scripts/dockerfiles/Dockerfile.init
vikasmb: Now, TARGETARCH value is used in the makefile to set the value of goarch

scripts/dockerfiles/Dockerfile.init
vikasmb: After setting `DOCKER_BUILDKIT=1`, if we execute go build using `BUILDPLATFORM`, we would have to pass additional arguments to `build-linux` target (to set the GOARCH value during go build).

However, such approach does mean that execution of `make build-linux` without docker commands would not work without manually / automatically setting GOARCH in the makefile in that invocation path (which would revert makefile to the approach in previous revision). Thus, it may be better to leave the dockerfiles as-is for simpler makefile logic. Thoughts ?

anguslees: > Thoughts?

I think trying to support builds using both buildx and older docker from the same Dockerfile is making everything complicated and messy.

Do we need to continue to support building with older docker?  If so, maybe we don't want to support buildx yet?

(This is a build-time dependency only, so we can be opinionated.)

vikasmb: So, with the makefile change, it will only mandate usage of docker buildkit which has been available in docker since version 18.09 (https://docs.docker.com/develop/develop-images/build_enhancements/).

Usage of docker buildx cli within the existing makefile commands isn't needed and the usual docker cli can be used with docker buildkit too.

scripts/dockerfiles/Dockerfile.metrics
```diff
@@ -2,7 +2,6 @@ ARG golang_image
 
 FROM $golang_image as builder
 WORKDIR /go/src/github.com/aws/amazon-vpc-cni-k8s
-ARG GOARCH
```
M00nF1sh, line 5: i think we should make this metrics-helper a multi-arch image as well to be consistent.

M00nF1sh (approved):

scripts/dockerfiles/Dockerfile.metrics
vikasmb: ack, will check and make changes to metrics-helper if needed in a followup PR.

Issue #1724, opened 2021-11-04
CNI does not wait for ENI and IP to be available before reporting ready when setting MINIMUM_IP_TARGET and WARM_IP_TARGET


dtnyn:
## What happened:
After setting `MINIMUM_IP_TARGET` and `WARM_IP_TARGET`. When scheduled on a new node, pods are stuck on `failed to assign an IP address to container` for a few minutes before being able to be created.

From my observation I can see that after setting these override, aws-node does not wait for its datastore to be populated first before reporting ready causing pods to try to be scheduled onto these nodes and goes into a crashing state as no IP can be assigned

Pod log

```
Normal   SuccessfulAttachVolume  3m16s              attachdetach-controller  AttachVolume.Attach succeeded for volume "XXXXXXXX"
  Warning  FailedCreatePodSandBox  3m12s              kubelet                  Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "22cbd7067d76c2ed509436f65384e1df59ec8949e7910d7bbbfb32de8a511f89": add cmd: failed to assign an IP address to container
  Warning  FailedCreatePodSandBox  3m                 kubelet                  Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "81a9fde8b5ea0da28959e8c08e08dc2ffcbe700c7f3067afd273e9ac67b37588": add cmd: failed to assign an IP address to container
  Warning  FailedCreatePodSandBox  2m48s              kubelet                  Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "a94f47242b45d3610a9bf25ee17de09d8fb6f77dcf1754deac13395020b1082f": add cmd: failed to assign an IP address to container
  Warning  FailedCreatePodSandBox  2m37s              kubelet                  Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "a6f2f67496e2c75cdc89963afc5d4f0b47fe3d6988d0356b4cd0e0ffb641400c": add cmd: failed to assign an IP address to container
  Warning  FailedCreatePodSandBox  2m24s              kubelet                  Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "de2c20851923cc290e187ade06472c2f77595c5d55f7ee7522603fb55e9ee943": add cmd: failed to assign an IP address to container
[...]

PR #1827, opened 2022-01-27
Custom networking optimization

jayanthvn:
**What type of PR is this?**
Enhancement

#1724 

With custom networking, secondary ENI will be allocated by the reconciler, until then datastore will be empty and node will marked as ready so pods will be stuck in creating. This PR will allocate an ENI in nodeInit path.

Yes

**Automation added to e2e**:

No

No

No

No
```release-note

```

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -417,20 +417,6 @@ func New(rawK8SClient client.Client, cachedK8SClient client.Client) (*IPAMContex
 		return nil, err
 	}
 
-	mac := c.awsClient.GetPrimaryENImac()
-
-	// retrieve security groups
-	if c.enableIPv4 && !c.disableENIProvisioning {
-		err = c.awsClient.RefreshSGIDs(mac)
-		if err != nil {
-			return nil, err
-		}
-
-		// Refresh security groups and VPC CIDR blocks in the background
-		// Ignoring errors since we will retry in 30s
-		go wait.Forever(func() { _ = c.awsClient.RefreshSGIDs(mac) }, 30*time.Second)
-	}
-
 	return c, nil
 }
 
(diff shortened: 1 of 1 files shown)


bryantbiggs:
@M00nF1sh are you able to review?

pkg/ipamd/ipamd.go
```diff
@@ -572,19 +571,9 @@ func (c *IPAMContext) nodeInit() error {
...
-			c.updateLastNodeIPPoolAction()
-		} else if err != nil {
-			if containsInsufficientCIDRsOrSubnetIPs(err) {
-				log.Errorf("Unable to attach IPs/Prefixes for the ENI, subnet doesn't seem to have enough IPs/Prefixes. Consider using new subnet or carve a reserved range using create-subnet-cidr-reservation")
-				c.lastInsufficientCidrError = time.Now()
-				return nil
-			}
-			return err
-		}
+		c.increaseDatastorePool(ctx)
```
M00nF1sh, line 576: this slightly changed the behavior. previously an error will return if any thing wrong when allocating IPs and prevent node from became ready. e.g. missing permission.

but now it will just try to allocate IPs and silently swallow the error, and node will be ready any way.

shall we make c.increaseDatastorePool returns errors and return the error during bootstrap, and ignore the error(only logs it) during the IP pool reconcile loop?

jayanthvn: Yeah that right, it would be good to fail. Instead of modifying c.increaseDatastorePool where the code flow is slightly different i.e if the AssignCidr returns (non-nil && subnet has space) we go ahead with creating an ENI however here we need it to fail. I am thinking we can move this logic - https://github.com/jayanthvn/amazon-vpc-cni-k8s/blob/d753d02410137571afa50adb63897206627baf49/pkg/ipamd/ipamd.go#L794-L808 to a separate function and call it in nodeInit and increaseDataStorePool. Any thoughts?

jayanthvn: On a second thought, we can modify increaseDatastorePool so that we can re-use the same function. Will test it out.

jayanthvn: nvm, I restructured the code.

jayanthvn: Also, I am thinking to limit this for customer networking and the first eni. Will discuss more offline.

achevuru: Yes, would be good to limit this for Custom networking case. 

Instead of modifying `increaseDataStorePool` to return an error (and then ignore it everywhere else except at this location), I think we can maybe check if the call resulted in a change in the datastore contents (i.e.,) ENIs (and maybe IPs) count and use that to determine if the call succeeded or not and return an error if it didn't.

sidewinder12s:
@jayanthvn Any update on this?

jayanthvn:
@sidewinder12s - Sorry couldn't get a chance to look into this since got tied up with other activities. Will spend some time next week and update the PR.

jdn5126:
Closing as this is being replaced by #2354

PR #1832, opened 2022-01-31
Updated go version in go.mod

cgchinmay:
**What type of PR is this?**
feature

Update go version

### DIFF
--- a/.github/workflows/build-multi-arch.yml
+++ b/.github/workflows/build-multi-arch.yml
@@ -12,10 +12,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
 
-    - name: Set up Go 1.16
+    - name: Set up Go 1.17
       uses: actions/setup-go@v2
       with:
-        go-version: '1.16'
+        go-version: '1.17'
       id: go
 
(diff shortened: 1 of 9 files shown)


jayanthvn:
Even arm and x86 builds look fine.

jayanthvn (approved):

orsenthil (approved):
I had a question if the upgrade golang, will need any integration testing before release.

I think,  Integration testing will be safe, since we are upgrading go.mod dependencies along with the runtime.

jayanthvn:
Yeah agreed. We would need to run the integration test suite before release.

cgchinmay:
> I had a question if the upgrade golang, will need any integration testing before release.
> 
> I think, Integration testing will be safe, since we are upgrading go.mod dependencies along with the runtime.

Integration and e2e tests did pass

orsenthil:
> Integration and e2e tests did pass

Nice. You could link to the success of the Integration test link in the SIM tickets, just for future reference.
VPC resource controller integration tests could be included too.

Thank you, @cgchinmay

PR #1846, opened 2022-02-04
Describe instance types across all regions when generating limits.

cartermckinnon:
**What type of PR is this?**

bug / feature

No public issue number, we had an internal request for `hpc6a` support in `amazon-eks-ami`.

We received a request to add `hpc6a.48xlarge` to [amazon-eks-ami/files/eni-max-pods.txt](https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt). This instance type is only available (commercially) in `us-east-2`; and therefore its limits could only be calculated with:
```sh
AWS_REGION=us-east-2 make generate-limits
```

However, some instance types available elsewhere are also not available in `us-east-2`.

This change modifies the logic to combine the instance type limits from all regions (that the invoking credentials are opted-into).

I used credentials that are opted-into all regions to execute the script.

** Will this PR introduce any new dependencies?**:

No.

Not applicable.

No.

No.

```release-note
Describe instance types across all regions when generating limits.
```


cartermckinnon:
@jayanthvn Can you kick-off the checks for this again? I had to fix a minor format issue.

M00nF1sh (approved):
/lgtm

jayanthvn (approved):
lgtm

PR #1848, opened 2022-02-04
Update eni-and-ip-target.md

jolcese:
**What type of PR is this?**

bug/documentation

Fixed the use case for MINIMUM_IP_TARGET=10 and pods=0 where showed Unused IPs as 5 but should be 10

Fixes documentation

Not applicable

**Automation added to e2e**:

Not applicable

No

No

No

No

```release-note

```


jayanthvn (approved):
lgtm. Thanks for fixing this.

PR #1870, opened 2022-02-16
Update Calico README for migration instructions

haouc:
**What type of PR is this?**
We want to provide some instructions to users who need migrate/upgrade their Calico resource after we deprecate Calico charts in this repo.

documentation
Providing instructions for users to migrate their Calico resources.

We need provide some instructions for future use case after we deprecate the Calico charts.

Yes. Calico resources were tested for migration/upgrading with Star Policy Demo
**Automation added to e2e**:

no

no

no

no
```release-note

```


jayanthvn (approved):
lgtm

PR #1871, opened 2022-02-16
Disable auto scale-in for the cluster autoscaler in perf tests

vikasmb:
- Disable auto scale-in as it led to longer execution times
  for 5000 pod test

**What type of PR is this?**
cleanup

Disable auto scale-in of nodes as it led to longer execution times
  for 5000 pod test

With auto scale-in, ASG capacity got lowered from initial value of 99 and re-scaling during 5000 pod takes longer time compared to previous test executions.
New config based on faq entry of https://github.com/kubernetes/autoscaler/blob/59f2eaa1d8fcb92bf7df97ec083fc02c6830205b/cluster-autoscaler/FAQ.md#how-can-i-prevent-cluster-autoscaler-from-scaling-down-non-empty-nodes

**Automation added to e2e**:

No

No

No

No
```release-note

```

### DIFF
--- a/.github/workflows/weekly-cron-test.yml
+++ b/.github/workflows/weekly-cron-test.yml
@@ -2,7 +2,7 @@ name: Weekly CNI tests
 
 on:
   schedule:
-    - cron: "0 22 * * 3"  # every Wednesday
+    - cron: "0 16 * * 4"  # every Thursday
 
 permissions:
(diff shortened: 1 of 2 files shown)


jayanthvn (commented):
lgtm

jayanthvn (approved):

Issue #1728, opened 2021-11-05
Change e2e tests to IMDSv2 only nodes


chlunde:
**What would you like to be added**:
Change e2e tests to IMDSv2 only nodes

**Why is this needed**:
Because IMDSv2 is recommended and we need to test for that (which also will work on IMDSv1+v2 setups)

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html
https://github.com/kubernetes/kops/pull/10324/files

Please add label "testing" as requested in the PR template, I can't do that.

Related: https://github.com/aws/amazon-vpc-cni-k8s/pull/1727

jayanthvn:
#1873 merged to master.

PR #1873, opened 2022-02-18
Use IMDSv2 for cni e2e tests

cgchinmay:
**What type of PR is this?**
feature 

#1728 

It enforces usage for IMDSv2 for e2e test nodes

Ran custom-networking, pod-eni and snat e2e tests and validated that all are passing 

No

No

No


jayanthvn (approved):
lgtm

Issue #1821, opened 2022-01-26
TestIPAMContext_nodeIPPoolTooLow is flaky, and relies on the order of the output for assertion


orsenthil:
**What happened**:

Exercising this test https://github.com/orsenthil/amazon-vpc-cni-k8s/blob/8c1be691b618ac97914e297a59f68395f75faed8/pkg/ipamd/ipamd_test.go#L1323 multiple times can give different results.

**Attach logs**

**What you expected to happen**:

Deterministic output from TestIPAMContext_nodeIPPoolTooLow test run.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- CNI Version
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):

vikasmb:
In my testing, I did not find `TestIPAMContext_nodeIPPoolTooLow` to be flaky. Refer to sample execution below.
```
while true; do go clean -testcache; /usr/local/go/bin/go test -timeout 30s -run ^TestIPAMContext_nodeIPPoolTooLow$ github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd; done
ok      github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd     0.015s
ok      github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd     0.015s
ok      github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd     0.015s
ok      github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd     0.016s
ok      github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd     0.015s
ok      github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd     0.015s
ok      github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd     0.015s
ok      github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd     0.015s
ok      github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd     0.015s
ok      github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd     0.015s
ok      github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd     0.015s
ok      github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd     0.015s
[...]

PR #1880, opened 2022-02-23
Fix flaky test in ipamd package

vikasmb:
- Also fixed kops version string

**What type of PR is this?**
cleanup

#1821 

Fixes assertion logic in TestIPAMContext_filterUnmanagedENIs so that it doesn't rely on a particular  order for comparing the expected and output slices. 
Also, fixes the kops version string

Executed unit tests
**Automation added to e2e**:

N/A

No
No

No

No
```release-note

```

### DIFF
--- a/pkg/ipamd/ipamd_test.go
+++ b/pkg/ipamd/ipamd_test.go
@@ -1304,7 +1304,7 @@ func TestIPAMContext_filterUnmanagedENIs(t *testing.T) {
 		{"Secondary/Tertiary ENI unmanaged", Test2TagMap, allENIs, primaryENIonly, []string{eni2.ENIID, eni3.ENIID}},
 		{"Secondary ENI unmanaged", Test3TagMap, allENIs, filteredENIonly, []string{eni2.ENIID}},
 		{"Secondary ENI unmanaged and Tertiary ENI CNI created", Test4TagMap, allENIs, filteredENIonly, []string{eni2.ENIID}},
-		{"Secondary ENI not CNI created and Tertiary ENI CNI created", Test5TagMap, allENIs, filteredENIonly, []string{eni2.ENIID}},
+		{"Secondary ENI not CNI created and Tertiary ENI CNI created", Test5TagMap, allENIs, filteredENIonly, nil},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
--- a/pkg/ipamd/ipamd_test.go
+++ b/pkg/ipamd/ipamd_test.go
@@ -1311,7 +1311,13 @@ func TestIPAMContext_filterUnmanagedENIs(t *testing.T) {
 			c := &IPAMContext{
 				awsClient:                mockAWSUtils,
 				enableManageUntaggedMode: true}
-			mockAWSUtils.EXPECT().SetUnmanagedENIs(tt.unmanagedenis).AnyTimes()
+
+			mockAWSUtils.EXPECT().SetUnmanagedENIs(gomock.Any()).
+				Do(func(args []string) {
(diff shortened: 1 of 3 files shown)


jayanthvn (approved):

pkg/ipamd/ipamd_test.go
```diff
@@ -1311,7 +1311,13 @@ func TestIPAMContext_filterUnmanagedENIs(t *testing.T) {
 			c := &IPAMContext{
 				awsClient:                mockAWSUtils,
 				enableManageUntaggedMode: true}
-			mockAWSUtils.EXPECT().SetUnmanagedENIs(tt.unmanagedenis).AnyTimes()
+
+			mockAWSUtils.EXPECT().SetUnmanagedENIs(gomock.Any()).
+				Do(func(args []string) {
+					sort.Strings(tt.unmanagedenis)
```
orsenthil, line 1317: This looks fine. Not sure why the unit tests are failing. Perhaps there is flake elsewhere too?

orsenthil: Had failed with this diff.
```
=== RUN   TestIPAMContext_filterUnmanagedENIs/Secondary_ENI_not_CNI_created_and_Tertiary_ENI_CNI_created
--- FAIL: TestIPAMContext_filterUnmanagedENIs (0.00s)
    --- PASS: TestIPAMContext_filterUnmanagedENIs/No_tags_at_all (0.00s)
    --- PASS: TestIPAMContext_filterUnmanagedENIs/Primary_ENI_unmanaged (0.00s)
    --- PASS: TestIPAMContext_filterUnmanagedENIs/Secondary/Tertiary_ENI_unmanaged (0.00s)
    --- PASS: TestIPAMContext_filterUnmanagedENIs/Secondary_ENI_unmanaged (0.00s)
    --- PASS: TestIPAMContext_filterUnmanagedENIs/Secondary_ENI_unmanaged_and_Tertiary_ENI_CNI_created (0.00s)
    assertions.go:262: 
        	Error Trace:	ipamd_test.go:1319
        	            				value.go:556
        	            				value.go:339
        	            				call.go:152
        	            				controller.go:221
        	            				awsutils_mocks.go:469
        	            				ipamd.go:296
        	            				ipamd_test.go:1321
        	Error:      	Not equal: 
        	            	expected: []string{"eni-00000001"}
        	            	actual  : []string(nil)
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,4 +1,2 @@
        	            	-([]string) (len=1) {
        	            	- (string) (len=12) "eni-00000001"
        	            	-}
        	            	+([]string) <nil>
        	            	 
        	Test:       	TestIPAMContext_filterUnmanagedENIs/Secondary_ENI_unmanaged_and_Tertiary_ENI_CNI_created
    --- FAIL: TestIPAMContext_filterUnmanagedENIs/Secondary_ENI_not_CNI_created_and_Tertiary_ENI_CNI_created (0.00s)```

vikasmb: Yes, it was due to a wrong expected value in the test case for `Secondary ENI not CNI created and Tertiary ENI CNI created`. Fixed it in the latest commit in this PR.

jayanthvn: @vikasmb as discussed offline we need to return `nil`. This test should have failed always since expected o/p was wrong but it was successful most of the time. So even after changing this to nil, the function might return nil or eni-1 and still fail.

I think we need to fix this - https://github.com/aws/amazon-vpc-cni-k8s/blob/5f79f942dcfcc694125db998b1ffa46bea72b8ee/pkg/ipamd/ipamd_test.go#L1318-L1329

PR #1884, opened 2022-02-24
Update README to highlight containerd.sock edge case with EKS AMI.

orsenthil:
**What type of PR is this?**

Documention

Customer were using EKS AMI, and followed these instructions to point to containerd runtime. 
They also tried to upgrade, and ran into surprising issues. 

* This clarifies, that while using EKS AMI, no need to point to containerd to cri.sock as AMI symlinks containerd.sock to dockershim.sock
* Also gives a link to upgrade documentation that customers can follow to upgrade customer to containerd runtime. As suggested by managed nodes group team.

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -620,10 +620,10 @@ The mountPath should be changed to `/var/run/cri.sock` and hostPath should be po
 `/var/run/containerd/containerd.sock` for containerd. If using helm chart, the flag `--set cri.hostPath.path=/var/run/containerd/containerd.sock`
 can set the paths for you.
 
-*Note*: 
+*Note*:
 
-* When using other container runtime instead of dockershim, make sure also setting kubelet in instances.
-* If you are using EKS provided AMI, there is no need to containerd.sock to cri.sock, as EKS AMI [symlinks containerd.sock to dockershim.sock](https://github.com/awslabs/amazon-eks-ami/blob/2289937fe9fb3933d066b6ae72a4299b2624e154/files/bootstrap.sh#L470). In addition, if planning upgrade to containerd runtime while using EKS provided AMI, please follow the instructions in our documentation.
+* When using a different container runtime instead of dockershim in VPC CNI, make sure kubelet is also configured to use the same CRI.
+* If you want to enable containerd runtime with the support provided by Amazon AMI, please follow the instructions in our documentation, [Enable the containerd runtime bootstrap flag](https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html#containerd-bootstrap)
 
 ### Notes
 
--- a/README.md
+++ b/README.md
@@ -620,7 +620,10 @@ The mountPath should be changed to `/var/run/cri.sock` and hostPath should be po
 `/var/run/containerd/containerd.sock` for containerd. If using helm chart, the flag `--set cri.hostPath.path=/var/run/containerd/containerd.sock`
 can set the paths for you.
 
-*Note*: When using other container runtime instead of dockershim, make sure also setting kubelet in instances.
+*Note*: 
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -594,7 +594,10 @@ The mountPath should be changed to `/var/run/cri.sock` and hostPath should be po
 `/var/run/containerd/containerd.sock` for containerd. If using helm chart, the flag `--set cri.hostPath.path=/var/run/containerd/containerd.sock`
 can set the paths for you.
 
-*Note*: When using other container runtime instead of dockershim, make sure also setting kubelet in instances.
+*Note*: 
+
+* When using other container runtime instead of dockershim, make sure also setting kubelet in instances.
```
vikasmb, line 599: "make sure also setting kubelet in instances." - this seems to be incomplete.

README.md
```diff
@@ -594,7 +594,10 @@ The mountPath should be changed to `/var/run/cri.sock` and hostPath should be po
...
+* If you are using EKS provided AMI, there is no need to containerd.sock to cri.sock, as EKS AMI [symlinks containerd.sock to dockershim.sock](https://github.com/awslabs/amazon-eks-ami/blob/2289937fe9fb3933d066b6ae72a4299b2624e154/files/bootstrap.sh#L470). In addition, if planning upgrade to containerd runtime while using EKS provided AMI, please follow the instructions in our documentation.
```
vikasmb, line 600: should this be "there is no need to mount containerd.sock to cri.sock"

README.md
achevuru: Yeah, agree on the above. Maybe something along the lines of - `When using a different container runtime instead of dockershim in VPC CNI, make sure `kubelet` is also configured to use the same CRI` and we can leave a link to `containerd` configuration flow - https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-ami.html#containerd-bootstrap

README.md
achevuru: While this is accurate, I feel we should encourage users to mount `containerd.sock` explicitly in to `cri.sock` when they are using `containerd`. Using `dockershim.sock` as the CRI socket name for `containerd` cases will be confusing. For ex - We can't really say if they are using `dockershim` or `containerd` as CRI by looking at the VPC CNI spec and creates unnecessary confusion in my opinion.

jayanthvn: Also we can put a note that pods have to be deleted prior to migration since IPs will leak.

orsenthil: Hi @achevuru , Given EKSAMI Symlink containerd.sock to dockershim.sock (https://github.com/awslabs/amazon-eks-ami/blob/2289937fe9fb3933d066b6ae72a4299b2624e154/files/bootstrap.sh#L470) - so, when the customer was running containerd, they were were confused and and MNG team was asserting that it was un-necessary since they symlink in AMI.  I think, this is just to avoid the confusion, until the symlink is removed?

--- 
The other link on document from mng team (which I should add) should explain about how to migrate from one runtime to another.

achevuru: So, we're now referring to the socket name VPC CNI should use. MNG symlink will help in general - I mean if an user has some other application relying on `dockershim.sock` name, migration will be seamless for them. But for VPC CNI, we can encourage users to mount the correct socket under `cri.sock` as a best practice recommendation. Maybe we can retain the existing wording but add an additional sentence stating our recommendation to mount it under cri.sock (?)

orsenthil:
Hi @achevuru , @jayanthvn - If this is okay, could you please approve this change, and we can continue to add further documentation, on multiple runtime support on top or as additional documentation?

jayanthvn (approved):
lgtm

PR #1902, opened 2022-03-02
Emit metrics to cloudwatch on test execution status

vikasmb:
**What type of PR is this?**
feature

Emits metrics to cloudwatch regarding test status.

Emits metrics to cloudwatch regarding test status.

Tested it locally.

**Automation added to e2e**:

No
No

No

No
```release-note

```


scripts/lib/integration.sh
```diff
@@ -23,6 +23,7 @@ function run_kops_conformance() {
     KOPS_TEST_DURATION=$((SECONDS - START))
     echo "TIMELINE: KOPS tests took $KOPS_TEST_DURATION seconds."
 
+    emit_cloudwatch_metric "kops_test_status" "1"
```
jayanthvn, line 26: Can we also add for bottlerocket, calico and perf tests?

vikasmb: Yes, updated it in next commit.

scripts/run-integration-tests.sh
```diff
@@ -35,6 +35,7 @@ __cluster_deprovisioned=0
 
 on_error() {
     echo "Error with exit code $1 occurred on line $2"
+    emit_cloudwatch_metric "error_occurred" "1"
```
jayanthvn, line 38: Here we won't know which test failed right? or do we get that granularity in CW?

jayanthvn: Synced up with Vikas on this.

jayanthvn (approved):
lgtm

PR #1903, opened 2022-03-03
IPv4 Randomize SNAT support for IPv6 pods

achevuru:
**What type of PR is this?**
Enhancement + Addresses few inconsistencies/bugs with IPv6 mode.

NA

- For Egress IPv4 flows from an IPv6 pod, we SNAT the node-local IPv4 address assigned to an IPv6 Pod to Node's Primary IPv4 address. We enable `random-fully` by default if the underlying iptables version supports it. PR extends support for `AWS_VPC_K8S_CNI_RANDOMIZESNAT` to the `egress-v4-cni-plugin` which will allow user to configure this.
- Addresses duplicate IPv4 route issue across CNI restarts in IPv6 mode.

 - Validated that `random-fully` option is only set if `AWS_VPC_K8S_CNI_RANDOMIZESNAT` is set to `prng` for egress IPv4 flows.
 - Validated that the node local IPv4 route isn't replicated across CNI restarts.
 - 
No

No

No

 - Extends `AWS_VPC_K8S_CNI_RANDOMIZESNAT` support to `egress-v4-cni-plugin`.


cmd/egress-v4-cni-plugin/snat/snat.go
```diff
@@ -43,13 +43,17 @@ func iptRules4(target, src net.IP, chain, comment string, useRandomFully bool) [
...
-func Snat4(target, src net.IP, chain, comment string) error {
+func Snat4(target, src net.IP, chain, comment, randomizeSNAT string) error {
 	ipt, err := iptables.NewWithProtocol(iptables.ProtocolIPv4)
 	if err != nil {
 		return fmt.Errorf("failed to locate iptables: %v", err)
 	}
 
-	rules := iptRules4(target, src, chain, comment, ipt.HasRandomFully())
+	useRandomFully := true
+	if randomizeSNAT == "none" {
```
M00nF1sh, line 53: 1. shall we support for "--random" as well? i.e. randomizeSNAT == "hashrandom"
2. we should check ipt.HasRandomFully() as well, and if it's not supported, fallback to either "--random" or none. (depends on question 1)

M00nF1sh (approved):
/lgtm

jayanthvn (approved):
lgtm, but I think we should add some documentation around this, I mean explaining how this works.

achevuru:
> lgtm, but I think we should add some documentation around this, I mean explaining how this works.
Behavior currently documented under `AWS_VPC_K8S_CNI_RANDOMIZESNAT` will still be correct. We're extending the env variable out to the secondary CNI plugin as well. We can add a note if required under that. Will track it for README updates prior to 1.11 release.

jayanthvn:
> > lgtm, but I think we should add some documentation around this, I mean explaining how this works.
> 
> 
> 
> Behavior currently documented under `AWS_VPC_K8S_CNI_RANDOMIZESNAT` will still be correct. We're extending the env variable out to the secondary CNI plugin as well. We can add a note if required under that. Will track it for README updates prior to 1.11 release.

Sounds good.

PR #1705, opened 2021-10-21
Add VlanId in the cmdAdd Result struct 

cgchinmay:
This VlanId will appear in the prevResult during cmdDel request
The cni can then fetch the vlanId and delete the network associated with vlanId
This will help prevent calls made to api server from ipamd to fetch the vlan device information 

#1644 
#1313 

Tested cni and ipamd integration tests
Test with e2e tests

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1907, opened 2022-03-04
support new SGPP standard mode

M00nF1sh:
**What type of PR is this?**
feature
cleanup

N/A

1. This PR introduces a new knob called `POD_SECURITY_GROUP_ENFORCING_MODE`,  which decides the traffic mode ["security group for pods"](https://docs.aws.amazon.com/eks/latest/userguide/security-groups-for-pods.html) feature.
The possible settings of  `POD_SECURITY_GROUP_ENFORCING_MODE` are:
    * `strict`:  all inbound/outbound traffic from pod with security group will be enforced by security group rules.
         * **this is the default mode, and is the existing behavior before this change**.
   * `standard`: the traffic of pod with security group behaves same as pods without security group, except that each pod occupies dedicated ENI
        *  inbound traffic to pod with security group from another host will be enforced by security group rules.
        *  outbound traffic from pod with security group to another host in same VPC will be enforced by security group rules.
        *  outbound traffic from pod with security group  to IP address outside VPC
            * if externalSNAT enabled, traffic won't be SNATed, thus will be enforced by security group rules.
            * if externalSNAT disabled, traffic will be SNATed via eth0, thus will only be enforced by security group associated with eth0.
        * inbound/bound traffic from another pod on same host or another service on same host(such as kubelet/nodeLocalDNS) won't be enforced by security group rules.
2. This PR did several refactor to make original code more clean and maintainable.

1. rewrote test cases for `driver.go`
    * increases test coverage from 34% to 95%.
    * actually test API calls instead of abuse `gomock.any`
    * fixes old broken tests

**Automation added to e2e**:

N/A

No, the default value of "POD_SECURITY_GROUP_ENFORCING_MODE" if not set is "standard".

```release-note
Introduce a new knob named POD_SECURITY_GROUP_ENFORCING_MODE to decide the traffic mode for security group for pods feature. 
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,4 @@
+
 # amazon-vpc-cni-k8s
 
 Networking plugin for pod networking in [Kubernetes](https://kubernetes.io/) using [Elastic Network Interfaces](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html) on AWS.
@@ -420,6 +421,31 @@ Any of the WARM targets do not impact the scale of the branch ENI pods so you wi
 
 
 **NOTE!** Toggling `ENABLE_POD_ENI` from `true` to `false` will not detach the Trunk ENI from instance. To delete/detach the Trunk ENI from instance, you need recycle the instance.
+
+
+---
+
+#### `POD_SECURITY_GROUP_ENFORCING_MODE` (v1.11.0+)
+
+Type: String
+
+Default: `strict`
+
+Valid Values: `strict`, `standard`
+
+Once `ENABLE_POD_ENI` is set to `true`, this value controls how the traffic of pods with security group behaves.
+
+  * `strict` mode: all inbound/outbound traffic from pod with security group will be enforced by security group rules. This is the **default** mode if POD_SECURITY_GROUP_ENFORCING_MODE is not set.
+
+  * `standard` mode: the traffic of pod with security group behaves same as pods without security group, except that each pod occupies a dedicated branch ENI.
+    * inbound traffic to pod with security group from another host will be enforced by security group rules.
+    * outbound traffic from pod with security group to another host in same VPC will be enforced by security group rules.
+    * inbound/outbound traffic from another pod on same host or another service on same host(such as kubelet/nodeLocalDNS) won't be enforced by security group rules.
+    *  outbound traffic from pod with security group to IP address outside VPC
+        * if externalSNAT enabled, traffic won't be SNATed, thus will be enforced by security group rules.
+        * if externalSNAT disabled, traffic will be SNATed via eth0, thus will only be enforced by security group associated with eth0.
+
+**NOTE!**: To make new behavior be in effect after switching the mode, existing pods with security group must be recycled. Alternatively you can restart the nodes as well.
 
 ---
 
--- a/cmd/routed-eni-cni-plugin/cni.go
+++ b/cmd/routed-eni-cni-plugin/cni.go
@@ -25,6 +25,10 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/cniutils"
+
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/sgpp"
+
 	"github.com/containernetworking/cni/pkg/skel"
 	"github.com/containernetworking/cni/pkg/types"
 	"github.com/containernetworking/cni/pkg/types/current"
@@ -45,7 +49,6 @@ import (
 
 const ipamdAddress = "127.0.0.1:50051"
 
-const vlanInterfacePrefix = "vlan"
 const dummyVlanInterfacePrefix = "dummy"
 
 var version string
@@ -61,6 +64,9 @@ type NetConf struct {
 
 	// MTU for eth0
 	MTU string `json:"mtu"`
+
+	// PodSGEnforcingMode is the enforcing mode for Security groups for pods feature
+	PodSGEnforcingMode sgpp.EnforcingMode `json:"podSGEnforcingMode"`
 
 	PluginLogFile string `json:"pluginLogFile"`
 
@@ -91,8 +97,9 @@ func init() {
 func LoadNetConf(bytes []byte) (*NetConf, logger.Logger, error) {
 	// Default config
 	conf := NetConf{
-		MTU:        "9001",
-		VethPrefix: "eni",
+		MTU:                "9001",
+		VethPrefix:         "eni",
+		PodSGEnforcingMode: sgpp.DefaultEnforcingMode,
 	}
 
 	if err := json.Unmarshal(bytes, &conf); err != nil {
@@ -208,9 +215,10 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 
 	// Non-zero value means pods are using branch ENI
 	if r.PodVlanId != 0 {
-		hostVethName = generateHostVethName(vlanInterfacePrefix, string(k8sArgs.K8S_POD_NAMESPACE), string(k8sArgs.K8S_POD_NAME))
-		err = driverClient.SetupPodENINetwork(hostVethName, args.IfName, args.Netns, v4Addr, v6Addr, int(r.PodVlanId), r.PodENIMAC,
-			r.PodENISubnetGW, int(r.ParentIfIndex), mtu, log)
+		hostVethNamePrefix := sgpp.BuildHostVethNamePrefix(conf.VethPrefix, conf.PodSGEnforcingMode)
+		hostVethName = generateHostVethName(hostVethNamePrefix, string(k8sArgs.K8S_POD_NAMESPACE), string(k8sArgs.K8S_POD_NAME))
+		err = driverClient.SetupBranchENIPodNetwork(hostVethName, args.IfName, args.Netns, v4Addr, v6Addr, int(r.PodVlanId), r.PodENIMAC,
+			r.PodENISubnetGW, int(r.ParentIfIndex), mtu, conf.PodSGEnforcingMode, log)
 
 		// This is a dummyVlanInterfaceName generated to identify dummyVlanInterface
 		// which will be created for PPSG scenario to pass along the vlanId information
@@ -226,7 +234,7 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 		// build hostVethName
 		// Note: the maximum length for linux interface name is 15
 		hostVethName = generateHostVethName(conf.VethPrefix, string(k8sArgs.K8S_POD_NAMESPACE), string(k8sArgs.K8S_POD_NAME))
-		err = driverClient.SetupNS(hostVethName, args.IfName, args.Netns, v4Addr, v6Addr, int(r.DeviceNumber), r.VPCv4CIDRs, r.UseExternalSNAT, mtu, log)
+		err = driverClient.SetupPodNetwork(hostVethName, args.IfName, args.Netns, v4Addr, v6Addr, int(r.DeviceNumber), mtu, log)
 	}
 
 	if err != nil {
@@ -314,42 +322,14 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 		return errors.Wrap(err, "del cmd: failed to load k8s config from args")
 	}
 
-	prevResult, ok := conf.PrevResult.(*current.Result)
-
-	// Try to use prevResult if available
-	// prevResult might not be availabe, if we are still using older cni spec < 0.4.0.
-	// So we should fallback to the old clean up method
-	if ok {
-		dummyVlanInterfaceName := generateHostVethName(dummyVlanInterfacePrefix, string(k8sArgs.K8S_POD_NAMESPACE), string(k8sArgs.K8S_POD_NAME))
-		for _, iface := range prevResult.Interfaces {
-			if iface.Name == dummyVlanInterfaceName {
-				podVlanId, err := strconv.Atoi(iface.Mac)
-				if err != nil {
-					log.Errorf("Failed to parse vlanId from prevResult: %v", err)
-					return errors.Wrap(err, "del cmd: failed to parse vlanId from prevResult")
-				}
-
-				// podVlanID can not be 0 as we add dummyVlanInterface only for ppsg
-				// if it is 0 then we should return an error
-				if podVlanId == 0 {
-					log.Errorf("Found SG pod:%s namespace:%s with 0 vlanID", k8sArgs.K8S_POD_NAME, k8sArgs.K8S_POD_NAMESPACE)
-					return errors.Wrap(err, "del cmd: found Incorrect 0 vlandId for ppsg")
-				}
-
-				if isNetnsEmpty(args.Netns) {
-					log.Infof("Ignoring TeardownPodENI as Netns is empty for SG pod:%s namespace: %s containerID:%s", k8sArgs.K8S_POD_NAME, k8sArgs.K8S_POD_NAMESPACE, k8sArgs.K8S_POD_INFRA_CONTAINER_ID)
-					return nil
-				}
-
-				err = cleanUpPodENI(podVlanId, log, args.ContainerID, driverClient)
-				if err != nil {
-					return err
-				}
-				log.Infof("Received del network response for pod %s namespace %s sandbox %s with vlanId: %v", string(k8sArgs.K8S_POD_NAME),
-					string(k8sArgs.K8S_POD_NAMESPACE), string(k8sArgs.K8S_POD_INFRA_CONTAINER_ID), podVlanId)
-				return nil
-			}
-		}
+	handled, err := tryDelWithPrevResult(driverClient, conf, k8sArgs, args.IfName, args.Netns, log)
+	if err != nil {
+		return errors.Wrap(err, "del cmd: failed to delete with prevResult")
+	}
+	if handled {
+		log.Infof("Handled CNI del request with prevResult: ContainerID(%s) Netns(%s) IfName(%s) PodNamespace(%s) PodName(%s)",
+			args.ContainerID, args.Netns, args.IfName, string(k8sArgs.K8S_POD_NAMESPACE), string(k8sArgs.K8S_POD_NAME))
+		return nil
 	}
 
 	// notify local IP address manager to free secondary IP
@@ -420,9 +400,9 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 				log.Infof("Ignoring TeardownPodENI as Netns is empty for SG pod:%s namespace: %s containerID:%s", k8sArgs.K8S_POD_NAME, k8sArgs.K8S_POD_NAMESPACE, k8sArgs.K8S_POD_INFRA_CONTAINER_ID)
 				return nil
 			}
-			err = driverClient.TeardownPodENINetwork(int(r.PodVlanId), log)
+			err = driverClient.TeardownBranchENIPodNetwork(addr, int(r.PodVlanId), conf.PodSGEnforcingMode, log)
 		} else {
-			err = driverClient.TeardownNS(addr, int(r.DeviceNumber), log)
+			err = driverClient.TeardownPodNetwork(addr, int(r.DeviceNumber), log)
 		}
 
 		if err != nil {
@@ -436,14 +416,43 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 	return nil
 }
 
-func cleanUpPodENI(podVlanId int, log logger.Logger, containerId string, driverClient driver.NetworkAPIs) error {
-	err := driverClient.TeardownPodENINetwork(podVlanId, log)
-	if err != nil {
-		log.Errorf("Failed on TeardownPodNetwork for container ID %s: %v",
-			containerId, err)
-		return errors.Wrap(err, "del cmd: failed on tear down pod network")
+// tryDelWithPrevResult will try to process CNI delete request without IPAMD.
+// returns true if the del request is handled.
+func tryDelWithPrevResult(driverClient driver.NetworkAPIs, conf *NetConf, k8sArgs K8sArgs, contVethName string, netNS string, log logger.Logger) (bool, error) {
+	// prevResult might not be available, if we are still using older cni spec < 0.4.0.
+	prevResult, ok := conf.PrevResult.(*current.Result)
+	if !ok {
+		return false, nil
 	}
-	return nil
+
+	dummyIfaceName := generateHostVethName(dummyVlanInterfacePrefix, string(k8sArgs.K8S_POD_NAMESPACE), string(k8sArgs.K8S_POD_NAME))
+	_, dummyIface, found := cniutils.FindInterfaceByName(prevResult.Interfaces, dummyIfaceName)
+	if !found {
+		return false, nil
+	}
+	podVlanID, err := strconv.Atoi(dummyIface.Mac)
+	if err != nil || podVlanID == 0 {
+		return true, errors.Errorf("malformed vlanID in prevResult: %s", dummyIface.Mac)
+	}
+	if isNetnsEmpty(netNS) {
+		log.Infof("Ignoring TeardownPodENI as Netns is empty for SG pod:%s namespace: %s containerID:%s", k8sArgs.K8S_POD_NAME, k8sArgs.K8S_POD_NAMESPACE, k8sArgs.K8S_POD_INFRA_CONTAINER_ID)
+		return true, nil
+	}
+
+	containerIfaceIndex, _, found := cniutils.FindInterfaceByName(prevResult.Interfaces, contVethName)
+	if !found {
+		return false, errors.Errorf("cannot find contVethName %s in prevResult", contVethName)
+	}
+	containerIPs := cniutils.FindIPConfigsByIfaceIndex(prevResult.IPs, containerIfaceIndex)
+	if len(containerIPs) != 1 {
+		return false, errors.Errorf("found %d containerIP for %v in prevResult", len(containerIPs), contVethName)
+	}
+	containerIP := containerIPs[0].Address
+
+	if err := driverClient.TeardownBranchENIPodNetwork(&containerIP, podVlanID, conf.PodSGEnforcingMode, log); err != nil {
+		return true, err
+	}
+	return true, nil
 }
 
 // Scope usage of this function to only SG pods scenario (https://harbinger.amazon.com/notices/65203)
--- a/cmd/routed-eni-cni-plugin/driver/driver.go
+++ b/cmd/routed-eni-cni-plugin/driver/driver.go
@@ -21,6 +21,8 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/sgpp"
+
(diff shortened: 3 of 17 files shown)


jayanthvn:
@M00nF1sh - Can you please run "make format"?

pkg/sgpp/constants.go
```diff
@@ -0,0 +1,15 @@
...
+type EnforcingMode string
+
+const (
+	EnforcingModeStrict   EnforcingMode = "strict"
+	EnforcingModeStandard EnforcingMode = "standard"
+)
+
+const (
+	// DefaultEnforcingMode is the default enforcing mode if not specified explicitly.
+	DefaultEnforcingMode EnforcingMode = EnforcingModeStrict
```
cgchinmay, line 12: wonder if we should rename "strict" -> "default" instead and "standard" -> something like "partial".  Since standard tends to make you feel it's the standard when actually it's not. @M00nF1sh your call, just my thought.

pkg/sgpp/utils.go
```diff
@@ -0,0 +1,35 @@
...
+	if !found {
+		return DefaultEnforcingMode
+	}
+	switch envVal {
+	case string(EnforcingModeStrict):
+		return EnforcingModeStrict
+	case string(EnforcingModeStandard):
+		return EnforcingModeStandard
+	default:
+		return DefaultEnforcingMode
```
cgchinmay, line 33: It appears that EnforcingModeStrict is same as DefaultEnforcingMode. If this is not going to change in future then I think we can just check for string(EnforcingModeStandard) and if not then return the DefaultEnforcingMode

pkg/sgpp/utils.go
```diff
@@ -0,0 +1,35 @@
...
+		return hostVethNamePrefix
+	default:
+		return vlanInterfacePrefix
+	}
+}
+
+// LoadEnforcingModeFromEnv tries to load the enforcing mode from environment variable and fall-back to DefaultEnforcingMode.
+func LoadEnforcingModeFromEnv() EnforcingMode {
+	envVal, found := os.LookupEnv(envEnforcingMode)
+	if !found {
```
cgchinmay, line 24: I think we can remove this if !found check as envVal will be empty and switch will return DefaultEnforcingMode ? 
https://pkg.go.dev/os#LookupEnv

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -434,14 +414,65 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
...
+
+	containerIfaceIndex, _, found := findCNIInterfaceByName(prevResult.Interfaces, contVethName)
+	if !found {
+		return false, nil
+	}
+	containerIPs := findCNIIPConfigsByIfaceIndex(prevResult.IPs, containerIfaceIndex)
+	if len(containerIPs) != 1 {
+		return false, nil
+	}
+	containerIP := containerIPs[0].Address
```
cgchinmay, line 450: Could you add a comment on what containerIP represent ? why is it always at index 0 ?

pkg/sgpp/constants.go
M00nF1sh: this was proposed by our PM. we prefer to use "standard" as the default mode in the future, that's why we have the naming. 
I'm with him as "standard" can be interpreted as similar behavior with "standard pods"(i.e. pods without security group)

pkg/sgpp/utils.go
M00nF1sh: we do prefer to make the standard the default mode in the future.  however, it depends on feedbacks and usage once this feature rolls out.

pkg/sgpp/utils.go
M00nF1sh: make sense, this can simplify it a bit. updated.

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -91,8 +95,9 @@ func init() {
 func LoadNetConf(bytes []byte) (*NetConf, logger.Logger, error) {
 	// Default config
 	conf := NetConf{
-		MTU:        "9001",
-		VethPrefix: "eni",
+		MTU:                "9001",
+		VethPrefix:         "eni",
+		PodSGEnforcingMode: sgpp.DefaultEnforcingMode,
```
achevuru, line 100: Why do we want SG mode in `NetConf` struct? Is this needed because we no longer can rely on `podVlanID != 0` check now that we have 2 SGPP modes?

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -434,14 +417,58 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
...
+func tryDelWithPrevResult(driverClient driver.NetworkAPIs, conf *NetConf, k8sArgs K8sArgs, contVethName string, log logger.Logger) (bool, error) {
+	// prevResult might not be available, if we are still using older cni spec < 0.4.0.
+	prevResult, ok := conf.PrevResult.(*current.Result)
+	if !ok {
+		return false, nil
 	}
-	return nil
+
+	dummyIfaceName := generateHostVethName(dummyVlanInterfacePrefix, string(k8sArgs.K8S_POD_NAMESPACE), string(k8sArgs.K8S_POD_NAME))
+	_, dummyIface, found := findCNIInterfaceByName(prevResult.Interfaces, dummyIfaceName)
```
achevuru, line 430: nit: I think this function can be named in generic fashion (i.e.,) `findInterfaceByName`

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -54,11 +53,16 @@ const (
...
 // NetworkAPIs defines network API calls
 type NetworkAPIs interface {
-	SetupNS(hostVethName string, contVethName string, netnsPath string, v4Addr *net.IPNet, v6Addr *net.IPNet, deviceNumber int, vpcCIDRs []string, useExternalSNAT bool, mtu int, log logger.Logger) error
-	TeardownNS(addr *net.IPNet, deviceNumber int, log logger.Logger) error
-	SetupPodENINetwork(hostVethName string, contVethName string, netnsPath string, v4Addr *net.IPNet, v6Addr *net.IPNet, vlanID int, eniMAC string,
-		subnetGW string, parentIfIndex int, mtu int, log logger.Logger) error
-	TeardownPodENINetwork(vlanID int, log logger.Logger) error
+	// SetupPodNetwork sets up pod network for normal ENI based pods
+	SetupPodNetwork(hostVethName string, contVethName string, netnsPath string, v4Addr *net.IPNet, v6Addr *net.IPNet, deviceNumber int, mtu int, log logger.Logger) error
+	// TeardownPodNetwork clean up pod network for normal ENI based pods
```
achevuru, line 58: nit: clean up -> cleans up

cmd/routed-eni-cni-plugin/cni.go
M00nF1sh: yes, it's needed since both SGPP mode relies on podVlanID.
the VPC resource controller is agnostic of the SGPP mode, it's purely client-side thing.

cmd/routed-eni-cni-plugin/cni.go
M00nF1sh: i used `findCNIInterfaceByName` since `findInterfaceByName` is too generic as this function is specific to CNI.

maybe we can move it to some dedicated package such as cniUtils.go?

cmd/routed-eni-cni-plugin/driver/driver.go
M00nF1sh: :D fixed

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -267,354 +271,358 @@ func WaitForAddressesToBeStable(ifName string, timeout time.Duration) error {
...
-			Gw:        gw,
-			Table:     vlanTableID,
-		},
+func (n *linuxNetwork) teardownIPBasedContainerRouteRules(containerAddr *net.IPNet, rtTable int, log logger.Logger) error {
+	toContainerRule := n.netLink.NewRule()
+	toContainerRule.Dst = containerAddr
+	toContainerRule.Priority = toContainerRulePriority
+	toContainerRule.Table = unix.RT_TABLE_MAIN
+	if err := n.netLink.RuleDel(toContainerRule); err != nil && !containsNoSuchRule(err) {
+		return errors.Wrapf(err, "failed to delete toContainer rule, containerAddr=%s, rtTable=%v", containerAddr.String(), "main")
```
achevuru, line 513: `DEL` is a best effort anyway. So, why do we want to return `error` from here? 

https://github.com/containernetworking/cni/blob/main/SPEC.md#del-remove-container-from-network-or-un-apply-modifications

I see, we currently log an error and move on if we fail to delete the route pointing to container IP to container's hostVeth (i.e.,) `toContainerRule` - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/cmd/routed-eni-cni-plugin/driver/driver.go#L545

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -267,354 +271,358 @@ func WaitForAddressesToBeStable(ifName string, timeout time.Duration) error {
...
 
+	hostVeth, err := n.setupVeth(hostVethName, contVethName, netnsPath, v4Addr, v6Addr, mtu, log)
 	if err != nil {
-		log.Errorf("Failed to add toContainer rule for %s err=%v, ", addrHostAddr.String(), err)
-		return errors.Wrap(err, "setupNS network: failed to add toContainer")
+		return errors.Wrapf(err, "SetupBranchENIPodNetwork: failed to setup veth pair")
 	}
 
-	log.Infof("Added toContainer rule for %s", addrHostAddr.String())
+	// clean up any previous hostVeth ip rule recursively. (when pod with same name are recreated multiple times)
```
achevuru, line 327: Do we still need this? We're now using `prevResult` to drive the deletion and so (ideally) shouldn't run in to the issue anymore - #1705

cmd/routed-eni-cni-plugin/driver/driver.go
M00nF1sh: The original code merely logs the error if it failed to delete the `toContainerRule`, but it will error out if failed to delete fromContainer rule: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/cmd/routed-eni-cni-plugin/driver/driver.go#L556

i normalized this two logic by check for the error condition(!containsNoSuchRule)  to ignore rule not exists error, and return err only if it's other errors.

cmd/routed-eni-cni-plugin/driver/driver.go
M00nF1sh: I keep this logic for safety purpose. but  i don't know why the original code explicitly delete this hostVeth but not the hostVlan rule here.

cmd/routed-eni-cni-plugin/driver/driver.go
achevuru: Ok. Since `toContainerRule` is in Main table and pointing to Container's `hostVeth` interface, the rule (ideally) should be purged once we delete the veth pair right? So, even if it fails here for whatever reason the veth deletion operation should take care of it. Am I missing something here? With deletes, since kubelet retries again ...we might be stuck in the delete loop..

`fromContainerRule` on the other hand is pointing to one of the secondary ENIs and the secondary ENI can hold multiple pods and will stay active even after the current pod is deleted.

cmd/routed-eni-cni-plugin/driver/driver.go
achevuru: Yes, if I remember correctly we can end up deleting the wrong Pod because of the race condition (i.e.,) by the time CNI makes an API Server call, the old pod might've been already deleted so it will not be able to derive the vlan id and upon retry it will get the pod spec of the new pod (with same name and ns) and uses it's vlan id to clean up the rules. Will result in CNI incorrectly deleting the vlan rules of the new pod instead of the old pod.

achevuru: Agree a comment here on why we were only deleting the `hostVeth` here would've been helpful.

cmd/routed-eni-cni-plugin/cni.go
achevuru: Understand, we're looking for interfaces created by VPC CNI but the function actually can work for any interface name (in the result struct) (i.e.,) the function itself is implemented in generic fashion and hence the name suggestion. I do like the suggestion of making it an util function.

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -267,354 +271,358 @@ func WaitForAddressesToBeStable(ifName string, timeout time.Duration) error {
...
-	hostVeth, err := setupVeth(hostVethName, contVethName, netnsPath, v4Addr, v6Addr, netLink, ns, mtu, procSys, log)
+	hostVeth, err := n.setupVeth(hostVethName, contVethName, netnsPath, v4Addr, v6Addr, mtu, log)
 	if err != nil {
-		return errors.Wrapf(err, "setupNS network: failed to setup veth pair.")
+		return errors.Wrapf(err, "SetupPodNetwork: failed to setup veth pair")
 	}
 
-	log.Debugf("Setup host route outgoing hostVeth, LinkIndex %d", hostVeth.Attrs().Index)
+	var containerAddr *net.IPNet
+	if v4Addr != nil {
```
achevuru, line 286: `v4Addr.IP.To4()` and `v6Addr.IP.To16()` are helpful to have here. For ex, if an IPv6 address ended being passed as `v4Addr` here - these checks can help catch them. `To4` will be helpful in particular to identify these mismatches.

README.md
```diff
@@ -421,6 +422,29 @@ Any of the WARM targets do not impact the scale of the branch ENI pods so you wi
...
+  * `strict` mode: all inbound/outbound traffic from pod with security group will be enforced by security group rules. This is the **default** mode if POD_SECURITY_GROUP_ENFORCING_MODE is not set.
+
+  * `standard` mode: the traffic of pod with security group behaves same as pods without security group, except that each pod occupies a dedicated branch ENI.
+    * inbound traffic to pod with security group from another host will be enforced by security group rules.
+    * outbound traffic from pod with security group to another host in same VPC will be enforced by security group rules.
+    * inbound/outbound traffic from another pod on same host or another service on same host(such as kubelet/nodeLocalDNS) won't be enforced by security group rules.
+    *  outbound traffic from pod with security group to IP address outside VPC
+        * if externalSNAT enabled, traffic won't be SNATed, thus will be enforced by security group rules.
+        * if externalSNAT disabled, traffic will be SNATed via eth0, thus will only be enforced by security group associated with eth0.
+
```
jayanthvn, line 447: Can we also call out on upgrade and downgrade steps?

cmd/routed-eni-cni-plugin/cni.go
jayanthvn: Yes it would be good if we can move it to utils folder. I had created for this PR - https://github.com/aws/amazon-vpc-cni-k8s/pull/1726/files

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -434,14 +417,58 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
...
+		return false, nil
+	}
+	podVlanID, err := strconv.Atoi(dummyIface.Mac)
+	if err != nil || podVlanID == 0 {
+		return true, errors.Errorf("malformed vlanID from prevResult: %s", dummyIface.Mac)
+	}
+
+	containerIfaceIndex, _, found := findCNIInterfaceByName(prevResult.Interfaces, contVethName)
+	if !found {
+		return false, nil
```
jayanthvn, line 441: Should we return "nil" here? It should be a hard failure right? Since we have the previous result but not able to find the interface.

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -434,14 +417,58 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
...
+	}
+	containerIPs := findCNIIPConfigsByIfaceIndex(prevResult.IPs, containerIfaceIndex)
+	if len(containerIPs) != 1 {
+		return false, nil
+	}
+	containerIP := containerIPs[0].Address
+
+	if err := driverClient.TeardownBranchENIPodNetwork(&containerIP, podVlanID, conf.PodSGEnforcingMode, log); err != nil {
+		return true, err
```
jayanthvn, line 450: Can we add a warning here? since this will return handled == true and print the log as handled with previous result.

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -267,354 +271,358 @@ func WaitForAddressesToBeStable(ifName string, timeout time.Duration) error {
...
-	if v4Addr != nil && v4Addr.IP.To4() != nil {
-		addrHostAddr = &net.IPNet{
-			IP:   v4Addr.IP,
-			Mask: net.CIDRMask(32, 32)}
-	} else if v6Addr != nil && v6Addr.IP.To16() != nil {
-		addrHostAddr = &net.IPNet{
-			IP:   v6Addr.IP,
-			Mask: net.CIDRMask(128, 128)}
+	rtTable := unix.RT_TABLE_MAIN
+	if deviceNumber > 0 {
```
jayanthvn, line 293: Before we used to add the container rule only if device number is > 0 - https://github.com/aws/amazon-vpc-cni-k8s/blob/8855418090a8c69b0c8561113322fa1191cbe5da/cmd/routed-eni-cni-plugin/driver/driver.go#L321-L331 but now we will continue to setup the rule. Is it added on purpose?

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -267,354 +271,358 @@ func WaitForAddressesToBeStable(ifName string, timeout time.Duration) error {
...
-	if isToContainer {
-		// Example: 512:	from all to 10.200.202.222 lookup main
-		containerRule.Dst = addr
-		containerRule.Priority = toContainerRulePriority
-	} else {
-		// Example: 1536:	from 10.200.202.222 to 10.200.0.0/16 lookup 2
-		containerRule.Src = addr
-		containerRule.Priority = fromContainerRulePriority
+		// note: older version CNI sets up multiple CIDR based from container rule, so we recursively delete them to be backwards-compatible.
+		if err := netLinkRuleDelAll(n.netLink, fromContainerRule); err != nil {
```
achevuru, line 524: Not following this comment. Here, we are dealing with `from a.b.c.d dev eth1` rule isn't it? How is CIDR playing a part here?

From what I see here - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/cmd/routed-eni-cni-plugin/driver/driver.go#L553 , we're only attempting to delete a rule that matches the Pod IP.

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -267,354 +271,358 @@ func WaitForAddressesToBeStable(ifName string, timeout time.Duration) error {
...
+	if err := n.teardownVlan(vlanID, log); err != nil {
+		return errors.Wrapf(err, "TeardownBranchENIPodNetwork: failed to teardown vlan")
+	}
+
+	// to handle the migration between different enforcingMode, we try to clean up rules under both mode since the pod might be setup with a different mode.
+	rtTable := vlanID + 100
+	if err := n.teardownIIFBasedContainerRouteRules(rtTable, log); err != nil {
+		return errors.Wrapf(err, "TeardownBranchENIPodNetwork: unable to teardown IIF based container routes and rules")
+	}
+	if err := n.teardownIPBasedContainerRouteRules(containerAddr, rtTable, log); err != nil {
```
jayanthvn, line 373: Correct me if I am wrong, should we return error here? because if this is not migration then we might not have one of the rules to delete right?

cmd/routed-eni-cni-plugin/driver/driver.go
M00nF1sh: @achevuru i can change it to tolerate the error to be safe to match old code, and log a warning instead only if it's not rule exists error. (i believe there shouldn't be cases trigger that)

cmd/routed-eni-cni-plugin/driver/driver.go
M00nF1sh: make sense, will add a comment here

cmd/routed-eni-cni-plugin/driver/driver.go
M00nF1sh: i believe the old code's To4 and To16 never triggers since the `addrHostAddr` will be nil and trigger nil-pointer exception if To4 and To16 fails. 
I removed to be align with our original PodENI code as well. 
```
var addrHostAddr *net.IPNet
	//We only support either v4 or v6 modes.
	if v4Addr != nil && v4Addr.IP.To4() != nil {
		addrHostAddr = &net.IPNet{
			IP:   v4Addr.IP,
			Mask: net.CIDRMask(32, 32)}
	} else if v6Addr != nil && v6Addr.IP.To16() != nil {
		addrHostAddr = &net.IPNet{
			IP:   v6Addr.IP,
			Mask: net.CIDRMask(128, 128)}
	}
```

README.md
M00nF1sh: make sense, will add

cmd/routed-eni-cni-plugin/cni.go
M00nF1sh: i want to fallback to old code if anything not expected here (thus it retains original behavior).

But i think it make sense to hard fail here as well, as cni spec should guarantee that.

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -434,14 +417,58 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
...
+		return true, errors.Errorf("malformed vlanID from prevResult: %s", dummyIface.Mac)
+	}
+
+	containerIfaceIndex, _, found := findCNIInterfaceByName(prevResult.Interfaces, contVethName)
+	if !found {
+		return false, nil
+	}
+	containerIPs := findCNIIPConfigsByIfaceIndex(prevResult.IPs, containerIfaceIndex)
+	if len(containerIPs) != 1 {
+		return false, nil
```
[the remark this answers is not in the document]
M00nF1sh: i want to fallback to old code if anything not expected here (thus it retains original behavior).

But i think it make sense to hard fail as well, will do some test about older CNI versions

cmd/routed-eni-cni-plugin/driver/driver.go
M00nF1sh: No, we will only add the "fromContainer" rule if the RTTable is not main.  (the check of deviceNumber > 0 is now replaced by rtTable != main inside the function been called )

cmd/routed-eni-cni-plugin/driver/driver.go
M00nF1sh: so previously on older CNI versions, the rules is like
```
from a.b.c.d dev to CIDR1 eth1
from a.b.c.d dev to CIDR2 eth1
```
That's why the code dels rules by SRC. 
I simplified the code to use the new introduced netLinkRuleDelAll to achieve same thing. which removes the hacky ipamd code dependency while remains backwards-compatible. (can delete multiple rules that was setup by older version of CNI)

cmd/routed-eni-cni-plugin/driver/driver.go
M00nF1sh: all the teardownIIFBasedContainerRouteRules and teardownIPBasedContainerRouteRules is designed to be safe to invoke if rules is not exists. 
so they can be called twice and return error safely :D

cmd/routed-eni-cni-plugin/driver/driver.go
M00nF1sh: after more thought, i believe we should still return an error here and let kubelet retry.
1. we already checked for ruleNotExists cases, the rule should be able to be deleted under all other case.
2. if the rule is not deleted, it won't be clean up automatically when veth is delete, only route will.

cmd/routed-eni-cni-plugin/cni.go
M00nF1sh: @achevuru @jayanthvn moved to a utils folder :D

cmd/routed-eni-cni-plugin/driver/driver.go
M00nF1sh: added a comment here.

README.md
M00nF1sh: added upgrade/downgrade steps

cmd/routed-eni-cni-plugin/cni.go
M00nF1sh: change it to hard fail :D

cmd/routed-eni-cni-plugin/cni.go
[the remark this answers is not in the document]
M00nF1sh: change it to hard fail :D

cmd/routed-eni-cni-plugin/cni.go
M00nF1sh: i think it's not needed as the error will be handled and logged. (handled is not used if err != nil)

cmd/routed-eni-cni-plugin/driver/driver.go
achevuru: I do agree that apart from rule not being present, it is extremely unlikely to get an error for this operation. If it did fail for some other reason, will retry help here though? If retry doesn't help, my concern is that we will be stuck forever in retry loop. Understand same can be said about `fromContainerrule` but I'm not sure  retrying the operation is going to help here. We can discuss.

cmd/routed-eni-cni-plugin/cni.go
achevuru: How will this work for upgrade cases if we hard fail here? Let's say, we created an SGPP pod with say v1.10.1, then we will not have the interface information filled up during add operation. Now let's say we move to v1.10.2(with cni spec 0.4.0) - and we now delete the SGPP pod, we will now receive `PrevResult` but the interface will be missing isn't it? and we should fall back to previous sgpp delete code flow. Am I missing something here?

cmd/routed-eni-cni-plugin/driver/driver.go
achevuru: Understood. Makes sense when we're moving from an older version that did setup rules per VPC CIDR.

cmd/routed-eni-cni-plugin/driver/driver.go
achevuru: So, it helps us catch if an incorrect address family IP address is passed in via v4Addr (or) v6Addr. Is the concern about a `nil` exception if these checks fail? I mean incorrect IP addresses being passed in here is an issue to begin with anyways. We can probably address the `nil` pointer concern instead.

M00nF1sh: chatted offline, we'll add a comment to assert that we expect v4/v6 to have correct IPFamily

jayanthvn (approved):

joebowbeer:
@M00nF1sh The description contains conflicting statements. I assume this one is wrong:

> >Does this change require updates to the CNI daemonset config files to work?:

> No, the default value of "POD_SECURITY_GROUP_ENFORCING_MODE" if not set is "standard".

I assume default is `strict` to agree with previous statement in description.

PR #1931, opened 2022-03-16
fix issues with IPv6 canary

abhipth:
**What type of PR is this?**
This addresses bunch of failures encountered during IPv6 Canary tests.

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/scripts/lib/add-on.sh
+++ b/scripts/lib/add-on.sh
@@ -8,6 +8,7 @@ function load_addon_details() {
 
   LATEST_ADDON_VERSION=$(echo "$DESCRIBE_ADDON_VERSIONS" | jq '.addons[0].addonVersions[0].addonVersion' -r)
   DEFAULT_ADDON_VERSION=$(echo "$DESCRIBE_ADDON_VERSIONS" | jq -r '.addons[].addonVersions[] | select(.compatibilities[0].defaultVersion == true) | .addonVersion')
+  EXISTING_SERVICE_ACCOUNT_ROLE_ARN=$(kubectl get serviceaccount -n kube-system aws-node -o json | jq '.metadata.annotations."eks.amazonaws.com/role-arn"' -r)
 }
 
 function wait_for_addon_status() {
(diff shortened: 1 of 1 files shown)


scripts/lib/add-on.sh
```diff
@@ -69,7 +82,13 @@ function install_add_on() {
...
 
   echo "installing addon $new_addon_version"
-  aws eks create-addon $ENDPOINT_FLAG --cluster-name "$CLUSTER_NAME" --addon-name $VPC_CNI_ADDON_NAME --resolve-conflicts OVERWRITE --addon-version $new_addon_version --region "$REGION"
+
+  # If installing the addon after deleting it once, we should retain the Service Account Role ARN of the Addon
+  if [ "$EXISTING_SERVICE_ACCOUNT_ROLE_ARN" != "null" ]; then
+     SA_ROLE_ARN_ARG="--service-account-role-arn $EXISTING_SERVICE_ACCOUNT_ROLE_ARN"
+  fi
+
+  aws eks create-addon $ENDPOINT_FLAG $SA_ROLE_ARN_ARG --cluster-name "$CLUSTER_NAME" --addon-name $VPC_CNI_ADDON_NAME --resolve-conflicts OVERWRITE --addon-version $new_addon_version --region "$REGION"
```
jayanthvn, line 91: if `$EXISTING_SERVICE_ACCOUNT_ROLE_ARN` is null then for the create-addon we would pass `$SA_ROLE_ARN_ARG` null, is that fine?

abhipth: I think $SA_ROLE_ARN_ARG would be an empty string which I believe should be fine, but I have not tested it :D

jayanthvn: right, it would be empty. Should be fine then.

jayanthvn (approved):
lgtm

scripts/lib/add-on.sh
achevuru: Yeah should count as an extra space

PR #1938, opened 2022-03-25
Fix cni integration test

sushrk:
**What type of PR is this?**
bug

CNI suite integration test failures

Ginkgo runs test in random order. The test namespace will be deleted in JustAfterEach block when PD tests are completed, and all other tests run after fail due to "namespace not found" error.
The test namespace creation & deletion is already done in BeforeSuite & AfterSuite, so this can be removed from PD test files. 

N/A

Run cni integration test

**Automation added to e2e**:
N/A

N/A

N/A

N/A

N/A


jayanthvn (approved):
lgtm

PR #1943, opened 2022-03-30
Timeout and reconcile when checking API server connectivity

prateekgogia:
**What type of PR is this?**
Bug

Some of the instance take ~2 minutes to become ready while other can connect in <30 seconds.

In the current implementation, request to the API server hangs if kube-proxy starts after aws-node pod, it doesn't time out. Container gets restarted after ~90-120 seconds and client reconnects again to the API server to check connectivity.
As part of this change kube client times out in 1 second if it fails to connect and will try to reconnect after 1 second. If its never able to connect to the API server container will be restarted like it does today.

I have tested this change on my cluster by adding ~20 nodes with this fix and without this fix. 

The following graph shows the time a given node stays in not ready state.

Test with CNI image - `amazon-k8s-cni:v1.10.2`, shows some of the instances were not ready for ~2 minutes in the API server
Screen Shot 2022-03-30 at 10 39 15 AM

Test image with this fix, shows all the nodes were ready in <30 seconds.
Screen Shot 2022-03-30 at 10 40 06 AM

**Automation added to e2e**:

No

No

No

No

No
```release-note

```

### DIFF
--- a/pkg/k8sapi/k8sutils.go
+++ b/pkg/k8sapi/k8sutils.go
@@ -2,10 +2,13 @@ package k8sapi
 
 import (
 	"fmt"
+	"os"
+	"time"
 
 	eniconfigscheme "github.com/aws/amazon-vpc-cni-k8s/pkg/apis/crd/v1alpha1"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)


pkg/k8sapi/k8sutils.go
```diff
@@ -85,15 +88,25 @@ func CheckAPIServerConnectivity() error {
 	if err != nil {
 		return err
 	}
-	clientSet, _ := kubernetes.NewForConfig(restCfg)
-
-	log.Infof("Testing communication with server")
-	version, err := clientSet.Discovery().ServerVersion()
+	restCfg.Timeout = 1 * time.Second
```
M00nF1sh, line 91: can we change this to be 5 second to be safe?
with some slow response cluster, 1 second might be too aggressive.

change to 5 second shouldn't impact the this optimization much.

prateekgogia: How about 2 second timeout? because if a API server is taking more than 2 seconds to respond to version query its too slow and we want to know why and fix the issue.

M00nF1sh (approved):
/lgtm

pkg/k8sapi/k8sutils.go
matti: apiserver will easily take more than 2 seconds when scaling 10,000 pods, 600 nodes

prateekgogia: Version call shouldn't be impacted by the load in the server. 
If it's happening we will root cause, try and fix it, feel free to cut a support ticket and we can help debug.

matti: ok, didn't read the actual code, just shouting random comments 👍

PR #1984, opened 2022-05-06
Multus release manifest

cgchinmay:
**What type of PR is this?**
feature

N/A

Manifest update to install latest Multus image

Verified against local cluster

**Automation added to e2e**:

No

No


jayanthvn (approved):
lgtm. Maybe in future we need to automate, I mean generate from jsonnet or helm if these manifests are not vended from upstream.

PR #1998, opened 2022-06-02
fix integration test script

sushrk:
**What type of PR is this?**
bug

1. Previously ginkgo timeout was not being honored for ipamd tests. After changing to run test in each folder separately, the issue has been fixed.
2. Add PENDING_RELEASE label to skip tests not included in any release yet.

Before this change, test was timing out after 1h(default ginkgo timeout) even when test timeout was set to 120m:
```
FAIL! - Interrupted by Timeout
--- FAIL: TestIPAMD (3626.82s)

Ginkgo ran 1 suite in 1h0m30.041509644s
```
ipamd test is run for >1h now:
```
Ginkgo ran 1 suite in 1h19m15.127402222s
```

**Automation added to e2e**:
N/A

N/A

N/A

N/A
N/A

### DIFF
--- a/scripts/run-cni-release-tests.sh
+++ b/scripts/run-cni-release-tests.sh
@@ -15,9 +15,8 @@
 set -e
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-TEST_DIR="$SCRIPT_DIR/../test"
-INTEGRATION_TEST_DIR="$TEST_DIR/integration"
-CALICO_TEST_DIR="$TEST_DIR/e2e/calico"
+INTEGRATION_TEST_DIR="$SCRIPT_DIR/../test/integration"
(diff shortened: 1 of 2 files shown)


orsenthil (dismissed):
LGTM.

I am not sure why that pod events is a version specific test case. I thought, it should run on all VPC CNI.

Minor suggestion. Instead of "PENDING RELEASE", I was thinking of  `[VERSION_SPECIFIC]` is a better annotation.

sushrk:
> 
> LGTM.
> 
> I am not sure why that pod events is a version specific test case. I thought, it should run on all VPC CNI.
> 
> Minor suggestion. Instead of "PENDING RELEASE", I was thinking of `[VERSION_SPECIFIC]` is a better annotation.

The changes required to raise aws-node pod events when permissions are missing is not released yet. So this test fails when run against the default CNI version in a new test cluster.

orsenthil (approved):
LGTM.

PR #2004, opened 2022-06-08
Enable Calico on ARM64 and add configureable flags for Calico installation

haouc:
Enable ARM for Calico integration tests on latest Calico version. Also fix a failing cause when running multiple times of the test on different settings.
- support ARM64 via Calico multi-arch image
- add flag InstallCalico 
- use instance-type flag to direct Calico tests

####Tests####
Tests were done and passed from dev cluster in us-west-2
- amd64
- arm64
- amd64 with prefix delegation
- arm64 with prefix delegation
```
Ran 22 of 22 Specs in 238.575 seconds
SUCCESS! -- 22 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS

Ginkgo ran 1 suite in 4m6.691385569s
Test Suite Passed
```

**What type of PR is this?**

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/scripts/run-integration-tests.sh
+++ b/scripts/run-integration-tests.sh
@@ -31,7 +31,7 @@ ARCH=$(go env GOARCH)
 : "${RUNNING_PERFORMANCE:=false}"
 : "${RUN_CALICO_TEST:=false}"
 : "${RUN_LATEST_CALICO_VERSION:=false}"
-: "${CALICO_VERSION:=3.22.0}"
+: "${CALICO_VERSION:=v3.23.0}"
 : "${RUN_CALICO_TEST_WITH_PD:=true}"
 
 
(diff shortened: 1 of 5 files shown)


scripts/run-integration-tests.sh
```diff
@@ -31,7 +31,7 @@ ARCH=$(go env GOARCH)
 : "${RUNNING_PERFORMANCE:=false}"
 : "${RUN_CALICO_TEST:=false}"
 : "${RUN_LATEST_CALICO_VERSION:=false}"
-: "${CALICO_VERSION:=3.22.0}"
+: "${CALICO_VERSION:=v3.23.0}"
```
jayanthvn, line 34: here there is v3.23.0 but in scripts/run-cni-release-tests.sh it is without a "v" so is it a typo maybe?

haouc: I have tested calico from running ```./scripts/run-integration-tests.sh```, but not tested ```run-cni-release-tests.sh```. Thus I didn't append the "v" as a prefix. Since they seem using the same test suite to run Calico test, I think it is ok to do the same for consistence. Let me add it as well. Thanks for pointing it out.

jayanthvn (approved):

PR #2029, opened 2022-07-11
Re-use logger instance

vikasmb:
**What type of PR is this?**
bug-fix

- Existing logger initialization constructed different logger
  instances upon call to Get() method.
- Fixed the initialization logic to re-use the logger instance.

- Logger instances were holding references to ipamd.log file even after rotation leading to leaked file handles.

Deployed this fix on a cluster and verified that leaked file handles for ipamd log file are no longer present.
**Automation added to e2e**:

No
No

No

No
```release-note

```

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -691,7 +691,7 @@ func (c *IPAMContext) decreaseDatastorePool(interval time.Duration) {
 
 // tryFreeENI always tries to free one ENI
 func (c *IPAMContext) tryFreeENI() {
-	if c.isTerminating() {
+	if c.isTerminating() || c.isNodeNonSchedulable() {
 		log.Debug("AWS CNI is terminating, not detaching any ENIs")
 		return
 	}
@@ -780,7 +780,7 @@ func (c *IPAMContext) increaseDatastorePool(ctx context.Context) {
 		}
 	}
 
-	if c.isTerminating() {
+	if c.isTerminating() || c.isNodeNonSchedulable() {
 		log.Debug("AWS CNI is terminating, will not try to attach any new IPs or ENIs right now")
 		return
 	}
@@ -1804,6 +1804,34 @@ func (c *IPAMContext) setTerminating() {
 
 func (c *IPAMContext) isTerminating() bool {
 	return atomic.LoadInt32(&c.terminating) > 0
+}
+
+func (c *IPAMContext) isNodeNonSchedulable() bool {
+	ctx := context.TODO()
+
+	request := types.NamespacedName{
+		Name: c.myNodeName,
+	}
+
+	node := &corev1.Node{}
+	// Find my node
+	err := c.cachedK8SClient.Get(ctx, request, node)
+	if err != nil {
+		log.Errorf("Failed to get node while determining schedulability: %v", err)
+		return false
+	}
+	log.Debugf("Node found %q - no of taints - %d", node.Name, len(node.Spec.Taints))
+	taintToMatch := &corev1.Taint{
+		Key:    "node.kubernetes.io/unschedulable",
+		Effect: corev1.TaintEffectNoSchedule,
+	}
+	for _, taint := range node.Spec.Taints {
+		if taint.MatchTaint(taintToMatch) {
+			return true
+		}
+	}
+
+	return false
 }
 
 // GetConfigForDebug returns the active values of the configuration env vars (for debugging purposes).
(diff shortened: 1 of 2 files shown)


orsenthil (approved):
Excellent Fix.

pkg/utils/logger/logger.go
```diff
@@ -47,22 +47,17 @@ type Logger interface {
 
 // Get returns an default instance of the zap logger
 func Get() Logger {
-	var logf = &structuredLogger{}
```
orsenthil, line 50: Why are we removing this structuredLogger reference? and isEmpty() call?

orsenthil (commented):
I have a question on the change too.

cmd/aws-k8s-agent/main.go
```diff
@@ -30,11 +30,7 @@ func main() {
 
 func _main() int {
 	//Do not add anything before initializing logger
-	logConfig := logger.Configuration{
-		LogLevel:    logger.GetLogLevel(),
-		LogLocation: logger.GetLogLocation(),
-	}
-	log := logger.New(&logConfig)
+	log := logger.Get()
```
achevuru, line 33: Since, we go through this only during initial bootup - the leaked file handles show up only if the `aws-node` pod on the node goes through a restart for some reason? I thought we've a known issue in the upstream logger lib and so curious why we are not running in to that? Would be good to expand on this in the PR description.

pkg/utils/logger/logger.go
jayanthvn: Here logf will always be empty, hence it was removed.

But I am thinking should we instead implement isEmpty() for the global variable and assign the local variable to global variable if the global is empty. Just to keep the code readable.

jayanthvn:
Also can we add the UT which you tried? Using that as base commit we can start adding more UTs as follow up.

pkg/utils/logger/logger.go
vikasmb: Existing logic was 
```
var logf = &structuredLogger{}
	if logf.isEmpty() {
```	

```
func (logf *structuredLogger) isEmpty() bool {
	return logf.zapLogger == nil
}
```

Through log messages, it was confirmed that each package was obtaining it's own logger instance:
Test execution logs (prior to the change)
```
"level":"info","ts":"2022-07-07T23:02:41.525Z","caller":"eniconfig/eniconfig.go:61","msg":"******** : Initialized new logger"}
{"level":"info","ts":"2022-07-07T23:02:41.526Z","caller":"networkutils/network.go:128","msg":"******** : Initialized new logger"}
{"level":"info","ts":"2022-07-07T23:02:41.531Z","caller":"awssession/session.go:39","msg":"******** : Initialized new logger"}
{"level":"info","ts":"2022-07-07T23:02:41.531Z","caller":"ec2wrapper/ec2wrapper.go:21","msg":"******** : Initialized new logger"}
{"level":"info","ts":"2022-07-07T23:02:41.537Z","caller":"k8sapi/k8sutils.go:23","msg":"******** : Initialized new logger"}
{"level":"info","ts":"2022-07-07T23:02:41.537Z","caller":"eventrecorder/eventrecorder.go:33","msg":"******** : Initialized new logger"}
{"level":"info","ts":"2022-07-07T23:02:41.538Z","caller":"awsutils/awsutils.go:81","msg":"******** : Initialized new logger"}
```

Thus, the extra method was removed and nil check was moved in-line, keeping the code simple and easy to read.

cmd/aws-k8s-agent/main.go
vikasmb: >>  the leaked file handles show up only if the aws-node pod on the node goes through a restart for some reason?
No, leaked file handles show up with the long running aws-node ipamd daemon. If it is restarted, leaked file handles would disappear.
Core of the issue is that different packages used to obtain different logger instances (pointing to the same underlying file) and upon rotation, file reference would be updated in one logger but not in others, leading to leaked file handle. This change fixes the situation by ensuring that a single logger instance is ever created and the rotation logic within lumberjack ensures that rotation and file closure is properly handled. Unit-test with a single logger instance doing file rotation correctly was confirmed by using delve debugger.

Regarding the upstream issue of https://github.com/natefinch/lumberjack/issues/81, it is not directly applicable in case of ipamd as transient goroutines holding on to logger instances were not observed (using dlv).

pkg/utils/logger/logger.go
orsenthil: Understood, and that was the only place it was used, so helps to remove it.

Thank you.

orsenthil:
LGTM, Vikas. All my comments are addressed.

jayanthvn (approved):
lgtm

PR #1749, opened 2021-11-13
Manifest Updates

achevuru:
**What type of PR is this?**
Enhancement

**Which issue does this PR fix**:
Increase liveness/readiness probe timeout values.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2038, opened 2022-07-22
Increase cpu requests limit

vikasmb:
- Porting changes from release-1.10 branch made in PR #1749

**What type of PR is this?**

cleanup
Porting CPU requests change made in 1.10 branch in #1749

Increases CPU requests to 25m

**Automation added to e2e**:

No
No

Yes

No
```release-note

```

### DIFF
--- a/charts/aws-vpc-cni/Chart.yaml
+++ b/charts/aws-vpc-cni/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: aws-vpc-cni
-version: 1.1.17
+version: 1.1.18
 appVersion: "v1.11.2"
 description: A Helm chart for the AWS VPC CNI
 icon: https://raw.githubusercontent.com/aws/eks-charts/master/docs/logo/aws.png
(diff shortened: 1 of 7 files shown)


charts/aws-vpc-cni/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: aws-vpc-cni
-version: 1.1.17
+version: 1.1.18
 appVersion: "v1.11.2"
```
achevuru, line 4: We need to bump the appVersion to 1.11.3 as well, so maybe we can update that and merge this PR as part of release process.

achevuru (approved):

charts/aws-vpc-cni/Chart.yaml
vikasmb: Synced up offline, appVersion change would be made in a separate PR.

jayanthvn:
Even for the cni-metrics helper we need to bump up the version.

vikasmb:
> Even for the cni-metrics helper we need to bump up the version - https://github.com/vikasmb/amazon-vpc-cni-k8s/blob/master/charts/cni-metrics-helper/Chart.yaml

@jayanthvn In previous change for this file - https://github.com/aws/amazon-vpc-cni-k8s/pull/1962/files, it would be changed along with image version, is that accurate ? If so, I will do that change while changing the app version in a separate PR.

PR #2039, opened 2022-07-22
fix ipamd integration failures and cleanup

sushrk:
**What type of PR is this?**
bug, cleanup

1. Fix ipamd integration test failures: Remove WARM_ENI_TARGET, WARM_IP_TARGET, MINIMUM_IP_TARGET and WARM_PREFIX_TARGET before running IPAMD warm target tests
2. Remove CNIMetricsHelperPolicy creation in test suite
3. Cleanup: Move test namespace creation and deletion to before suite & after suite respectively to avoid doing for each test case

N/A

```
ginkgo -v -timeout 100m --fail-on-pending -- \
 --aws-region=$AWS_REGION \
 --aws-vpc-id=$VPC_ID \
 --ng-name-label-val=$NG_NAME_LABEL_VAL
Running Suite: VPC IPAMD Test Suite - /Users/ravsushm/go/src/github.com/aws/amazon-vpc-cni-k8s/test/integration/ipamd
=====================================================================================================================
Random Seed: 1658486649

Will run 26 of 26 specs
<REDACTED FOR VERBOSITY>

Ran 26 of 26 Specs in 5244.027 seconds
FAIL! -- 25 Passed | 1 Failed | 0 Pending | 0 Skipped
--- FAIL: TestIPAMD (5244.03s)
FAIL

Ginkgo ran 1 suite in 1h27m45.209972174s

Test Suite Failed
```

The one failed test above is testing unauthorized event on aws-node pod when VPC_CNI policy is missing. This test is not applicable to current release and can be ignored for now.

N/A

N/A

N/A

N/A

### DIFF
--- a/test/framework/resources/aws/services/iam.go
+++ b/test/framework/resources/aws/services/iam.go
@@ -37,6 +37,7 @@ type IAM interface {
 	CreatePolicy(policyName string, policyDocument string) (*iam.CreatePolicyOutput, error)
 	DeletePolicy(policyARN string) error
 	GetInstanceProfile(instanceProfileName string) (*iam.GetInstanceProfileOutput, error)
+	ListPolicies(scope string) (*iam.ListPoliciesOutput, error)
 }
 
 type defaultIAM struct {
@@ -82,6 +83,13 @@ func (d *defaultIAM) GetInstanceProfile(instanceProfileName string) (*iam.GetIns
 		InstanceProfileName: aws.String(instanceProfileName),
 	}
 	return d.IAMAPI.GetInstanceProfile(getInstanceProfileInput)
+}
+
+func (d *defaultIAM) ListPolicies(scope string) (*iam.ListPoliciesOutput, error) {
+	listPolicyInput := &iam.ListPoliciesInput{
+		Scope: aws.String(scope),
+	}
+	return d.IAMAPI.ListPolicies(listPolicyInput)
 }
 
 func NewIAM(session *session.Session) IAM {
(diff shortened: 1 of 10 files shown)


jayanthvn (commented):
But on master the event test should have been a pass right?

sushrk:
> But on master the event test should have been a pass right?

Test was run on 1.11.2 image, did not update to master image

test/framework/resources/aws/services/iam.go
```diff
@@ -84,6 +85,13 @@ func (d *defaultIAM) GetInstanceProfile(instanceProfileName string) (*iam.GetIns
 	return d.IAMAPI.GetInstanceProfile(getInstanceProfileInput)
 }
 
+func (d *defaultIAM) ListPolicies(scope string) (*iam.ListPoliciesOutput, error) {
+	listPolicyInput := &iam.ListPoliciesInput{
+		Scope: aws.String(scope),
+	}
+	return d.IAMAPI.ListPolicies(listPolicyInput)
```
vikasmb, line 92: should we return errors here as well ? https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListPolicies.html#API_ListPolicies_ResponseElements

sushrk: iam interface ListPolicies has two return values so this should work as is: 
https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.40.6/service/iam/iamiface#IAMAPI.ListPolicies

jayanthvn (dismissed):
lgtm

jayanthvn:
Can we ensure these tests are enabled in the workflow and those are not impacted?

vikasmb (dismissed):
lgtm

vikasmb (approved):
lgtm

Issue #2041, opened 2022-07-25
issue with data_store.total value while enabling prefix delegation


aglover-zendesk:
**What happened**:

Shortly after setting `ENABLE_PREFIX_DELEGATION` to true, the `awscni_total_ip_addresses` metric started showing very low values (and even negative values for many nodes, see screenshot below). This occurred while the underlying nodes had a mix of /28 IP prefixes and "normal" secondary IPs that were provisioned and attached before enabling prefix delegation.

![image](https://user-images.githubusercontent.com/25018339/180850613-e18ff429-0cef-4bc3-8cfd-7fcfe7bc9c9e.png)

Note that the nodes are eventually replaced by our automation after they report abnormally low values for total IPs, hence the metrics stop reporting after a time.

In addition to the issue with `awscni_total_ip_addresses`, it seems like the data coming back from the `/v1/enis` API endpoint shares the same issue. While investigating, I queried a few nodes and discovered that the `TotalIPs` statistic didn't seem to match the more detailed information in the payload. For example, see the following response from one of our nodes:

<details>
  <summary>`/v1/enis` response JSON</summary> 

  ```json
  {
    "TotalIPs": 46,
    "AssignedIPs": 49,
    "ENIs": {
      "eni-06499138db120979f": {
        "ID": "eni-06499138db120979f",
        "IsPrimary": false,
        "IsTrunk": false,
        "IsEFA": false,
        "DeviceNumber": 2,
        "AvailableIPv4Cidrs": {
          "10.192.73.16/28": {
            "Cidr": {
              "IP": "10.192.73.16",
              "Mask": "////8A=="
            },
            "IPAddresses": {
              "10.192.73.16": {
                "Address": "10.192.73.16",
                "IPAMKey": {
                  "networkName": "aws-cni",
                  "containerID": "c6b1629b21fcae359e7eed2d9f20b5f4a467f041544032c98d286edca24de90e",
                  "ifName": "eth0"
                },
                "IPAMMetadata": {
                  "k8sPodNamespace": "<redacted>",
                  "k8sPodName": "<redacted>"
                },
                "AssignedTime": "2022-07-22T04:37:34.903573754Z",
                "UnassignedTime": "0001-01-01T00:00:00Z"
              },
              "10.192.73.17": {
                "Address": "10.192.73.17",
                "IPAMKey": {
                  "networkName": "aws-cni",
[...]

PR #2042, opened 2022-07-25
Remove extra decrement of totalIP count

jayanthvn:
**What type of PR is this?**
Bug

Wrong computation of totalIPs

The issue is because we decrement the total here - https://github.com/aws/amazon-vpc-cni-k8s/blob/c7443ae30f40037adaa997c26b66fc5912737958/pkg/ipamd/datastore/data_store.go#L1224 and decrement again here https://github.com/aws/amazon-vpc-cni-k8s/blob/c7443ae30f40037adaa997c26b66fc5912737958/pkg/ipamd/datastore/data_store.go#L655 while we force delete the secondary IP because we are in PD mode and we don't need the IP/32 anymore attached to the instance.

#2041 

Repro'ed the issue and fix is holding up

Secondary IP mode and 1 pod running ->

```
"TotalIPs":10,"AssignedIPs":1
```

PD mode toggle ->

```
"TotalIPs":17,"AssignedIPs":1
```

Deleted the pod behind secondary IP ->

```
"TotalIPs":16,"AssignedIPs":0
```
**Automation added to e2e**:

No, will add it up

No

No

No

No
```release-note

```

### DIFF
--- a/pkg/ipamd/datastore/data_store.go
+++ b/pkg/ipamd/datastore/data_store.go
@@ -1219,10 +1219,7 @@ func (ds *DataStore) UnassignPodIPAddress(ipamKey IPAMKey) (e *ENI, ip string, d
 		return nil, "", 0, err
 	}
 	addr.UnassignedTime = time.Now()
-	if ds.isPDEnabled && !availableCidr.IsPrefix {
-		ds.log.Infof("Prefix delegation is enabled and the IP is from secondary pool hence no need to update prefix pool")
-		ds.total--
-	}
+
 	//Update prometheus for ips per cidr
 	ipsPerCidr.With(prometheus.Labels{"cidr": availableCidr.Cidr.String()}).Dec()
 	ds.log.Infof("UnassignPodIPAddress: sandbox %s's ipAddr %s, DeviceNumber %d",


pkg/ipamd/datastore/data_store_test.go
```diff
@@ -96,6 +96,57 @@ func TestDeleteENI(t *testing.T) {
 	assert.Error(t, err)
 	err = ds.RemoveENIFromDataStore("eni-1", true)
 	assert.NoError(t, err)
+
+}
+
+func TestDeleteENIwithPDEnabled(t *testing.T) {
```
jdn5126, line 102: nit: this is nearly the same test as above. Can you share the test above and pass a `PDEnabled` boolean?

pkg/ipamd/datastore/data_store_test.go
```diff
@@ -96,6 +96,57 @@ func TestDeleteENI(t *testing.T) {
...
+	assert.Equal(t, "10.0.0.0", ip)
+	assert.Equal(t, 4, device)
+
+	// Test force removal.  The first call fails because eni-1 has an IP with a pod assigned to it,
+	// but the second call force-removes it and succeeds.
+	err = ds.RemoveENIFromDataStore("eni-4", false)
+	assert.Error(t, err)
+	err = ds.RemoveENIFromDataStore("eni-4", true)
+	assert.NoError(t, err)
+
```
jdn5126, line 149: I think it is worth checking datastore count here, then draining all ENIs and verifying count goes to 0.

jdn5126: Should this test also be checking `ds.Total`?

pkg/ipamd/datastore/data_store_test.go
```diff
@@ -141,6 +192,49 @@ func TestAddENIIPv4Address(t *testing.T) {
...
+	assert.Equal(t, len(ds.eniPool["eni-1"].AvailableIPv4Cidrs), 2)
+	assert.Equal(t, len(ds.eniPool["eni-2"].AvailableIPv4Cidrs), 1)
+
+	ipv4Addr = net.IPNet{IP: net.ParseIP("40.0.0.0"), Mask: net.IPv4Mask(255, 255, 255, 240)}
+	err = ds.AddIPv4CidrToStore("dummy-eni", ipv4Addr, true)
+	assert.Error(t, err)
+	assert.Equal(t, ds.total, 48)
+	assert.Equal(t, len(ds.eniPool["eni-1"].AvailableIPv4Cidrs), 2)
+	assert.Equal(t, len(ds.eniPool["eni-2"].AvailableIPv4Cidrs), 1)
+
```
jdn5126, line 235: nit: can you add a case of adding the same ENI,addr (not the update case) and verify that the operation is idempotent?

pkg/ipamd/datastore/data_store_test.go
```diff
@@ -177,6 +271,42 @@ func TestGetENIIPs(t *testing.T) {
 	assert.Error(t, err)
 }
 
+func TestGetENIIPsWithPDEnabled(t *testing.T) {
```
jdn5126, line 274: What does this test cover that `TestAddENIIPv4AddressWithPDEnabled` does not?

pkg/ipamd/datastore/data_store_test.go
```diff
@@ -235,6 +365,143 @@ func TestDelENIIPv4Address(t *testing.T) {
 	assert.Equal(t, len(ds.eniPool["eni-1"].AvailableIPv4Cidrs), 1)
 }
 
+func TestDelENIIPv4AddressWithPDEnabled(t *testing.T) {
```
jdn5126, line 368: nit: this also feels like it should be calling `TestDelENIIPv4Address` and passing a PD boolean

pkg/ipamd/datastore/data_store_test.go
```diff
@@ -491,6 +932,58 @@ func TestGetIPStatsV4(t *testing.T) {
 	)
 }
 
+func TestGetIPStatsV4WithPD(t *testing.T) {
```
jdn5126, line 935: nit: can this call `TestGetIPStatsV4` with PD boolean to simplify testing?

pkg/ipamd/datastore/data_store_test.go
jayanthvn: This test case is to check force removal of ENIs. The total count is checked in the next test case.

pkg/ipamd/datastore/data_store_test.go
jayanthvn: Line 210, does add the same (ENI,Addr) and expected to error out.

pkg/ipamd/datastore/data_store_test.go
jayanthvn: This get UTs are to test get APIs, there is overlap with add but we do want to cover get tests separately.

pkg/ipamd/datastore/data_store_test.go
jayanthvn: Same as above, to avoid conditional checks.

pkg/ipamd/datastore/data_store_test.go
jayanthvn: Same as above, to avoid conditional checks.

jdn5126 (approved):

PR #2068, opened 2022-08-25
fix warm target tests in ipamd integration test

sushrk:
**What type of PR is this?**
cleanup, bug

n/a

Fix failures in warm target integration tests.

We require to not have any pods scheduled on the node for running warm target tests. Previously, the primary node we choose could have a coredns pod running on the node and was handled by accounting for the coredns pods in available IP/ENI calculations as seen in [test/integration/ipamd/warm_target_test_PD_enabled.go](https://github.com/aws/amazon-vpc-cni-k8s/pull/2068/files#diff-237a40a86a7004620c7e78d94f0145b9efa67db6e730fd5e2fa4970166ea9114). 
This was missed in test/integration/ipamd/warm_target_test.go. After adding similar changes in warm_target_test.go, some failures were fixed, but still few tests were flaky. 

Changes in this PR: Force coredns pods to be scheduled on a single node by adding nodeSelector label in coredns deployment. Then, we pick a different node with no pods scheduled to run ipamd tests. This avoids additional calculations and the failures we've seen before with warm target tests since we ensure that the test is run on node with no pods.

n/a

Run integration test

**Automation added to e2e**:
n/a

No

No

No

No

### DIFF
--- a/test/integration/ipamd/ipamd_suite_test.go
+++ b/test/integration/ipamd/ipamd_suite_test.go
@@ -22,7 +22,15 @@ import (
 	"github.com/aws/amazon-vpc-cni-k8s/test/framework/utils"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	v1 "k8s.io/api/apps/v1"
 )
+
+const (
+	CoreDNSDeploymentName      = "coredns"
+	CoreDNSDeploymentNameSpace = "kube-system"
+)
+
+var coreDNSDeploymentCopy *v1.Deployment
 
 func TestIPAMD(t *testing.T) {
 	RegisterFailHandler(Fail)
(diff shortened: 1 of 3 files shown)


test/integration/ipamd/ipamd_suite_test.go
```diff
@@ -43,22 +51,54 @@ var _ = BeforeSuite(func() {
...
+	By("getting node with no pods scheduled to run tests")
+	coreDNSDeployment, err := f.K8sResourceManagers.DeploymentManager().GetDeployment(CoreDNSDeploymentName,
+		CoreDNSDeploymentNameSpace)
+	Expect(err).ToNot(HaveOccurred())
+
+	// Copy the deployment to restore later
+	coreDNSDeploymentCopy = coreDNSDeployment.DeepCopy()
+
+	coreDNSDeployment.Spec.Template.Spec.NodeSelector = map[string]string{
+		"kubernetes.io/hostname": *primaryInstance.PrivateDnsName,
```
sushrk, line 70: Should we also set coredns replica to 1?

orsenthil: What difference will it make?

sushrk: Just wondering since it's not best practice to have both replicas on same node generally. 

I think it is okay in this case, not much of a difference.

orsenthil:
Hi @sushrk , could you be update the PR description to state what this change is doing to fix the test? I am getting a vague idea, but will like to read a description.

orsenthil (dismissed):
LGTM. 

Thanks for adding the details to the PR.

orsenthil (approved):

Issue #2079, opened 2022-09-08
aws-node container failing to restart with SELinux enabled


sahilbadla:
**What happened**:

aws-node container is having problems restarting after SELinux enforcing mode is applied. This happens because the file '/host/etc/cni/net.d/10-aws.conflist' was created by an older container(first aws-node container on the node) with a different context than the new container initializing. That puts the new container in CrashLoopBackOff. 

Here's an example from one of the nodes:

![image](https://user-images.githubusercontent.com/3950985/189212226-10bf8117-dd92-46ca-851c-3259e2bd4dc9.png)

Now when i delete this pod and a new pod(aws-node-rxcq6) comes up, here's the error message from the aws-node container:

![image](https://user-images.githubusercontent.com/3950985/189212581-5387983d-12ee-4d89-966a-2829a650c988.png)

Detailed exception:

![image](https://user-images.githubusercontent.com/3950985/189212764-2ce24eb6-2130-4a5e-b151-e2ab01278686.png)

As you can see the new container(with context c220,c982) is not able to "mv" the file "10-aws.conflist" that was created with different context (c615,c828). This is the line of code where it fails https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/entrypoint.sh#L199

Looks like if we change mv command to cp instead, it should work with selinux. Reference https://danwalsh.livejournal.com/56534.html

**Attach logs**

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:
- Deploy cni v1.11.2
- Set Selinux to enforcing mode

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): v1.21.14-eks-18ef993
- CNI Version : v1.11.2
- OS (e.g: `cat /etc/os-release`): 
```
-NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
```
- Kernel (e.g. `uname -a`): ip-10-134-188-70.vpc.internal 5.4.204-113.362.amzn2.x86_64

jdn5126:
Since the move to a Golang entrypoint, we no longer use the shell `mv` command. Instead, we copy the file to the destination path named `<file>.tmp`, then call `os.Rename`: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/utils/cp/cp.go#L48

This should resolve the SELinux issue in enforcing mode, so closing this as fixed by #2146

PR #2080, opened 2022-09-09
selinux permission denied, change mv to cp to maintain file context 

sahilbadla:
**What type of PR is this?**
bug 

#2079 

This fixes the permissions issue experienced when SELinux enforcing mode is enabled. Entrypoint.sh script uses **_mv_** command([link](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/entrypoint.sh#L199)) to write to host file system. When the aws-node container is killed or restarted it tries to write to the same file while has a different context(from the old container) and its not allowed by policies in SeLinux. To make it work with SeLinux, changing the **_mv_** command to **_cp_** works as **_cp_** maintains the context of file while writing.

Unit testing
Tested with custom container image for amazon-k8s-cni 

**Automation added to e2e**:

No

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -196,7 +196,7 @@ if [[ "$ENABLE_BANDWIDTH_PLUGIN" == "true" ]]; then
     mv "$TMP_AWS_BW_CONFLIST_FILE" "$TMP_AWS_CONFLIST_FILE" 
 fi
 
-mv "$TMP_AWS_CONFLIST_FILE" "$HOST_CNI_CONFDIR_PATH/10-aws.conflist"
+cp "$TMP_AWS_CONFLIST_FILE" "$HOST_CNI_CONFDIR_PATH/10-aws.conflist" && rm -rf "$TMP_AWS_CONFLIST_FILE"
 
 log_in_json info "Successfully copied CNI plugin binary and config file."
 


scripts/entrypoint.sh
```diff
@@ -196,7 +196,7 @@ if [[ "$ENABLE_BANDWIDTH_PLUGIN" == "true" ]]; then
     mv "$TMP_AWS_BW_CONFLIST_FILE" "$TMP_AWS_CONFLIST_FILE" 
 fi
 
-mv "$TMP_AWS_CONFLIST_FILE" "$HOST_CNI_CONFDIR_PATH/10-aws.conflist"
+cp "$TMP_AWS_CONFLIST_FILE" "$HOST_CNI_CONFDIR_PATH/10-aws.conflist" && rm -rf "$TMP_AWS_CONFLIST_FILE"
```
anusha-ragunathan, line 199: Can you please a comment here explaining why we need to do this for SELinux? That way, folks looking at this later on wont optimize `cp && rm` for a `mv` and break SELinux

sahilbadla:
looks like this solutions unblocks the crashloop problem but issue still persists. Even though the pods get restarted fine, the file /host/etc/cni/net.d/10-aws.conflist never gets replaced. Logs are something like this:
` type=AVC msg=audit(1663054943.599:343637): avc:  denied  { write } for  pid=851535 comm="cp" name="10-aws.conflist" dev="nvme0n1p1" ino=36708256 scontext=system_u:system_r:container_t:s0:c880,c964 tcontext=system_u:object_r:container_file_t:s0:c241,c852 tclass=file permissive=0`
that means even cp is not working with enforcing mode as the file to be written is in the host. Closing this PR.

PR #2101, opened 2022-10-04
Introducing new AWS_EC2_ENDPOINT env variable option

lxpbl:
**What type of PR is this?**
feature

N/A

This PR adds ability to override default EC2 API endpoint - (https://ec2.<region>.amazonaws.com), useful in case when it's preferred to use private VPC endpoint URLs instead of "public" address (and in cases when private DNS doesn't help due to complex potentially complex corporate DNS setup). Also, to set this plugin features on par of another aws plugin - https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/pkg/cloud/cloud.go#L250

N/A

Unit tests / successful use in DropBox e2e infra testing.

**Automation added to e2e**:
N/A

No

No

No

Adding new env variable option (default variant works as before), doesn't change anything else

```release-note

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -573,6 +573,16 @@ configured to operate in IPv6 mode. Prefix delegation is only supported on nitro
 
 ---
 
+#### `AWS_EC2_ENDPOINT` (v1.11.5+)
+
+Type: String
+
+Default: None
+
+Overrides default (`https://ec2.<region>.amazonaws.com`) EC2 API endpoint, useful in case when it's preferred to use private VPC endpoint URLs instead of "public" address.
+
+---
+
 ### VPC CNI Feature Matrix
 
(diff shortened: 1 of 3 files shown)


jdn5126:
@lxpbl instead of using an environment variable, an approach like https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/2179 would be better.

lxpbl:
@jdn5126 thank you for looking into that and sorry for the late response. I basically follow the style aws-cni uses now (params passed in as env variables). The aws load balancer seems to use command line args way, which is different from what we have here.

jdn5126:
@lxpbl yeah, adding this functionality is acceptable, but for the implementation, we are trying to avoid additional environment variables for VPC CNI to manage, and the command-line argument approach is more flexible and auditable.

If you cannot take this up, feel free to file an issue and we can add it to the roadmap.

PR #2105, opened 2022-10-13
VPC CNI Integration Test Fixes

jdn5126:
**What type of PR is this?**
Cleanup

N/A

This PR fixes three issues in VPC CNI integration tests.

1. Calico tests are removed from `run-cni-release-tests.sh`. Calico integration tests are already covered by the GitHub runner that invokes `run-integration-tests.sh`, and this provides the appropriate level of CNI regression testing. Calico tests would need to be refactored to run against Windows nodes in order to return to `run-cni-release-tests.sh`. The GitHub runner invokes them against Linux nodes.

2. CNI host networking tests are refactored to increase the `WARM_IP_TARGET`. This value was increased so that the test better mirrored a real customer setup. The value could only be changed for host networking tests as other CNI integration tests depend on the value set in the suite setup. The flakiness of this test led to https://github.com/kubernetes/kubernetes/issues/113073 being filed.

3. Ginkgo was upgraded to v2.3.1. This is necessary to resolve issues caused by running a ginkgo CLI client >= 2.3.0 against this repo (there is an incompatibility between versions < 2.3.0 and >= 2.3.0). The ginkgo issue is described in https://github.com/onsi/ginkgo/issues/1050

N/A

Validated that `run-cni-release-tests.sh` passes.

**Automation added to e2e**:
N/A

No

No

No

No

```release-note

```


test/integration/cni/host_networking_test.go
```diff
@@ -36,13 +36,27 @@ const (
 	NEW_VETH_PREFIX            = "veth"
 	DEFAULT_MTU_VAL            = "9001"
 	DEFAULT_VETH_PREFIX        = "eni"
+	WARM_IP_TARGET             = "WARM_IP_TARGET"
```
sushrk, line 39: This can be removed, should be good to use "WARM_IP_TARGET" string directly in argument below.

test/integration/cni/host_networking_test.go
```diff
@@ -36,13 +36,27 @@ const (
...
+	DEFAULT_WARM_IP_TARGET     = "3"
```
sushrk, line 40: Can we move this to `pod_networking_suite_test.go` and use it in the `BeforeSuite` function too? 
In case the default needs to be updated in future.

test/integration/cni/host_networking_test.go
jdn5126: Since the other variables used global const strings, I figured it was best to keep it consistent. Then only one line can be changed in the future

test/integration/cni/host_networking_test.go
jdn5126: Putting it in `BeforeSuite` would change the values for all tests, and there are other tests that depend on the `WARM_IP_TARGET` being low. So I changed it so that it is set for just these tests and is reset after. It actually changes an additional (unnecessary) time, but that is because ginkgo doesn't have all "before all tests in this file only"/"after all tests in this file only" option

sushrk: My bad, I meant, let's move the `DEFAULT_WARM_IP_TARGET` declaration to the `pod_networking_suite_test.go` and use this here too https://github.com/aws/amazon-vpc-cni-k8s/pull/2105/files#diff-86c2f600945618a77023a9e8ecd685ddd43b0f4847fb9b3bb5955d61429aa667R91

I assume this is where the value `3` is coming from? So that in the future if this default value changes, it will be updated both in `BeforeSuite` and `BeforeEach`

jdn5126: Ah got it, will do

test/integration/cni/host_networking_test.go
sushrk: I think we use ENV names like `WARM_IP_TARGET` as strings in other places too. Not required to use a const for it here.

sushrk (approved):
Thank you, lgtm!

PR #2108, opened 2022-10-18
Update GitHub runners to install ginkgo from `go.mod` in test directory

jdn5126:
**What type of PR is this?**
Cleanup

N/A

This PR updates GitHub runners to install the correct ginkgo version. It does so by installing the ginkgo version specified in `go.mod` in the test directory.

N/A

Coverage will be achieved through GitHub runner executing

**Automation added to e2e**:
N/A

No

N/A

No

No

```release-note

```

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -26,9 +26,12 @@ jobs:
           go-version: "1.19"
       - name: Set up tools
         run: |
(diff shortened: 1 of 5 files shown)


M00nF1sh (approved):
/lgtm

personally i think we should just use the go.mod in root directory and remove the go.mod in test directory and keep the go.mod in root directory updated.

PR #2110, opened 2022-10-19
switch to use state file for IP allocation pool management

M00nF1sh:
**What type of PR is this?**
feature

N/A

1. Switch to use state file instead of CRI socket when restoring the IP allocation pool upon CNI restart.
We need it to remove the CRI socket dependency.
2. Adds a soak test suit to keep spin up and teardown pods on a chosen node, which can be executed for given time or forever via ginkgo(--until-it-fails).
3. Did a few improvements to our test framework, mainly:
    * switch to use the REAL k8s client instead of cache one. which is one of the main source of flakes. 
       * we have a lot cases where we create a deployment and wait for it to become ready, and this often flakes due to the "cache" deployment was ready but the real one wasn't.
       * I don't see a need to use a cached client for testing purpose, uses real one simplifies our test logic and reduces flakes.
    * added context based wait utils into k8s resource generators, which allows external code control timeouts easily.

* Spin up/down pods on a given node for several hours, and monitor the state file is consistent with pods on node.

**Automation added to e2e**:

Added a new soak test suit that keep spin up and teardown pods and compare the state file with pods on nodes.
The suit can be executed via:
```
MAX_POD_PER_NODE=<max-pods-for-your-instance-type> ginkgo -v --until-it-fails -- --cluster-kubeconfig=/Users/yyyng/.kube/config --cluster-name=<clusterName> --aws-region=us-west-2 --aws-vpc-id=<vpc-id> --ng-name-label-key="kubernetes.io/os" --ng-name-label-val="linux"
```

No new dependencies as the file is already written.

This will break upgrades when upgrade from CNI < 1.7.0, as no state file were wrotten.
I have tested upgrade from CNI 1.10.0, and it works fine

Yes, the CRI socket mount is no longer needed.

```release-note
We have switched to use state file instead of CRI socket for IP allocation pool restore.
```


pkg/ipamd/datastore/data_store.go
```diff
@@ -1538,3 +1545,39 @@ func (ds *DataStore) CheckFreeableENIexists() bool {
 	}
 	return false
 }
+
+// NormalizeCheckpointDataByPodVethExistence will normalize checkpoint data by remove allocations that don't have a corresponding pod veth.
+// This shouldn't happen unless container runtimes have bugs that failed to invoke the cmdDel for died pods.
+// we use this reconciliation as a safety mechanism during transition from CRI to state file.
+func (ds *DataStore) normalizeCheckpointDataByPodVethExistence(checkpoint CheckpointData) (CheckpointData, error) {
```
jdn5126, line 1552: nit: can you add a test case for this logic just so we have coverage?

M00nF1sh: on it :D

pkg/ipamd/datastore/data_store.go
```diff
@@ -1538,3 +1546,39 @@ func (ds *DataStore) CheckFreeableENIexists() bool {
...
+	hostNSLinks, err := ds.netLink.LinkList()
```
jdn5126, line 1554: nit: I am not sure how critical performance is here, but I noticed that we get the list of interfaces here, then we loop over this list for every allocation. If performance is critical, we can pre-process this list and create a hash map that `validateAllocationByPodVethExistence` can use to just hash the `linkNameSuffix` into.

jdn5126 (commented):
Changes LGTM, just one performance concern and I will give others a chance to review

pkg/ipamd/datastore/data_store.go
M00nF1sh: @jdn5126 
The performance don't matter here as it's not in critical path such as pod launch(this check only happens during IPAMD restart). Also, the iterations here is only matters of 250 * 250 at most.

I used this instead of the make it a map to avoid depends on the fact that the linkSuffix is always 11 length. We can change this if you have concerns :D

pkg/ipamd/datastore/data_store.go
```diff
@@ -432,24 +439,25 @@ func (ds *DataStore) ReadBackingStore(isv6Enabled bool) error {
...
-			return nil
-		} else if err != nil {
-			return fmt.Errorf("datastore: error reading backing store: %v", err)
+		ds.log.Infof("Begin ipam state recovery from backing store")
+
+		if err := ds.backingStore.Restore(&data); err != nil {
+			// Assume that no file == no containers are currently in use, e.g. a fresh reboot just cleared everything out.
+			// This is ok, and no-op.
+			if os.IsNotExist(err) {
+				ds.log.Debugf("backing store don't exists, assuming bootstrap on a new node")
```
jayanthvn, line 448: nit: don't -> doesn't

pkg/ipamd/datastore/data_store.go
```diff
@@ -1538,3 +1546,39 @@ func (ds *DataStore) CheckFreeableENIexists() bool {
 	}
 	return false
 }
+
+// NormalizeCheckpointDataByPodVethExistence will normalize checkpoint data by remove allocations that don't have a corresponding pod veth.
```
jayanthvn, line 1550: nit: remove -> removing

jayanthvn (commented):
Just few nits. Can we also run the soak test on bottlerocket? Rest lgtm.

pkg/ipamd/datastore/data_store.go
jayanthvn: Yeah even I was thinking if we need to hash the links and use it for validation but as Yang mentioned this will be run only once during restart and worst case you will have max-pods number of loop.

pkg/ipamd/datastore/data_store.go
M00nF1sh: thanks :D fixed

pkg/ipamd/datastore/data_store.go
M00nF1sh: thanks :D fixed

jayanthvn (dismissed):
/lgtm

pkg/ipamd/datastore/data_store.go
M00nF1sh: To add some data, with 250 pods, under worse case , the loop take 0.3ms to finish. So i still prefer it than couple the code with a fixed 11 length suffix.

func BenchmarkDummy(b *testing.B) {
	for i := 0; i < b.N; i++ {
		count := 250
		var allocations []string
		var names []string
		for i := 0; i < count; i++ {
			allocations = append(allocations, "dummy1233")
			names = append(names, "abc333333333")
		}
		for _, allocation := range allocations {
			for _, name := range names {
				strings.HasSuffix(name, allocation)
			}
		}
	}
}```

Issue #2103, opened 2022-10-05
Pods cannot talk to cluster IPs on Ubuntu 2204


olemarkus:
**What happened**:

After upgrading clusters to use Ubuntu 22.04 by default, the kOps e2e tests started failing for this CNI: https://testgrid.k8s.io/kops-network-plugins#kops-aws-cni-amazon-vpc

What seems to happen is that Pods do receive IPs, but they fail to talk across nodes. Calling e.g a ClusterIP service from the _host_ works, but not from a Pod. Kube-proxy therefore should be working just fine.

I cannot see anything wrong in any logs. But what I do see is that there are AWS-related rules in the legacy iptables, while kube-proxy uses nftables. So my guess is that this is the cause of this behavior. nft and legacy iptables must not be mixed anyway.

**Attach logs**
Example logs here: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/e2e-kops-aws-cni-amazon-vpc/1577618499142946816/artifacts/i-0d90e121da8bff687/

**How to reproduce it (as minimally and precisely as possible)**:

```
kops create cluster --name test.kops-dev.srsandbox.io --cloud aws --networking=amazonvpc --zones=eu-central-1a,eu-central-1b,eu-central-1c --channel=alpha --master-count=3 --yes --kubernetes-version 1.25.0 --discovery-store=$KOPS_STATE_STORE/discovery --image=099720109477/ubuntu/images/hvm-ssd/ubuntu-jammy-22.04-amd64-server-20220921.1
```

jayanthvn:
This looks similar to this issue - https://github.com/aws/amazon-vpc-cni-k8s/issues/1847#issuecomment-1120371691.

This workaround - https://github.com/aws/amazon-vpc-cni-k8s/issues/1847#issuecomment-1040878745 has helped.

As suggested by @achevuru - Amazon Linux 2 images use iptables-legacy by default as well. We will check and update if there is something we can do to address this scenario.

olemarkus:
Let me try that workaround. If it works, I think it would be helpfull if an iptables-nft image could be published. I imagine it wouldn't be too much work to do that.

jayanthvn:
Thanks, please let us know if it works.

olemarkus:
Unfortunately, no luck. The workaround does remove the rules from iptables-legacy and I do see them now in nftables. Pods still cannot talk to cluster IPs.

olemarkus:
[...]

PR #2118, opened 2022-10-25
[driver] refactor pod network configuration not to use static ARP entry

veshij:
**What type of PR is this?**
bugfix

Fixes: https://github.com/aws/amazon-vpc-cni-k8s/issues/2103

This PR changes pod network confuguration so static ARP entry for host's veth mac address is no longer need.
**Previous configuration:**
* host veth doesn't have an IP address asigned
* pod veth has a direct route to fake router IP address
* pod veth has default route via fake router IP address
* fake router IP address has static arp entry in pod namespace pointed to host system's veth interface
```
$ ip addr sho dev enidbd62a3316e
18037: enidbd62a3316e@if3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1450 qdisc noqueue state UP group default
    link/ether 46:a9:d3:46:96:82 brd ff:ff:ff:ff:ff:ff link-netns  cni-9fc9c11a-2553-b00c-4ba8-98228b5a846a
    inet6 fe80::98df:3bff:feb0:dcf2/64 scope link

$ sudo ip netns exec cni-9fc9c11a-2553-b00c-4ba8-98228b5a846a ip ro
default via 169.254.1.1 dev eth0
169.254.1.1 dev eth0 scope link
$ sudo ip netns exec cni-9fc9c11a-2553-b00c-4ba8-98228b5a846a arp -n
Address                  HWtype  HWaddress           Flags Mask            Iface
169.254.1.1              ether   46:a9:d3:46:96:82   CM                    eth0
```

**Updated configuration:**
* host veth has link-local address assigned (all interfaces have same address)
* pod has an onlink default route to host's veth interface
```
$ sudo ip netns exec cni-0e566121-8e96-b032-a3e3-15ca97a87f01 ip ro
default via 169.254.1.1 dev eth0 onlink

$ ip addr sho dev enibd675d21895
18038: enibd675d21895@if3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1450 qdisc noqueue state UP group default
    link/ether a2:f3:e2:0c:5b:8b brd ff:ff:ff:ff:ff:ff link-netns cni-978889f6-92c0-8067-14a7-f509f39fa7f9
    inet 169.254.1.1/32 scope global enibd675d21895
    inet6 fe80::1057:1eff:fe7d:fe0b/64 scope link
```

More details in https://github.com/aws/amazon-vpc-cni-k8s/issues/2103
TLDR: on Ubuntu 22.04+ udevd assigns permanent mac address to host system veth interface once it's moved to host network namespace.  Since it happens after the pod network is configured with non-permanent mac address - pod has no network connectivity.
**Why not fix existing pod network configuration?**
I was not able to find a good way to monitor the udev-induced mac address change. The only option would be to add a sleep with some arbitrary timeout which either will slow-down pod allocation or will not be enough on heavy-loaded systems.

Verified updated configuration works on k8s cluster running on AWS EC2.

**Automation added to e2e**:
-

No

Should not/No

No

It changes pod network configuration as well as adds link-local address to host system's veth interface.

```release-note

```

### DIFF
--- a/cmd/routed-eni-cni-plugin/driver/driver.go
+++ b/cmd/routed-eni-cni-plugin/driver/driver.go
@@ -187,16 +187,15 @@ func (createVethContext *createVethPairContext) run(hostNS ns.NetNS) error {
 		}
 	}
 
-	// Add an onlink route to a dummy next hop (169.254.1.1 or fe80::1)
-	// # ip route show
-	// default via 169.254.1.1 dev eth0 onlink
 	vethConfig := createVethContext.vethConfiguration()
 	if err = createVethContext.netLink.AddrAdd(contVeth, vethConfig.contVethAddress); err != nil {
 		return errors.Wrapf(err, "setup NS network: failed to add IP addr to %q", createVethContext.contVethName)
 	}
 
-	// Add a default route via dummy next hop(169.254.1.1 or fe80::1). Then all outgoing traffic will be routed by this
-	// default route via dummy next hop (169.254.1.1 or fe80::1)
+	// Add a default onlink route via dummy next hop(169.254.1.1 or fe80::1).
+	// Then all outgoing traffic will be routed by this default route via dummy next hop (169.254.1.1 or fe80::1)
+	// # ip route show
+	// default via 169.254.1.1 dev eth0 onlink
 	defaultRoute := &netlink.Route{
 		LinkIndex: contVeth.Attrs().Index,
 		Scope:     netlink.SCOPE_UNIVERSE,
--- a/cmd/routed-eni-cni-plugin/driver/driver.go
+++ b/cmd/routed-eni-cni-plugin/driver/driver.go
@@ -106,6 +106,29 @@ func newCreateVethPairContext(contVethName string, hostVethName string, v4Addr *
 	}
 }
 
+type vethConfiguration struct {
+	dummyRouterAddress net.IP
+	maskLen            int
+	contVethAddress    *netlink.Addr
+	defaultDestination *net.IPNet
+}
+
+func (createVethContext *createVethPairContext) vethConfiguration() vethConfiguration {
+	var configuration vethConfiguration
+	if createVethContext.v4Addr != nil {
+		configuration.dummyRouterAddress = net.IPv4(169, 254, 1, 1)
+		configuration.maskLen = 8 * net.IPv4len
+		configuration.contVethAddress = &netlink.Addr{IPNet: createVethContext.v4Addr}
+		configuration.defaultDestination = &net.IPNet{IP: net.IPv4zero, Mask: net.CIDRMask(0, 8*net.IPv4len)}
+	} else if createVethContext.v6Addr != nil {
+		configuration.dummyRouterAddress = net.IP{0xfe, 0x80, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1}
+		configuration.maskLen = 8 * net.IPv6len
+		configuration.contVethAddress = &netlink.Addr{IPNet: createVethContext.v6Addr}
+		configuration.defaultDestination = &net.IPNet{IP: net.IPv6zero, Mask: net.CIDRMask(0, 8*net.IPv6len)}
+	}
+	return configuration
+}
+
 // run defines the closure to execute within the container's namespace to create the veth pair
 func (createVethContext *createVethPairContext) run(hostNS ns.NetNS) error {
 	veth := &netlink.Veth{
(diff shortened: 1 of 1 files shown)


olemarkus:
I tried running this patch (+ nftables) in our e2e setup and it seems to be working:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/14491/pull-kops-e2e-cni-amazonvpc/1588816070515363840

veshij:
@olemarkus thanks for confirming!
I'm trying to find someone who could review this PR.

jdn5126:
@veshij we are working on testing this internally. Will provide a timeframe later this week after we get this assigned

veshij:
> @veshij we are working on testing this internally. Will provide a timeframe later this week after we get this assigned

thanks for update!

jaydeokar:
> > @veshij we are working on testing this internally. Will provide a timeframe later this week after we get this assigned
> 
> thanks for update!

@veshij I'll be working on testing this internally. We plan to finish the internal testing by end of next week

jaydeokar:
> @veshij I'll be working on testing this internally. We plan to finish the internal testing by end of next week

I'm investigating few issues around Security group per pod feature which seems to be affected by this change

veshij:
I can help with troubleshooting if you could provide some details on the issue.

jaydeokar:
@veshij 
There are two issues as of now that we are tracking 
Ref [Security Group for Pod](https://docs.aws.amazon.com/eks/latest/userguide/security-groups-for-pods.html), [Security group Enforcing Mode](https://github.com/aws/amazon-vpc-cni-k8s#pod_security_group_enforcing_mode-v1110)


`strict` Mode
- Pod to Pod within the same security group fails even with allow all access on the security group. In the above example, curl from one pod to another fails when it is expected to allow the traffic, since they are using same security group and no block rules are added.

`standard` Mode
- Pod to Pod communication works in this mode. But communication between the pod to other host VM IPs in the cluster fails

jdn5126:
> @veshij There are two issues as of now that we are tracking Ref [Security Group for Pod](https://docs.aws.amazon.com/eks/latest/userguide/security-groups-for-pods.html), [Security group Enforcing Mode](https://github.com/aws/amazon-vpc-cni-k8s#pod_security_group_enforcing_mode-v1110)
> 
> `strict` Mode
> 
> * Pod to Pod within the same security group fails even with allow all access on the security group. In the above example, curl from one pod to another fails when it is expected to allow the traffic, since they are using same security group and no block rules are added.
> 
> `standard` Mode
> 
> * Pod to Pod communication works in this mode. But communication between the pod to other host VM IPs in the cluster fails

The TL;DR here is that in SGPP strict mode, the container is unable to resolve the ARP binding for its gateway addr (link-local address). In SGPP standard mode and non-SGPP, the container gets an ARP reply mapping the link-local IP to the host-side veth MAC.

In SGPP strict mode, we configure IIF-based IP route rules: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/cmd/routed-eni-cni-plugin/driver/driver.go#L557 , and we are trying to determine why the kernel does generate ARP reply in this case. Still digging...

veshij:
uh-oh, I somehow forgot about it. 
Trying to repro and the issue to fix it.

benedikt-bartscher:
Any news on this one?

Deshke:
any update on this ? @jdn5126 @veshij

jdn5126:
> any update on this ? @jdn5126 @veshij

I am going to have to defer to @veshij here. As it stands, we cannot merge this until the SGPP strict mode issue is understood

veshij:
ugh, totally forgot about this one. 
I'll be on PTO next week, will take a look after that.

heybronson:
@veshij any updates here? Can this be merged?

jayanthvn:
Not stale..need to debug the SGPP failures.

OverStruck:
bump

jdn5126:
I picked this up again, and I still cannot determine why in SGPP strict mode, the pod is unable to ARP for its host veth:
```
$ sudo tcpdump -vvven -i vlan873aabf4bea
tcpdump: listening on vlan873aabf4bea, link-type EN10MB (Ethernet), capture size 262144 bytes
20:07:16.391175 06:59:cf:7d:01:cb > Broadcast, ethertype ARP (0x0806), length 42: Ethernet (len 6), IPv4 (len 4), Request who-has 169.254.1.1 tell 192.168.65.161, length 28
```
In SGPP strict, we have a rule for pod traffic to force everything through the branch ENI:
```
$ ip rule show
10:     from all iif vlan873aabf4bea lookup 110
20:     from all lookup local
```
And that routing table is:
```
$ ip route show table 101
default via 192.168.64.1 dev vlan.eth.1
192.168.64.1 dev vlan.eth.1 scope link
192.168.86.180 dev vlan605fbc0e45b scope link
```
The host veth still has `169.254.1.1` assigned:
```
$ ifconfig vlan873aabf4bea
vlan873aabf4bea: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 9001
        inet 169.254.1.1  netmask 255.255.255.255  broadcast 0.0.0.0
        inet6 fe80::d839:a9ff:fee7:9ade  prefixlen 64  scopeid 0x20<link>
        ether da:39:a9:e7:9a:de  txqueuelen 0  (Ethernet)
        RX packets 0  bytes 0 (0.0 B)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 0  bytes 0 (0.0 B)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0
```
And I don't think any sysctls are preventing this from succeeding. Still digging

jdn5126:
Closing this pull request as we cannot implement this approach without breaking the Security Groups for Pods architecture. The recommendation for host operating systems is to set `MACAddressPolicy=none` in `/usr/lib/systemd/network/99-default.link`, as the official EKS AMI does for AL2023: https://github.com/awslabs/amazon-eks-ami/blob/master/scripts/install-worker.sh#L104

PR #2122, opened 2022-10-27
v1.12 Helm chart changes and release manifests

M00nF1sh:
**What type of PR is this?**
release

N/A

This PR adds the release manifests and helm changes for v1.12
The helm chart version was bumped to v1.2.0 since we no longer have CRI mounts since CNI v1.12

Manually tested canary tests with generated YAML.

**Automation added to e2e**:

Yes. This helm chart is only compatible with CNI version >= v1.12

Yes, the CNI's CRI mount is removed. Since CNI v1.12, we switched to use state file instead of rely on CRI mount.

```release-note
"cri.hostPath" is removed since helm chart v1.2.0, which ships with CNI v1.12.0 and no longer depends on CRI socket.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -622,16 +622,16 @@ and the kubelet respectively if you are making use of this tag.
 
 ### Container Runtime
 
-Currently, IPAMD uses dockershim socket to pull pod sandboxes information upon its starting. The runtime can be set to others.
-The mountPath should be changed to `/var/run/cri.sock` and hostPath should be pointed to the wanted socket, such as
-`/var/run/containerd/containerd.sock` for containerd. If using helm chart, the flag `--set cri.hostPath.path=/var/run/containerd/containerd.sock`
-can set the paths for you.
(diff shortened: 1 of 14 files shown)


jdn5126:
Changes LGTM. Can you also update the README in the root directory to reflect the dockershim removal? The "Container Runtime" section, that is

jayanthvn (commented):
lgtm but we need to update the readme - https://github.com/aws/amazon-vpc-cni-k8s#container-runtime

orsenthil (dismissed):
LGTM. 

Please call out on the readme that this cannot be utilized with older versions of CNI.

M00nF1sh:
updated the readme about container runtime :D

orsenthil (approved):
LGTM

orsenthil:
A post merge comment. Can we prevent installation of these in < EKS 1.24 cluster.

Issue #2048, opened 2022-08-01
ip rule leaked under some conditions 


cdemonchy:
Hi,

**What happened**:

Sometimes the ip rule  associated to a pod are not deleted when the pod is deleted. The leaked rules are never deleted but the ip is free in the ipam.

The ip can be reallocated on another node in the cluster. When this happens, the pods on the node with the leaked rules can't reach the pod with the reallocated ip. 

We've the case when cluster autoscaler fails to evict a node : 
- cluster autoscaler wants to evict a node
- pods on the nodes are stopped calling the cni plugin. The pod `aws-node` is stopped and restart quickly. 
- for some pods the cni plugin can't call ipamd before the container is deleted. In this case the rule are leaked. 
- After 2 minutes,  as it didn't success to move all the pod quickly (due to a very conservative pdb). The eviction is stopped and the node came back in the pool. ([cluster austoscaler faq](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#does-ca-work-with-poddisruptionbudget-in-scale-down))

**Attach logs**

The log for a pod during the delete phase : 
```
{"level":"debug","ts":"2022-07-31T11:35:11.678Z","caller":"routed-eni-cni-plugin/cni.go:303","msg":"Prev Result: &{0.4.0 [{Name:eni054dbc5e65c Mac: Sandbox:} {Name:eth0 Mac: Sandbox:/proc/12533/ns/net}] [{Version:4 Interface:0xc000270280 Address:{IP:10.109.167.233 Mask:ffffffff} Gateway:<nil>}] [] {[]  [] []}}\n"}
[...]

PR #2125, opened 2022-10-28
[CNI]: Teardown pod networking resources without IPAMD when possible 

jdn5126:
**What type of PR is this?**
bug

https://github.com/aws/amazon-vpc-cni-k8s/issues/2048

This PR resolves an issue in which IP rules were leaked by the CNI. When processing a pod deletion, the CNI would wait for IPAMD response before tearing down pod networking resources. If IPAMD could not be reached, CNI would return error and wait for kubelet to retry the delete. If IPAMD were restarted, the state for this pod would be cleared without CNI tearing down the associated networking resources. The trigger for the linked issue was the k8s cluster autoscaler evicting the `aws-node` daemonset pod before other pods and then later cancelling the pod evictions. https://github.com/kubernetes/autoscaler/issues/5240 was filed to ask for k8s cluster autoscaler to change its behavior.

With the changes in this PR, we now store state for cleanup in `PrevResult` for pods whether they use branch ENIs (SGPP mode) or not. During cleanup, we tear down networking resources whenever possible. If CNI cannot reach IPAMD, we still return error in non-SGPP mode and expect kubelet to retry the delete. Therefore, this delete is purely opportunistic in non-SGPP mode. It will only occur for pods created after the CNI version in which this change merges. This was done to avoid backward compatibility issues and preserve existing semantics, while also making an improvement.

Note that this also makes the CNI less dependent on IPAMD. The only behavior change to call out is that `tryDelWithPrevResult` now returns a `skipIpamd` boolean to indicate whether IPAMD request can definitely be skipped.

Added more test cases to `cni_test.go` and verified that all CNI and IPAMD integration tests pass with this change. Also manually verified that fix with IPv4 and IPv6 clusters.

**Automation added to e2e**:
N/A

No

Upgrades and downgraded will not be broken. A running cluster has been tested.

No

Yes

```release-note
Cleanup pod networking resources without IPAMD.
```

### DIFF
--- a/cmd/routed-eni-cni-plugin/cni.go
+++ b/cmd/routed-eni-cni-plugin/cni.go
@@ -47,7 +47,7 @@ import (
 
 const ipamdAddress = "127.0.0.1:50051"
 
-const dummyVlanInterfacePrefix = "dummy"
+const dummyInterfacePrefix = "dummy"
 
 var version string
 
@@ -187,11 +187,11 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 	log.Infof("Received add network response from ipamd for container %s interface %s: %+v",
 		args.ContainerID, args.IfName, r)
 
-	//We will let the values in result struct guide us in terms of IP Address Family configured.
+	// We will let the values in result struct guide us in terms of IP Address Family configured.
 	var v4Addr, v6Addr, addr *net.IPNet
 	var addrFamily string
 
-	//We don't support dual stack mode currently so it has to be either v4 or v6 mode.
+	// We don't support dual stack mode currently so it has to be either v4 or v6 mode.
 	if r.IPv4Addr != "" {
 		v4Addr = &net.IPNet{
 			IP:   net.ParseIP(r.IPv4Addr),
@@ -209,7 +209,10 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 	}
 
 	var hostVethName string
-	var dummyVlanInterface *current.Interface
+	var dummyInterface *current.Interface
+
+	// The dummy interface is purely virtual and is stored in the prevResult struct to assist in cleanup during the DEL command.
+	dummyInterfaceName := networkutils.GeneratePodHostVethName(dummyInterfacePrefix, string(k8sArgs.K8S_POD_NAMESPACE), string(k8sArgs.K8S_POD_NAME))
 
 	// Non-zero value means pods are using branch ENI
 	if r.PodVlanId != 0 {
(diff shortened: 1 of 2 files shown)


cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -311,17 +306,23 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 		return errors.Wrap(err, "del cmd: failed to load k8s config from args")
 	}
 
-	handled, err := tryDelWithPrevResult(driverClient, conf, k8sArgs, args.IfName, args.Netns, log)
-	if err != nil {
+	// Try to delete allocated network resources with data stored in PrevResult.
+	handled, skipIpamd, err := tryDelWithPrevResult(driverClient, conf, k8sArgs, args.IfName, args.Netns, log)
+	// On error, return immediately if pod was created with branch ENI. Otherwise, let IPAMD handle cleanup. Note that we cannot
+	// check conf.PodSGEnforcingMode as the pod we are deleting may have been added before the mode switch.
+	if err != nil && skipIpamd {
```
M00nF1sh, line 313: what if `err != nil and !skipIpamd`, seems the error is ignored.

jdn5126: The error still gets printed in `tryDelWithPrevResult`, but this is a case where we would still want to go through IPAMD for backwards compatibility. If we cannot delete from `PrevResult` and this pod was not created with a branch ENI (or we are not sure), then we give IPAMD a chance to do the delete.

In theory, this should never happen, but I figured we wanted to preserve backwards compatibility.

jdn5126: The error is effectively ignored, as we are assuming that IPAMD would return the same error if there is a real error.

M00nF1sh (approved):

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -407,41 +413,58 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
...
+		var deviceNumber int
+		deviceNumber, err = strconv.Atoi(dummyIface.Sandbox)
+		if err != nil {
+			return false, branchEni, errors.Errorf("malformed deviceNumber in prevResult: %s", dummyIface.Sandbox)
+		}
+		err = driverClient.TeardownPodNetwork(&containerIP, deviceNumber, log)
+	}
+
+	if err != nil {
+		return true, branchEni, err
```
M00nF1sh, line 465: since we return true here, does it mean we'll call ipamd to release IP but will skip teardownPodNetwork, and CNI won't retry due to we returned nil. seems we could leak some IP rules if TeardownPodNetwork failed for some reason.

```
if handled {
		return nil
	}
```

jdn5126: The `TeardownPodNetwork` call that we make based on the IPAMD response will be the same as the call we made here, but it does seem safer to try again. I will change this to return `branchEni, branchEni, err`.

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -311,17 +306,23 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
...
+	// check conf.PodSGEnforcingMode as the pod we are deleting may have been added before the mode switch.
+	if err != nil && skipIpamd {
 		return errors.Wrap(err, "del cmd: failed to delete with prevResult")
 	}
 	if handled {
 		log.Infof("Handled CNI del request with prevResult: ContainerID(%s) Netns(%s) IfName(%s) PodNamespace(%s) PodName(%s)",
 			args.ContainerID, args.Netns, args.IfName, string(k8sArgs.K8S_POD_NAMESPACE), string(k8sArgs.K8S_POD_NAME))
-		return nil
+		// When deleting a pod with a branch ENI configured, IPAMD does not need to be notified, so we can return early.
+		if skipIpamd {
```
jayanthvn, line 320: If handled we used to return `nil`. Why is the additional check?

jdn5126: We still need to call into IPAMD to release the IP allocation in the non-branch ENI case

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -407,41 +413,58 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
...
+		return false, branchEni, errors.Errorf("found %d containerIP for %v in prevResult", len(containerIPs), contVethName)
 	}
 	containerIP := containerIPs[0].Address
 
-	if err := driverClient.TeardownBranchENIPodNetwork(&containerIP, podVlanID, conf.PodSGEnforcingMode, log); err != nil {
-		return true, err
+	if branchEni {
+		if isNetnsEmpty(netNS) {
+			log.Infof("Ignoring TeardownPodENI as Netns is empty for SG pod: %s namespace: %s containerID: %s", k8sArgs.K8S_POD_NAME, k8sArgs.K8S_POD_NAMESPACE, k8sArgs.K8S_POD_INFRA_CONTAINER_ID)
+			return true, branchEni, nil
```
achevuru, line 451: If`netns` is already empty for a BranchENI pod, why don't we let IPAMD get the vlan information from API Server using the old flow (if the pod is still lingering around in etcd). If the current flow can get that for us, we can use that to clear the sgpp rules (if any). Will be a corner case and deferring it to IPAMD to get the vlan info might not be a bad idea for such cases...

jdn5126: Sure, I was only carrying this check over to preserve the existing behavior. Going through IPAMD when netns is already empty makes sense to me.

cmd/routed-eni-cni-plugin/cni.go
jayanthvn: I feel we just need to keep either handled or skipIpamd, returning 2 conditions with `tryDelWithPrevResult` can be changed.

achevuru:
General comment: We're changing the delete flow (i.e.,) tear down the pod network before deciding whether to call IPAMD or not, to release the IP address from the assigned list. Current flow is the recommended sequence, so I believe we should stick to that and fall back to `PrevResult` for Non branch ENI pods only if `IPAMD` call fails. It makes sense to do this for Branch ENI pods as `IPAMD` is not the IP address manager for them anyways, so there is not much value in invoking IPAMD. By sticking to the current flow sequence for non Branch ENI pods, we can also do away with `skipIpamd` check as this will purely be a fall back mechanism for some edge scenarios....

jayanthvn:
Agreed to what Apurup mentioned, we should keep the current flow and have a fall back when IPAMD call fails. The returning of 2 conditions from `tryDelWithPrevResult` would not be required then.

jdn5126:
> General comment: We're changing the delete flow (i.e.,) tear down the pod network first before deciding whether to call IPAMD or not, to release the IP address from the assigned list. Current flow is the recommended sequence, so I believe we should stick to that and fall back to `PrevResult` for Non branch ENI pods only if `IPAMD` call fails. It makes sense to do this for Branch ENI pods as `IPAMD` is not the IP address manager for them anyways, so there is not much value in invoking IPAMD. By sticking to the current flow sequence for non Branch ENI pods, we can also do away with `skipIpamd` check as this will purely be a fall back mechanism for some edge scenarios....

I did consider that, but that would involve calling in three cases:
1. IPAMD is unreachable
2. IPAMD returns error
3. IPAMD returns failure (!success)

I am fine with that approach, what do you think?

achevuru:
But is there a need to treat them all different? In all the above scenarios, we clean up what we can based on the PrevResult - similar to what you're trying to do in the PR before reaching out to IPAMD.  If the IP is missing in the state file, IPAMD anyways gives us a relevant error message and we return `success` in such a scenario back to kubelet and for any other error scenarios, we just let kubelet retry..

jdn5126:
> 

I thought about this, and I think the cleanest solution is to only call `TeardownPodNetwork` using `PrevResult` in the case where we cannot establish a connection with IPAMD. This means `tryDelWithPrevResult` would not be changed other than not returning an error when `podVlanId == 0`.

All IPAMD error conditions will cause kubelet to retry, but it is only the case where IPAMD is truly not running that the delete is critical to prevent stale rules. I will make these changes.

Issue #2109, opened 2022-10-18
Move go.mod in test directory to root go.mod


jdn5126:
**What would you like to be added**:
Move `go.mod` in test directory to use `go.mod` in root directory.

**Why is this needed**:
Managing dependencies is easier when they are all defined in one place. This is also closer to Golang recommended organization. 

Note that moving `go.mod` will require updating GitHub runners and (potentially) some import paths.

royninja:
but go.mod is already in root directory

jdn5126:
> but go.mod is already in root directory

There is an additional `go.mod` in the test directory, which can be removed, and its dependencies moved into the `go.mod` in the root directory.

jdn5126:
Closing as fixed

PR #2128, opened 2022-11-01
Downgrade go modules in test/go.mod to match root go.mod

jerryhe1999:
**What type of PR is this?**
cleanup

This is the first step in fixing #2109 

This PR downgrade the go modules in test/go.mod to match root go.mod.
This is needed because we want to test the same module version as we used in product code, and we eventually want to merge to a single module file.

N/A

Manually verified the test binaries build, unit tests passed, integration tests passed.

**Automation added to e2e**:
N/A

None

This will not break upgrades or downgrades. A running cluster has been tested.

No

No

```release-note

```


orsenthil:
> Which issue does this PR fix:
> This is the first step in fixing https://github.com/aws/amazon-vpc-cni-k8s/issues/2109

Thanks for the context and the comment regarding the replacement. Since these are test folder changes and the dependencies are now matching with the root go.mod - the changes look good to me.

In the next step, we can avoid doing any manual changes to go.mod

orsenthil (approved):
LGTM. Please get the approval from module owners too. thank you.

PR #2131, opened 2022-11-07
skip addon installation when addon info is not available

sushrk:
**What type of PR is this?**
bug

N/A

Adding changes to handle ipv4/ipv6 canary test script failure when addon information is not available. In this case, we will skip installing the EKS-managed addon, and tests will be run against the default self-managed addon. 

N/A

Yes, tested ```install_add_on()``` locally

**Automation added to e2e**:
No.

No.

No.

No

No.


orsenthil (approved):
LGTM.

orsenthil (approved):

Issue #2075, opened 2022-09-05
aws-vpc-cni: CRDs needs to be in crds folder


vadim-a-nesterov:
Hi,

we having issues with installing aws-vpc-cni and having race condition with ENIConfig CRDs, because helmwave is trying to create CRDs before CRD definition is applied. 

The aws-vpc-cni chart should be fixed, CRDs must be in separate folder like here
[](https://github.com/aws/eks-charts/tree/master/stable/appmesh-controller/crds)

```
STDERR:
  Error: unable to build kubernetes objects from release manifest: [resource mapping not found for name: "eu-west-1a" namespace: "" from "": no matches for kind "ENIConfig" in version "crd.k8s.amazonaws.com/v1alpha1"
  ensure CRDs are installed first, resource mapping not found for name: "eu-west-1b" namespace: "" from "": no matches for kind "ENIConfig" in version "crd.k8s.amazonaws.com/v1alpha1"
  ensure CRDs are installed first]

COMBINED OUTPUT:
  Release "aws-vpc-cni" does not exist. Installing it now.
  Error: unable to build kubernetes objects from release manifest: [resource mapping not found for name: "eu-west-1a" namespace: "" from "": no matches for kind "ENIConfig" in version "crd.k8s.amazonaws.com/v1alpha1"
  ensure CRDs are installed first, resource mapping not found for name: "eu-west-1b" namespace: "" from "": no matches for kind "ENIConfig" in version "crd.k8s.amazonaws.com/v1alpha1"
  ensure CRDs are installed first]
in ./helmfile.yaml: failed processing release aws-vpc-cni: command "/usr/local/bin/helm" exited with non-zero status:

```

jayanthvn:
Thanks, we will update and also update the installation doc to install CRDs and then run helm upgrade.

jayanthvn:
This is been actively worked on.

PR #2144, opened 2022-11-17
Helm dir update

jaydeokar:
**What type of PR is this?** 
cleanup

https://github.com/aws/amazon-vpc-cni-k8s/issues/2075

Defining chart folder structure as per helm v3 standard. Putting CRDs into `crds` folder also ensures they get installed before the chart is rendered and avoid any failures.

**Testing done on this change**: Yes tested install and upgrade on 1.23 EKS cluster. Helm v2 users will have to migrate to v3 using the documented steps provided by Helm.

**Automation added to e2e**: 
N/A

N/A

No break for upgrade/downgrade. Yes

No

No

### DIFF
--- a/charts/aws-vpc-cni/README.md
+++ b/charts/aws-vpc-cni/README.md
@@ -5,6 +5,7 @@ This chart installs the AWS CNI Daemonset: https://github.com/aws/amazon-vpc-cni
 ## Prerequisites
 
 - Kubernetes 1.11+ running on AWS
+- Helm v3
 
 ## Installing the Chart
 
@@ -21,6 +22,8 @@ $ helm install --name aws-vpc-cni --namespace kube-system eks/aws-vpc-cni
 ```
 
 To install into an EKS cluster where the CNI is already installed, see [this section below](#adopting-the-existing-aws-node-resources-in-an-eks-cluster)
+
+To migrate helm release for aws-vpc-cni chart from v2 to v3, see [ Migrate from helm v2 to helm v3 ](#migrate-from-helm-v2-to-helm-v3)
 
 ## Configuration
 
@@ -69,7 +72,6 @@ The following table lists the configurable parameters for this chart and their d
 | `serviceAccount.annotations` | Specifies the annotations for ServiceAccount       | `{}`                                |
 | `livenessProbe`         | Livenness probe settings for daemonset                  | (see `values.yaml`)                 |
 | `readinessProbe`        | Readiness probe settings for daemonset                  | (see `values.yaml`)                 |
-| `crd.create`            | Specifies whether to create the VPC-CNI CRD             | `true`                              |
 | `tolerations`           | Optional deployment tolerations                         | `[]`                                |
 | `updateStrategy`        | Optional update strategy                                | `type: RollingUpdate`               |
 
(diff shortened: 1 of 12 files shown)


charts/aws-vpc-cni/crds/customresourcedefinition.yaml
```diff
@@ -1,10 +1,9 @@
-{{- if .Values.crd.create -}}
```
jdn5126, line 1: Can you remove this field from `values.yaml` file? Once this is removed, I can approve

charts/aws-vpc-cni/crds/customresourcedefinition.yaml
```diff
@@ -1,10 +1,9 @@
...
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: eniconfigs.crd.k8s.amazonaws.com
   labels:
-{{ include "aws-vpc-cni.labels" . | indent 4 }}
```
jdn5126, line 7: Does this YAML file need all of the labels defined by `aws-vpc-cni.labels`? I think we are fine just keeping `k8s-app: aws-node` here

jdn5126 (commented):
Can you also run `make generate-cni-yaml` to update the manifests in `config/master/aws-vpc-cni*`?

charts/aws-vpc-cni/crds/customresourcedefinition.yaml
jaydeokar: We can get rid of those extra helm labels, they aren't required

jayanthvn:
Lets be careful before removing the labels because someone might be using it :)

charts/aws-vpc-cni/crds/customresourcedefinition.yaml
jdn5126: Unresolving as we decided it is safer to keep these labels

kishorj:
Also verify you can update the crds via `kubectl apply -k` option similar to other charts.

charts/aws-vpc-cni/crds/customresourcedefinition.yaml
```diff
@@ -1,10 +1,9 @@
...
+    k8s-app: aws-node
```
kishorj, line 6: any reason to add this label?

charts/aws-vpc-cni/README.md
```diff
@@ -81,7 +83,7 @@ $ helm install --name aws-vpc-cni --namespace kube-system eks/aws-vpc-cni --valu
...
 
-If you do not want to delete the existing aws-node resources in your cluster that run the aws-vpc-cni and then install this helm chart, you can adopt the resources into a release instead. This process is highlighted in this [PR comment](https://github.com/aws/eks-charts/issues/57#issuecomment-628403245). Once you have annotated and labeled all the resources this chart specifies, enable the `originalMatchLabels` flag, and also set `crd.create` to false on the helm release and run an update. If you have been careful this should not diff and leave all the resources unmodified and now under management of helm.
+If you do not want to delete the existing aws-node resources in your cluster that run the aws-vpc-cni and then install this helm chart, you can adopt the resources into a release instead. This process is highlighted in this [PR comment](https://github.com/aws/eks-charts/issues/57#issuecomment-628403245). Once you have annotated and labeled all the resources this chart specifies, enable the `originalMatchLabels` flag. If you have been careful this should not diff and leave all the resources unmodified and now under management of helm.
```
kishorj, line 86: Please consider including the information as part of the README instead of pointing to the PR comment.

charts/aws-vpc-cni/README.md
```diff
@@ -5,6 +5,7 @@ This chart installs the AWS CNI Daemonset: https://github.com/aws/amazon-vpc-cni
 ## Prerequisites
 
 - Kubernetes 1.11+ running on AWS
```
kishorj, line 7: is k8s 1.11 still the oldest version supported?

charts/aws-vpc-cni/crds/customresourcedefinition.yaml
jaydeokar: I thought we are using this label, seems like not. I'll be removing this and keep it consistent with the crd that gets installed by default on a cluster

charts/aws-vpc-cni/README.md
jaydeokar: This is already added in the README, I've removed the link since we anyway have this process defined in the script present in README

charts/aws-vpc-cni/README.md
jaydeokar: I am not sure on this one. I don't know if we have a way to determine the versions which are actively supported
@jayanthvn

kishorj:
LGTM

mindw:
From Helm [documentation](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/):

> Some caveats (and explanations)
There is no support at this time for upgrading or deleting CRDs using Helm. 

The above translates to special upgrade instructions whenever a CRD is changed. It also doesn't solve the chicken and egg problem (see rest of `Some caveats` above).
.
Please consider reverting the PR and suggesting option 2 as for #2075.

jdn5126:
@mindw we did consider this problem and decided that a `crds` folder within the `aws-vpc-cni` helm chart is the best solution. At this time, we only have one CRD, `ENIConfig`, and the definition is not something that will change frequently, if ever. If we ever add another CRD or change the definition for `ENIConfig`, we will include special instructions in the release notes for how to install the new CRDs.

The `crds` folder is still the recommended method and operator standard. If we get more CRDs and this is not as simple to manage, then we will use two charts (method #2).

PR #2145, opened 2022-11-17
[CNI]: Teardown pod network when IPAMD connection fails

jdn5126:
**What type of PR is this?**
bug

#2048 

Note that this PR replaces https://github.com/aws/amazon-vpc-cni-k8s/pull/2125

This PR resolves an issue in which IP rules were leaked by the CNI. When processing a pod deletion, the CNI would wait for IPAMD response before tearing down pod networking resources. If IPAMD could not be reached, CNI would return error and wait for kubelet to retry the delete. If IPAMD were restarted, the state for this pod would be cleared without CNI tearing down the associated networking resources. The trigger for the linked issue was the k8s cluster autoscaler evicting the `aws-node` daemonset pod before other pods and then later cancelling the pod evictions. https://github.com/kubernetes/autoscaler/issues/5240 was filed to ask for k8s cluster autoscaler to change its behavior.

The changes in this PR are two-fold:
1. non-branch ENI pods now store state in PrevResult for later cleanup
2. when IPAMD connection fails (two cases), we try to cleanup pod network using PrevResult

There is a lot of duplication between `teardownPodNetworkWithPrevResult` and `tryDelWithPrevResult`. I kept them separate to avoid unnecessarily complicating `tryDelWithPrevResult` and to make it clear that `teardownPodNetworkWithPrevResult` is a fallback mechanism.

Added more test cases to `cni_test.go` and verified that all CNI and IPAMD integration tests pass with this change. Also manually verified the fix with IPv4 and IPv6 clusters.

**Automation added to e2e**:
N/A

No

This will not break upgrades or downgraded. A running cluster has been tested.

No

No

```release-note
Cleanup pod networking resources when IPAMD is unreachable to prevent rule leaking.
```

### DIFF
--- a/cmd/routed-eni-cni-plugin/cni.go
+++ b/cmd/routed-eni-cni-plugin/cni.go
@@ -47,7 +47,7 @@ import (
 
 const ipamdAddress = "127.0.0.1:50051"
 
-const dummyVlanInterfacePrefix = "dummy"
+const dummyInterfacePrefix = "dummy"
 
 var version string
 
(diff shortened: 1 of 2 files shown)


M00nF1sh (approved):
/lgtm

M00nF1sh (approved):

PR #2137, opened 2022-11-11
Reduce Startup Latency by moving the init container to a regular container

bwagner5:
**What type of PR is this?**
 - feature

**Which issue does this PR fix**:
https://github.com/awslabs/amazon-eks-ami/issues/1099

**What does this PR do / Why do we need it**:
 - The VPC CNI controls when a node becomes usable for pod scheduling and execution so the startup time of the CNI should be optimized to bring nodes online faster. This PR improves the startup time of the VPC CNI by >50% (from 9 sec to 4 sec) by moving the init container to a regular container in the pod and synchronizing the vpc-cni container via a shared emptyDir volume. This allows the containers to initialize in parallel and avoids the kubelet latency of switching between the init container and the regular container. 

**If an issue # is not available please add repro steps and logs from IPAMD/CNI showing the issue**:

**Testing done on this change**:

I installed the changes into my cluster (1.22) via helm and ran tests on the current production version (v1.12.0) and my change. Below are representative samples across 10 runs. 

The "VPC CNI Plugin Initialized" event is when the CNI config file is moved to the `/etc/cni/net.d` dir, signaling to the container runtime that networking is ready. 

**Previous aws-node startup timing (9 secs):**
```
|     Event                     | Timestamp | t  |
|-------------------------------|-----------|----|
| VPC CNI Init Container Starts | 05:29:42  | 0s |
| AWS Node Container Starts     | 05:29:49  | 7s |
| VPC CNI Plugin Initialized    | 05:29:51  | 9s |
```

**After this change startup timing (4 secs):**
```
|     Event                     | Timestamp | t  |
|-------------------------------|-----------|----|
| AWS Node Container Starts     | 05:30:30  | 0s |
| VPC CNI Init Container Starts | 05:30:32  | 2s |
| VPC CNI Plugin Initialized    | 05:30:34  | 4s |
```

**Automation added to e2e**:

**Will this PR introduce any new dependencies?**:

No

**Will this break upgrades or downgrades. Has updating a running cluster been tested?**:

I have not tested updating a running cluster, but the changes are only on improving the startup sequence.

**Does this change require updates to the CNI daemonset config files to work?**:

Yes, this change requires moving the previous init-container under the regular containers list in the DaemonSet and involves adding an emptyDir volume and mapping in both containers. 

**Does this PR introduce any user-facing change?**:

No

```release-note

```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2146, opened 2022-11-21
VPC-CNI minimal image builds

jdn5126:
**What type of PR is this?**
Enhancement

N/A

This pull request migrates the VPC-CNI and VPC-CNI init images to use EKS minimal build image. This PR applied the diff from #1726 and updated it to handle recent changes. It was also tested to be compatible with #2137 , but those changes were left commented out as #2137 is not yet complete.

The VPC CNI image is based on `public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-iptables:latest.2`, which installs `iptables-legacy` and `iptables-nft`, but sets `iptables-legacy` as the default using `update-alternatives`.

The VPC CNI init image is based on `public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-glibc:latest.2`. The reason I could not use the base image is that the binaries are built in `pie` mode for security reasons, and building with `pie` requires the loader to be loaded despite the binary being static.

I considered splitting this into two PRs: one to add the Golang entrypoints and one to do the move, but concluded that it is easiest to do as one PR, and if there are any issues, it is simple to just revert the move.

N/A

I manually verified that all integration tests pass against this change. Output for init container:
```
$ kubectl logs aws-node-rg6rm -n kube-system -c aws-vpc-cni-init
time="2022-11-21T19:35:43Z" level=info msg="Copying CNI plugin binaries ..."
time="2022-11-21T19:35:43Z" level=info msg="Copied all CNI plugin binaries to /host/opt/cni/bin\n"
Installed /host/opt/cni/bin/loopback
Installed /host/opt/cni/bin/portmap
Installed /host/opt/cni/bin/bandwidth
Installed /host/opt/cni/bin/aws-cni-support.sh
time="2022-11-21T19:35:43Z" level=info msg="Found primaryMAC 0a:55:51:01:a4:d1"
time="2022-11-21T19:35:43Z" level=info msg="Found primaryIF eth0"
time="2022-11-21T19:35:43Z" level=info msg="Updated net/ipv4/conf/eth0/rp_filter to 2"
time="2022-11-21T19:35:43Z" level=info msg="Updated net/ipv4/tcp_early_demux to 1"
time="2022-11-21T19:35:43Z" level=info msg="CNI init container done"
```

Output for VPC-CNI:
```
$ kubectl logs aws-node-rg6rm -n kube-system -c aws-node
time="2022-11-21T19:35:43Z" level=info msg="Install CNI binaries.."
Installed /host/opt/cni/bin/aws-cni
time="2022-11-21T19:35:43Z" level=info msg="Starting IPAM daemon... "
time="2022-11-21T19:35:43Z" level=info msg="Checking for IPAM connectivity... "
Installed /host/opt/cni/bin/egress-v4-cni
time="2022-11-21T19:35:44Z" level=info msg="Copying config file... "
time="2022-11-21T19:35:44Z" level=info msg="Successfully copied CNI plugin binary and config file."
```

**Automation added to e2e**:
N/A

No, it will actually reduce dependencies.

This will not break upgrades or downgrades. Upgrade and downgrade has been tested with a running cluster.

No

The image base is different.

```release-note
Move VPC CNI and VPC CNI init images to use EKS minimal base image.
```


bwagner5:
Nice! Do you know the full image size differences?

jdn5126:
> Nice! Do you know the full image size differences?

Comparing upstream (v1.12.0) to my personal ECR:
VPC-CNI Init: 277.24 MB -> 19.90MB
VPC-CNI: 293.37 MB -> 47.67MB

cmd/aws-vpc-cni-init/main.go
```diff
@@ -0,0 +1,177 @@
...
+	os.Exit(_main())
+}
+
+func _main() int {
+	log.Debug("Started Initialization")
+	pluginBins := []string{"loopback", "portmap", "bandwidth", "aws-cni-support.sh"}
+	var err error
+	for _, plugin := range pluginBins {
+		if _, err = os.Stat(plugin); err != nil {
+			log.WithError(err).Fatalf("Required executable: %s not found\n", plugin)
```
M00nF1sh, line 55: we don't need the \n in logs.

cmd/aws-vpc-cni-init/main.go
```diff
@@ -0,0 +1,177 @@
...
+	vpcCniInitDonePath    = "/vpc-cni-init/done"
+	metadataLocalIP       = "local-ipv4"
+	metadataMAC           = "mac"
+
+	envDisableIPv4TcpEarlyDemux = "DISABLE_TCP_EARLY_DEMUX"
+	envEnableIPv6               = "ENABLE_IPv6"
+	envHostCniBinPath           = "HOST_CNI_BIN_PATH"
+)
+
+func getEnv(env, def string) string {
```
M00nF1sh, line 38: nit: i prefer to avoid use non-standard abbreviation and simply use `default` :D

cmd/aws-vpc-cni-init/main.go
```diff
@@ -0,0 +1,177 @@
...
+	entry = "net/ipv4/tcp_early_demux"
+	if disableIPv4EarlyDemux == "true" {
+		err = sys.SetSysctl(entry, 0)
+		if err != nil {
+			log.WithError(err).Fatalf("Failed to disable tcp_early_demux\n")
+			return 1
+		}
+	} else {
+		err = sys.SetSysctl(entry, 1)
+		if err != nil {
```
M00nF1sh, line 123: the original shell-based script have '-e'(Use this option to ignore errors about unknown keys), while Golang version don't. wondering why we have it in the shell version

cmd/aws-vpc-cni-init/main.go
```diff
@@ -0,0 +1,177 @@
...
+		err = sys.SetSysctl(entry, 2)
+		if err != nil {
+			log.WithError(err).Fatalf("Failed to enable ipv6 accept_ra\n")
+			return 1
+		}
+		val, _ = sys.GetSysctl(entry)
+		log.Infof("Updated %s to %d", entry, val)
+	}
+
+	// TODO: In order to speed up pod launch time, VPC CNI init container is not a Kubernetes init container.
```
M00nF1sh, line 164: just a thought, why we make it two container and sync using filePath?
how about just one container, that runs the init first then rest? (if we want parallel processing, we can use goroutings)

M00nF1sh: also nit:
since we are using golang, i prefer to split this "main" into smaller functions, such as `findNetLinkByMac`  `configureRPFilter()`, `configureIPv6Settings`

cmd/aws-vpc-cni-init/main.go
jdn5126: `default` is a reserved keyword in golang, so this was abbreviated. Do you think `defaultVal` is better?

cmd/aws-vpc-cni-init/main.go
jdn5126: Ah nice catch, looks like this comes from https://github.com/aws/amazon-vpc-cni-k8s/issues/1241 . Not sure if we still need to support kernels this old anymore, but it seems worth catching this error

cmd/aws-vpc-cni-init/main.go
jdn5126: That's a good point. Now that #2137 is removing the init container, we can actually just merge the two. I thought the init container was separate as it needed escalated privileges for setup, but it looks like that is not the case. I think we should do this in stages, though, i.e. this should merge with init container support, then #2137 can be rebased to merge with a single container.

cmd/aws-vpc-cni-init/main.go
M00nF1sh: defaultVal sounds good to me :D      `def ` is more like short for definition

cmd/aws-vpc-cni-init/main.go
jdn5126: Did some more digging, and it looks like we cannot merge the two containers as the `init` container requires privileged  permissions to set `/proc/sys/net/ipv4/conf/eth0/rp_filter` to loose mode.

I see you had a suggestion in https://github.com/aws/amazon-vpc-cni-k8s/issues/796 to use `src_valid_mark` instead, so I am trying that now

jdn5126: Oh nvm, setting that requires the same level of privilege, so yeah privilege is the limitation for the two containers

cmd/aws-vpc-cni-init/main.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	val, _ = sys.GetSysctl(entry)
+	log.Infof("Updated %s to %d", entry, val)
+
+	// Enable IPv6 when environment variable is set
+	// Note that IPv6 is not disabled when environment variable is unset. This is omitted to preserve default host semantics.
+	enableIPv6 := getEnv(envEnableIPv6, "false")
+	if enableIPv6 == "true" {
+		entry = "net/ipv6/conf/all/disable_ipv6"
```
kishorj, line 136: use sysctl format instead of the path, for example:
net.ipv6.conf.all.disable_ipv6 and do the necessary translation from the sysctl package. It will be easier to correlate with the sysctl documentation.

cmd/aws-vpc-cni-init/main.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+	}
+
+	if primaryIF == "" {
+		log.Errorf("Failed to retrieve primary IF")
+		return 1
+	}
+	log.Infof("Found primaryIF %s", primaryIF)
+
+	sys := sysctl.New()
```
kishorj, line 100: use a different name - sys sounds confusing - maybe sysctlUtil

utils/sysctl/sysctl.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+const (
+	sysctlBase = "/proc/sys"
+)
+
+// Interface is an injectable interface for running sysctl commands.
+type Interface interface {
+	// GetSysctl returns the value for the specified sysctl setting
+	GetSysctl(sysctl string) (int, error)
```
kishorj, line 19: why not Get and Set instead?

cmd/aws-vpc-cni-init/main.go
jdn5126: In the latest diff, I broke up `aws-vpc-cni-init/main.go:main` into several parts. For `aws-vpc-cni/main.go:main`, I think it was already split up pretty well

scripts/dockerfiles/Dockerfile.init
```diff
@@ -7,16 +7,18 @@ ARG TARGETARCH
...
 COPY . ./
+RUN make build-aws-vpc-cni-init
 
-# Build the architecture specific container image:
-FROM public.ecr.aws/amazonlinux/amazonlinux:2
-RUN yum update -y && \
-    yum install -y iproute procps-ng && \
-    yum clean all
+# Build from EKS minimal base + glibc
+FROM public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-glibc:latest.2
```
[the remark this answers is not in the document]
jdn5126: I am fine either way, but the reason I was thinking `latest` is ok here is that updates to this base repo are highly scrutinized and rarely happen, so we get the benefit of automatically getting fixes and updates without much stability risk or churn

cmd/aws-vpc-cni-init/main.go
jdn5126: Our `sysctl` library actually writes to that file path, i.e. calls `WriteFile` on `proc/sys/net/ipv6/conf/all/disable_ipv6`, so I think the readability improvement of passing it in dotted format just to string replace back to path format may not be worth the computational overhead. I am fine either way, though

cmd/aws-vpc-cni-init/main.go
jdn5126: will do

utils/sysctl/sysctl.go
jdn5126: No reason, I was just copying what was originally done. Changing to `Get` and `Set` now

cmd/aws-vpc-cni-init/main.go
jayanthvn: Ref for init-container use case - https://github.com/aws/amazon-vpc-cni-k8s/issues/796

cmd/aws-vpc-cni-init/main.go
```diff
@@ -0,0 +1,195 @@
...
+	if err != nil {
+		log.WithError(err).Errorf("Failed to configure IPv6 settings")
+		return 1
+	}
+
+	// TODO: In order to speed up pod launch time, VPC CNI init container is not a Kubernetes init container.
+	// The VPC CNI container blocks on the existence of vpcCniInitDonePath
+	//err = cp.TouchFile(vpcCniInitDonePath)
+	//if err != nil {
+	//	log.WithError(err).Errorf("Failed to set VPC CNI init done")
```
jayanthvn, line 186: Lets delete these lines.

cmd/aws-vpc-cni-init/main.go
```diff
@@ -0,0 +1,195 @@
...
+	//	return 1
+	//}
+
+	log.Infof("CNI init container done")
+
+	// TODO: Since VPC CNI init container is a real container, it never exits
```
jayanthvn, line 192: Same as above, lets clean it up.

cmd/aws-vpc-cni-init/main.go
jdn5126: These are going to be uncommented for #2137 , so I was leaving them to make it easier for that PR to rebase

cmd/aws-vpc-cni/main.go
```diff
@@ -0,0 +1,409 @@
...
+		cmd.Stdout = &outb
+		cmd.Run()
+		if outb.String() == "" {
+			return true
+		}
+	}
+}
+
+// Wait for vpcCniInitDonePath to exist (maximum wait time is 60 seconds)
+func waitForInit() error {
```
jayanthvn, line 179: This function is anyways commented, we can delete it.

jdn5126: same as above, this is temporarily commented to simplify rebase for #2137

cmd/aws-vpc-cni-init/main.go
jayanthvn: I see, sure.

scripts/dockerfiles/Dockerfile.init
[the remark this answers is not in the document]
jayanthvn: Lets keep it as latest similar to what we have today. Anyways for CVEs we do a rebuild and get all the latest packages.

cmd/aws-vpc-cni-init/main.go
bryantbiggs: This will make it harder since it would in theory cause a conflict as opposed to a net new addition in the follow on PR

jdn5126: These files do not exist in the new PR, so will not cause any conflict, and the entrypoint scripts the new PR is modifying are deleted by this PR, so will cause a conflict, but that was unavoidable and the resolution is easy (accept delete of file). #2137 needs cleanup anyway before it can be merged, so the goal is for the diff in these files to just be the uncommented lines

bryantbiggs: Ok, that's just a bit odd but ok

cmd/aws-vpc-cni-init/main.go
```diff
@@ -0,0 +1,195 @@
...
+		return val
+	}
+	return defaultVal
+}
+
+func getNodePrimaryIF() (string, error) {
+	var primaryIF string
+	primaryMAC, err := imds.GetMetaData("mac")
+	if err != nil {
+		return primaryIF, errors.Wrap(err, "Failed to get primary MAC from IMDS")
```
M00nF1sh, line 50: nit: prefer to just return `""` when an error is returned.

cmd/aws-vpc-cni-init/main.go
```diff
@@ -0,0 +1,195 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "Failed to set rp_filter for %s", primaryIF)
+	}
+	val, _ := sysctlUtil.Get(entry)
+	log.Infof("Updated %s to %d", entry, val)
+
+	// Enable or disable TCP early demux based on environment variable
+	// Note that older kernels may not support tcp_early_demux, so we must first check that it exists.
+	entry = "net/ipv4/tcp_early_demux"
+	if _, err := sysctlUtil.Get(entry); err != nil {
```
M00nF1sh, line 85: shouldn't this be `err == nil`?

cmd/aws-vpc-cni-init/main.go
```diff
@@ -0,0 +1,177 @@
...
+	return def
+}
+
+func main() {
+	os.Exit(_main())
+}
+
+func _main() int {
+	log.Debug("Started Initialization")
+	pluginBins := []string{"loopback", "portmap", "bandwidth", "aws-cni-support.sh"}
```
achevuru: We need `host-local` plugin for IPv6  - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/misc/10-aws.conflist#L23

cmd/aws-vpc-cni-init/main.go
jdn5126: If the path to `tcp_early_demux` does not exist, we return error. If an error is returned, we do not want to write to `tcp_early_demux`

cmd/aws-vpc-cni-init/main.go
```diff
@@ -0,0 +1,177 @@
...
+
+const (
+	defaultHostCNIBinPath = "/host/opt/cni/bin"
+	vpcCniInitDonePath    = "/vpc-cni-init/done"
+	metadataLocalIP       = "local-ipv4"
+	metadataMAC           = "mac"
+
+	envDisableIPv4TcpEarlyDemux = "DISABLE_TCP_EARLY_DEMUX"
+	envEnableIPv6               = "ENABLE_IPv6"
+	envHostCniBinPath           = "HOST_CNI_BIN_PATH"
```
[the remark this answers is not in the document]
jdn5126: `HOST_CNI_BIN_PATH` was copied from `scripts/init.sh`. The default value is `/host/opt/cni/bin`, but we do currently allow for it to be overridden. We do not advertise this env var, though

cmd/aws-vpc-cni-init/main.go
jdn5126: I do see `host-local` in entrypoint script here: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/entrypoint.sh#L154 , but it is not present in init script: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/init.sh#L25 . I assume it should be in both?

Also, it does seem like IPv6 test suite passes without this. I will add it, though.

cmd/aws-vpc-cni/main.go
```diff
@@ -0,0 +1,410 @@
...
+		log.Errorf("AWS_VPC_K8S_CNI_VETHPREFIX cannot be longer than 4 characters")
+		return false
+	}
+
+	if vethPrefix == "eth" || vethPrefix == "lo" || vethPrefix == "vlan" {
+		log.Errorf("AWS_VPC_K8S_CNI_VETHPREFIX cannot be set to reserved values 'eth', 'vlan', or 'lo'")
+		return false
+	}
+
+	podSGEnforcingMode := getEnv(envPodSGEnforcingMode, defaultPodSGEnforcingMode)
```
achevuru, line 302: Should this check be paired with `ENABLE_POD_ENI`? Why make this a mandatory check for all modes?

cmd/aws-vpc-cni-init/main.go
jayanthvn: I think we can keep it only in init script. It is confusing to have it in two places. We can take this chance to cleanup the code.

jayanthvn: btw @achevuru , some of my code was implemented before v6 hence it is a good callout to test v6 before merging.

jdn5126: but when someone sets `AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER` to true, which is the default, then it is main container that should install this binary

cmd/aws-vpc-cni/main.go
jdn5126: I was just copying the check from the existing `entrypoint` script, but yeah, makes sense to check this only when `ENABLE_POD_ENI` is set.

cmd/aws-vpc-cni-init/main.go
jdn5126: Oh, nvm, you are right here. I was not being thorough in testing after this change

cmd/aws-vpc-cni/main.go
```diff
@@ -0,0 +1,410 @@
...
+
+	err = cp.CopyFile(tmpAWSconflistFile, defaultHostCNIConfDirPath+awsConflistFile)
+	if err != nil {
+		log.WithError(err).Errorf("Failed to update 10-awsconflist")
+		return 1
+	}
+
+	log.Infof("Successfully copied CNI plugin binary and config file.")
+
+	awsConfFile := defaultHostCNIConfDirPath + "/aws.conf"
```
achevuru, line 393: I wonder if we really need this anymore. v1.6.0+ anyways migrated to `10-aws.conflist` and it has been 2+ years already.

cmd/aws-vpc-cni-init/main.go
jayanthvn: Init container will always be called right? so one place should be fine. I think it was added initially when init container was first supported..

cmd/aws-vpc-cni/main.go
jdn5126: The only thing we are doing here is deleting this file if it exists. But since no one consumes it and we do not allow upgrades from version < `v1.6.0` to `v1.12.1`, it does seem like this can be deleted. I'll test it out now

cmd/aws-vpc-cni-init/main.go
jdn5126: I guess the question then is whether we even support `AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER` being configured as `"false"`. Currently, both the init container and the main container installs these binaries. Since the init container does it unconditionally, there is no need for the main container to do it either way

jdn5126: nvm, it looks like `AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER` is used here: https://github.com/aws/amazon-vpc-cni-k8s/blob/f21b587fdf6370a01b181ddce797398eb7951a23/pkg/networkutils/network.go#L298 . So really, this env variable does not make sense as a gate for the main entrypoint installing the same binaries that init entrypoint already did

cmd/aws-vpc-cni-init/main.go
jdn5126: fixed in latest revision

cmd/aws-vpc-cni-init/main.go
jdn5126: Ran all IPv6 and IPv4 test suites and verified that they passed. Modified entrypoint scripts such that only init container installs external plugins. This also allowed me to simplify dockerfiles.

cmd/aws-vpc-cni/main.go
jdn5126: removed in latest revision

achevuru (approved):

bryantbiggs:
awesome work @jdn5126 🎉

PR #2149, opened 2022-11-29
Using events to signal the enabled Security Group for Pods feature

haouc:
**What type of PR is this?**
The PR is changing how CNI send signals to [amazon-vpc-resource-controller](https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/157) for enabled feature of Security Group for Pods. Amazon VPC CNI is labeling nodes object to send the signal to the controller when the feature is enabled by user. We should remove the permission labeling node and use alternatives to handle the process.

feature
N/A

This PR changes the way sending signal to another Amazon service.

N/A

Tested in EKS clusters to test if the nodes can be enabled and pods using Security Group can be created successfully.
```
% k get events | grep AwsNodeNotificationToRc
18m         Normal    AwsNodeNotificationToRc   node/ip-192-168-x-x.us-west-2.compute.internal    vpc.amazonaws.com/eniConfig=us-west-2a
18m         Normal    AwsNodeNotificationToRc   node/ip-192-168-x-x.us-west-2.compute.internal    vpc.amazonaws.com/has-trunk-attached=false
18m         Normal    AwsNodeNotificationToRc   node/ip-192-168-x-x.us-west-2.compute.internal    vpc.amazonaws.com/has-trunk-attached=false
18m         Normal    AwsNodeNotificationToRc   node/ip-192-168-x-x.us-west-2.compute.internal   vpc.amazonaws.com/eniConfig=us-west-2b
18m         Normal    AwsNodeNotificationToRc   node/ip-192-168-x-x.us-west-2.compute.internal   vpc.amazonaws.com/has-trunk-attached=false
18m         Normal    AwsNodeNotificationToRc   node/ip-192-168-x-x.us-west-2.compute.internal   vpc.amazonaws.com/has-trunk-attached=false
```
**Automation added to e2e**:

This PR doesn't change functionality and no need to add new e2e modules.

No
Yes. This has to be used with new version of amazon-vpc-resource-controller. 

not necessary. but if downgrade VPC CNI without adding back node update permission, security group for pods will not work with older version of VPC CNI.

Yes, nodes will no longer be modified by VPC CNI and new node events will be emitted.
```release-note

```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -248,6 +248,8 @@ type EC2InstanceMetadataCache struct {
 
 	imds   TypedIMDS
 	ec2SVC ec2wrapper.EC2
+
+	eventRecorder *eventrecorder.EventRecorder
 }
 
 // ENIMetadata contains information about an ENI
@@ -394,7 +396,7 @@ func (i instrumentedIMDS) GetMetadataWithContext(ctx context.Context, p string) 
 }
 
 // New creates an EC2InstanceMetadataCache
-func New(useCustomNetworking, disableENIProvisioning, v4Enabled, v6Enabled bool) (*EC2InstanceMetadataCache, error) {
+func New(useCustomNetworking, disableENIProvisioning, v4Enabled, v6Enabled bool, eventRecorder *eventrecorder.EventRecorder) (*EC2InstanceMetadataCache, error) {
 	//ctx is passed to initWithEC2Metadata func to cancel spawned go-routines when tests are run
 	ctx := context.Background()
 
@@ -408,6 +410,7 @@ func New(useCustomNetworking, disableENIProvisioning, v4Enabled, v6Enabled bool)
 	cache.imds = TypedIMDS{instrumentedIMDS{ec2Metadata}}
 	cache.clusterName = os.Getenv(clusterNameEnvVar)
 	cache.additionalENITags = loadAdditionalENITags()
+	cache.eventRecorder = eventRecorder
 
 	region, err := ec2Metadata.Region()
 	if err != nil {
@@ -432,8 +435,6 @@ func New(useCustomNetworking, disableENIProvisioning, v4Enabled, v6Enabled bool)
 	if err != nil {
 		return nil, err
 	}
-	// event recorder to raise events for failed EC2 API calls
-	eventRecorder = eventrecorder.Get()
 
 	// Clean up leaked ENIs in the background
 	if !disableENIProvisioning {
--- a/pkg/awsutils/awsutils_test.go
+++ b/pkg/awsutils/awsutils_test.go
@@ -31,6 +31,13 @@ import (
 	"github.com/aws/aws-sdk-go/aws/awserr"
 
 	mock_ec2wrapper "github.com/aws/amazon-vpc-cni-k8s/pkg/ec2wrapper/mocks"
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/eventrecorder"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/events"
+	testclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
 const (
@@ -73,6 +80,7 @@ const (
 	eni2v6Prefix         = "2001:db8::/64"
 	eni2ID               = "eni-12341234"
 	metadataVPCIPv4CIDRs = "192.168.0.0/16	100.66.0.0/1"
+	myNodeName           = "testNodeName"
 )
 
 func testMetadata(overrides map[string]interface{}) FakeIMDS {
@@ -131,6 +139,28 @@ func setup(t *testing.T) (*gomock.Controller,
 		mock_ec2wrapper.NewMockEC2(ctrl)
 }
 
+func setupEventRecorder(t *testing.T) *eventrecorder.EventRecorder {
+	ctx := context.Background()
+	fakeRecorder := events.NewFakeRecorder(3)
+	k8sSchema := runtime.NewScheme()
+	clientgoscheme.AddToScheme(k8sSchema)
+
+	mockEventRecorder := &eventrecorder.EventRecorder{
+		Recorder:        fakeRecorder,
+		RawK8SClient:    testclient.NewClientBuilder().WithScheme(k8sSchema).Build(),
+		CachedK8SClient: testclient.NewClientBuilder().WithScheme(k8sSchema).Build(),
+	}
+
+	fakeNode := v1.Node{
+		TypeMeta:   metav1.TypeMeta{Kind: "Node"},
+		ObjectMeta: metav1.ObjectMeta{Name: myNodeName},
+		Spec:       v1.NodeSpec{},
+		Status:     v1.NodeStatus{},
+	}
+	_ = mockEventRecorder.CachedK8SClient.Create(ctx, &fakeNode)
+	return mockEventRecorder
+}
+
 func TestInitWithEC2metadata(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Millisecond)
 	defer cancel()
@@ -139,15 +169,15 @@ func TestInitWithEC2metadata(t *testing.T) {
 	defer ctrl.Finish()
 	mockMetadata := testMetadata(nil)
 
-	ins := &EC2InstanceMetadataCache{imds: TypedIMDS{mockMetadata}, ec2SVC: mockEC2}
-	err := ins.initWithEC2Metadata(ctx)
+	cache := &EC2InstanceMetadataCache{imds: TypedIMDS{mockMetadata}, ec2SVC: mockEC2}
+	err := cache.initWithEC2Metadata(ctx)
 	if assert.NoError(t, err) {
-		assert.Equal(t, az, ins.availabilityZone)
-		assert.Equal(t, localIP, ins.localIPv4.String())
-		assert.Equal(t, ins.instanceID, instanceID)
-		assert.Equal(t, ins.primaryENImac, primaryMAC)
-		assert.Equal(t, ins.primaryENI, primaryeniID)
-		assert.Equal(t, subnetID, ins.subnetID)
+		assert.Equal(t, az, cache.availabilityZone)
+		assert.Equal(t, localIP, cache.localIPv4.String())
+		assert.Equal(t, cache.instanceID, instanceID)
+		assert.Equal(t, cache.primaryENImac, primaryMAC)
+		assert.Equal(t, cache.primaryENI, primaryeniID)
+		assert.Equal(t, subnetID, cache.subnetID)
 	}
 }
 
@@ -168,11 +198,11 @@ func TestInitWithEC2metadataErr(t *testing.T) {
 			key: fmt.Errorf("An error with %s", key),
 		})
 
-		ins := &EC2InstanceMetadataCache{imds: TypedIMDS{mockMetadata}, ec2SVC: mockEC2}
+		cache := &EC2InstanceMetadataCache{imds: TypedIMDS{mockMetadata}, ec2SVC: mockEC2}
 
 		// This test is a bit silly.  We expect broken metadata to result in an err return here.  But if the code is resilient and _succeeds_, then of course that's ok too.  Mostly we just want it not to panic.
 		assert.NotPanics(t, func() {
-			_ = ins.initWithEC2Metadata(ctx)
+			_ = cache.initWithEC2Metadata(ctx)
 		}, "Broken metadata %s resulted in panic", key)
 	}
 }
@@ -186,8 +216,8 @@ func TestGetAttachedENIs(t *testing.T) {
 		metadataMACPath + eni2MAC + metadataIPv4s:      eni2PrivateIP,
 	})
 
-	ins := &EC2InstanceMetadataCache{imds: TypedIMDS{mockMetadata}}
-	ens, err := ins.GetAttachedENIs()
+	cache := &EC2InstanceMetadataCache{imds: TypedIMDS{mockMetadata}}
+	ens, err := cache.GetAttachedENIs()
 	if assert.NoError(t, err) {
 		assert.Equal(t, len(ens), 2)
 	}
@@ -203,8 +233,8 @@ func TestGetAttachedENIsWithPrefixes(t *testing.T) {
 		metadataMACPath + eni2MAC + metadataIPv4Prefixes: eni2Prefix,
 	})
 
-	ins := &EC2InstanceMetadataCache{imds: TypedIMDS{mockMetadata}}
-	ens, err := ins.GetAttachedENIs()
+	cache := &EC2InstanceMetadataCache{imds: TypedIMDS{mockMetadata}}
+	ens, err := cache.GetAttachedENIs()
 	if assert.NoError(t, err) {
 		assert.Equal(t, len(ens), 2)
 	}
@@ -217,8 +247,8 @@ func TestAWSGetFreeDeviceNumberOnErr(t *testing.T) {
 	// test error handling
 	mockEC2.EXPECT().DescribeInstancesWithContext(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, errors.New("error on DescribeInstancesWithContext"))
 
-	ins := &EC2InstanceMetadataCache{ec2SVC: mockEC2}
-	_, err := ins.awsGetFreeDeviceNumber()
+	cache := &EC2InstanceMetadataCache{ec2SVC: mockEC2}
+	_, err := cache.awsGetFreeDeviceNumber()
 	assert.Error(t, err)
 }
 
@@ -240,8 +270,8 @@ func TestAWSGetFreeDeviceNumberNoDevice(t *testing.T) {
 
 	mockEC2.EXPECT().DescribeInstancesWithContext(gomock.Any(), gomock.Any(), gomock.Any()).Return(result, nil)
 
-	ins := &EC2InstanceMetadataCache{ec2SVC: mockEC2}
-	_, err := ins.awsGetFreeDeviceNumber()
+	cache := &EC2InstanceMetadataCache{ec2SVC: mockEC2}
+	_, err := cache.awsGetFreeDeviceNumber()
 	assert.Error(t, err)
 }
 
@@ -300,8 +330,8 @@ func TestGetENIAttachmentID(t *testing.T) {
 	for _, tc := range testCases {
 		mockEC2.EXPECT().DescribeNetworkInterfacesWithContext(gomock.Any(), gomock.Any(), gomock.Any()).Return(tc.output, tc.awsErr)
 
-		ins := &EC2InstanceMetadataCache{ec2SVC: mockEC2}
-		id, err := ins.getENIAttachmentID("test-eni")
+		cache := &EC2InstanceMetadataCache{ec2SVC: mockEC2}
+		id, err := cache.getENIAttachmentID("test-eni")
 		assert.Equal(t, tc.expErr, err)
 		assert.Equal(t, tc.expID, id)
 	}
@@ -343,8 +373,8 @@ func TestDescribeAllENIs(t *testing.T) {
 
 	for _, tc := range testCases {
 		mockEC2.EXPECT().DescribeNetworkInterfacesWithContext(gomock.Any(), gomock.Any(), gomock.Any()).Times(tc.n).Return(result, tc.awsErr)
-		ins := &EC2InstanceMetadataCache{imds: TypedIMDS{mockMetadata}, ec2SVC: mockEC2}
-		metaData, err := ins.DescribeAllENIs()
+		cache := &EC2InstanceMetadataCache{imds: TypedIMDS{mockMetadata}, ec2SVC: mockEC2}
+		metaData, err := cache.DescribeAllENIs()
 		assert.Equal(t, tc.expErr, err, tc.name)
 		assert.Equal(t, tc.exptags, metaData.TagMap, tc.name)
 	}
@@ -380,11 +410,13 @@ func TestAllocENI(t *testing.T) {
 	mockEC2.EXPECT().AttachNetworkInterfaceWithContext(gomock.Any(), gomock.Any(), gomock.Any()).Return(attachResult, nil)
 	mockEC2.EXPECT().ModifyNetworkInterfaceAttributeWithContext(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, nil)
 
-	ins := &EC2InstanceMetadataCache{
-		ec2SVC: mockEC2,
-		imds:   TypedIMDS{mockMetadata},
+	cache := &EC2InstanceMetadataCache{
+		ec2SVC:        mockEC2,
+		imds:          TypedIMDS{mockMetadata},
+		eventRecorder: setupEventRecorder(t),
 	}
-	_, err := ins.AllocENI(false, nil, "")
+
+	_, err := cache.AllocENI(false, nil, "")
 	assert.NoError(t, err)
 }
 
@@ -413,11 +445,13 @@ func TestAllocENINoFreeDevice(t *testing.T) {
 	mockEC2.EXPECT().DescribeInstancesWithContext(gomock.Any(), gomock.Any(), gomock.Any()).Return(result, nil)
 	mockEC2.EXPECT().DeleteNetworkInterfaceWithContext(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, nil)
 
-	ins := &EC2InstanceMetadataCache{
-		ec2SVC: mockEC2,
-		imds:   TypedIMDS{mockMetadata},
+	cache := &EC2InstanceMetadataCache{
+		ec2SVC:        mockEC2,
+		imds:          TypedIMDS{mockMetadata},
+		eventRecorder: setupEventRecorder(t),
 	}
-	_, err := ins.AllocENI(false, nil, "")
+
+	_, err := cache.AllocENI(false, nil, "")
 	assert.Error(t, err)
 }
 
@@ -448,11 +482,13 @@ func TestAllocENIMaxReached(t *testing.T) {
 	mockEC2.EXPECT().AttachNetworkInterfaceWithContext(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, errors.New("AttachmentLimitExceeded"))
 	mockEC2.EXPECT().DeleteNetworkInterfaceWithContext(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, nil)
 
-	ins := &EC2InstanceMetadataCache{
-		ec2SVC: mockEC2,
-		imds:   TypedIMDS{mockMetadata},
+	cache := &EC2InstanceMetadataCache{
+		ec2SVC:        mockEC2,
+		imds:          TypedIMDS{mockMetadata},
+		eventRecorder: setupEventRecorder(t),
 	}
-	_, err := ins.AllocENI(false, nil, "")
+
+	_, err := cache.AllocENI(false, nil, "")
 	assert.Error(t, err)
 }
 
@@ -468,8 +504,12 @@ func TestFreeENI(t *testing.T) {
 	mockEC2.EXPECT().DetachNetworkInterfaceWithContext(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, nil)
 	mockEC2.EXPECT().DeleteNetworkInterfaceWithContext(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, nil)
 
-	ins := &EC2InstanceMetadataCache{ec2SVC: mockEC2}
-	err := ins.freeENI("test-eni", time.Millisecond, time.Millisecond)
+	cache := &EC2InstanceMetadataCache{
+		ec2SVC:        mockEC2,
+		eventRecorder: setupEventRecorder(t),
+	}
+
+	err := cache.freeENI("test-eni", time.Millisecond, time.Millisecond)
 	assert.NoError(t, err)
 }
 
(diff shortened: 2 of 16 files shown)


pkg/awsutils/awsutils_test.go
```diff
@@ -139,15 +169,15 @@ func TestInitWithEC2metadata(t *testing.T) {
...
-		assert.Equal(t, ins.instanceID, instanceID)
-		assert.Equal(t, ins.primaryENImac, primaryMAC)
-		assert.Equal(t, ins.primaryENI, primaryeniID)
-		assert.Equal(t, subnetID, ins.subnetID)
+		assert.Equal(t, az, cache.availabilityZone)
+		assert.Equal(t, localIP, cache.localIPv4.String())
+		assert.Equal(t, cache.instanceID, instanceID)
+		assert.Equal(t, cache.primaryENImac, primaryMAC)
+		assert.Equal(t, cache.primaryENI, primaryeniID)
+		assert.Equal(t, subnetID, cache.subnetID)
```
jdn5126, line 180: nit: should we check the value of `cache.eventRecorder` here?

pkg/awsutils/awsutils_test.go
```diff
@@ -380,11 +410,12 @@ func TestAllocENI(t *testing.T) {
 	mockEC2.EXPECT().AttachNetworkInterfaceWithContext(gomock.Any(), gomock.Any(), gomock.Any()).Return(attachResult, nil)
 	mockEC2.EXPECT().ModifyNetworkInterfaceAttributeWithContext(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, nil)
 
-	ins := &EC2InstanceMetadataCache{
+	cache := &EC2InstanceMetadataCache{
 		ec2SVC: mockEC2,
 		imds:   TypedIMDS{mockMetadata},
 	}
-	_, err := ins.AllocENI(false, nil, "")
+	cache.eventRecorder = setupEventRecorder(t)
```
jdn5126, line 417: nit: can we set `eventRecorder` in constructor above? Seems easier to read that way. Same with the ones below

pkg/ipamd/ipamd.go
```diff
@@ -374,9 +376,15 @@ func New(rawK8SClient client.Client, cachedK8SClient client.Client) (*IPAMContex
...
+	eventrecorder, err := eventrecorder.New(rawK8SClient, cachedK8SClient)
+	if err != nil {
+		return nil, errors.Wrap(err, "ipamd: can not initialize event recorder")
+	}
+	c.eventRecorder = eventrecorder
+
 	c.disableENIProvisioning = disablingENIProvisioning()
 
-	client, err := awsutils.New(c.useCustomNetworking, c.disableENIProvisioning, c.enableIPv4, c.enableIPv6)
+	client, err := awsutils.New(c.useCustomNetworking, c.disableENIProvisioning, c.enableIPv4, c.enableIPv6, eventrecorder)
```
jdn5126, line 387: nit: `c.eventRecorder`

pkg/ipamd/ipamd.go
```diff
@@ -1319,13 +1311,8 @@ func (c *IPAMContext) nodeIPPoolReconcile(ctx context.Context, interval time.Dur
...
 
 		if c.enablePodENI && metadataResult.TrunkENI != "" {
-			// Label the node that we have a trunk
-			err = c.SetNodeLabel(ctx, "vpc.amazonaws.com/has-trunk-attached", "true")
-			if err != nil {
-				podENIErrInc("askForTrunkENIIfNeeded")
-				log.Errorf("Failed to set node label for trunk. Aborting reconcile", err)
-				return
-			}
+			// Signal event to VPC RC that the node has a trunk already
```
jdn5126, line 1314: nit: we are not signaling anything to VPC RC here, right? Just logging locally as the annotation is already present

pkg/ipamd/ipamd.go
```diff
@@ -1936,7 +1880,7 @@ func (c *IPAMContext) GetPod(podName, namespace string) (*corev1.Pod, error) {
 	}
 	err := c.rawK8SClient.Get(ctx, podKey, &pod)
 	if err != nil {
-		return nil, fmt.Errorf("Error while trying to retrieve Pod Info: %s", err)
+		return nil, fmt.Errorf("Error while trying to retrieve Pod Info: %s", err.Error())
```
jdn5126, line 1883: nit: `%v", err`

pkg/ipamd/ipamd_test.go
```diff
@@ -1850,6 +1850,26 @@ func TestIPAMContext_askForTrunkENIIfNeeded(t *testing.T) {
...
+	fakeRecorder := events.NewFakeRecorder(3)
+	k8sSchema := runtime.NewScheme()
+	clientgoscheme.AddToScheme(k8sSchema)
+
+	mockEventRecorder := &eventrecorder.EventRecorder{
+		Recorder:        fakeRecorder,
+		RawK8SClient:    testclient.NewClientBuilder().WithScheme(k8sSchema).Build(),
+		CachedK8SClient: testclient.NewClientBuilder().WithScheme(k8sSchema).Build(),
+	}
+	eventrecorder.MyNodeName = myNodeName
```
jdn5126, line 1862: nit: this is strange to read, can you add a comment that this is setting global var in `eventrecorder` package

test/framework/framework.go
```diff
@@ -53,9 +60,28 @@ func New(options Options) *Framework {
 	eniConfig.AddToScheme(k8sSchema)
 	sgp.AddToScheme(k8sSchema)
 
-	k8sClient, err := client.New(config, client.Options{Scheme: k8sSchema})
+	stopChan := ctrl.SetupSignalHandler()
```
jdn5126, line 63: nit: can you add some comments here? A but hard to follow what is happening

jdn5126 (commented):
Changes overall LGTM, just some nits and I will give @jayanthvn a chance to review

pkg/ipamd/ipamd.go
```diff
@@ -1201,15 +1196,12 @@ func (c *IPAMContext) askForTrunkENIIfNeeded(ctx context.Context) {
...
 			return
 		}
 		// We need to signal that VPC Resource Controller needs to attach a trunk ENI
-		err := c.SetNodeLabel(ctx, "vpc.amazonaws.com/has-trunk-attached", "false")
-		if err != nil {
-			podENIErrInc("askForTrunkENIIfNeeded")
-			log.Errorf("Failed to set node label", err)
-		}
+		log.Debug("VPC CNI is asking RC to initialize trunk interface")
+		c.eventRecorder.SendNodeEvent(corev1.EventTypeNormal, eventrecorder.EventReason, "NeedTrunk", "vpc.amazonaws.com/has-trunk-attached=false")
```
jdn5126, line 1204: If node does not support `has-trunk-attached`, can we avoid sending this event and log error for readability?

pkg/ipamd/ipamd.go
haouc: Right, this was leftover after I removed the call to send event during updating some logic flow. Thanks for catching this.

pkg/ipamd/ipamd.go
haouc: Why should we log the entire object instead of just error string for information?

pkg/ipamd/ipamd.go
haouc: It is a bit harder for now because we haven't verified if we can validate node type for the feature during runtime. But this is planned to be done for the next release of VPC resource controller if APIs are available from EC2. You can refer to t[his open issue](https://github.com/aws/amazon-vpc-resource-controller-k8s/issues/98).
Otherwise we have to maintain a hard copy of file listing all supported types, which is exactly what we want to deprecate.

pkg/ipamd/ipamd.go
```diff
@@ -538,37 +550,20 @@ func (c *IPAMContext) nodeInit() error {
...
-	if metadataResult.TrunkENI != "" {
-		// Signal to VPC Resource Controller that the node has a trunk already
-		err := c.SetNodeLabel(ctx, "vpc.amazonaws.com/has-trunk-attached", "true")
-		if err != nil {
-			log.Errorf("Failed to set node label", err)
-			podENIErrInc("nodeInit")
-			// If this fails, we probably can't talk to the API server. Let the pod restart
-			return err
+		if metadataResult.TrunkENI != "" {
+			// Signal event to VPC RC that the node has a trunk already
```
jayanthvn, line 561: Don't we want to signal the event to VPC RC about `trunk attached = true` or since it is found in metadataResult assumption is the node is already managed in VPC RC? If it is not needed let's remove the comment.

haouc: Right, I think we don't have to send a confirmation back to VPC CNI since CNI has it' own check on trunk interface. I will remove the meaningless comment here after all comments are collect for the current revision. Thanks @jayanthvn

pkg/awsutils/awsutils.go
```diff
@@ -433,7 +436,6 @@ func New(useCustomNetworking, disableENIProvisioning, v4Enabled, v6Enabled bool)
 		return nil, err
 	}
 	// event recorder to raise events for failed EC2 API calls
```
jdn5126, line 438: nit: comment can be deleted

jdn5126 (commented):
Changes LGTM, will give Jayanth a chance to finish review

pkg/ipamd/ipamd.go
jdn5126: Good point, this is fine

pkg/eniconfig/eniconfig.go
```diff
@@ -141,10 +142,9 @@ func GetNodeSpecificENIConfigName(ctx context.Context, k8sClient client.Client)
 	}
 
 	eniConfigName = val
-	if val != eniConfigDefault {
-		labels := node.GetLabels()
-		labels["vpc.amazonaws.com/eniConfig"] = eniConfigName
-		node.SetLabels(labels)
+	if val != eniConfigDefault && generate_event {
+		// Signal event to VPC RC that the custom networking is enabled
+			eventRecorder.SendNodeEvent(corev1.EventTypeNormal, "AwsNodeNotification", "CustomNetworkingEnabled", eniConfigLabel+"="+eniConfigName)
```
achevuru: Why generate the event from here though? Why not from the IPAMD init flow where we currently check the env variable and are currently labeling the node? - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/ipamd.go#L543. We can avoid the eventRecorder references in the controller code flow here...

pkg/ipamd/ipamd.go
```diff
@@ -374,9 +376,15 @@ func New(rawK8SClient client.Client, cachedK8SClient client.Client) (*IPAMContex
 	c.enableIPv4 = isIPv4Enabled()
 	c.enableIPv6 = isIPv6Enabled()
 
+	eventRecorder, err := eventrecorder.New(rawK8SClient, cachedK8SClient)
+ 	if err != nil {
+ 		return nil, errors.Wrap(err, "ipamd: can not initialize event recorder")
```
achevuru: nit: `can not` --> `unable to`. `can not` sounds like a permission related error to me.

pkg/ipamd/ipamd.go
```diff
@@ -538,37 +547,22 @@ func (c *IPAMContext) nodeInit() error {
...
-		if err != nil {
-			log.Errorf("Failed to delete eniConfig node label", err)
-			podENIErrInc("nodeInit")
-			return err
+	if c.enablePodENI {
+		if c.useCustomNetworking {
+			eniConfigName, err := eniconfig.GetNodeSpecificENIConfigName(ctx, c.cachedK8SClient, c.eventRecorder, true)
+			if err == nil && eniConfigName != "default" {
+				// Signal event to VPC RC that the custom networking is enabled
+				c.eventRecorder.SendNodeEvent(corev1.EventTypeNormal, "AwsNodeNotification", "CustomNetworkingEnabled", vpcENIConfigLabel+"="+eniConfigName)
```
achevuru: I see that we are generating the event for Custom networking from here as well...Why do we need to do it again from `GetNodeSpecificENIConfigName`?

pkg/ipamd/ipamd.go
achevuru: But, this is done only once on bootup isn't it? Also, how will VPC RC rebuild the cache upon a restart?

pkg/ipamd/ipamd.go
```diff
@@ -834,7 +829,7 @@ func (c *IPAMContext) tryAllocateENI(ctx context.Context) error {
 	var subnet string
 
 	if c.useCustomNetworking {
-		eniCfg, err := eniconfig.MyENIConfig(ctx, c.cachedK8SClient)
+		eniCfg, err := eniconfig.MyENIConfig(ctx, c.cachedK8SClient, c.eventRecorder, c.enablePodENI)
```
achevuru, line 833: Same comment as above. Do we need to do this?

pkg/ipamd/ipamd.go
haouc: Do you mean VPC CNI check trunk? In current code, VPC CNI does reconcile this every 30s in update IP resources and rely on datastore for trunk info. Upon restart of VPC RC, it iterates through all nodes and manages all nodes which have valid labels. Their trunk interfaces are also described and cached.

achevuru: Right now, yes. But once we transition to events based approach, VPC RC would've to go through the events against the node to do the same, isn't it?

haouc: Once we change to event, RC relies on event and node label. The workflow looks like this

1. CNI emits events continuously till finding trunk interface on node 
2. RC reconcile events and label node if event is valid
3. RC provision interfaces based on node label
4. if RC restarts, RC looks for node label for building cache

If events are still alive during RC booting up, RC will 

1. if the node is in cache, RC will skip the node
2. if the node hasn't been added in cache, RC will retry to label the node which should be safe

pkg/ipamd/ipamd.go
haouc: This is used to tell VPC RC to look for specific subnet when custom networking is enabled. CNI does this work for RC before events.

pkg/eniconfig/eniconfig.go
haouc: Thanks, Apurup. I have move the code to ipamd instead.

pkg/ipamd/ipamd.go
```diff
@@ -1319,13 +1365,7 @@ func (c *IPAMContext) nodeIPPoolReconcile(ctx context.Context, interval time.Dur
...
 
 		if c.enablePodENI && metadataResult.TrunkENI != "" {
-			// Label the node that we have a trunk
-			err = c.SetNodeLabel(ctx, "vpc.amazonaws.com/has-trunk-attached", "true")
-			if err != nil {
-				podENIErrInc("askForTrunkENIIfNeeded")
-				log.Errorf("Failed to set node label for trunk. Aborting reconcile", err)
-				return
-			}
+			log.Debugf("Trunk interface (%s) has been added to the node already.", metadataResult.TrunkENI)
```
achevuru, line 1368: Are we making the assumption here that the event would've been already generated for this node, since the trunk is already present?

pkg/ipamd/ipamd.go
```diff
@@ -1947,7 +1944,7 @@ func (c *IPAMContext) AnnotatePod(podName, podNamespace, key, val string) error
 	var err error
 
 	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		pod := &corev1.Pod{}
+		var pod *corev1.Pod
```
achevuru, line 1947: Why the change here?

pkg/sgpp/constants.go
```diff
@@ -3,8 +3,12 @@ package sgpp
 type EnforcingMode string
 
 const (
-	EnforcingModeStrict   EnforcingMode = "strict"
-	EnforcingModeStandard EnforcingMode = "standard"
+	EnforcingModeStrict               EnforcingMode = "strict"
+	EnforcingModeStandard             EnforcingMode = "standard"
+	VpcCNINodeEventActionForTrunk     string        = "NeedTrunk"
+	TrunkEventNote                    string        = "vpc.amazonaws.com/has-trunk-attached=false"
```
achevuru, line 9: Are we not generating event for `vpc.amazonaws.com/has-trunk-attached=true`? I believe we were trying to address that for nodes that come up with a trunk interface attached.

pkg/utils/eventrecorder/eventrecorder.go
```diff
@@ -16,84 +16,107 @@ package eventrecorder
...
-		e.recorder.Event(&pod, eventType, reason, message)
+		e.Recorder.Eventf(&pod, nil, eventType, reason, "", message)
 	}
 }
+
+func (e *EventRecorder) findMyNode(ctx context.Context) (corev1.Node, error) {
+	var node corev1.Node
+	// Find my node
+	err := e.CachedK8SClient.Get(ctx, types.NamespacedName{Name: MyNodeName}, &node)
+	log.Debugf("Node found %s - labels - %d", node.Name, len(node.Labels))
```
achevuru, line 98: We're not checking for error here before we print this debug message. Will be misleading.

pkg/ipamd/ipamd.go
```diff
@@ -538,36 +553,12 @@ func (c *IPAMContext) nodeInit() error {
...
-		err := c.SetNodeLabel(ctx, "vpc.amazonaws.com/has-trunk-attached", "true")
-		if err != nil {
-			log.Errorf("Failed to set node label", err)
-			podENIErrInc("nodeInit")
-			// If this fails, we probably can't talk to the API server. Let the pod restart
-			return err
-		}
-	} else {
-		// Check if we want to ask for one
+		// we want to ask VPC RC once during node init to cover special cases
```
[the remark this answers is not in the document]
haouc: The error comes from sending node events. If error happens, we don't have to retry since the CNI will resend another event in 30 s. We will just let it fail through.

pkg/ipamd/ipamd.go
```diff
@@ -655,7 +646,15 @@ func (c *IPAMContext) StartNodeIPPoolManager() {
 }
 
 func (c *IPAMContext) updateIPPoolIfRequired(ctx context.Context) {
-	c.askForTrunkENIIfNeeded(ctx)
+	if c.enablePodENI {
+		c.askForTrunkENIIfNeeded(ctx)
```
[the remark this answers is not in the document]
haouc: For the same reason, we let CNI tries to send another event in 30s instead of retry or handle error at this place since the error should come from `SendNodeEvent()`.

pkg/ipamd/ipamd.go
```diff
@@ -1197,20 +1197,66 @@ func (c *IPAMContext) logPoolStats(dataStoreStats *datastore.DataStoreStats) {
...
+	}
+
+	return err
+}
+
+func (c *IPAMContext) getNodeLabels(ctx context.Context) (map[string]string, error) {
+	var node corev1.Node
+	err := c.cachedK8SClient.Get(ctx, types.NamespacedName{Name: c.myNodeName}, &node)
+
+	return node.GetLabels(), err
```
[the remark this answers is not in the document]
haouc: Yes, this is a problem. We should handle the failure on node GET.

pkg/ipamd/ipamd.go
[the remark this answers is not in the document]
achevuru: So, we're banking on the reconciler triggering `updateIPPoolIfRequired` every 30s?

pkg/ipamd/ipamd.go
haouc: Yes, since this is in reconciling, we won't send event if the trunk has been attached to the node.

pkg/ipamd/ipamd.go
haouc: This was suggested to change by go-staticcheck as `this value of pod is never used`.

haouc: I can add a nil check for the pointer if that is the concern.

pkg/sgpp/constants.go
haouc: We do check `c.dataStore.GetTrunkENI() == "" || !hasTrunkLabel` to send an event if a node has no trunk or has no label. We don't check label value and only check label key that means we check for `vpc.amazonaws.com/has-trunk-attached` not the boolean value.

```
if labels != nil {
		_, hasTrunkLabel = labels[trunkInterfaceLabel]
	}
```

pkg/ipamd/ipamd.go
achevuru: Interesting. We're using it here though - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/ipamd.go#L1955 .

haouc: I would think it was because this `pod, err = c.GetPod(podName, podNamespace)`

pkg/ipamd/ipamd.go
haouc: We do check `c.dataStore.GetTrunkENI() == "" || !hasTrunkLabel` to send an event if a node has no trunk or has no label.

pkg/ipamd/ipamd.go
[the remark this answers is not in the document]
haouc: yes, Apurup. We can discuss on this offline

pkg/utils/eventrecorder/eventrecorder.go
haouc: Makes sense. I will add an error handle here.

pkg/ipamd/ipamd.go
```diff
@@ -655,7 +658,21 @@ func (c *IPAMContext) StartNodeIPPoolManager() {
 }
 
 func (c *IPAMContext) updateIPPoolIfRequired(ctx context.Context) {
-	c.askForTrunkENIIfNeeded(ctx)
+	if c.enablePodENI {
+		if labels, err := c.getNodeLabels(ctx); err == nil {
```
haouc, line 662: `getNodeLabels()` will be called every 30 seconds to ensure the correct node labels are added to nodes. This shouldn't be expensive considering it calls to cached client. We call this method only once for two labeling checks (trunk and eniConfig). One note to this change can be that we should revisit this change if we are seeing any performance impact to API server in large clusters due to this new call from every aws-node replica.

pkg/ipamd/ipamd.go
```diff
@@ -538,37 +553,25 @@ func (c *IPAMContext) nodeInit() error {
...
-		err := c.SetNodeLabel(ctx, vpcENIConfigLabel, "")
-		if err != nil {
-			log.Errorf("Failed to delete eniConfig node label", err)
-			podENIErrInc("nodeInit")
-			return err
+
+		// if node labels is nil in rare cases, we will log the info instead of failing VPC CNI startup
+		if labels == nil {
+			// if labels is nil, ipamd logs the warning for debugging
+			log.Warn("node labels are nil during ipamd starts up, may need investigation before using the feature of security group for pods")
```
jayanthvn, line 566: nit: "ipamd starts up" ->"ipamd init"

pkg/ipamd/ipamd.go
```diff
@@ -538,37 +553,25 @@ func (c *IPAMContext) nodeInit() error {
...
-		// Signal to VPC Resource Controller that the node has a trunk already
-		err := c.SetNodeLabel(ctx, "vpc.amazonaws.com/has-trunk-attached", "true")
-		if err != nil {
-			log.Errorf("Failed to set node label", err)
-			podENIErrInc("nodeInit")
-			// If this fails, we probably can't talk to the API server. Let the pod restart
-			return err
+		if c.useCustomNetworking {
+			// not doing retry here, don't have to handle error since error has been logged in the called method
+			c.askForENIConfigIfNeeded(ctx, labels)
```
jayanthvn, line 571: Why do we need to proceed if labels are `nil`. Maybe we can just skip the rest of the steps?

pkg/ipamd/ipamd.go
```diff
@@ -655,7 +658,21 @@ func (c *IPAMContext) StartNodeIPPoolManager() {
 }
 
 func (c *IPAMContext) updateIPPoolIfRequired(ctx context.Context) {
-	c.askForTrunkENIIfNeeded(ctx)
+	if c.enablePodENI {
+		if labels, err := c.getNodeLabels(ctx); err == nil {
+			if labels == nil {
+				// if labels is nil, ipamd logs the warning for debugging
+				log.Warn("labels are nil during ipamd reconciling resources, may need attention and investigation before using the feature of security group for pods")
```
jayanthvn, line 665: Same as during nodeinit, why do we need to continue with `askForTrunkENIIfNeeded` and/or `askForENIConfigIfNeeded`

pkg/ipamd/ipamd.go
```diff
@@ -1197,20 +1215,74 @@ func (c *IPAMContext) logPoolStats(dataStoreStats *datastore.DataStoreStats) {
 	log.Debugf("%s: %s, c.maxIPsPerENI = %d", prefix, dataStoreStats, c.maxIPsPerENI)
 }
 
-func (c *IPAMContext) askForTrunkENIIfNeeded(ctx context.Context) {
-	if c.enablePodENI && c.dataStore.GetTrunkENI() == "" {
+func (c *IPAMContext) askForTrunkENIIfNeeded(ctx context.Context, labels map[string]string) error {
+	var err error
+	hasTrunkLabel := false
```
jayanthvn, line 1220: I see here if labels are `nil`, hasTrunkLabel is false so we end up requesting a trunk ENI, how is it handled in VPC RC if trunk is attached and for whatever reason node label is missing? Do we just retry adding the label?

pkg/ipamd/ipamd.go
```diff
@@ -1197,20 +1215,74 @@ func (c *IPAMContext) logPoolStats(dataStoreStats *datastore.DataStoreStats) {
...
+		log.Debug("VPC CNI is asking RC to initialize trunk interface")
+		err = c.eventRecorder.SendNodeEvent(corev1.EventTypeNormal, eventrecorder.EventReason, sgpp.VpcCNINodeEventActionForTrunk, sgpp.TrunkEventNote)
 		if err != nil {
-			podENIErrInc("askForTrunkENIIfNeeded")
-			log.Errorf("Failed to set node label", err)
+			log.Error("Failed sending a node event to VPC RC for enabling trunk interface")
 		}
 	}
+
+	return err
```
jayanthvn, line 1244: Btw, err is not initialized.

pkg/ipamd/ipamd.go
```diff
@@ -1197,20 +1215,74 @@ func (c *IPAMContext) logPoolStats(dataStoreStats *datastore.DataStoreStats) {
...
+			// Signal event to VPC RC that the custom networking is enabled
+			err = c.eventRecorder.SendNodeEvent(corev1.EventTypeNormal, eventrecorder.EventReason, sgpp.VpcCNINodeEventActionForEniConfig, vpcENIConfigLabel+"="+eniConfigName)
+			if err == nil {
+				log.Infof("Send an event to notify RC custom networking is enabled. Config name should be labelled on node as %s", eniConfigName)
+			} else {
+				log.Errorf("Failed sending an event to notify RC custom networking is enabled. Error is %s", err.Error())
+			}
+		}
+	}
+	return err
```
jayanthvn, line 1285: Same here, initialize `err`.

pkg/ipamd/ipamd.go
```diff
@@ -1197,20 +1215,74 @@ func (c *IPAMContext) logPoolStats(dataStoreStats *datastore.DataStoreStats) {
...
+	} else {
+		return node.GetLabels(), err
+	}
+}
+
+func (c *IPAMContext) askForENIConfigIfNeeded(ctx context.Context, labels map[string]string) error {
+	var err error
+	hasLabelled := false
+
+	if err != nil {
```
jayanthvn, line 1260: Why is this here? Missed during merge or cleanup?

pkg/ipamd/ipamd.go
```diff
@@ -538,37 +553,25 @@ func (c *IPAMContext) nodeInit() error {
...
-			return err
+		if c.useCustomNetworking {
+			// not doing retry here, don't have to handle error since error has been logged in the called method
+			c.askForENIConfigIfNeeded(ctx, labels)
 		}
-	} else {
-		// Check if we want to ask for one
-		c.askForTrunkENIIfNeeded(ctx)
+		// we want to ask VPC RC once during node init to cover special cases
+		c.askForTrunkENIIfNeeded(ctx, labels)
```
jayanthvn, line 574: Do we skip adding events for unsupported instances? I didn't see a check or I might have missed during code walk thru.

haouc: We won't. VPC resource controller will check if the instance is supported. Since the instance type is checked from a static file in resource controller and so far we don't have a API to check during runtime, VPC CNI can't check and send event.

pkg/ipamd/ipamd.go
haouc: oh, this should have been removed after I combined the getNodeLabels into caller instead of making two calls in trunk and eniconfig methods. I will remove this. Thanks.

pkg/ipamd/ipamd.go
haouc: I think we don't have to initialize an error variable and error is nil when we declare it as an error.

pkg/ipamd/ipamd.go
haouc: I think we don't have to init error and error is nil when we declare it as an error?

pkg/ipamd/ipamd.go
haouc: Yes, I am adding a check in resource controller to handle this. I don't think we handled the nil map in labels in current codes. I am thinking to either retry in resource controller in case API server error, or add a map to pod as labels and add the label.

pkg/ipamd/ipamd.go
haouc: makes sense, will update.

pkg/ipamd/ipamd.go
jayanthvn: Yes defaulted to `nil`

jayanthvn (commented):
Also once we merge this, we need to add few more integration tests to name a few.. CNI's behavior when label is not present, restart scenario, node not supported, vpc rc failing to add label and so on.

haouc:
> 

Makes sense. To adding the new tests will have to wait for the control plane side update to enable the new resource controller. Thanks.

pkg/ipamd/ipamd.go
haouc: Yes. I will update README and also check if that can be feasible to use a dummy interface on unsupported instance types.

haouc:
I added not-supported label value check to help CNI stopping sending events from unsupported nodes. Please check the VPC resource controller PR as reference 
https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/164

pkg/ipamd/ipamd.go
```diff
@@ -134,6 +134,8 @@ const (
 
 	// trunkInterfaceLabel is used by CNI to check if correct label was added by the VPC resource controller
 	trunkInterfaceLabel = "vpc.amazonaws.com/has-trunk-attached"
+	// VPC resource controller label nodes as not-supported if the EC2 instance is not supported yet
```
achevuru, line 137: nit: `if the EC2 instance is not supported yet` --> `if the EC2 instance doesn't support Branch ENI feature`

haouc: Makes sense. Updated the comment.

achevuru (approved):

jayanthvn (approved):
lgtm

PR #2153, opened 2022-12-07
Deprecate AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER and remove no-op setter

jdn5126:
**What type of PR is this?**
Cleanup

None

This PR removes an unnecessary rp_filter setter in `pkg/networkutils/network.go` and deprecates the `AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER` environment variable. During IPAMD node init, we were setting `rp_filter` on primary interface to "loose" mode in certain scenarios. But `rp_filter` is already set to "loose" mode on primary interface unconditionally by init container.

N/A

Manually verified that canary tests pass and that `rp_filter` is set before this code even executes.

**Automation added to e2e**:
N/A

No

Will not affect upgrade or downgrades. A running cluster has been tested.

No

No

```release-note
Deprecating AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER environment variable.
```


jdn5126:
The unconditional setting of `rp_filter` to "loose" mode in init container happens here btw: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/cmd/aws-vpc-cni-init/main.go#L73

jayanthvn:
Toggling `AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER` is tested?

This is called out in docs -

```
Specifies whether ipamd should configure rp filter for primary interface. Setting this to false will require rp filter to be configured through init container
```

jdn5126:
> 

Yeah, so that environment variable becomes a no-op with this change. I was going to remove it, but wanted to wait first. I should probably just remove it as part of this change.

Previously, VPC CNI could be run without the init container, but today, we require the init container to do required setup. So since the init container is required and unconditionally sets `rp_filter` to `2`, this logic is not needed.

pkg/networkutils/network.go
```diff
@@ -273,39 +270,7 @@ func (n *linuxNetwork) SetupHostNetwork(vpcv4CIDRs []string, primaryMAC string,
...
-		//   the RPF check, it looks up the route using the pod IP as the source.
-		// - Thus, it finds the source-based route that leaves via the secondary ENI.
-		// - In "strict" mode, the RPF check fails because the return path uses a different interface to the incoming
-		//   packet. In "loose" mode, the check passes because some route was found.
-		//
-		// The same applies when performing SNAT on Pod traffic.
-		primaryIntfRPFilter := "net/ipv4/conf/" + primaryIntf + "/rp_filter"
-		const rpFilterLoose = "2"
-
-		if n.shouldConfigureRpFilter {
```
kishorj, line 298: Also cleanup shoudConfigureRpFilter attribute, the underlying env variable and documentation.

README.md
```diff
@@ -390,13 +390,15 @@ string, tag addition will be ignored.
 
 ---
 
-#### `AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER`
```
jaydeokar, line 393: Would this be 1.12.1 ?

jdn5126: good call, fixed

Issue #2152, opened 2022-12-06
Fix weekly tests


jayanthvn:
**What happened**:

Weekly tests are failing due to infra issues.

**Attach logs**

**What you expected to happen**:
Successful runs

**How to reproduce it (as minimally and precisely as possible)**:
N/A

**Anything else we need to know?**:
N/A

**Environment**:
- Kubernetes version (use `kubectl version`):
- CNI Version
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):

PR #2154, opened 2022-12-07
Fix Weekly Cron Tests

jdn5126:
**What type of PR is this?**
test cleanup

#2152 

This PR fixes the weekly cron tests. Three issues were addressed:
1. Assumption of `OSS_TEST_ROLE_ARN` was increased from 4 hours to 8 hours. Tests late in the pipeline would fail whenever > 4 hours had already passed.
2. CANARY integration tests against default CNI version now only run against `integration-tests` and `nightly-cron-tests` workflow. They no longer run against `weekly-cron-tests` or `pr-manual-tests` workflow.
3. The permissions for `OSS_TEST_ROLE_ARN` were updated to prevent s3 bucket creation failure in kops test. This was done external to this PR.

Also, I updated the integration test script to reuse `check_ds_rollout`.

N/A

Coverage will be attained by running weekly cron job action against this branch.

**Automation added to e2e**:
None

No

N/A

No

No

```release-note

```

### DIFF
--- a/scripts/run-integration-tests.sh
+++ b/scripts/run-integration-tests.sh
@@ -11,6 +11,7 @@ source "$DIR"/lib/common.sh
 source "$DIR"/lib/aws.sh
 source "$DIR"/lib/cluster.sh
 source "$DIR"/lib/integration.sh
+source "$DIR"/lib/k8s.sh
 source "$DIR"/lib/performance_tests.sh
 
 # Variables used in /lib/aws.sh
@@ -26,6 +27,7 @@ ARCH=$(go env GOARCH)
 : "${RUN_CONFORMANCE:=false}"
 : "${RUN_TESTER_LB_ADDONS:=false}"
 : "${RUN_KOPS_TEST:=false}"
+: "${RUN_INTEGRATION_DEFAULT_CNI:=true}"
 : "${RUN_BOTTLEROCKET_TEST:=false}"
 : "${RUN_PERFORMANCE_TESTS:=false}"
 : "${RUNNING_PERFORMANCE:=false}"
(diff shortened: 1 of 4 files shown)


scripts/run-integration-tests.sh
```diff
@@ -210,50 +210,17 @@ grep -r -q $TEST_IMAGE_VERSION $TEST_CONFIG_PATH
...
     export KUBECONFIG=~/.kube/config
-fi
-
-if [[ $RUN_KOPS_TEST == true ]]; then
     run_kops_conformance
 fi
-ADDONS_CNI_IMAGE=$($KUBECTL_PATH describe daemonset aws-node -n kube-system | grep Image | cut -d ":" -f 2-3 | tr -d '[:space:]')
-
-echo "*******************************************************************************"
-echo "Running integration tests on default CNI version, $ADDONS_CNI_IMAGE"
```
sushrk, line 224: The nightly & integration tests are run using the same script, so removing it here will exclude the tests against default CNI in all cases. 
We should skip this conditionally, ie, for weekly tests only.

scripts/run-integration-tests.sh
```diff
@@ -210,50 +210,17 @@ grep -r -q $TEST_IMAGE_VERSION $TEST_CONFIG_PATH
...
-    UPDATED_PODS_COUNT=$($KUBECTL_PATH get pods -A --field-selector=status.phase=Running -l k8s-app=aws-node --no-headers=true | wc -l)
-    let RETRY_ATTEMPT=RETRY_ATTEMPT+1
-    sleep 5
-    echo "Waiting for cni daemonset update. $UPDATED_PODS_COUNT are in ready state in retry attempt $RETRY_ATTEMPT"
-done
-echo "Updated!"
-
-CNI_IMAGE_UPDATE_DURATION=$((SECONDS - START))
-echo "TIMELINE: Updating CNI image took $CNI_IMAGE_UPDATE_DURATION seconds."
+check_ds_rollout "aws-node" "kube-system" "4m"
```
sushrk, line 223: Nice!

scripts/run-integration-tests.sh
```diff
@@ -26,6 +27,7 @@ ARCH=$(go env GOARCH)
 : "${RUN_CONFORMANCE:=false}"
 : "${RUN_TESTER_LB_ADDONS:=false}"
 : "${RUN_KOPS_TEST:=false}"
+: "${RUN_DEFAULT_CANARY:=false}"
```
sushrk, line 30: Can we set this to `true` by default and only skip for the weekly tests? 
Also, might want to rename to like `RUN_INTEGRATION_DEFAULT_CNI` as the plan is to run all integration tests in the future, not just canary tests.

scripts/run-integration-tests.sh
```diff
@@ -210,50 +212,35 @@ grep -r -q $TEST_IMAGE_VERSION $TEST_CONFIG_PATH
...
-    UPDATED_PODS_COUNT=$($KUBECTL_PATH get pods -A --field-selector=status.phase=Running -l k8s-app=aws-node --no-headers=true | wc -l)
-    let RETRY_ATTEMPT=RETRY_ATTEMPT+1
-    sleep 5
-    echo "Waiting for cni daemonset update. $UPDATED_PODS_COUNT are in ready state in retry attempt $RETRY_ATTEMPT"
-done
-echo "Updated!"
-
-CNI_IMAGE_UPDATE_DURATION=$((SECONDS - START))
-echo "TIMELINE: Updating CNI image took $CNI_IMAGE_UPDATE_DURATION seconds."
+check_ds_rollout "aws-node" "kube-system" "4m"
```
sushrk, line 243: Wondering if 4min timeout is sufficient as there could be upto 6 `aws-node` pods in the test cluster. 
Let's trigger an integration test before merging the PR.

sushrk (changes requested):

scripts/run-integration-tests.sh
jdn5126: Done. I noticed that for the weekly job, since we call the script 4 times, we will run integration tests against latest CR four times. We can definitely improve this in the future.

scripts/run-integration-tests.sh
jdn5126: Good call, updated to 10m, which I think should be sufficient for now

sushrk:
Triggered test against PR https://github.com/aws/amazon-vpc-cni-k8s/actions/runs/3651859414/jobs/6169572552

jdn5126:
Manual integration tests against PR passed at https://github.com/aws/amazon-vpc-cni-k8s/actions/runs/3652979006

jayanthvn (approved):
/lgtm

sushrk (approved):

PR #2176, opened 2022-12-22
Integration test fixes

jdn5126:
**What type of PR is this?**
Cleanup

N/A

This PR fixes issues in existing integration tests. The fixes are:
1. Custom networking tests - use `100.64.0.0/16` as CIDR to attach to existing VPC CIDR
2. SNAT tests - instead of using self-managed node group template from `amazon-eks-ami` repo, vendor template in `testdata/`
3. CNI Metrics Helper - pass `AWS_CLUSTER_ID` value to helm chart. Skip test case for validating EC2 API request increase. This test case was flaky (not all instance types will make EC2 call to allocate new ENIs), and making it non-flaky is non-trivial. The purpose of this test is to validate that metrics can be published and incremented, so just checking `addReqCnt` is sufficient.

N/A

Validated that all modified tests pass.

**Automation added to e2e**:
N/A

No

N/A

No

No

```release-note

```

### DIFF
--- a/scripts/run-cni-release-tests.sh
+++ b/scripts/run-cni-release-tests.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# script to run integration and calico tests tests(no cluster creation & deletion/installing addons)
+# script to run integration tests (no cluster creation & deletion/installing addons)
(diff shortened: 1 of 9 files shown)


jaydeokar (approved):
LGTM

PR #2182, opened 2022-12-23
Return AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER to chart and manifest

jdn5126:
**What type of PR is this?**
bug

N/A

In #2153 , `AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER` was deprecated. The current helm chart and master manifests use VPC CNI version v1.12.0, which requires `AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER` to be set to false in order for `aws-node` pod to start.

This PR restores `AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER` in helm chart and manifests. The environment variable will be removed after v1.12.1 releases.

This PR also increases the rollout waiting period in `scripts/update-cni-images.sh` from 4 minutes to 10 minutes. 4 minutes is not enough time for larger clusters.

N/A

Manually verified that `config/master/aws-k8s-cni.yaml` can be applied successfully after adding env var.

**Automation added to e2e**:
N/A

No

No, this fixes an upgrade/downgrade issue using the manifest. A running cluster has been tested.

Restoring env var in daemonset.

No

```release-note

```

### DIFF
--- a/charts/aws-vpc-cni/Chart.yaml
+++ b/charts/aws-vpc-cni/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: aws-vpc-cni
-version: 1.2.2
+version: 1.2.1
 appVersion: "v1.12.0"
 description: A Helm chart for the AWS VPC CNI
 icon: https://raw.githubusercontent.com/aws/eks-charts/master/docs/logo/aws.png
--- a/scripts/update-cni-images.sh
+++ b/scripts/update-cni-images.sh
@@ -30,4 +30,4 @@ fi
 echo "Applying amazon-vpc-cni-k8s/config/master/aws-k8s-cni.yaml manifest"
 kubectl apply -f $AWS_K8S_CNI_MANIFEST
 
-check_ds_rollout "aws-node" "kube-system" "4m"
+check_ds_rollout "aws-node" "kube-system" "10m"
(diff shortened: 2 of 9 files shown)


charts/aws-vpc-cni/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: aws-vpc-cni
-version: 1.2.2
+version: 1.2.1
```
jayanthvn, line 3: Is this just to revert back the charts version?

scripts/update-cni-images.sh
```diff
@@ -30,4 +30,4 @@ fi
 echo "Applying amazon-vpc-cni-k8s/config/master/aws-k8s-cni.yaml manifest"
 kubectl apply -f $AWS_K8S_CNI_MANIFEST
 
-check_ds_rollout "aws-node" "kube-system" "4m"
+check_ds_rollout "aws-node" "kube-system" "10m"
```
jayanthvn, line 33: 10m seems to be a lot. Instead we should consider setting this value based on the number of nodes or aws node pod instances in the cluster. For now it is fine but we should fine tune it in future.

jayanthvn (approved):
Minor comments but Lgtm.

charts/aws-vpc-cni/Chart.yaml
jdn5126: Yeah, this is a revert to keep in sync with what was shipped with the current `appVersion: "v1.12.0"`. No chart was ever released with version `1.2.2`, so I figured it was best to revert back to the previous state.

scripts/update-cni-images.sh
jdn5126: This doesn't wait 10 minutes by default, it waits up to 10 minutes before calling the operation a failure. This just ends up calling `kubectl rollout status ds/aws-node -n kube-system --timeout=10m`. So it is no slower than what is already there

PR #2183, opened 2022-12-27
Fix crypto package vulnerability

jaydeokar:
**What type of PR is this?**
Bug 

N/A

Bumping up the version to address crypto package vulnerability. 

N/A

**Automation added to e2e**:
N/A

N/A

N/A

N/A


jayanthvn (dismissed):

go.mod
```diff
@@ -83,13 +83,11 @@ require (
 	github.com/go-openapi/jsonreference v0.19.3 // indirect
 	github.com/go-openapi/spec v0.19.3 // indirect
 	github.com/go-openapi/swag v0.19.5 // indirect
-	github.com/go-task/slim-sprig v0.0.0-20210107165309-348f09dbbbc0 // indirect
```
jdn5126, line 86: This needs to stay

go.mod
```diff
@@ -83,13 +83,11 @@ require (
...
 	github.com/gobwas/glob v0.2.3 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
 	github.com/golang/groupcache v0.0.0-20200121045136-8c9f03a8e57e // indirect
 	github.com/google/btree v1.0.0 // indirect
 	github.com/google/gofuzz v1.1.0 // indirect
-	github.com/google/pprof v0.0.0-20210407192527-94a9f03dee38 // indirect
```
jdn5126, line 92: I believe this needs to stay as well. Good to check last PR for the link to that integration run

jaydeokar: Ah good catch. I kept it in go.sum file but not in this. Fixed

jayanthvn (approved):

PR #2186, opened 2022-12-28
Enhancements to run integration suite on kOps cluster

jaydeokar:
**What type of PR is this?**
feature 

N/A

Run integration suites on kOps cluster so that it can used to test against upstream k8s versions

N/A

Yes, ran integration suites on cluster created using kOps and eksctl 

**Automation added to e2e**:
 N/A

N/A

N/A

N/A

N/A

### DIFF
--- a/scripts/lib/cluster.sh
+++ b/scripts/lib/cluster.sh
@@ -56,8 +56,10 @@ function up-test-cluster() {
 function up-kops-cluster {
     KOPS_S3_BUCKET=kops-cni-test-eks-$AWS_ACCOUNT_ID
     echo "Using $KOPS_S3_BUCKET as kops state store"
-    aws s3api create-bucket --bucket $KOPS_S3_BUCKET --region $AWS_DEFAULT_REGION --create-bucket-configuration LocationConstraint=$AWS_DEFAULT_REGION
-    kops_version="v1.22.3"
+    if ! aws s3api head-bucket --bucket $KOPS_S3_BUCKET 2>/dev/null; then
+        aws s3api create-bucket --bucket $KOPS_S3_BUCKET --region $AWS_DEFAULT_REGION --create-bucket-configuration LocationConstraint=$AWS_DEFAULT_REGION
+    fi
+    kops_version="v1.25.3"
     echo "Using kops version $kops_version"
     curl -LO https://github.com/kubernetes/kops/releases/download/$kops_version/kops-linux-amd64
     chmod +x kops-linux-amd64
@@ -77,13 +79,16 @@ function up-kops-cluster {
     fi
 
     $KOPS_BIN create cluster \
+    --cloud aws \
     --zones ${AWS_DEFAULT_REGION}a,${AWS_DEFAULT_REGION}b \
-    --networking amazon-vpc-routed-eni \
-    --container-runtime docker \
+    --networking amazonvpc \
+    --container-runtime containerd \
     --node-count 2 \
+    --node-size c5.xlarge \
     --ssh-public-key=~/.ssh/devopsinuse.pub \
     --kubernetes-version ${K8S_VERSION} \
     ${CLUSTER_NAME}
+
     $KOPS_BIN update cluster --name ${CLUSTER_NAME} --yes
     sleep 100
     $KOPS_BIN export kubeconfig --admin
@@ -96,6 +101,13 @@ function up-kops-cluster {
         let RETRY_ATTEMPT=RETRY_ATTEMPT+1
         echo "In attempt# $RETRY_ATTEMPT, waiting for cluster validation"
     done
+
+    # kOps cluster by default comes with ebs-csi-node and coredns-autoscaler as an add-on which cannot be excluded on cluster creation
+    # Since CNI tests don't really need any of these components for running tests, we delete these to avoid any flakiness
+
+    kubectl delete daemonset ebs-csi-node -n kube-system
+    kubectl delete deployment coredns-autoscaler -n kube-system
+
     kubectl apply -f https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/${MANIFEST_CNI_VERSION}/config/master/cni-metrics-helper.yaml
 }
 
--- a/scripts/lib/integration.sh
+++ b/scripts/lib/integration.sh
@@ -24,12 +24,7 @@ function run_kops_conformance() {
     echo "TIMELINE: KOPS tests took $KOPS_TEST_DURATION seconds."
 
     sleep 240 #Workaround to avoid ENI leakage during cluster deletion: https://github.com/aws/amazon-vpc-cni-k8s/issues/1223
-    START=$SECONDS
-    down-kops-cluster
-    emit_cloudwatch_metric "kops_test_status" "1"
-    DOWN_KOPS_DURATION=$((SECONDS - START))
-    echo "TIMELINE: Down KOPS cluster took $DOWN_KOPS_DURATION seconds."
-    exit 0
+
 }
 
 function run_calico_test() {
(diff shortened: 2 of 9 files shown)


jdn5126 (commented):
Changes overall LGTM, just a few nits

test/framework/resources/k8s/resources/node.go
```diff
@@ -44,10 +44,21 @@ func (d *defaultNodeManager) GetNodes(nodeLabelKey string, nodeLabelVal string)
...
 	nodeList := v1.NodeList{}
+
 	err := d.k8sClient.List(ctx, &nodeList, client.MatchingLabels{
 		nodeLabelKey: nodeLabelVal,
 	})
-	return nodeList, err
+
+	var workerNodeList v1.NodeList
+
+	// Filter control plane nodes from the list of nodes
```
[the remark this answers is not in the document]
jaydeokar: Yes, it creates master instances in the same subnet and it shows the control plane nodes when you run `kubectl get nodes`. And sure, I'll add few lines on it

test/integration/ipamd/ipamd_suite_test.go
```diff
@@ -67,20 +69,31 @@ var _ = BeforeSuite(func() {
...
 	// Add nodeSelector label to coredns deployment so coredns pods are scheduled on 'primary' node
 	coreDNSDeployment.Spec.Template.Spec.NodeSelector = map[string]string{
-		"kubernetes.io/hostname": *primaryInstance.PrivateDnsName,
+		"kubernetes.io/hostname": primaryNode.Name,
 	}
 	err = f.K8sResourceManagers.DeploymentManager().UpdateAndWaitTillDeploymentIsReady(coreDNSDeployment,
 		utils.DefaultDeploymentReadyTimeout)
 	Expect(err).ToNot(HaveOccurred())
 
+	coreDNSAutoscalerDeployment, err := f.K8sResourceManagers.DeploymentManager().GetDeployment(CoreDNSAutoscalerDeploymentName,
```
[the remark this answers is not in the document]
jaydeokar: I think we can remove `core-dns-autoscaler` and `ebs-csi-node` daemonset in the automation scripts instead of defining it in our ginkgo tests, since they are not being used anyway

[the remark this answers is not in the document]
jdn5126: Good call, thanks

test/framework/resources/k8s/resources/node.go
```diff
@@ -44,10 +44,25 @@ func (d *defaultNodeManager) GetNodes(nodeLabelKey string, nodeLabelVal string)
...
 	err := d.k8sClient.List(ctx, &nodeList, client.MatchingLabels{
 		nodeLabelKey: nodeLabelVal,
 	})
-	return nodeList, err
+
+	/*
+		Filtering control plane nodes from the list of nodes. kOps creates control plane nodes in the
+		same subnet as worker nodes. Control plane nodes have the label `node-role.kubernetes.io/control-plane`
+		defined, which can be used to filter out the control plane nodes
+	*/
```
jdn5126, line 56: big nit: but go recommends multi-line comments still just use `// ...` for each line

jdn5126 (dismissed):
Changes LGTM

scripts/lib/cluster.sh
```diff
@@ -56,8 +56,10 @@ function up-test-cluster() {
 function up-kops-cluster {
     KOPS_S3_BUCKET=kops-cni-test-eks-$AWS_ACCOUNT_ID
     echo "Using $KOPS_S3_BUCKET as kops state store"
-    aws s3api create-bucket --bucket $KOPS_S3_BUCKET --region $AWS_DEFAULT_REGION --create-bucket-configuration LocationConstraint=$AWS_DEFAULT_REGION
-    kops_version="v1.22.3"
+    if ! aws s3api head-bucket --bucket $KOPS_S3_BUCKET; then
+        aws s3api create-bucket --bucket $KOPS_S3_BUCKET --region $AWS_DEFAULT_REGION --create-bucket-configuration LocationConstraint=$AWS_DEFAULT_REGION
+    fi
+    kops_version="v1.25.3"
```
jdn5126, line 62: This looks like it is ignoring the environment variable passed by the GitHub workflow above. Is that intentional?

scripts/lib/cluster.sh
```diff
@@ -96,6 +101,10 @@ function up-kops-cluster {
         let RETRY_ATTEMPT=RETRY_ATTEMPT+1
         echo "In attempt# $RETRY_ATTEMPT, waiting for cluster validation"
     done
+
+    kubectl delete daemonset ebs-csi-node -n kube-system || true
+    kubectl delete deployment coredns-autoscaler -n kube-system || true
```
jdn5126, line 106: nit: no need to "or" with true. If you want to log if this fails, then you should "or" with a log statement

scripts/lib/integration.sh
```diff
@@ -24,12 +24,7 @@ function run_kops_conformance() {
     echo "TIMELINE: KOPS tests took $KOPS_TEST_DURATION seconds."
 
     sleep 240 #Workaround to avoid ENI leakage during cluster deletion: https://github.com/aws/amazon-vpc-cni-k8s/issues/1223
-    START=$SECONDS
-    down-kops-cluster
```
jdn5126, line 28: Is there a reason we no longer tear down the cluster here? Does that happen elsewhere?

test/framework/resources/k8s/resources/node.go
```diff
@@ -44,10 +44,23 @@ func (d *defaultNodeManager) GetNodes(nodeLabelKey string, nodeLabelVal string)
...
+
 	err := d.k8sClient.List(ctx, &nodeList, client.MatchingLabels{
 		nodeLabelKey: nodeLabelVal,
 	})
-	return nodeList, err
+
+	//Filtering control plane nodes from the list of nodes. kOps creates control plane nodes in the
+	//same subnet as worker nodes. Control plane nodes have the label `node-role.kubernetes.io/control-plane`
+	//defined, which can be used to filter out the control plane nodes
+
```
jdn5126, line 55: bit nit: we need space between `//` and the text for this to pass stricter `go fmt` checks. Also need to remove empty newline before variable definitions

scripts/lib/integration.sh
jaydeokar: Earlier we only used to run the upstream e2e tests. For the version automation, we wanted to run cni tests periodically. So instead of finishing the workflow here, it goes and also runs the tests defined in the script

scripts/lib/cluster.sh
jaydeokar: The version variable here is the binary version that we are using for kops. The environment variable defined in workflow is for defining the kubernetes cluster version we want to create

scripts/lib/cluster.sh
jaydeokar: I'll remove the "or" statements since they are not needed now

scripts/lib/cluster.sh
```diff
@@ -96,6 +101,10 @@ function up-kops-cluster {
         let RETRY_ATTEMPT=RETRY_ATTEMPT+1
         echo "In attempt# $RETRY_ATTEMPT, waiting for cluster validation"
     done
+
+    kubectl delete daemonset ebs-csi-node -n kube-system
```
jdn5126, line 105: nit: let's add a comment here explaining why these are being deleted

scripts/lib/integration.sh
jdn5126: Got it, just wanted to make sure that all resources are cleaned up

jdn5126 (dismissed):

scripts/lib/cluster.sh
jaydeokar: Added comment here explaining the reason

jdn5126 (approved):

orsenthil (approved):
LGTM.

Issue #2143, opened 2022-11-16
Create a readme for cni-metrics-helper charts


jayanthvn:
**What would you like to be added**:
Readme for cni-metrics-helper charts - https://github.com/aws/amazon-vpc-cni-k8s/tree/master/charts/cni-metrics-helper

Similar to https://github.com/aws/amazon-vpc-cni-k8s/blob/master/charts/aws-vpc-cni/README.md

**Why is this needed**:
Ease of use and better customer experience.

0xquark:
Hi 👋 @jayanthvn  I would like to work on this issue

0xquark:
Hi @jdn5126 , Sorry for the delay! Been studying for the end sems lately, Now I'm up for working on the issue.Before writing readme, I wanted to install this chart. I am not sure about which repo has the cni-metrics-helper chart so that i can add it in helm and then install it.Could you help me with that?

jdn5126:
@0xquark no worries, I had some free time the other day but then got pulled into something else. Currently, the chart is only present in this repo at https://github.com/aws/amazon-vpc-cni-k8s/tree/master/charts/cni-metrics-helper , so that's where the README would go. The helm installation instructions would be to install from that directory, i.e. `helm install cni-metrics-helper ./charts/cni-metrics-helper`

jdn5126:
All env variables should be called out, i.e. `USE_CLOUDWATCH`, `METRIC_UPDATE_INTERVAL`, `AWS_REGION`, `AWS_CLUSTER_ID`, and all fields in https://github.com/aws/amazon-vpc-cni-k8s/blob/master/charts/cni-metrics-helper/values.yaml .

Bonus points would be adding a `resources` section to the `values.yaml`. https://github.com/aws/amazon-vpc-cni-k8s/blob/master/charts/aws-vpc-cni/README.md is a good reference.

PR #2187, opened 2022-12-28
Added cni-metrics-helper docs

0xquark:
**What type of PR is this?**
This PR adds README.md to cni-metrics-helper as part of the docs

#2143 

Ease of use and better customer experience.

**Automation added to e2e**:

```release-note
Added resources field to cni-metrics-helper chart
```

### DIFF
--- a/charts/cni-metrics-helper/README.md
+++ b/charts/cni-metrics-helper/README.md
@@ -48,11 +48,9 @@ $ helm install my-release ./amazon-vpc-cni-k8s/charts/cni-metrics-helper --set u
 
 ## Resources
 
-You can specify resource limits for the cni-metrics-helper pods by setting the following values:
-
 | Parameter                 | Description                                    | Default |
 |---------------------------|------------------------------------------------|---------|
-| resources.limits.cpu      | CPU limit for the cni-metrics-helper pods      | 25m     |
+|    resources              | Resources for the pods.                        |   `{}`  |
 
 
 for example, to set a CPU limit of 200m and a memory limit of 256Mi for the cni-metrics-helper pods, you can use the following command:
--- a/charts/cni-metrics-helper/README.md
+++ b/charts/cni-metrics-helper/README.md
@@ -1,0 +1,64 @@
+# CNI METRICS HELPER
+
+This chart provides a Kubernetes deployment for the Amazon VPC CNI Metrics Helper, which is used to collect metrics for the Amazon VPC CNI plugin for Kubernetes.
+
+## Prerequisites
+
+- Kubernetes 1.11+ running on AWS
+- Helm 3.0+
+
+## Installing the Chart
+
+Clone the Amazon VPC CNI for Kubernetes repository to your local machine.
+
+```shell
+$ git clone https://github.com/aws/amazon-vpc-cni-k8s.git
+```
+Use the helm install command to install the chart into your Kubernetes cluster
+
+```shell
+$ helm install cni-metrics-helper ./amazon-vpc-cni-k8s/charts/cni-metrics-helper
+```
+
+## Configuration
+
+The following table lists the configurable parameters for this chart and their default values.
+
+| Parameter                  | Description                                                   | Default            |
+|----------------------------|---------------------------------------------------------------|--------------------|
+| fullnameOverride           | Override the fullname of the chart                            | cni-metrics-helper |
+| image.region               | ECR repository region to use. Should match your cluster       | us-west-2          |
+| image.tag                  | Image tag                                                     | v1.11.4            |
+| image.account              | ECR repository account number                                 | 602401143452       |
+| image.domain               | ECR repository domain                                         | amazonaws.com      |
+| env.USE_CLOUDWATCH         | Whether to export CNI metrics to CloudWatch                   | true               |
+| env.AWS_CLUSTER_ID         | ID of the cluster to use when exporting metrics to CloudWatch | default            |
+| metricUpdateInterval       | Interval at which to update CloudWatch metrics, in seconds    | 60                 |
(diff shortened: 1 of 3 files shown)


0xquark:
Hi @jdn5126 ! Thanks for the help. I've made a PR on  this issue. Few things i want you to take a look is :
1) I've used this git clone way to install the cni-metrics-helper. We can do the same with `helm repo add amazon-vpc-cni-k8s https://aws.github.io/amazon-vpc-cni-k8s/` and then using helm install. But this gave me an error of chart not existing.So the above method i used to install cni-metrics-helper is fine?
2) I've added the Resources section in the README.md , is that what you meant by it? or do i need to add a column for resources in the values.yaml file? ( I've used the default values for the resources as available on internet)
3) I've added all the parameters in the readme which were present in values.yaml file.
Let me know if there are any more parameters to add or any way i can improve the documentation.

charts/cni-metrics-helper/README.md
```diff
@@ -0,0 +1,64 @@
...
+| Parameter                  | Description                                                   | Default            |
+|----------------------------|---------------------------------------------------------------|--------------------|
+| fullnameOverride           | Override the fullname of the chart                            | cni-metrics-helper |
+| image.region               | ECR repository region to use. Should match your cluster       | us-west-2          |
+| image.tag                  | Image tag                                                     | v1.11.4            |
+| image.account              | ECR repository account number                                 | 602401143452       |
+| image.domain               | ECR repository domain                                         | amazonaws.com      |
+| env.USE_CLOUDWATCH         | Whether to export CNI metrics to CloudWatch                   | true               |
+| env.AWS_CLUSTER_ID         | ID of the cluster to use when exporting metrics to CloudWatch | default            |
+| metricUpdateInterval       | Interval at which to update CloudWatch metrics, in seconds    | 60                 |
```
jdn5126, line 36: nit: `env.METRIC_UPDATE_INTERVAL`, and the default is 30

I would update the description (will take multiple lines) to:
```
Interval at which to update CloudWatch metrics, in seconds
Metrics are published to CloudWatch at 2*METRIC_UPDATE_INTERVAL
```

charts/cni-metrics-helper/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+
+Specify each parameter using the `--set key=value[,key=value]` argument to `helm install` or provide a YAML file containing the values for the above parameters:
+
+```shell
+$ helm install my-release ./amazon-vpc-cni-k8s/charts/cni-metrics-helper --set useCloudwatch=false --values values.yaml
+
+```
+
+## Resources
```
jdn5126, line 49: For this to work properly, you will need to add a `resources` section to `values.yaml` in the existing helm chart. Note that the existing deployment already handles when resources are set: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/charts/cni-metrics-helper/templates/deployment.yaml#L23

So the steps would be:
1. Copy resources from https://github.com/aws/amazon-vpc-cni-k8s/blob/master/charts/aws-vpc-cni/values.yaml#L116
2. Increment helm chart version: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/charts/cni-metrics-helper/Chart.yaml#L18
3. Update README to capture `resources` defaults

jdn5126 (commented):
Thanks so much for doing this! Just a few nits and then this is good to go

jdn5126:
> 

Oh sorry, I missed these questions before reviewing:
1. Yeah, the method you used is fine. Eventually, this chart will be published at https://github.com/aws/eks-charts , and that will be the recommended `helm repo add`.
2. So we need the `resources` section to be present in the `values.yaml` file for this to work properly. I added some comments in the review, but feel free to reach out if you have any questions.
3. The documentation looks excellent, thanks again!

charts/cni-metrics-helper/README.md
0xquark: Done!

charts/cni-metrics-helper/README.md
0xquark: About the 3rd point , since we have set the defaults for cpu only in the resources section ( 25m ) So i need to have only this parameter? You can check the latest commit and let me know about it.

jdn5126: Good call out, so technically we just need the top-level `resources` field to be present, and then anyone can override it. I think we should follow what our integration tests do and just define it as an empty dictionary: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/test/helm/charts/cni-metrics-helper/values.yaml#L41

Very strange that this was missing in the first place. But yeah, then we can just add the `resources` entry to the README like:
```
| `resources`             | Resources for the pods                                  | `{}`                 |
```
then we don't have to worry about defaults

0xquark: Should i also remove the example ? or is it correct?

jdn5126: Yeah, let's update
```
resources:
  requests:
    cpu: 25m
```
in `values.yaml` to
```
resources: {}
```
and then I think this is good to go! Since this is a user-facing change, can you also update the release note in the PR description to something like "added resources field to cni-metrics-helper chart"

Thanks for all of the help!

0xquark: Done!

jdn5126 (approved):

jdn5126:
Merged, thanks again @0xquark !

0xquark:
> Merged, thanks again @0xquark !

Thank you so much @jdn5126 for helping out! It was greatly appreciated. 
Looking forward to contributing more in future! 🚀 🚀 🚀

PR #2188, opened 2022-12-28
Update Ginkgo download in workflows so go mod tidy runs

jdn5126:
**What type of PR is this?**
cleanup

N/A

This PR changes the way `ginkgo` is installed in integration test workflows. The previous method, `go install`, would error if extra packages were not present in `go.mod`/`go.sum`. Using `go mod download` should resolve this, and we can now run `go mod tidy` without fear of breaking workflows.

N/A

Kicking off manual PR workflow to verify fix

**Automation added to e2e**:
N/A

No

N/A

No

No

```release-note

```

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -26,8 +26,8 @@ jobs:
           go-version: "1.19"
       - name: Set up tools
         run: |
-          # Install ginkgo version from go.mod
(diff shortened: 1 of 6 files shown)


jdn5126:
"Manual Pull Request test" fails against this PR as it uses the old workflow file. Trying to determine how to make it use the new workflow file.

jayanthvn:
Lgtm, needs branch update.

jayanthvn (approved):

Issue #2071, opened 2022-08-26
Expose metrics for cni-metrics-helper for prometheus to scrape it


aravindhkudiyarasan:
**What would you like to be added**: Kindly expose metrics for [cni-metrics-helper](https://github.com/aws/amazon-vpc-cni-k8s/tree/master/cmd/cni-metrics-helper)

**Why is this needed**: Currently prometheus is unable to scrape metrics from cni-metrics-helper. hence cannot view cni related metrics in prometheus

haouc:
@aravindhkudiyarasan if you are interested to contribute, you are welcome to create a PR.

balajisa09:
@haouc  I would like to work on this, please assign this to me

balajisa09:
@aravindhkudiyarasan  @haouc  I can see that ipamD metrics are already exposed at port 61678 at path /metrics for prometheus to scrape. I am not clear about what other metrics you need. Can you please help on clarifying this?

balajisa09:
@haouc  any inputs on this ?

aravindhkudiyarasan:
Its not just about ipmad. The ultimate solidarity we need is to export
cni-metrics-helper pod to expose metrics directly so tat Prometheus can
scrape it.

Currently it needs to push it to cloudwatch then we have to export it

balajisa09:
got it. Thanks

balajisa09:
@haouc  can you review this. This is a draft code and i am looking for suggestions to improve it.

balajisa09:
@jdn5126  it is scrap able.  The 61678 port is exposed by the aws-vpc-cni application. There is a cni-metrics-helper app that scrapes the metrics from all the replicas of aws-vpc-cni app and aggregate those metrics ,do some post processing which is then being pushed to cloudwatch aws service. so, now you can see the overall vpc-cni app metrics only in cloudwatch.  

we are try to expose those aggregated metrics to prometheus , instead of pushing it to cloudwatch.

jdn5126:
Got it, thank you for explaining. So the design is for cni-metrics-helper pod to expose all metrics that it already exports to CloudWatch at a port that prometheus can pull from. And provide configuration by which customers can publish to CloudWatch, expose for prometheus, or both

balajisa09:
yeah. That's right.

jdn5126:
@balajisa09 thank you, I will start reviewing this soon

xyfleet:
[...]

PR #2192, opened 2023-01-02
Expose cni-metrics-helper to prometheus to scrape

balajisa09:
**What type of PR is this?**
feature

Expose metrics for cni-metrics-helper for prometheus to scrape it #2071

Expose cni-metrics-helper metrics directly to prometheus, previously it was being scraped and pushed to cloudwatch

**Testing done on this change**: NA

**Automation added to e2e**:  NA

**Will this break upgrades or downgrades. Has updating a running cluster been tested?**: No

```release-note

```


cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -413,18 +413,75 @@ func metricsListGrabAggregateConvert(ctx context.Context, t metricsTarget) (map[
...
-
+	
 	return families, interestingMetrics, resetDetected, nil
 }
 
 // Handler grabs metrics from target, aggregates the metrics and convert them into cloudwatch metrics
 func Handler(ctx context.Context, t metricsTarget) {
 	families, interestingMetrics, resetDetected, err := metricsListGrabAggregateConvert(ctx, t)
 
+	updatePromMetrics(interestingMetrics)
```
jdn5126, line 424: I think we should update prometheus metrics after CloudWatch metrics are updated, and we should do so only when env var is true. We should also add a comment explaining the call.

cmd/cni-metrics-helper/metrics/metrics.go
```diff
@@ -413,18 +413,75 @@ func metricsListGrabAggregateConvert(ctx context.Context, t metricsTarget) (map[
...
+
 	if err != nil || resetDetected {
 		t.getLogger().Infof("Skipping 1st poll after reset, error: %v", err)
 	}
 
 	cw := t.getCWMetricsPublisher()
 	produceCloudWatchMetrics(t, families, interestingMetrics, cw)
 }
+
+func updatePromMetrics(interestingMetrics map[string]metricsConvert){
```
jdn5126, line 434: Instead of manually calling metrics, we should loop over all metrics similar to how `produceCloudWatchMetrics` function does. Maybe we can move both into the same function and use env var to control updates.

jdn5126: This will also simplify the process of adding new metrics.

cmd/cni-metrics-helper/metrics/prometheus.go
```diff
@@ -0,0 +1,179 @@
+package metrics
+
+import (
+	"net/http"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/prometheus/common/log"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
```
jdn5126, line 10: Since these are the exact same metrics defined in https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/awsutils/awsutils.go#L86 and https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/datastore/data_store.go#L89 , I think we can move the initialization of these variables to a common package. This will simplify adding new metrics in the future and keep everything in one place.

cmd/cni-metrics-helper/metrics/prometheus.go
```diff
@@ -0,0 +1,179 @@
...
+func StartPrometheusMetricsServer(){
+	log.Info("Starting prometehus metrics server for cni-metrics-helper")
+	http.ListenAndServe("localhost:2112",nil)
+}
+
+func init(){
+	reg := prometheus.NewRegistry()
+	promHandler := promhttp.HandlerFor(reg,promhttp.HandlerOpts{})
+
+	reg.MustRegister(awsAPILatency)
```
jdn5126, line 157: Let's see if we can share this registration with ipamd via a helper function

cmd/cni-metrics-helper/metrics/prometheus.go
```diff
@@ -0,0 +1,179 @@
...
+	reg.MustRegister(podENIErr)
+	reg.MustRegister(enis)
+	reg.MustRegister(totalIPs)
+	reg.MustRegister(assignedIPs)
+	reg.MustRegister(forceRemovedENIs)
+	reg.MustRegister(forceRemovedIPs)
+	reg.MustRegister(totalPrefixes)
+	reg.MustRegister(ipsPerCidr)
+
+	http.Handle("/metrics",promHandler)
```
jdn5126, line 178: This belongs in the `setupMetricsServer` function mentioned above

pkg/k8sapi/k8sutils.go
```diff
@@ -18,6 +18,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/oidc"
```
jdn5126, line 21: remove this

PR #2201, opened 2023-01-06
Refactor cni-metrics-helper chart for eks charts release

jdn5126:
**What type of PR is this?**
cleanup

N/A

This PR cleans up cni-metrics-helper chart so that it can be published to eks-charts: https://github.com/aws/eks-charts

The eks-charts PR is https://github.com/aws/eks-charts/pull/870

N/A

Manually verified that chart can be installed via helm

**Automation added to e2e**:
N/A

No

This will not break upgrades or downgrades. A running cluster has been tested.

No

No

```release-note

```

### DIFF
--- a/charts/cni-metrics-helper/Chart.yaml
+++ b/charts/cni-metrics-helper/Chart.yaml
@@ -1,23 +1,19 @@
 apiVersion: v2
 name: cni-metrics-helper
-description: A Helm chart for Kubernetes
-
-# A chart can be either an 'application' or a 'library' chart.
-#
-# Application charts are a collection of templates that can be packaged into versioned archives
-# to be deployed.
-#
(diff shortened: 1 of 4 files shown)


charts/cni-metrics-helper/Chart.yaml
```diff
@@ -1,23 +1,19 @@
-apiVersion: v2
+apiVersion: v1
```
jaydeokar, line 1: Do we need to change this to upload to eks-charts or just because we don't mandate the helm version to be 3 and above for installation?

jdn5126: I was thinking about helm2 compatibility, but since this chart is new and being released after we specified helm3 is required, I will put this back to `v2`.

jaydeokar (approved):
lgtm

PR #2202, opened 2023-01-09
[Test] Wait for ipamd to reconcile enis and prefixes before starting new test

jaydeokar:
**What type of PR is this?**
feature

N/A

With warm target tests, we don't wait for ipamd to reconcile back to the old state before beginning new test. This sometimes results in flaky test as the enis/ips from the last run are not yet recycled. With this change, we poll the APIs every 30 sec (with timeout for 2 mins) to check if the original state has been reconciled and then move forward to the test.

N/A

Yes, tested this on 1.24 cluster

**Automation added to e2e**:
N/A

N/A

N/A

N/A

N/A

### DIFF
--- a/test/integration/common/util.go
+++ b/test/integration/common/util.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"time"
 
 	"github.com/aws/amazon-vpc-cni-k8s/test/agent/pkg/input"
 	"github.com/aws/amazon-vpc-cni-k8s/test/framework"
@@ -199,4 +200,19 @@ func ValidateTraffic(f *framework.Framework, serverDeploymentBuilder *manifest.D
 	Expect(err).ToNot(HaveOccurred())
 	Expect(successRate).Should(BeNumerically(">=", succesRate))
 
+}
+
+func WaitToReconcileInitialState(f *framework.Framework, primaryInstance *ec2.Instance, defaultEniCount int, defaultIpsPerEni int, DefaultPrefixPerEni int) {
+	By("Verifying number of enis, ips and ip prefixes before updating the parameters")
+	Eventually(func(g Gomega) {
+		primaryInstance, err := f.CloudServices.
+			EC2().DescribeInstance(*primaryInstance.InstanceId)
+
+		g.Expect(err).ToNot(HaveOccurred())
+		g.Expect(len(primaryInstance.NetworkInterfaces)).Should(Equal(defaultEniCount))
+		g.Expect(len(primaryInstance.NetworkInterfaces[0].PrivateIpAddresses)).
+			Should(Equal(defaultIpsPerEni))
+		g.Expect(len(primaryInstance.NetworkInterfaces[0].Ipv4Prefixes)).
+			Should(Equal(DefaultPrefixPerEni))
+	}).WithTimeout(time.Minute * 3).WithPolling(30 * time.Second).Should(Succeed())
 }
--- a/test/integration/ipamd/ipamd_suite_test.go
+++ b/test/integration/ipamd/ipamd_suite_test.go
@@ -30,10 +30,13 @@ import (
 var primaryInstance *ec2.Instance
 var f *framework.Framework
 var err error
+var defaultEniCount int
+var defaultIpsPerEni int
 
 const (
 	CoreDNSDeploymentName = "coredns"
 	KubeSystemNamespace   = "kube-system"
+	DefaultPrefixPerEni   = 0
 )
 
 var coreDNSDeploymentCopy *v1.Deployment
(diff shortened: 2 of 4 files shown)


test/integration/ipamd/ipamd_suite_test.go
```diff
@@ -34,6 +34,9 @@ var err error
 const (
 	CoreDNSDeploymentName = "coredns"
 	KubeSystemNamespace   = "kube-system"
+	DefaultEniCount       = 1
+	DefaultIpsPerEni      = 15
```
jayanthvn, line 38: Is this hardcoded based on the instance type we are using? Can we instead make this dynamic? Since if we are using locally and have a different instance type then this test will fail.

jaydeokar: > Is this hardcoded based on the instance type we are using? Can we instead make this dynamic? Since if we are using locally and have a different instance type then this test will fail.

Yes, I assumed we use the same instance type across our tests. But yes I agree, I'll make the values dynamic

jayanthvn: During release tests, we will be running on our cluster so better to be dynamic. Thanks.

jaydeokar: @jayanthvn I've updated it to pull dynamically based on the instance type of primary node. 

Assumptions here: 
- Every network adapter has one ENI at the start
- There's no prefix assigned to any interface at the start

test/integration/ipamd/warm_target_test.go
```diff
@@ -34,6 +34,22 @@ var _ = Describe("test warm target variables", func() {
 		var maxENI int
 
 		JustBeforeEach(func() {
+
+			By("Verify number of enis and ips before updating the parameters")
```
jdn5126, line 38: Since this is called in multiple places, let's turn it into a helper function. Also, I think 2 mins may not be enough. Let's start with 5 mins

test/integration/ipamd/ipamd_suite_test.go
```diff
@@ -88,6 +91,12 @@ var _ = BeforeSuite(func() {
 	primaryInstance, err = f.CloudServices.EC2().DescribeInstance(instanceID)
 	Expect(err).ToNot(HaveOccurred())
 
+	primaryInstanceDefaults, err := f.CloudServices.EC2().DescribeInstanceType(*primaryInstance.InstanceType)
+	Expect(err).ToNot(HaveOccurred())
+
+	defaultEniCount = len(primaryInstanceDefaults[0].NetworkInfo.NetworkCards)
+	defaultIpsPerEni = int(*primaryInstanceDefaults[0].NetworkInfo.Ipv4AddressesPerInterface)
```
jdn5126, line 98: Quick question: if test cases run before this (like the CNI suite) and have not cleaned up yet, will this default count be off?

jaydeokar: No, this is being pulled from `instance-type` metadata. So it should be consistent irrespective of what tests were run

test/integration/ipamd/warm_target_test.go
jaydeokar: I was expecting few more conditions to check for different tests. I can extract it into a function for now

jaydeokar: Also, I was under the impression if recycle duration is 30s then it should ideally finish the cleanup in about a minute right for the worst case ?  That's why I had made it to 2 mins just to leave some headroom. I can surely increase it to 5 mins if it's necessary.

jdn5126: Ah, I'm not sure what the duration is. If it is 30s, then this sounds reasonable. We can start with 2m knowing that if it takes longer, it is something we want to investigate.

jaydeokar: Increased it to 3 mins just to be on safer side and not have into flaky tests.

test/integration/common/util.go
```diff
@@ -200,3 +201,23 @@ func ValidateTraffic(f *framework.Framework, serverDeploymentBuilder *manifest.D
 	Expect(successRate).Should(BeNumerically(">=", succesRate))
 
 }
+
+func WaitToReconcileInitialState(f *framework.Framework, primaryInstance *ec2.Instance, defaultEniCount int, defaultIpsPerEni int, DefaultPrefixPerEni int) {
+	By("Verifying number of enis, ips and ip prefixes before updating the parameters")
+	Eventually(func(g Gomega) {
+
```
jdn5126, line 208: nit: I would remove a lot of these empty newlines so that the related calls are connected

test/integration/ipamd/warm_target_test.go
```diff
@@ -34,6 +35,10 @@ var _ = Describe("test warm target variables", func() {
 		var maxENI int
 
 		JustBeforeEach(func() {
+
```
jdn5126, line 38: nit: these newlines seem unnecessary

jdn5126 (approved):
style nits, otherwise this looks good to me

jdn5126 (approved):

Issue #2212, opened 2023-01-13
'aws-node' pod's not running (CrashLoopBackOff ) with internal error


Argal87:
**What happened**:

The last commit 'a81e07f8e9c642f9e7370b44ca02d0fd60b17ed1' don't run 'aws-node' pod.

**Attach logs**

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:
kubectl apply -f https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/master/config/master/aws-k8s-cni.yaml

**Anything else we need to know?**:
when I downloaded the 'aws-k8s-cni.yaml'  file from previous commit and ran it with kubectl apply, that was worked.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.26.0
- CNI Version
- OS (e.g: `cat /etc/os-release`): CentOS Linux 7 (AltArch)
- Kernel (e.g. `uname -a`): 4.18.0-348.20.1.el7.aarch64

jdn5126:
@Argal87 thanks for catching this! The commit you highlighted modified `config/master/*` in a backward-incompatible way, as those changes are not compatible with `v1.12.1` images used in YAML file. The changes to `config/master/*` need to wait until the next release image is available (`v1.12.2` or `v1.13.0`). I will revert changes in that directory now.

jayanthvn:
PR is merged, closing the issue.

PR #2213, opened 2023-01-13
Revert manifest changes from #2149

jdn5126:
**What type of PR is this?**
bug

#2212

This PR reverts the manifest changes made in #2149 . Those manifest changes are not compatible with the `v1.12.1` image. The chart changes from that PR are fine, as the official chart is published on eks-charts, but we need to hold off on updating `config/master/aws-k8s-cni*` until a compatible image is present in manifest (`v1.12.2` or `v1.13.0`).

N/A

Verified that manifest can be applied and aws-node pods come up

**Automation added to e2e**:
N/A

N/A

This will not break upgrades or downgrades, and a running cluster has been tested.

Yes, reverting change.

No

```release-note

```

### DIFF
--- a/config/master/aws-k8s-cni-cn.yaml
+++ b/config/master/aws-k8s-cni-cn.yaml
@@ -61,7 +61,7 @@ rules:
   - apiGroups: [""]
     resources:
       - nodes
-    verbs: ["list", "watch", "get"]
+    verbs: ["list", "watch", "get", "update"]
   - apiGroups: ["extensions"]
     resources:
       - '*'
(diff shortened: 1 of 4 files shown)


jayanthvn:
Since we generate the manifests from the charts during the release we should be good but can we also open an issue and milestone it for the release?

jayanthvn (approved):
/lgtm

jdn5126:
> Since we generate the manifests from the charts during the release we should be good but can we also open an issue and milestone it for the release?

Filed #2214

PR #2221, opened 2023-01-17
Fix integration test workflows installing ginkgo

jdn5126:
**What type of PR is this?**
Bug

N/A

This PR fixes integration test workflows that need to install ginkgo. https://github.com/aws/amazon-vpc-cni-k8s/pull/2188 switched `go install` to `go mod download`, as `go install` would fail unless `go get` was run first to pull in some extra packages that should not be needed.

Now, we use `go install -mod=mod`, which should succeed and skip those extra, unneeded packages. If this does not work, I will move back to `go get && go install`.

N/A

N/A

**Automation added to e2e**:
N/A

No

N/A

No

No

```release-note

```

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -26,8 +26,8 @@ jobs:
           go-version: "1.19"
       - name: Set up tools
         run: |
-          # Download ginkgo version from go.mod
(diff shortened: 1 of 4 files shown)


jayanthvn (approved):

jayanthvn (approved):
lgtm

PR #2233, opened 2023-01-24
Update image tags to v1.12.1-eksbuild.2

jdn5126:
**What type of PR is this?**
cleanup

N/A

This PR updates the image tags used in `release-1.12` branch to `v1.12.1-eksbuild.2`.

N/A

Verified that image works.

**Automation added to e2e**:
N/A

No

N/A

Yes, updating image tags

N/A

```release-note

```

### DIFF
--- a/charts/aws-vpc-cni/values.yaml
+++ b/charts/aws-vpc-cni/values.yaml
@@ -8,7 +8,7 @@ nameOverride: aws-node
 
 init:
   image:
-    tag: v1.12.1
+    tag: v1.12.1-eksbuild.2
     region: us-west-2
     account: "602401143452"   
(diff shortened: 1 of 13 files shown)


charts/aws-vpc-cni/values.yaml
```diff
@@ -8,7 +8,7 @@ nameOverride: aws-node
 
 init:
   image:
-    tag: v1.12.1
+    tag: v1.12.1-eksbuild.2
```
jayanthvn, line 11: Let's sync up on this. We generally don't update eksbuild tags here.

jdn5126:
Discussed offline and we will not update the tags like this on GitHub

PR #2231, opened 2023-01-24
Update CNI metrics helper README

jdn5126:
**What type of PR is this?**
documentation

**Which issue does this PR fix**:
N/A

**What does this PR do / Why do we need it**:
This PR updates the CNI metrics helper README to recommend installing from helm chart or latest manifest in `config/master`.

**If an issue # is not available please add repro steps and logs from IPAMD/CNI showing the issue**:
N/A

**Testing done on this change**:
N/A

**Automation added to e2e**:
N/A

**Will this PR introduce any new dependencies?**:
N/A

**Will this break upgrades or downgrades. Has updating a running cluster been tested?**:
N/A

**Does this change require updates to the CNI daemonset config files to work?**:
No

**Does this PR introduce any user-facing change?**:
No

```release-note

```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2238, opened 2023-01-27
add backticks to cni-metric port

### LINKED ISSUE #2208: Update cni-metric-helper doc with requirement about node ports
**What happened**:

I deployed a cluster using the terraform module [terraform-aws-eks](https://github.com/terraform-aws-modules/terraform-aws-eks/) by default the managed node groups are created with a minimal set of security group ingress rules with tcp ports 443 (cluster API),10250(kubelet), 53(coredns) as noted [here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/node_groups.tf#L74-L106)

From what I understand how cni-metric-helper communicates with ipamd/vpc-cni pods that have the endpoint on tcp port :61678/metrics

The cni-metric-helper doesn't try to reach the pod IP directly and instead use the kube-api to proxy, the cni-metric-helper will hit the cluster kube-api node on port 443 (which is very common configured node->cluster) then a connection is made from cluster master node to node on port 61678.

In my case the cni-metric-helper was not able to communicate and printed the following logs
```
Failed to grab CNI endpoint: the server is currently unable to handle the request (get pods aws-node-ktpbn:61678)"}
Failed to grab CNI endpoint: the server is currently unable to handle the request (get pods aws-node-4tjv5:61678)"}
```

We should update the docs to have this information as part of installation instructions.

Deploy a EKS with recommended [Restricting cluster traffic](https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html#security-group-default-rules), then install cni-metric-helper following [docs](https://github.com/terraform-aws-modules/terraform-aws-eks/)

I opened an issue on terraform repo https://github.com/terraform-aws-modules/terraform-aws-eks/issues/2394

csantanapr:
**What type of PR is this?**

documentation

Resolves #2208

Fix format of doc string

### DIFF
--- a/cmd/cni-metrics-helper/README.md
+++ b/cmd/cni-metrics-helper/README.md
@@ -13,7 +13,7 @@ The following diagram shows how `cni-metrics-helper` works in a cluster:
 
 ![](../../docs/images/cni-metrics-helper.png)
 
-As you can see in the diagram, the `cni-metrics-helper` connects to the API Server over https (`tcp/443`), and another connection is created from the API Server to the worker node over http (tcp/61678). If you deploy Amazon EKS with recommended [Restricting cluster traffic](https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html#security-group-restricting-cluster-traffic), then make sure that a security group is in place that allows the inbound connection from the API Server to the worker nodes over `tcp/61678`.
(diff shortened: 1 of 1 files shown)


csantanapr:
already covered by another PR

jdn5126:
Sorry for the confusion here. Discussed offline and I will fix this nit in #2231

PR #2240, opened 2023-01-27
using instance id instead of node name for event regarding UID

haouc:
**What type of PR is this?**
This PR is changing the node event to VPC resource controller to use instance id as event regarding UID.

feature
n/a

This will enable VPC resource controller to cache on events for better performance.
Referring to this PR
https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/160

manual tests were done
**Automation added to e2e**:

e2e will be followed up to wait for the other controller being available in EKS.

no
no

no

no
```release-note

```

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -390,12 +390,6 @@ func New(rawK8SClient client.Client, cachedK8SClient client.Client) (*IPAMContex
 	c.enableIPv4 = isIPv4Enabled()
 	c.enableIPv6 = isIPv6Enabled()
 
-	eventrecorder, err := eventrecorder.New(rawK8SClient, cachedK8SClient)
-	if err != nil {
-		return nil, errors.Wrap(err, "ipamd: unable to initialize event recorder")
-	}
-	c.eventRecorder = eventrecorder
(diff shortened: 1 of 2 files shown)


pkg/ipamd/ipamd.go
```diff
@@ -452,6 +452,9 @@ func New(rawK8SClient client.Client, cachedK8SClient client.Client) (*IPAMContex
 		go wait.Forever(func() { _ = c.awsClient.RefreshSGIDs(mac) }, 30*time.Second)
 	}
 
+	// setup Host instanceID in event recorder
+	eventrecorder.HostID = c.awsClient.GetInstanceID()
```
jayanthvn, line 456: Can you move this up in line 397 but make sure to also reorder the `awsClient` in 405. You might need to move the event recorder init below 405..

jayanthvn (approved):
lgtm

pkg/ipamd/ipamd.go
haouc: Makes sense. I have updated the PR. Please review. Thanks.

jayanthvn (approved):

Issue #2102, opened 2022-10-04
Source Environment Variables from ConfigMap


jdn5126:
**What would you like to be added**:
CNI environment variables to be sourced from ConfigMap

**Why is this needed**:
To allow environment variables to persist during CNI upgrade. Current mechanism is to run `kubectl set env` during upgrade.

jdn5126:
Reference #1088

dschunack:
Still waiting in this feature

dschunack:
Still waiting on this

warezWally:
Would also like this feature, to allow setting of the "ENI_CONFIG_LABEL_DEF" environment variable in the aws-node pods. Without it my pods are not being assigned an IP

Issue #1910, opened 2022-03-05
Some pods unable to reach service backed by public IP of ec2 instance


prateekgogia:
**What happened**:
When running a service on EKS which is backed by public IP address of the instance its reachable by some of the pods running on the same node. Pods which are assigned ip from primary ENI are able to reach this service, however, pods with IPs assigned from an attached ENI are not able to reach.

`my-service` is backed by nginx pod running in host network namespace, and 3.4.5.6(changed) is the public IP of the instance 
```
➜  k get endpoints my-service
NAME         ENDPOINTS            AGE
my-service   3.4.5.6:8443   82m

➜  k get service my-service
NAME         TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
my-service   ClusterIP   10.100.205.85   <none>        8443/TCP   83m

➜ k get pods nginx -o wide
NAME    READY   STATUS    RESTARTS   AGE   IP             NODE                                         NOMINATED NODE   READINESS GATES
nginx   1/1     Running   0          36m   192.168.1.80   ip-192-168-1-80.us-west-2.compute.internal   <none>           <none>
```

Created 3 pods, of which 2 pods (jv4cm and x82fw) are able to curl the service IP and one is not able to curl, it timesout.
```
curl -I http://10.100.205.85:8443
HTTP/1.0 200 OK
Server: SimpleHTTP/0.6 Python/2.7.15
Date: Sat, 05 Mar 2022 20:30:49 GMT
Content-type: text/html; charset=UTF-8
Content-Length: 710
```
```
kubectl get pods -o wide
NAME                               READY   STATUS    RESTARTS   AGE   IP               NODE                                         NOMINATED NODE   READINESS GATES
dummy-deployment-6f8877777-jv4cm   1/1     Running   0          23m   192.168.28.203   ip-192-168-1-80.us-west-2.compute.internal   <none>           <none>
dummy-deployment-6f8877777-k7jwn   1/1     Running   0          22m   192.168.27.187   ip-192-168-1-80.us-west-2.compute.internal   <none>           <none>
dummy-deployment-6f8877777-x82fw   1/1     Running   0          34m   192.168.7.122    ip-192-168-1-80.us-west-2.compute.internal   <none>           <none>
```
Two ENIs are attached to the node with the following IPs-

```
Primary ENI with public IP attached has following IPs
192.168.7.122
192.168.28.203
192.168.30.91

Secondary ENI has IPs -
[...]

PR #2243, opened 2023-01-31
Support routing to external IPs behind service

jdn5126:
**What type of PR is this?**
bug

#1910 

This PR adds a new environment variable, `AWS_EXTERNAL_SERVICE_CIDRS`, where customers can specify CIDRs outside of VPC that are reachable via service IP. This is necessary to route to these endpoints via service IP from secondary ENIs (#1910).

For each CIDR, an IP rule is installed (with a priority higher than the from-container rule) to route to the CIDR via main routing table. An example rule is:
```
1535:   from all to A.B.C.D lookup main
```

Currently, every time `AWS_EXTERNAL_SERVICE_CIDRS` is changed, the `aws-node` pod restarts, which is why the implementation for this function is in IPAMD node initialization. In the future, if a ConfigMap is used (#2102 is implemented), then operators will need to manually restart the `aws-node` container after modifying the environment variable.

If a solution can be implemented one day that dynamically detects this case and ensures proper routing, this environment variable will be deprecated.

N/A

Unit tests are added to cover new functions. I manually verified functionality and ensured that all existing integration tests pass. I considered adding an integration test case, but decided against it as it would require setting up communication outside of VPC or requiring node IPs to be public, both of which we want to avoid. The unit tests cover that we add and delete the IP rules as expected, and we know that routing works when IP rules are present. 

**Automation added to e2e**:
N/A

No

This PR will not break upgrades or downgrades. A running cluster has been tested.

No

Yes

```release-note
Added AWS_EXTERNAL_SERVICE_CIDRS to resolve issues routing to endpoints outside of VPC backed by service.
```


README.md
```diff
@@ -602,6 +601,15 @@ VPC CNI uses `iptables-legacy` by default. Setting `ENABLE_NFTABLES` to `true` w
 **Note:** VPC CNI image contains `iptables-legacy` and `iptables-nft`. Switching between them is done via `update-alternatives`. It is *strongly* recommended that the iptables mode matches that which is used by the base OS and `kube-proxy`.
 Switching modes while pods are running or rules are installed will not trigger reconciliation. It is recommended that rules are manually updated or nodes are drained and cordoned before updating. If reloading node, ensure that previous rules are not set to be persisted.
 
+#### `AWS_EXTERNAL_SERVICE_CIDRS` (v1.13.0+)
+
+Type: String
+
+Default: empty
+
+Specify a comma-separated list of IPv4 and/or IPv6 CIDRs that *must* be routed via main routing table. This is required for secondary ENIs to reach endpoints outside of VPC that are backed by a service.
```
achevuru, line 610: We don't need to support this for IPv6. We currently only support IPv6 in PD mode and so we will only need one IPv6 prefix attached to Primary ENI for Pod IPs and so we will not be affected by symmetric routing fix in IPv6 mode.

README.md
```diff
@@ -602,6 +601,15 @@ VPC CNI uses `iptables-legacy` by default. Setting `ENABLE_NFTABLES` to `true` w
...
+For every item in the list, an `ip rule` will be created with a priority greater than the from container rule. If an item is not a valid IPv4 or IPv6 CIDR, it will be skipped.
```
achevuru, line 611: nit: `from container rule` is probably too technical for someone who is not familiar with the code base. Maybe we can change it to `priority greater than ip rule capturing the egress traffic from the container` or something similar..

pkg/ipamd/ipamd.go
```diff
@@ -632,6 +632,14 @@ func (c *IPAMContext) configureIPRulesForPods() error {
 			log.Warnf("UpdateRuleListBySrc in nodeInit() failed for IP %s: %v", info.IP, err)
 		}
 	}
+
+	// Program IP rules for external service CIDRs and cleanup stale rules.
+	extServiceCIDRs := c.networkClient.GetExternalServiceCIDRs()
+	err = c.networkClient.UpdateExternalIpRules(extServiceCIDRs)
+	if err != nil {
+		log.Warnf("UpdateExternalServiceIpRules in nodeInit() failed")
```
achevuru, line 640: Function name in the log message and the actual function name don't match. I think the one in log message is more clear and intuitive.

pkg/networkutils/network.go
```diff
@@ -67,11 +72,15 @@ const (
 	// be installed and will be removed if they are already installed. Defaults to false.
 	envExternalSNAT = "AWS_VPC_K8S_CNI_EXTERNALSNAT"
 
-	// This environment is used to specify a comma separated list of ipv4 CIDRs to exclude from SNAT. An additional rule
+	// This environment is used to specify a comma-separated list of IPv4 CIDRs to exclude from SNAT. An additional rule
 	// will be written to the iptables for each item. If an item is not an ipv4 range it will be skipped.
 	// Defaults to empty.
 	envExcludeSNATCIDRs = "AWS_VPC_K8S_CNI_EXCLUDE_SNAT_CIDRS"
 
+	// This environment is used to specify a comma-separated list of IPv4 and IPv6 CIDRs that require routing lookup in
```
achevuru, line 80: Let's remove IPv6 and make this env variable a no-op in IPv6 mode

achevuru (dismissed):

PR #2260, opened 2023-02-09
Fixing the log message to be meaningful

rajeeshckr:
**What type of PR is this?**
Update the debug log message.

When CNI pod was launching on a non schedulable node, we were seeing this error message. 
`{"level":"debug","ts":"2023-01-23T03:13:23.608Z","caller":"ipamd/ipamd.go:660","msg":"AWS CNI is terminating, will not try to attach any new IPs or ENIs right now"}`. 

This was confusing, we had to look in the code to understand this is because of node not schedulable in our case. 

Updated the logic to log valid debug message when node is not schedulable.

`{"level":"debug","ts":"2023-01-23T03:13:23.608Z","caller":"ipamd/ipamd.go:660","msg":"AWS CNI is terminating, will not try to attach any new IPs or ENIs right now"}`. 

No

No

No

No

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -739,8 +739,13 @@ func (c *IPAMContext) decreaseDatastorePool(interval time.Duration) {
 
 // tryFreeENI always tries to free one ENI
 func (c *IPAMContext) tryFreeENI() {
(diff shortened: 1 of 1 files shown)


jayanthvn (approved):
lgtm, thanks for the PR.

PR #2263, opened 2023-02-09
Remove timestamp from eni-max-pods.txt

cartermckinnon:
**What type of PR is this?**

cleanup/feature

N/A

I'd like to automate the generation of `eni-max-pods.txt` for `awslabs/amazon-eks-ami`. Currently, every time the file is generated, it changes; because a timestamp is included in the contents. Removing this timestamp allows an empty `diff` of the file when none of the max pods values have changed. I don't think the timestamp provides much value anyway, because equivalent timestamps are already tracked by `git`.

N/A

N/A

**Automation added to e2e**:

N/A

No

No

No

No


jdn5126 (dismissed):

jdn5126:
Good call, thanks for the fix @cartermckinnon !

jdn5126 (approved):

cartermckinnon:
The format checks always seem to fail on my PR's 😂 @jdn5126 do you have a guess as to why?

jdn5126:
> The format checks always seem to fail on my PR's 😂 @jdn5126 do you have a guess as to why? I can't reproduce it locally.

Huh, looks like it is complaining about the spacing in struct definition. I use the `vim-go` plugin and it auto-formatted for me, so this diff passed:
```
diff --git a/scripts/gen_vpc_ip_limits.go b/scripts/gen_vpc_ip_limits.go
index 6ccb1873..9f769576 100644
--- a/scripts/gen_vpc_ip_limits.go
+++ b/scripts/gen_vpc_ip_limits.go
@@ -101,13 +101,11 @@ func main() {
                log.Fatalf("Failed to create file: %s\n", err)
        }
        err = eksMaxPodsTemplate.Execute(f, struct {
-               Timestamp string
-               ENIPods   []string
-               Regions   []string
+               ENIPods []string
+               Regions []string
        }{
-               Timestamp: time.Now().Format(time.RFC3339),
-               ENIPods:   eniPods,
-               Regions:   regions,
+               ENIPods: eniPods,
+               Regions: regions,
        })
        if err != nil {
                log.Fatalf("Failed to generate template: %v\n", err)
@@ -239,7 +237,6 @@ var limitsTemplate = template.Must(template.New("").Parse(`// Copyright Amazon.c
 // permissions and limitations under the License.

 // Code generated by go generate; DO NOT EDIT.
-// This file was generated at {{ .Timestamp }}
 //
 // The regions queried were:
 {{- range $region := .Regions}}
```

cartermckinnon:
Ah, duh; sorry about that. Fixed.

jdn5126 (approved):

PR #2288, opened 2023-02-27
increase tcp connect timeout

jaydeokar:
**What type of PR is this?**
cleanup

N/A

IPv6 test fails for the tcp client because of timeout. This increases the timeout value to 5 secs

N/A

N/A

**Automation added to e2e**:
N/A

N/A

N/A

N/A

N/A


jdn5126 (approved):

orsenthil (approved):
LGTM

PR #2291, opened 2023-02-28
using aws-node as event regarding object

haouc:
**What type of PR is this?**
Change SGP events' regarding object from node to pod aws-node and also include related object to the event. The related object is for the node.
Populate node and pod only once when the event recorder is created by IPAMD instead of creating every time when sending event. They won't change since IPAMD starts.

N/A

N/A

Installed aws-node through yaml and helm install.
Tested with SGP enabled and checked events from describing pod aws-node
```
{"level":"debug","ts":"2023-02-28T04:35:54.875Z","caller":"ipamd/ipamd.go:1313","msg":"Sent sgp event: eventType: Normal, reason: AwsNodeNotificationToRc, message: vpc.amazonaws.com/has-trunk-attached=false, action NeedTrunk: 8e742703-2504-48d5-9c46-8ac3f07382d4"}
```
**Automation added to e2e**:

N/A

NO
NO

Updated

Yes, events will be populated to pod events
```release-note

```

### DIFF
--- a/cmd/aws-k8s-agent/main.go
+++ b/cmd/aws-k8s-agent/main.go
@@ -19,7 +19,6 @@ import (
 
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/k8sapi"
-	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/eventrecorder"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/version"
 )
@@ -58,8 +57,6 @@ func _main() int {
 		log.Errorf("Failed to create cached kube client: %s", err)
 		return 1
 	}
-
-	eventrecorder.New(rawK8SClient, cacheK8SClient)
 
 	ipamContext, err := ipamd.New(rawK8SClient, cacheK8SClient)
 	if err != nil {
--- a/pkg/utils/eventrecorder/eventrecorder.go
+++ b/pkg/utils/eventrecorder/eventrecorder.go
@@ -21,10 +21,13 @@ import (
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/k8sapi"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/sgpp"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
+	"github.com/aws/amazon-vpc-cni-k8s/test/framework/utils"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/uuid"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/events"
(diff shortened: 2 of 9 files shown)


haouc:
Sending pod event from aws-node has these benefits
- the event will be sent to kube-system namespace instead of default, which can help VPC resource controller filters event on system namespace. Customers could have many workloads in their default namespace.
- we can see much more node events than aws-node pod events. Creating/Updating/Deleting Pods into the node for example can create many events although we can filter them later from watcher.

Note: Using pod event is not a must. But we feel it has more benefits.

cmd/aws-k8s-agent/main.go
```diff
@@ -59,8 +58,6 @@ func _main() int {
 		return 1
 	}
 
-	eventrecorder.New(rawK8SClient, cacheK8SClient)
```
jdn5126, line 62: Nice find!

pkg/utils/eventrecorder/eventrecorder.go
```diff
@@ -65,6 +71,14 @@ func New(rawK8SClient, cachedK8SClient client.Client) (*EventRecorder, error) {
 	eventRecorder.RawK8SClient = rawK8SClient
 	eventRecorder.CachedK8SClient = cachedK8SClient
 
+	if eventRecorder.hostNode, err = findMyNode(eventRecorder.CachedK8SClient); err != nil {
+		log.Errorf("Failed to find host node: %s", err)
+	}
+
+	if eventRecorder.hostPod, err = findMyPod(eventRecorder.CachedK8SClient); err != nil {
```
jdn5126, line 78: Good call on caching this, this will be helpful for refactoring `BroadcastEvent` as well

jdn5126 (commented):
Changes overall LGTM. Can we also check pod in associated unit test? https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/utils/eventrecorder/eventrecorder_test.go#L152

pkg/utils/eventrecorder/eventrecorder.go
```diff
@@ -104,24 +118,32 @@ func (e *EventRecorder) findMyNode(ctx context.Context) (corev1.Node, error) {
...
+			Namespace: e.hostNode.Namespace,
+			UID:       types.UID(e.HostID),
+		},
+	}
 
-	e.Recorder.Eventf(nodeCopy, nil, eventType, reason, action, message)
-	log.Debugf("Sent node event: eventType: %s, reason: %s, message: %s, action %s", eventType, reason, message, action)
+	// make the event unique avoid to be aggregated by event broadcaster
+	action = action + ": " + string(uuid.NewUUID())
+	e.Recorder.Eventf(&e.hostPod, instance, eventType, reason, action, message)
```
jayanthvn, line 145: The function name is SendNodeEvent, but this will be pod event right? Am I missing anything?

jdn5126 (approved):

pkg/utils/eventrecorder/eventrecorder.go
haouc: I didn't change the name since the event is still for node label as 
```
new event{
  Regarding `aws-node pod`
  Related to 'node' 
}
```
I have thought if I need rename but was not sure if that is necessary. Thoughts?

jayanthvn (approved):

jdn5126:
> The failed action doesn't look like a blocking issue
> 
> ```
> ERROR: failed to solve: public.ecr.aws/eks-distro-build-tooling/golang:1.19.6-3-gcc-al2: pulling from host public.ecr.aws failed with status code [manifests 1.19.6-3-gcc-al2]: 400 Bad Request
> ```
> 
> I will retry the task and merge this PR after the task finishes.

Yeah, this error message is typically a result of throttling pulling build image

Issue #2293, opened 2023-02-28
Error: "Failed to generate 10-awsconflist" error="json: cannot unmarshal number into Go struct field NetConf.plugins.mtu of type string"


tomsucho:
**What happened**:

On EKS 1.23 updated AWS VPC CNI addon to 1.12.2-eksbuild.1 as recommended in AWS docs 

**Attach logs**

the log from the aws-node pod says:
```
$ kubectl -n kube-system logs aws-node-jtlzf
Installed /host/opt/cni/bin/aws-cni
Installed /host/opt/cni/bin/egress-v4-cni
time="2023-02-28T13:56:09Z" level=info msg="Starting IPAM daemon... "
time="2023-02-28T13:56:09Z" level=info msg="Checking for IPAM connectivity... "
time="2023-02-28T13:56:10Z" level=info msg="Copying config file... "
time="2023-02-28T13:56:10Z" level=error msg="Failed to generate 10-awsconflist" error="json: cannot unmarshal number into Go struct field NetConf.plugins.mtu of type string"
```

**What you expected to happen**:
The aws-node pod to come up fine.

**How to reproduce it (as minimally and precisely as possible)**:
Tried to update AWS VPC CNI addon to latest as per the AWS docs. 

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.3", GitCommit:"c92036820499fedefec0f847e2054d824aea6cd1", GitTreeState:"clean", BuildDate:"2021-10-27T18:41:28Z", GoVersion:"go1.16.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"23+", GitVersion:"v1.23.16-eks-48e63af", GitCommit:"e6332a8a3feb9e0fe3db851878f88cb73d49dd7a", GitTreeState:"clean", BuildDate:"2023-01-24T19:18:15Z", GoVersion:"go1.19.5", Compiler:"gc", Platform:"linux/amd64"}
```
- CNI Version 
from: `amazon-k8s-cni:v1.11.4-eksbuild.1` to: `amazon-k8s-cni:v1.12.2-eksbuild.1`
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):

**UPDATE:** this only happens when this is set in cfg:
```
[...]

PR #2295, opened 2023-03-01
Fix MTU parameter in egress-v4-cni plugin

jdn5126:
**What type of PR is this?**
bug

#2293 

This PR updates the default `10-aws.conflist` so that the MTU set for `egress-v4-cni` plugin is a string rather than an integer. This was causing an issue when `ENABLE_BANDWIDTH_PLUGIN` environment variable was set, as we call `json.Unmarshal` on the default conflist and fail to parse `9001` as a string.

I added a unit test in `cmd/aws-vpc-cni/` to cover this specific case. I explicitly avoided using `gomock` for this test case to keep things simple.

N/A

Added a unit test to validate json marshaling of conflist.

**Automation added to e2e**:
N/A

No

No, Yes

No

Yes

```release-note
Fix conflist parsing when ENABLE_BANDWIDTH_PLUGIN is set.
```

### DIFF
--- a/cmd/aws-vpc-cni/main.go
+++ b/cmd/aws-vpc-cni/main.go
@@ -217,16 +217,9 @@ func isValidJSON(inFile string) error {
 	return json.Unmarshal([]byte(inFile), &result)
 }
 
-func generateJSON(jsonFile string, outFile string) error {
+func generateJSON(jsonFile string, outFile string, nodeIP string) error {
 	byteValue, err := ioutil.ReadFile(jsonFile)
 	if err != nil {
-		return err
-	}
-
-	var nodeIP string
-	nodeIP, err = getNodePrimaryV4Address()
-	if err != nil {
-		log.Errorf("Failed to get Node IP")
 		return err
 	}
 
@@ -391,8 +384,16 @@ func _main() int {
 	//	return 1
 	//}
 
+	// Get node IP for conflist
+	var nodeIP string
+	nodeIP, err = getNodePrimaryV4Address()
+	if err != nil {
+		log.Errorf("Failed to get Node IP, error: %v", err)
+		return 1
+	}
+
 	log.Infof("Copying config file... ")
-	err = generateJSON(defaultAWSconflistFile, tmpAWSconflistFile)
+	err = generateJSON(defaultAWSconflistFile, tmpAWSconflistFile, nodeIP)
 	if err != nil {
 		log.WithError(err).Errorf("Failed to generate 10-awsconflist")
 		return 1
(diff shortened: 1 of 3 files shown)


cmd/aws-vpc-cni/main.go
```diff
@@ -217,14 +217,14 @@ func isValidJSON(inFile string) error {
...
-func generateJSON(jsonFile string, outFile string) error {
+func generateJSON(jsonFile string, outFile string, getNodePrimaryAddress func() (string, error)) error {
 	byteValue, err := ioutil.ReadFile(jsonFile)
 	if err != nil {
 		return err
 	}
 
 	var nodeIP string
-	nodeIP, err = getNodePrimaryV4Address()
+	nodeIP, err = getNodePrimaryAddress()
```
orsenthil, line 227: Do we send anything other than `getNodePrimaryV4Address` function to this ?

jdn5126: No, I made the function a parameter only to simplify mock testing, and I dropped the V4 only to have a different name, and since Go does not support default function parameters.

orsenthil: Adding this as a parameter to `generateJSON` function is breaking the modularity too.
I assume this was done writing the test case.

The the best approach will be to introduce another function and keep the generateJSON to taking an "infile" and "outfile" - it just much to reason about while reading the code.

jdn5126: Yeah this was done for mocking. Not sure I follow the other function approach, can you expand? As long as `generateJSON` makes an imds call, it needs mocking

orsenthil: I missed seeing  `generateJSON` is calling `getNodePrimaryV4Address` even now.

- In that case, moving it parameter is fine for testing.

Should we break that call out of the generateJSON itself, and do the call before calling generateJSON and send node-ip if the call was successful?

```
	var nodeIP string
	nodeIP, err = getNodePrimaryV4Address()
	if err != nil {
		log.Errorf("Failed to get Node IP")
		return err
	}

    generateJSON(nodeIp string, sonFile string, outFile string)
```


Either this, or the the current approach of sending a function parameter is fine.
But former looks a bit more readable to me.

jdn5126: Yeah, that makes these more testable separately, so I will make that change

jdn5126: updated in latest commit

jayanthvn (approved):

PR #2297, opened 2023-03-01
Set Kubeconfig path for running kops tests 

jaydeokar:
**What type of PR is this?**
bug

N/A

The weekly test fails with variable undefined error. This explicitly sets the KUBECONFIG path and passes it to the test

N/A
N/A

**Automation added to e2e**:
N/A

N/A

N/A

N/A

N/A

### DIFF
--- a/.github/workflows/weekly-cron-tests.yaml
+++ b/.github/workflows/weekly-cron-tests.yaml
@@ -1,6 +1,7 @@
 name: Weekly Cron tests
 
 on:
+  workflow_dispatch: {}
   schedule:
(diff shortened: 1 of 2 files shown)


orsenthil:
Could you point to the full run with the fix?

jaydeokar:
> 

I cannot run actions on my forked repo since actions are not allowed on it. Let me know if there's a way to test this change. I had checked running locally with the previous change as well, but I had the variable set already in my environment variables which is why this one got missed

orsenthil (approved):
LGTM

PR #2304, opened 2023-03-06
Use GET for IAM Permissions event; update controller-runtime from 0.13.1 to 0.14.4 and client-go from v0.25.5 to v0.26.1

jdn5126:
**What type of PR is this?**
cleanup

N/A

This PR updates base go modules like client-go and helm in order to resolve outstanding dependabot alerts. It also migrates the "MissingIAMPermissions" event to use client GET instead of LIST.

This PR also updates the default CNI metrics helper tag used for integration tests to v1.12.5.

N/A

Manually verified that integration tests pass. Also kicking off GitHub action to run full integration tests.

**Automation added to e2e**:
N/A

No

No, Yes

No

Yes

```release-note
Update Go modules, use client GET for IAM permissions event
```


jdn5126:
GitHub PR integration tests: https://github.com/aws/amazon-vpc-cni-k8s/actions/runs/4347987317

pkg/utils/eventrecorder/eventrecorder.go
```diff
@@ -83,27 +77,12 @@ func New(rawK8SClient, cachedK8SClient client.Client) (*EventRecorder, error) {
...
-
-	// Get aws-node pod objects with label & field selectors
-	labelSelector := labels.SelectorFromSet(labels.Set{labelK8sapp: awsNode})
-	fieldSelector := fields.SelectorFromSet(fields.Set{specNodeName: MyNodeName})
-	listOptions := client.ListOptions{
-		LabelSelector: labelSelector,
-		FieldSelector: fieldSelector,
-	}
+// BroadcastPodEvent will raise event on aws-node with given type, reason, & message
+func (e *EventRecorder) BroadcastPodEvent(eventType, reason, message string) {
```
sushrk, line 81: nit: maybe we can rename as 'SendPodEvent' to be consistent.

go.mod
```diff
@@ -21,45 +21,45 @@ require (
...
-	k8s.io/cli-runtime v0.25.5
-	k8s.io/client-go v0.25.5
+	helm.sh/helm/v3 v3.11.1
+	k8s.io/api v0.26.1
+	k8s.io/apimachinery v0.26.1
+	k8s.io/cli-runtime v0.26.0
+	k8s.io/client-go v0.26.1
 	k8s.io/cri-api v0.23.1
-	sigs.k8s.io/controller-runtime v0.13.1
+	sigs.k8s.io/controller-runtime v0.14.4
```
sushrk, line 41: For reference, would be good to add which versions were updated manually in the PR description. This will help to differentiate with those updated automatically because of dependencies.

sushrk (changes requested):
Few minor comments, looks good overall. Thanks for fixing this.

pkg/utils/eventrecorder/eventrecorder.go
jdn5126: Good call, this sounds better to me

go.mod
jdn5126: good call, will do

sushrk (approved):
Thanks, lgtm!

PR #2305, opened 2023-03-07
Update readme for the use of privileged mode in init container

jaydeokar:
**What type of PR is this?**
documentation

N/A

Adds details about the requirement of using `privileged` mode in vpc-cni init containers

N/A

N/A

**Automation added to e2e**:
N/A

N/A

N/A

N/A

N/A


README.md
```diff
@@ -93,6 +93,11 @@ For example, a m4.4xlarge node can have up to 8 ENIs, and each ENI can have up t
 
 For a detailed explanation, see [`WARM_ENI_TARGET`, `WARM_IP_TARGET` and `MINIMUM_IP_TARGET`](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/docs/eni-and-ip-target.md).
 
+
+## Privileged mode
+
+VPC CNI makes use of privileged mode (`privileged: true`) in the manifest only for init container. This elevated privilege is required to set the networking kernel parameter `rp_filter` to `loose` mode.
```
jdn5126, line 99: It is also needed for setting IPV6 kernel parameters. I think we should just say something like:
```
VPC CNI makes use of privileged mode (`privileged: true`) in the manifest for its init container. This elevated privilege is required to set the networking kernel parameters. I would also move this 
```

jdn5126 (dismissed):

jdn5126 (approved):

PR #2307, opened 2023-03-07
Remove old checkpoint migration logic; update containerd version

jdn5126:
**What type of PR is this?**
cleanup

N/A

This PR removes old code from the now completed (as of v1.12.0) datastore checkpoint migration. The motivation for this cleanup is the deprecation of the `k8s.io/cri-api/pkg/apis/runtime/v1alpha2` package used by `pkg/cri/cri.go` and the desire to update containerd version > 1.5.x.

In order to move containerd version to v1.6.6, I had to pin `github.com/containernetworking/cni` and `github.com/containernetworking/plugins` to versions less than 1.0.0. The reason is that moving to versions >= 1.0.0 forces dependencies on CNI spec 1.0.0, which VPC CNI is not yet ready to move to.

This old code is fine to remove as we already have a callout in the README: https://github.com/aws/amazon-vpc-cni-k8s#container-runtime and already require version updates to be one major version at a time (i.e. 1.11.x -> 1.12.x).

N/A

Manually verified that all integration tests passed. Will kickoff integration test action against this job for further coverage.

**Automation added to e2e**:
N/A

No

No, Yes

No

Yes

```release-note
Update containerd version
```


orsenthil (approved):
LGTM

go.mod
```diff
@@ -165,6 +164,7 @@ replace github.com/emicklei/go-restful => github.com/emicklei/go-restful/v3 v3.8
 
 replace golang.org/x/crypto => golang.org/x/crypto v0.1.0
 
-// Containerd package 1.6.x has a dependency of containernetworking/cni > 1.0.0 and which also has a CNI change spec 1.0.0
-// To keep it backward compatible to the spec 0.4.0, we will pin it to v1.5.x
-replace github.com/containerd/containerd => github.com/containerd/containerd v1.5.18
+// Containernetworking packages must remain pinned until VPC CNI moves to CNI spec 1.0.0
+replace github.com/containernetworking/cni => github.com/containernetworking/cni v0.8.1
```
M00nF1sh, line 168: why we update `github.com/containernetworking/cni v1.1.1`

jdn5126: The callout for k8s 1.26 is to use a containerd version >= 1.6.0. Containerd is an indirect dependency for us, but it moving to v1.6.6 moves `containernetworking` packages past 1.0.0. We do not want `containernetworking` to move past 1.0.0, as that forces us to use CNI spec 1.0.0, which we cannot support on older EKS versions. Trying to use `types/040` from `containernetworking/cni` does not fully work as other packages use `types/100` and there is no way to cast forward

pkg/ipamd/datastore/data_store.go
```diff
@@ -1197,21 +1117,6 @@ func (ds *DataStore) UnassignPodIPAddress(ipamKey IPAMKey) (e *ENI, ip string, d
...
-		// migration code is finally removed.  Leaving a
-		// compile dependency here to make that obvious :P
-		var _ = checkpointMigrationPhase
-
-		// If the entry was discovered by querying CRI at
-		// restart rather than by observing an ADD operation
-		// directly, then we won't have captured the true
-		// networkname/ifname.
-		ds.log.Debugf("UnassignPodIPAddress: Failed to find IPAM entry under full key, trying CRI-migrated version")
-		ipamKey.NetworkName = backfillNetworkName
```
M00nF1sh, line 1211: this section cannot be removed. as customer upgrading from old CNI will still hit this branch. (since old pods is still have backfillNetworkName in their ipamd files)

jdn5126: I see, so this comment means that once we guarantee existing pods were not created by VPC CNI version running checkpoint < 2, we can remove this `if` block?

Will that ever be true? We do only support customers upgrading one major version at a time, i.e. in 1.13, we will guarantee that we are upgrading from minimum 1.12, but pods could have been created before that upgrade.

M00nF1sh: that will not be true. I don't know why the if statement were there(it's too optimistic :D) 

as long as we support the upgrade, there will always have old ipamd entries in the past with backfillNetworkName

jdn5126: Got it, I'll just remove the comment then

pkg/ipamd/datastore/data_store.go
```diff
@@ -1198,24 +1122,15 @@ func (ds *DataStore) UnassignPodIPAddress(ipamKey IPAMKey) (e *ENI, ip string, d
...
-		// restart rather than by observing an ADD operation
-		// directly, then we won't have captured the true
-		// networkname/ifname.
+		// If the entry is not present in state file, check if it is present under placeholder value.
+		// This scenario could happen if the pod was created by an older CNI version back when CRI read was done.
 		ds.log.Debugf("UnassignPodIPAddress: Failed to find IPAM entry under full key, trying CRI-migrated version")
 		ipamKey.NetworkName = backfillNetworkName
 		ipamKey.IfName = backfillNetworkIface
 		eni, availableCidr, addr = ds.eniPool.FindAddressForSandbox(ipamKey)
-	}
```
M00nF1sh, line 1214: why is this if removed..

jdn5126: The if condition below had the exact same check, so we don't need two if blocks, just one

jdn5126: Ah sorry, fixed now

M00nF1sh (approved):
/approve
/lgtm

PR #2308, opened 2023-03-08
Fix Calico integration tests

jdn5126:
**What type of PR is this?**
testing bug

N/A

This PR fixes the calico integration test suite, which was failing due to changes in installation requirements in calico 1.23+. Before installing the calico chart from helm, we must create the `tigera-operator` namespace. Also, the chart URL changed.

N/A

Manually verified that calico integration tests pass after this change. Note that because calico-system resources are not cleaned up between test runs, the test MUST be run last. Subsequent test runs need to pass `--install-calico=false`.

**Automation added to e2e**:
N/A

No

No, Yes

No

No

```release-note

```


jaydeokar (approved):
/lgtm
Have you checked it doesn't leak any resources post deletion of cluster ?

jdn5126:
> /lgtm Have you checked it doesn't leak any resources post deletion of cluster ?

Yeah, everything created by tigera-operator is within the cluster (daemonsets, CRDs, deployments, etc). No additional resources are provisioned in cloud provider account

PR #2322, opened 2023-03-17
Run cni release test for deveks

jaydeokar:
**What type of PR is this?**
feature

N/A

To get cluster information during integration tests on deveks cluster

N/A

Yes

**Automation added to e2e**:
N/A

N/A

N/A

N/A

N/A

### DIFF
--- a/scripts/lib/cluster.sh
+++ b/scripts/lib/cluster.sh
@@ -7,6 +7,14 @@ function load_cluster_details() {
   K8S_VERSION=$(echo "$DESCRIBE_CLUSTER_OP" | jq .cluster.version -r)
 }
 
+function load_deveks_cluster_details() {
+
+  echo "loading cluster details $CLUSTER_NAME"
+  PROVIDER_ID=$(kubectl get nodes --kubeconfig $KUBE_CONFIG_PATH -ojson | jq -r '.items[0].spec.providerID')
+  INSTANCE_ID=${PROVIDER_ID##*/}
+  VPC_ID=$(aws ec2 describe-instances --instance-ids ${INSTANCE_ID} --no-cli-pager | jq -r '.Reservations[].Instances[].VpcId')
+}
+
 function down-test-cluster() {
     echo -n "Deleting cluster  (this may take ~10 mins) ... "
     eksctl delete cluster $CLUSTER_NAME >>$CLUSTER_MANAGE_LOG_PATH 2>&1 ||
--- a/scripts/run-cni-release-tests.sh
+++ b/scripts/run-cni-release-tests.sh
@@ -9,6 +9,7 @@
 # KUBE_CONFIG_PATH: path to the kubeconfig file, default ~/.kube/config
 # NG_LABEL_KEY: nodegroup label key, default "kubernetes.io/os"
 # NG_LABEL_VAL: nodegroup label val, default "linux"
+# RUN_DEVEKS_TEST: Set this variable for tests to run on a deveks cluster
 # CNI_METRICS_HELPER: cni metrics helper image tag, default "602401143452.dkr.ecr.us-west-2.amazonaws.com/cni-metrics-helper:v1.12.6"
 
 set -e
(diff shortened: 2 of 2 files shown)


scripts/run-cni-release-tests.sh
```diff
@@ -29,8 +30,8 @@ function run_integration_test() {
   echo "cni test took $((SECONDS - START)) seconds."
   echo "Running ipamd integration tests"
   START=$SECONDS
-  # NOTE: skipping ipamd_event_test.go until it can be triaged further
-  cd $INTEGRATION_TEST_DIR/ipamd && CGO_ENABLED=0 ginkgo $EXTRA_GINKGO_FLAGS --skip-file=ipamd_event_test.go -v -timeout 90m --no-color --fail-on-pending -- --cluster-kubeconfig="$KUBE_CONFIG_PATH" --cluster-name="$CLUSTER_NAME" --aws-region="$REGION" --aws-vpc-id="$VPC_ID" --ng-name-label-key="$NG_LABEL_KEY" --ng-name-label-val="$NG_LABEL_VAL" || TEST_RESULT=fail
+
+  cd $INTEGRATION_TEST_DIR/ipamd && CGO_ENABLED=0 ginkgo $EXTRA_GINKGO_FLAGS -v -timeout 90m --no-color --fail-on-pending -- --cluster-kubeconfig="$KUBE_CONFIG_PATH" --cluster-name="$CLUSTER_NAME" --aws-region="$REGION" --aws-vpc-id="$VPC_ID" --ng-name-label-key="$NG_LABEL_KEY" --ng-name-label-val="$NG_LABEL_VAL" || TEST_RESULT=fail
```
jdn5126, line 34: Why is this being removed? I think `ipamd_event_test` is still flaky

jaydeokar: Oh I thought we fixed it.. I can revert this

test/framework/resources/k8s/resources/deployment.go
```diff
@@ -60,6 +60,11 @@ func (d *defaultDeploymentManager) CreateAndWaitTillDeploymentIsReady(deployment
 func (d *defaultDeploymentManager) DeleteAndWaitTillDeploymentIsDeleted(deployment *v1.Deployment) error {
 	ctx := context.Background()
 	err := d.k8sClient.Delete(ctx, deployment)
+
+	if errors.IsNotFound(err) {
```
jdn5126, line 64: why this change? This looks like it will skip the normal return below

jaydeokar: So I was trying to make sure we end up cleaning all the resources we create during a test and not stop the cleanup just because of resource doesn't exist error. 
Say if a deployment doesn't exist ( eg: we got an error before the deployment was even created), we shouldn't error out during the cleanup process. 
Specifically, in the After each step, we check the status of delete deployment call and if it returns an not found error, the rest of the cleanup get's missed. (eg - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/test/integration/cni/pod_traffic_test.go#L150)

jdn5126: I see. So I think the function should still return an error, but we should do all of the cleanup regardless of whether an earlier step returns an error, similar to what we do here: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/test/integration/custom-networking/custom_networking_suite_test.go#L175

Basically, we append errors and check them at the end

scripts/lib/cluster.sh
```diff
@@ -7,6 +7,14 @@ function load_cluster_details() {
   K8S_VERSION=$(echo "$DESCRIBE_CLUSTER_OP" | jq .cluster.version -r)
 }
 
+function load_deveks_cluster_details() {
+
+  echo "loading cluster details $CLUSTER_NAME"
+  PROVIDER_ID=$(kubectl get nodes --kubeconfig $KUBE_CONFIG_PATH -ojson | jq -r '.items[0].spec.providerID')
+  INSTANCE_ID=${PROVIDER_ID##*/}
```
orsenthil, line 14: Why do we have `##*/` here? Perhaps for a comment?

jaydeokar: It's a way to fetch the instance id from the string provider id. It removes all the characters upto the last '/'
aws:///us-west-2a/i-xxxxxxxxxxxxx -> i-xxxxxxxxxxxxx

jdn5126 (approved):

orsenthil (commented):
LGTM

Issue #2321, opened 2023-03-17
Non-SGPP pods created in v1.12.1+ cannot be deleted in v1.11.4


jdn5126:
**What happened**:
Due to https://github.com/aws/amazon-vpc-cni-k8s/pull/2145, pods created in v1.12.1+ create a PrevResult with a VLAN ID of 0. In v1.11.4 (really anything < v1.12.1), in `tryDelWithPrevResult`, if PrevResult exists and has a VLAN ID of 0, we return error and do not delete.

**Attach logs**
```
error killing pod: failed to "KillPodSandbox" for "8d5583d3-535d-4106-b3be-d42a6d735056" with KillPodSandboxError: "rpc error: code = Unknown desc = networkPlugin cni failed to teardown pod \"test-app-5dc4bf8b96-j6rx5_default\" network: del cmd: failed to delete with prevResult: malformed vlanID in prevResult: 0"
```

**What you expected to happen**:
Pod to be deleted

**How to reproduce it (as minimally and precisely as possible)**:
1. Create pod with VPC CNI v1.12.1+
2. Downgrade VPC CNI to v1.11.4
3. Delete pod

**Anything else we need to know?**:
No

**Environment**:
- Kubernetes version (use `kubectl version`): all
- CNI Version: v1.11.4
- OS (e.g: `cat /etc/os-release`): AL2
- Kernel (e.g. `uname -a`): Linux 5.4

jdn5126:
Closing as fixed by #2323 . Fix will ship in v1.11.5

PR #2323, opened 2023-03-17
Handle pod deletion when PrevResult has VLAN 0

jdn5126:
**What type of PR is this?**
bug

#2321 

This PR handles the case where a pod being deleted has a PrevResult entry with a VLAN of 0. This fix is needed in the 1.11 release train so that if a customer creates a pod with v1.12.1+ and then decides they need to downgrade to 1.11, the pod can be deleted without issue. 

N/A

Manually created pods with v1.12.1+, then downgraded to an image with this change and verified that pods could be deleted. Also verified that integration tests pass.

**Automation added to e2e**:
N/A

N/A

No, Yes

No

Yes

```release-note
Handle pod deletion when PrevResult has VLAN 0 in 1.11 release train
```

### DIFF
--- a/cmd/routed-eni-cni-plugin/cni.go
+++ b/cmd/routed-eni-cni-plugin/cni.go
@@ -431,8 +431,12 @@ func tryDelWithPrevResult(driverClient driver.NetworkAPIs, conf *NetConf, k8sArg
 		return false, nil
 	}
 	podVlanID, err := strconv.Atoi(dummyIface.Mac)
-	if err != nil || podVlanID == 0 {
+	if err != nil {
 		return true, errors.Errorf("malformed vlanID in prevResult: %s", dummyIface.Mac)
+	}
+	// If VLAN value is 0, return without handling so that normal delete logic can occur.
+	if podVlanID == 0 {
+		return false, nil
 	}
 	if isNetnsEmpty(netNS) {
 		log.Infof("Ignoring TeardownPodENI as Netns is empty for SG pod:%s namespace: %s containerID:%s", k8sArgs.K8S_POD_NAME, k8sArgs.K8S_POD_NAMESPACE, k8sArgs.K8S_POD_INFRA_CONTAINER_ID)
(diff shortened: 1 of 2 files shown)


cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -431,9 +431,13 @@ func tryDelWithPrevResult(driverClient driver.NetworkAPIs, conf *NetConf, k8sArg
 		return false, nil
 	}
 	podVlanID, err := strconv.Atoi(dummyIface.Mac)
-	if err != nil || podVlanID == 0 {
+	if err != nil {
 		return true, errors.Errorf("malformed vlanID in prevResult: %s", dummyIface.Mac)
 	}
+	// If VLAN value is 0, return without handling so that normal delete logic can occur.
```
jayanthvn, line 437: Can we make `0` a constant something like `reservedVlanId` or anything.. and additionally (might be there already) if during add, for whatever reason vlan id is 0 for SGPP , we need to bail out.

jdn5126: Yeah, ee already guard against VLAN ID of 0 during add in older releases: https://github.com/aws/amazon-vpc-cni-k8s/blob/release-1.11/cmd/routed-eni-cni-plugin/cni.go#L217 . As for making this a constant, I think that might fit better in the master branch: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/cmd/routed-eni-cni-plugin/cni.go#L448 . I am a bit wary of changing this older release branch more than just to fix this bug

achevuru (approved):

cmd/routed-eni-cni-plugin/cni.go
jayanthvn: Yeah we can change on both since it will be just a const..

jayanthvn:
lgtm

cmd/routed-eni-cni-plugin/cni.go
jdn5126: Added in latest revision

jayanthvn (approved):

PR #2325, opened 2023-03-20
Add generic script to run integration tests

jdn5126:
**What type of PR is this?**
testing

N/A

This PR adds a generic script that can be used to run any integration test suite by setting the `SUITE_NAME` environment variable. This is useful for kicking off tests in prow environment.

N/A

Manually verified that script runs as expected

**Automation added to e2e**:
N/A

No

N/A

No

No

```release-note

```


scripts/run-ginkgo-integration-suite.sh
```diff
@@ -0,0 +1,29 @@
...
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+GINKGO_TEST_BUILD="$SCRIPT_DIR/../test/build"
+SKIP_MAKE_TEST_BINARIES=1
+
+source "$SCRIPT_DIR"/lib/add-on.sh
+source "$SCRIPT_DIR"/lib/cluster.sh
+source "$SCRIPT_DIR"/lib/canary.sh
+
+function run_ginkgo_test() {
+  (CGO_ENABLED=0 ginkgo $EXTRA_GINKGO_FLAGS -v --timeout 30m --no-color --fail-on-pending $GINKGO_TEST_BUILD/$SUITE_NAME.test -- --cluster-kubeconfig="$KUBE_CONFIG_PATH" --cluster-name="$CLUSTER_NAME" --aws-region="$REGION" --aws-vpc-id="$VPC_ID" --ng-name-label-key="kubernetes.io/os" --ng-name-label-val="linux")
```
orsenthil, line 18: I know we use `KUBE_CONFIG_PATH`. but at some point in time, we should change this to `KUBECONFIG` which is more standard convention.

scripts/run-ginkgo-integration-suite.sh
```diff
@@ -0,0 +1,29 @@
...
+}
+
+load_cluster_details
+load_addon_details
+
+echo "Running $SUITE_NAME integration tests against the latest VPC CNI addon version"
+
+install_add_on "$LATEST_ADDON_VERSION"
```
orsenthil, line 26: *important* - We can leave out add-on installation from this script. Let the script's focus only be running the tests. Doing add-on installation was conflating the responsibilities (even earlier) So we can remove this.

scripts/run-ginkgo-integration-suite.sh
jdn5126: yeah, it would be good to set both at the caller and then transition away from `KUBE_CONFIG_PATH`

scripts/run-ginkgo-integration-suite.sh
jdn5126: Yep, leaving it to the one creating/configuring the cluster makes sense to me. Removed in latest revision.

orsenthil (approved):
LGTM.

PR #2330, opened 2023-03-27
Update integration test resources to use termination grace period of 0 by default

jdn5126:
**What type of PR is this?**
test cleanup

N/A

This PR updates k8s resource creation in CNI tests to use a default termination grace period of 1 second. A value of 0 is special in that the resource is immediately removed on the API server instead of waiting for kubelet deletion to complete. This can lead to resource leakage, which can cause later tests to fail. Ways to handle resource leakage is being addressed separately. 

N/A

Manually verified that CNI and IPAMD full integration tests pass after this change (except for ipamd event test, which needs to be refactored).

**Automation added to e2e**:
N/A

No

N/A

No

No

```release-note

```


orsenthil (approved):
Interesting change. Lets see if this improves the reliability in our tests.

Issue #2311, opened 2023-03-11
Make all the boolean values case insensitive in aws vpc cni environmental variables. 


vtgspk:
**What would you like to be added**:
ENABLE_IPV6=FALSE is enabling the EGRESSV4PLUGIN and as a result an additional interface is added to the EKS pod. Changing from "FALSE" to "false" is fixing the issue. I believe this is because of this [code] I suppose (https://github.com/aws/amazon-vpc-cni-k8s/blob/master/cmd/egress-v4-cni-plugin/cni.go#L360).

So please make the boolean values of ENV variables in aws vpc cni  wherever applicable as case insensitive. 
**Why is this needed**:
Good coding standards to have case insensitive values for Boolean.  Had to spend 2+ weeks to figure out this case sensitivity was causing the problem.

jdn5126:
Thanks for the callout @vtgspk, this is definitely something we want to handle in future releases. Will work on getting this assigned. Also marking as a good first issue in case anyone from the community wants to take it before that

PR #2334, opened 2023-03-29
Make all the aws vpc cni environmental variables case insensitive

jerryhe1999:
**What type of PR is this?**
feature

#2311 

Originally, we have several aws-vpc-cni environmental variables with the type Boolean as a String which against good coding standards. Having case insensitive values for Boolean can avoid users meet situation that spend 2+ weeks to figure out the case sensitivity cause the problem.

Manually tested the code change in testing cluster with setting environmental variables in case insensitivity. Both unit tests and integration tests passed

**Automation added to e2e**:
N/A

No

No

No

No

### DIFF
--- a/cmd/aws-vpc-cni-init/main.go
+++ b/cmd/aws-vpc-cni-init/main.go
@@ -18,6 +18,7 @@ import (
 	"os"
 
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/procsyswrapper"
+	"github.com/aws/amazon-vpc-cni-k8s/utils"
 	"github.com/aws/amazon-vpc-cni-k8s/utils/cp"
 	"github.com/aws/amazon-vpc-cni-k8s/utils/imds"
 	"github.com/pkg/errors"
@@ -26,22 +27,17 @@ import (
 )
 
 const (
-	defaultHostCNIBinPath = "/host/opt/cni/bin"
-	vpcCniInitDonePath    = "/vpc-cni-init/done"
-	metadataLocalIP       = "local-ipv4"
-	metadataMAC           = "mac"
+	defaultHostCNIBinPath           = "/host/opt/cni/bin"
+	vpcCniInitDonePath              = "/vpc-cni-init/done"
+	metadataLocalIP                 = "local-ipv4"
+	metadataMAC                     = "mac"
+	defaultDisableIPv4TcpEarlyDemux = false
+	defaultEnableIPv6               = false
 
 	envDisableIPv4TcpEarlyDemux = "DISABLE_TCP_EARLY_DEMUX"
 	envEnableIPv6               = "ENABLE_IPv6"
 	envHostCniBinPath           = "HOST_CNI_BIN_PATH"
 )
-
-func getEnv(env, defaultVal string) string {
-	if val, ok := os.LookupEnv(env); ok {
-		return val
-	}
-	return defaultVal
-}
 
 func getNodePrimaryIF() (string, error) {
 	var primaryIF string
@@ -83,8 +79,8 @@ func configureSystemParams(procSys procsyswrapper.ProcSys, primaryIF string) err
 	// Note that older kernels may not support tcp_early_demux, so we must first check that it exists.
 	entry = "net/ipv4/tcp_early_demux"
 	if _, err := procSys.Get(entry); err == nil {
-		disableIPv4EarlyDemux := getEnv(envDisableIPv4TcpEarlyDemux, "false")
-		if disableIPv4EarlyDemux == "true" {
+		disableIPv4EarlyDemux := utils.GetBoolAsStringEnvVar(envDisableIPv4TcpEarlyDemux, defaultDisableIPv4TcpEarlyDemux)
+		if disableIPv4EarlyDemux {
 			err = procSys.Set(entry, "0")
 			if err != nil {
 				return errors.Wrap(err, "Failed to disable tcp_early_demux")
@@ -105,8 +101,8 @@ func configureIPv6Settings(procSys procsyswrapper.ProcSys, primaryIF string) err
 	var err error
 	// Enable IPv6 when environment variable is set
 	// Note that IPv6 is not disabled when environment variable is unset. This is omitted to preserve default host semantics.
-	enableIPv6 := getEnv(envEnableIPv6, "false")
-	if enableIPv6 == "true" {
+	enableIPv6 := utils.GetBoolAsStringEnvVar(envEnableIPv6, defaultEnableIPv6)
+	if enableIPv6 {
 		entry := "net/ipv6/conf/all/disable_ipv6"
 		err = procSys.Set(entry, "0")
 		if err != nil {
@@ -150,7 +146,7 @@ func _main() int {
 	}
 
 	log.Infof("Copying CNI plugin binaries ...")
-	hostCNIBinPath := getEnv(envHostCniBinPath, defaultHostCNIBinPath)
+	hostCNIBinPath := utils.GetEnv(envHostCniBinPath, defaultHostCNIBinPath)
 	err = cp.InstallBinaries(pluginBins, hostCNIBinPath)
 	if err != nil {
 		log.WithError(err).Errorf("Failed to install binaries")
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -1,0 +1,30 @@
+package utils
+
+import (
+	"os"
+	"strconv"
+
+	log "github.com/sirupsen/logrus"
+)
+
+// Parse environment variable and return boolean representation of string, or default value if environment variable is unset
+func GetBoolAsStringEnvVar(env string, defaultVal bool) bool {
+	if val, ok := os.LookupEnv(env); ok {
+		parsedVal, err := strconv.ParseBool(val)
+		if err == nil {
+			return parsedVal
+		}
+		log.Errorf("Failed to parse variable %s with value %s as boolean", env, val)
+		return defaultVal
+	}
+	// Environment variable is not set, so return default value
+	return defaultVal
+}
+
+// If environment variable is set, return set value, otherwise return default value
+func GetEnv(env, defaultVal string) string {
+	if val, ok := os.LookupEnv(env); ok {
+		return val
+	}
+	return defaultVal
+}
--- a/utils/utils_test.go
+++ b/utils/utils_test.go
@@ -1,0 +1,47 @@
+package utils
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	defaultPathEnv = "/host/opt/cni/bin"
+	defaultBoolEnv = false
+
+	envPath = "Path"
+	envBool = "Bool"
+)
+
+// Validate that GetBoolAsStringEnvVar runs against acceptable format input without error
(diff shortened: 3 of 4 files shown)


jdn5126 (changes requested):
This helps environment variables read in `cmd/aws-vpc-cni/main.go`, but we need to be more thorough and do case-insensitive compare everywhere a boolean environment variable is accessed.

Also, this same function is in `cmd/aws-vpc-cni-init/main.go`

You'll need to go through each "Boolean as a String" environment variable in https://github.com/aws/amazon-vpc-cni-k8s#cni-configuration-variables and ensure that everywhere it is read in the code does a case-insensitive compare.

cmd/aws-vpc-cni-init/main.go
```diff
@@ -16,10 +16,12 @@ package main
 
 import (
 	"os"
+	"strings"
```
jdn5126, line 19: this import should be removed

utils/getenv/getenv.go
```diff
@@ -0,0 +1,18 @@
+package getenv
```
jdn5126, line 1: nit: let's choose a different name to make this more extensible in the future. You can just create a file called `utils.go` in the `utils` directory and then import it with `"github.com/aws/amazon-vpc-cni-k8s/utils"`

utils/getenv/getenv.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"os"
+	"strconv"
+	"strings"
+)
+
+func parseEnvVar(env, defaultVal string) string {
+	if val, ok := os.LookupEnv(env); ok {
+		_, err := strconv.ParseBool(val)
+		if err == nil {
```
jdn5126, line 12: This will not work in all cases. You can use a separate function for "boolean as a string" to make it easier. Also, you will need to add unit tests for this in a file called `utils_test.go`

jdn5126 (changes requested):

utils/utils.go
```diff
@@ -0,0 +1,27 @@
+package utils
+
+import (
+	"os"
+	"strconv"
+	"strings"
+)
+
+// Retrieves the specified environmental variable with type of Boolean as string, set it with defualt variable if not exist or unacceptable format
+func getBoolAsStringEnvVar(env, defaultVal string) string {
```
jdn5126, line 10: this function should just return a boolean. No need to cast it back to a string after we already have it in boolean format

utils/utils.go
```diff
@@ -0,0 +1,27 @@
+package utils
+
+import (
+	"os"
+	"strconv"
+	"strings"
+)
+
+// Retrieves the specified environmental variable with type of Boolean as string, set it with defualt variable if not exist or unacceptable format
```
jdn5126, line 9: nit: maybe "Parse environment variable and return boolean representation of string, or default value if environment variable is unset"

utils/utils_test.go
```diff
@@ -0,0 +1,58 @@
...
+const (
+	defaultHostCNIBinPath = "/host/opt/cni/bin"
+	defaultVethPrefix     = "eni"
+	defaultMTU            = "9001"
+	defaultEnablePodEni   = "false"
+
+	envHostCniBinPath = "HOST_CNI_BIN_PATH"
+	envVethPrefix     = "AWS_VPC_K8S_CNI_VETHPREFIX"
+	envEniMTU         = "AWS_VPC_ENI_MTU"
+	envEnablePodEni   = "ENABLE_POD_ENI"
```
jdn5126, line 18: You should just use a fake environment variable and default value for it here. One should be enough for testing, and a fake value keeps it simple

jdn5126 (changes requested):

cmd/aws-vpc-cni-init/main.go
```diff
@@ -83,8 +79,11 @@ func configureSystemParams(procSys procsyswrapper.ProcSys, primaryIF string) err
 	// Note that older kernels may not support tcp_early_demux, so we must first check that it exists.
 	entry = "net/ipv4/tcp_early_demux"
 	if _, err := procSys.Get(entry); err == nil {
-		disableIPv4EarlyDemux := getEnv(envDisableIPv4TcpEarlyDemux, "false")
-		if disableIPv4EarlyDemux == "true" {
+		disableIPv4EarlyDemux, boolenv := utils.GetBoolAsStringEnvVar(envDisableIPv4TcpEarlyDemux, defaultDisableIPv4TcpEarlyDemux)
+		if boolenv != err {
+			log.WithError(boolenv).Errorf("Failed with incorrect environmental value")
+		}
```
jdn5126, line 85: If the function returns an error, then the variable name should be `err`, not `boolenv`. The comparison at line 83 does not make any sense. Also, did you try passing the log to the function so that it does not have to return an error?

Whatever log you add should say something like: `"Failed to parse variable %s with value %s as boolean", envDisableIPv4TcpEarlyDemux, value`, which is why I think the logging makes more sense in the function

jdn5126: Also, anywhere "environmental" is used needs to be changed to "environment"

utils/utils.go
```diff
@@ -0,0 +1,29 @@
+package utils
+
+import (
+	"fmt"
+	"os"
+	"strconv"
+)
+
+// Parse environment variable and return boolean representation of string, or default value if environment variable is unset
+func GetBoolAsStringEnvVar(env string, defaultVal bool) (bool, error) {
```
jdn5126, line 10: nit: as mentioned above, after thinking about this more, I think the caller code will be much cleaner if you pass the logger as an argument and log here when `ParseBool` returns an error

utils/utils.go
```diff
@@ -0,0 +1,29 @@
...
+	if val, ok := os.LookupEnv(env); ok {
+		parsedVal, err := strconv.ParseBool(val)
+		if err == nil {
+			return parsedVal, nil
+		}
+		// Environment flag variable exist but with unacceptable format
```
jdn5126, line 16: nit: this comment is unnecessary

utils/utils.go
```diff
@@ -0,0 +1,29 @@
...
+		return defaultVal, fmt.Errorf("Environmental flag set but with unacceptable format: %s", err.Error())
+	}
+	// Environment flag variable not exist
```
jdn5126, line 19: nit: "Environment variable is not set, so return default value"

utils/utils.go
```diff
@@ -0,0 +1,29 @@
...
+	return defaultVal, nil
+}
+
+// Retrieves the value of specified environment variable, set it with defualt variable if not exist
```
jdn5126, line 23: nit: "If environment variable is set, return set value, otherwise return default value"

utils/utils_test.go
```diff
@@ -0,0 +1,48 @@
...
+	tmp, err = GetBoolAsStringEnvVar(envBool, defaultBoolEnv)
+	assert.NoError(t, err)
+	assert.Equal(t, tmp, true)
+
+	// Test basic Boolean as string set with unacceptable format
+	_ = os.Setenv("Bool", "TrUe")
+	tmp, err = GetBoolAsStringEnvVar(envBool, defaultBoolEnv)
+	assert.Error(t, err)
+	assert.Equal(t, tmp, false)
+}
```
jdn5126, line 36: At the end of this function, you should unset the environment variables that you set using. After the set calls, you can add `defer os.Unsetenv(envBool)`

utils/utils_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+// Validate that GetBoolAsStringEnvVar runs against acceptable format input without error
+func TestGetBoolAsStringEnvVar(t *testing.T) {
+	// Test environmental flag variable not set
+	tmp, err := GetBoolAsStringEnvVar("Bool", defaultBoolEnv)
+	assert.NoError(t, err)
+	assert.Equal(t, tmp, defaultBoolEnv)
+
+	// Test basic Boolean as string set with acceptable format
+	_ = os.Setenv("Bool", "True")
```
jdn5126, line 26: Let's use the global variable you defined above

utils/utils_test.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// Validate that GetEnv runs without error against environmental variable with type other than boolean as string
+func TestGetEnv(t *testing.T) {
+	// Test environmental flag variable not set
+	tmp := GetEnv("Path", defaultPathEnv)
+	assert.Equal(t, tmp, defaultPathEnv)
+
+	// Test environmental flag variable set
+	_ = os.Setenv("Path", "/host/opt/cni/bin/test")
```
jdn5126, line 45: As mentioned above, use the global variable you defined and unset the environment variable after the function runs

jdn5126 (changes requested):

utils/utils.go
```diff
@@ -0,0 +1,30 @@
...
+func GetBoolAsStringEnvVar(env string, defaultVal bool) bool {
+	if val, ok := os.LookupEnv(env); ok {
+		parsedVal, err := strconv.ParseBool(val)
+		if err == nil {
+			return parsedVal
+		}
+		log.Errorf("Failed to parse variable %s with value %s as boolean", env, val)
+		return defaultVal
+	}
+	log.Errorf("Environment variable %s is not set, so return default value", env)
```
jdn5126, line 20: Let's remove this log, as it will be too noisy. You can just add a comment here

utils/utils_test.go
```diff
@@ -0,0 +1,47 @@
...
+	defaultBoolEnv = false
+
+	envPath = "Path"
+	envBool = "Bool"
+)
+
+// Validate that GetBoolAsStringEnvVar runs against acceptable format input without error
+func TestGetBoolAsStringEnvVar(t *testing.T) {
+	// Test environment flag variable not set
+	tmp := GetBoolAsStringEnvVar("Bool", defaultBoolEnv)
```
jdn5126, line 21: nit: use `envBool` here

utils/utils_test.go
```diff
@@ -0,0 +1,47 @@
...
+	assert.Equal(t, tmp, defaultBoolEnv)
+
+	// Test basic Boolean as string set with acceptable format
+	_ = os.Setenv(envBool, "True")
+	tmp = GetBoolAsStringEnvVar(envBool, defaultBoolEnv)
+	assert.Equal(t, tmp, true)
+
+	// Test basic Boolean as string set with unacceptable format
+	_ = os.Setenv("Bool", "TrUe")
```
jdn5126, line 30: nit: use `envBool` here

utils/utils_test.go
```diff
@@ -0,0 +1,47 @@
...
+	defer os.Unsetenv(envBool)
+	tmp = GetBoolAsStringEnvVar(envBool, defaultBoolEnv)
+	assert.Equal(t, tmp, false)
+}
+
+// Validate that GetEnv runs without error against environment variable with type other than boolean as string
+func TestGetEnv(t *testing.T) {
+	// Test environment flag variable not set
+	tmp := GetEnv("Path", defaultPathEnv)
```
jdn5126, line 39: nit: use `envPath` here

jdn5126 (commented):
Just a few more nits and then this should be good to go. @jerryhe1999 can you also rebase and squash this into one commit?

jdn5126 (approved):

Issue #1925, opened 2022-03-15
IPv6 egress support for IPv4 EKS clusters


asheldon:
**What would you like to be added**:

The ability to egress from an IPv4-only pod to an IPv6-only endpoint

**Why is this needed**:

When a resource is available only via IPv6, it cannot be accessed from an IPv4-only EKS cluster, except via host networking.

achevuru:
[asheldon](https://github.com/asheldon) We will look in to it and get back to you. However, interested in understanding the use case a bit more. Are you more interested in reaching external IPv6 only services (or) IPv6 services hosted on different clusters with in the same VPC?

Have you considered using a [dual stack NLB/ALB](https://aws.amazon.com/about-aws/whats-new/2021/11/application-load-balancer-network-load-balancer-end-to-end-ipv6-support/) for this purpose? (i.e.,) let the dual stack NLB handle the IPv4 to IPv6 transition between the client and the server. AWS ALB controller now supports IPv6 targets as well (IP Mode).

asheldon:
@achevuru Thanks for the reply. Snapchat runs a [service mesh](https://eng.snap.com/monolith-to-multicloud-microservices-snap-service-mesh) based on [Envoy](https://envoyproxy.io/), so for typical communication between services, we do not use traditional load balancers. This is especially true in scenarios where we use routing techniques, like ring hashing, that don't treat every server in a fleet as stateless and interchangeable. Dual-stack NLB/ALB is a great technology, but it doesn't meet all our needs.

Reaching IPv6 resources within the VPC is more important than reaching IPv6 resources outside the VPC, but we'd like to be able to do both.

asheldon:
There may be some relationship with #956

achevuru:
[asheldon](https://github.com/asheldon) We're exploring a solution that is similar to our current Egress IPv4 support for IPv6 pods. Will provide an update in few weeks.

jdn5126:
Closing as complete by https://github.com/aws/amazon-vpc-cni-k8s/pull/2361. Feature will ship in v1.13.0 release

PR #2339, opened 2023-04-04
Enable IPv6 egress traffic IPv4 eks cluster

wanyufe:
**What type of PR is this?**
feature

#1925 

IPv6 egress support for IPv4 EKS cluster

Manual testing - unit testing has not finished
This is a draft PR request, a basic function to meet the requirements is included. 
But code needs to be refactored
1. Put some util function code to pkg/utils/cni 
2. combine egress-v4-cni and egres-v6-cni into one binary

```
amazon-vpc-cni-k8s git:(main) ✗ k exec -it multitool-4 -- bash                     
bash-5.1# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
    inet6 ::1/128 scope host 
3: eth0@if5: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9001 qdisc noqueue state UP group default 
    link/ether b6:c6:db:06:8c:ec brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 192.168.31.136/32 scope global eth0
    inet6 fd00::2/8 scope global 
    inet6 fe80::b4c6:dbff:fe06:8cec/64 scope link 
bash-5.1# ip -6 r
fd00::/8 dev eth0 proto kernel metric 256 pref medium
fd00::/8 dev eth0 metric 1024 pref medium
fe80::/64 dev eth0 proto kernel metric 256 pref medium
default via fe80::eccb:1aff:fe3d:c14c dev eth0 metric 1024 pref medium
bash-5.1# ping6 2600::
PING 2600::(2600::) 56 data bytes
64 bytes from 2600::: icmp_seq=1 ttl=43 time=36.2 ms
64 bytes from 2600::: icmp_seq=2 ttl=43 time=36.2 ms
64 bytes from 2600::: icmp_seq=3 ttl=43 time=36.1 ms
64 bytes from 2600::: icmp_seq=4 ttl=43 time=36.2 ms
64 bytes from 2600::: icmp_seq=5 ttl=43 time=36.1 ms
64 bytes from 2600::: icmp_seq=6 ttl=43 time=36.2 ms
```

**Automation added to e2e**:

No

No

no

```release-note

```


cmd/egress-v6-cni-plugin/cni.go
```diff
@@ -0,0 +1,608 @@
...
+
+	// delete container IPv6 route
+	err = tearDownContainerIPv6Route(netns, &containerVethIPv6, args.IfName)
+	if err != nil {
+		log.Debugf("tear down container IPv6 route failed: %v", err)
+		return err
+	}
+	log.Debugf("container %s IPv6 route deleted", args.Netns)
+
+	if err := ipam.ExecDel(netConf.IPAM.Type, args.StdinData); err != nil {
```
wanyufe, line 600: This piece of code may not working - need manual testing to confirm IPv6 address is freed

cmd/egress-v6-cni-plugin/cni.go
```diff
@@ -0,0 +1,608 @@
...
+		return nil
+	})
+}
+
+func enableHostIPv6Forwarding() error {
+	var hostPrimaryIf string
+	err := ip.EnableIP6Forward()
+	if err != nil {
+		return err
+	}
```
jdn5126, line 357: This should already be covered during container init: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/cmd/aws-vpc-cni-init/main.go#L108

You just need to enable these IPv6 settings whenever the new environment variable you add is set as well

cmd/egress-v6-cni-plugin/cni.go
```diff
@@ -0,0 +1,608 @@
...
+
+	if len(hostPrimaryIf) > 0 {
+		var entry = "/proc/sys/net/ipv6/conf/" + hostPrimaryIf + "/accept_ra"
+
+		if content, err := os.ReadFile(entry); err == nil {
+			if bytes.Equal(bytes.TrimSpace(content), []byte("2")) {
+				return nil
+			}
+		}
+		return os.WriteFile(entry, []byte("2"), 0644)
```
jdn5126, line 386: This can also be handled during container init: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/cmd/aws-vpc-cni-init/main.go#L126

cmd/egress-v6-cni-plugin/cni.go
```diff
@@ -0,0 +1,608 @@
...
+		Scope:     netlink.SCOPE_HOST,
+		Dst:       &net.IPNet {
+			IP: containerIPv6,
+			Mask: net.CIDRMask(128, 128),
+		},
+	})
+}
+
+func setupHostIPv6Snat(containerIPv6 net.IP, hostPrimaryIfIPv6 net.IP, chain, comment, randomizeSNAT string) error {
+	return  snat.Snat6(hostPrimaryIfIPv6, containerIPv6, chain, comment, randomizeSNAT)
```
jdn5126, line 426: The first argument should come from `netConf.NodeIP`. Covered more below

cmd/egress-v6-cni-plugin/cni.go
```diff
@@ -0,0 +1,608 @@
...
+	if err != nil {
+		log.Debugf("IPv6 address not found on host primary interface which is needed for IPv6 egress: %v", err)
+		return err
+	}
+
+	hostPrimaryIfIPv6 := net.ParseIP(hostPrimaryIfIPv6AddrStr)
+	if hostPrimaryIfIPv6 == nil || !hostPrimaryIfIPv6.IsGlobalUnicast() {
+		return fmt.Errorf("gobal unicast IPv6 address is needed on host primary interface for IPv6 egress")
+	}
+	log.Debugf("IPv6 address retrieved: %s", hostPrimaryIfIPv6AddrStr)
```
jdn5126, line 470: The IPv6 address assigned to the primary ENI should be derived during container startup and then set in `NetConf.NodeIP`: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/cmd/aws-vpc-cni/main.go#L389 , similar to what is done for IPv4 egress.

The reason I think we should do this during container startup is that if the customer does not configure an IPv6 address on the primary ENI, we can log an error early and set the chained plugin to `Disabled`. That way, the plugin will never be enabled unless the customer config meets that precondition.

For resiliency, if the IPv6 address happens to be removed on the primary ENI while the plugin is running, the result should be that we just create an SNAT rule that won't actually forward. And have no crash.

cmd/egress-v6-cni-plugin/cni.go
```diff
@@ -0,0 +1,608 @@
...
+		return err
+	}
+	log.Debugf("container route set up successfully")
+
+	err = enableHostIPv6Forwarding()
+	if err != nil {
+		log.Debugf("enableHostIPv6Forwarding failed: %v", err)
+		return err
+	}
+	log.Debugf("enable host IPv6 forwarding successfully")
```
jdn5126, line 503: note: this will be removed as IPv6 forwarding should already be enabled for all newly created interfaces due to the setting in init container

cmd/egress-v6-cni-plugin/cni.go
```diff
@@ -0,0 +1,608 @@
...
+
+	err = setupHostIPv6Route(netns, args.IfName, containerIPNetV6.IP)
```
jdn5126, line 505: nit: can you eventually add a comment here or at the top of this function definition explaining that this adds an IPv6 route in host network namespace for traffic destined to pod IPv6 address?

cmd/egress-v6-cni-plugin/cni.go
```diff
@@ -0,0 +1,608 @@
...
+	if err != nil {
+		log.Debugf("setupHostIPv6Route failed: %v", err)
+		return err
+	}
+	log.Debugf("host IPv6 route set up successfully")
+
+	err = setupHostIPv6Snat(containerIPNetV6.IP, hostPrimaryIfIPv6, chain, comment, netConf.RandomizeSNAT)
```
jdn5126, line 512: nit: can you eventually add a comment here or at the top of this function definition explaining that this adds an iptables entries to NAT from pod IPv6 address to node IPv6 address assigned to primary ENI?

cmd/egress-v6-cni-plugin/cni.go
```diff
@@ -0,0 +1,608 @@
...
+	}
+
+	log.Debugf("Received an ADD request for: conf=%v; Plugin enabled=%s", netConf, netConf.Enabled)
+	log.Debugf("PrevResult: %v", result)
+	// We will not be vending out this as a separate plugin by itself, and it is only intended to be used as a
+	// chained plugin to VPC CNI in IPv4 mode.
+	// We only need this plugin to kick in if v6 is NOT enabled in VPC CNI. So, the
+	// value of an env variable in VPC CNI determines whether this plugin should be enabled and this is an attempt to
+	// pass through the variable configured in VPC CNI.
+	if netConf.Enabled == "false" {
```
jdn5126, line 456: nit: we should do a case-insensitive compare here

cmd/egress-v6-cni-plugin/cni.go
```diff
@@ -0,0 +1,608 @@
...
+
+	netConf, log, err := loadConf(args.StdinData)
+	if err != nil {
+		return fmt.Errorf("failed to parse config: %v", err)
+	}
+
+	log.Debugf("Received Del request for: conf=%v; Plugin enabled=%s", netConf, netConf.Enabled)
+
+	// We only need this plugin to kick in if v6 is enabled
+	if netConf.Enabled == "false" {
```
jdn5126, line 541: nit: we should do a case-insensitive compare here

cmd/egress-v6-cni-plugin/cni.go
```diff
@@ -0,0 +1,608 @@
...
+	}
+
+	result, err := current.GetResult(netConf.PrevResult)
+	if err != nil {
+		log.Debugf("get PrevResult failed: %v", err)
+		return err
+	}
+	log.Debugf("PrevResult: %v", result)
+
+	// figure out container IPv6 address
```
jdn5126, line 557: So at this point, the main VPC CNI plugin has already run, but since it does not delete the `eth0` interface, we can still derive the IPv6 ULA address that we assigned to it. Thinking out loud, the only concern here is if the main VPC CNI plugin changes its functionality someday to delete the `eth0` interface. We can discuss more offline about how we can protect from this regression ever happening.

cmd/egress-v6-cni-plugin/cni.go
```diff
@@ -0,0 +1,608 @@
...
+	var containerVethIPv6 net.IPNet
+	for _, ipc := range result.IPs {
+		if ipc.Version == "6" && ipc.Address.IP.IsGlobalUnicast() {
```
jdn5126, line 560: nit: is there a more specific check we can do for ULA?

cmd/egress-v6-cni-plugin/cni.go
```diff
@@ -0,0 +1,608 @@
...
+	}
+	log.Debugf("host IPv6 route for container %s IPv6 %s deleted", args.Netns, containerVethIPv6.String())
+
+	// delete container IPv6 route
+	err = tearDownContainerIPv6Route(netns, &containerVethIPv6, args.IfName)
+	if err != nil {
+		log.Debugf("tear down container IPv6 route failed: %v", err)
+		return err
+	}
+	log.Debugf("container %s IPv6 route deleted", args.Netns)
```
jdn5126, line 598: This delete should not be necessary, as everything in the container networking namespace will be deleted after this returns success

cmd/egress-v6-cni-plugin/cni.go
```diff
@@ -0,0 +1,608 @@
...
+		log.Debugf("failed to open netns %q: %v", args.Netns, err)
+		return fmt.Errorf("failed to open netns %q: %v", args.Netns, err)
+	}
+	defer netns.Close()
+
+	// delete host route - return traffic route for container's IPv6 traffic
+	err = tearDownHostIPv6Route(netns, args.IfName, containerVethIPv6.IP)
+	if err != nil {
+		log.Debugf("tear down host IPv6 route for container IPv6: $s failed: %v", containerVethIPv6.String(), err)
+		return err
```
jdn5126, line 588: CNI deletes should be best effort, so we don't want to return error here. We can only log and continue

cmd/egress-v6-cni-plugin/cni.go
jdn5126: Freed in the `host-local` IPAM range, you mean? I think we can find the `host-local` logs somewhere to confirm. Upstream docs are here: https://www.cni.dev/plugins/current/ipam/host-local/

Actually, it looks like we can look in `/var/lib/cni/networks/`

cmd/egress-v6-cni-plugin/snat/snat.go
```diff
@@ -0,0 +1,130 @@
...
+			existingChains[chain] = true
+		}
+	}
+
+	for _, rule := range rules {
+		chain := rule[0]
+		if err := ipt.AppendUnique("nat", chain, rule[1:]...); err != nil {
+			return err
+		}
+	}
```
jdn5126, line 92: Why do we loop twice here? I'm not sure I understand this loop

cmd/egress-v6-cni-plugin/snat/snat.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil && !isNotExist(err) {
+		return err
+	}
+
+	return nil
+}
+
+// isNotExist returns true if the error is from iptables indicating
+// that the target does not exist.
+func isNotExist(err error) bool {
```
jdn5126, line 124: nit: this probably belongs in some shared utils file

go.mod
```diff
@@ -92,6 +92,7 @@ require (
 	github.com/jmoiron/sqlx v1.3.5 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
+	github.com/karampok/diktyo v0.0.0-20180308145956-65d3cb09123e // indirect
```
jdn5126, line 95: Is this needed? I don't see anything you imported relying on this

jdn5126 (commented):
This is awesome progress, thanks for posting this draft @wanyufe ! I went through and the overall implementation looks great, so I just added comments around areas we might want to refactor or discuss further

cmd/egress-v6-cni-plugin/snat/snat.go
wanyufe: The loop before line 87 is to add any chains that does not exist yet. Basically, we cannot add rules to a non-exist chain.

cmd/aws-vpc-cni-init/main.go
```diff
@@ -34,6 +34,7 @@ const (
 	envDisableIPv4TcpEarlyDemux = "DISABLE_TCP_EARLY_DEMUX"
 	envEnableIPv6               = "ENABLE_IPv6"
 	envHostCniBinPath           = "HOST_CNI_BIN_PATH"
+	envEnEgress                 = "ENABLE_ENGRESS"
```
pmotyka, line 37: @wanyufe should this be ENABLE_EGRESS?

jdn5126: nit: let's go with `ENABLE_V6_EGRESS`

cmd/aws-vpc-cni-init/main.go
```diff
@@ -114,13 +115,16 @@ func configureIPv6Settings(procSys procsyswrapper.ProcSys, primaryIF string) err
 		}
 		val, _ := procSys.Get(entry)
 		log.Infof("Updated %s to %s", entry, val)
-
-		entry = "net/ipv6/conf/all/forwarding"
+	}
+	// Check if IPv6 egress supporting is enabled in IPv4 cluster
```
jdn5126, line 119: nit: "Check if IPv6 forwarding should be enabled for IPv4 cluster"

jdn5126: Also, don't we need to set `net/ipv6/conf/all/disable_ipv6` to 0 as well?

cmd/aws-vpc-cni/main.go
```diff
@@ -64,8 +64,14 @@ const (
 	defaultPodSGEnforcingMode    = "strict"
 	defaultPluginLogFile         = "/var/log/aws-routed-eni/plugin.log"
 	defaultEgressV4PluginLogFile = "/var/log/aws-routed-eni/egress-v4-plugin.log"
```
jdn5126, line 66: nit: this can be removed

cmd/aws-vpc-cni/main.go
```diff
@@ -80,12 +86,14 @@ const (
 	envPluginLogFile         = "AWS_VPC_K8S_PLUGIN_LOG_FILE"
 	envPluginLogLevel        = "AWS_VPC_K8S_PLUGIN_LOG_LEVEL"
 	envEgressV4PluginLogFile = "AWS_VPC_K8S_EGRESS_V4_PLUGIN_LOG_FILE"
```
jdn5126, line 88: nit: let's remove this

cmd/aws-vpc-cni/main.go
```diff
@@ -200,36 +208,82 @@ func waitForInit() error {
...
-		if err != nil {
-			log.WithError(err).Fatalf("aws-vpc-cni failed")
-			return "", err
-		}
-		if hostIP != "" {
-			return hostIP, nil
-		}
+	hostIP, err = cniutils.GetNodeMetadata("local-ipv4")
+	if err != nil {
+		log.WithError(err).Fatalf("aws-vpc-cni failed")
```
jdn5126, line 213: nit: let's make this "failed to get node primary v4 address for egress plugin"

cmd/aws-vpc-cni/main.go
```diff
@@ -200,36 +208,82 @@ func waitForInit() error {
...
+	}
+	return hostIP, nil
+}
+
+func getNodePrimaryV6Address() (string, error) {
+	var hostIP string
+	var err error
+	hostIP, err = cniutils.GetNodeMetadata("ipv6")
+	if err != nil {
+		log.WithError(err).Fatalf("aws-vpc-cni failed")
```
jdn5126, line 224: nit: let's make this "failed to get node primary v6 address for egress plugin"

cmd/egress-cni-plugin/main.go
```diff
@@ -0,0 +1,164 @@
...
+		chainPrefix = "E4-"
+	}
+	chain := utils.MustFormatChainNameWithPrefix(netConf.Name, args.ContainerID, chainPrefix)
+	comment := utils.FormatComment(netConf.Name, args.ContainerID)
+
+	if isIPv6Egress {
+		netConf.IfName = netconf.EgressIPv6InterfaceName
+		return cni.CmdDelEgressV6(args.Netns, netConf.IfName, chain, comment, log)
+	}
+	netConf.IfName = netconf.EgressIPv4InterfaceName
```
jdn5126, line 162: nit: let's put this in an else block for readability

jdn5126 (commented):
Changes are looking great, just some nits

jdn5126:
Closing this as individual PRs are present for each step of merging this

Issue #2331, opened 2023-03-28
Increasing resource requests leads to loss of cni on nodes


steveizzle:
**What happened**:

I want to increase the memory requests for the CNI DaemonSet (currently from 0 eksctl utils default to 50MB) on existing Clusters which have the workload distributed efficiently, so a lot of >95% memory requested nodes. 

Now 

1. the old **aws-node** Pod gets deleted on the node 
2. The new **aws-node** pod evicts _other pod_ ,which is not so important, to free up memory 
3. The _other pod_ can not get deleted, because there is no aws-node cni anymore (seems to be necessary in vpc-cni when pods are evicted from nodes)
4. The new **aws-node** can never get deployed onto the node 
5. The _other pod_ can never get removed because there is no aws-node
6. Node has no cni anymore and is therefore "broken"

**Attach logs**

describe new **aws-node**
`Warning  FailedScheduling  18s   default-scheduler  0/4 nodes are available: 1 Insufficient memory.`

allocated resources on the node:
```
Allocated resources:
  (Total limits may be over 100 percent, i.e., overcommitted.)
  Resource                    Requests      Limits
  --------                    --------      ------
  cpu                         1081m (72%)   1500m (100%)
  memory                      2949Mi (99%)  6174Mi (208%)
```

get _other pod_ (which is chosen for eviction):

`flux-system   notification-controller-7797cd9fb7-5pjsx                     0/1     Terminating   0          4h22m   100.64.102.5    ip-10-46-65-53.eu-central-1.compute.internal   <none>           <none>`

describe _other pod_:
```
Events:
  Type     Reason         Age                From               Message
  ----     ------         ----               ----               -------
  Normal   Preempted      94s                default-scheduler  Preempted by a pod on node ip-10-46-65-53.eu-central-1.compute.internal
  Normal   Killing        94s                kubelet            Stopping container manager
[...]

PR #2350, opened 2023-04-11
Return success from CNI DEL when IPAMD is unreachable

jdn5126:
**What type of PR is this?**
bug

#2331 

This PR modifies the VPC CNI pod deletion logic to return no error to the container-runtime when IPAMD is unreachable. This is needed for three primary reasons:
1. The CNI spec specifies that delete operations should generally complete without error, even when resources are missing.
2. The CNI cannot count on kubelet to retry deletes until IPAMD is available. If a user deletes a pod with a termination grace period of 0 seconds, the pod is deleted from the API server before kubelet returns success, so kubelet will give up after a few iterations.
3. If the aws-node daemonset pod cannot be scheduled until other pods are deleted, pod deletion needs to happen without IPAMD.

CNI Changes:
As before, the CNI will try to delete non-branch-ENI pods using PrevResult when IPAMD is not reachable. The difference is that CNI will now return no error in this case. Note that PrevResult will only be available for pods created with VPC CNI v1.12.1+. For pods created before this, IP rules will be leaked, and it is IPAMD's responsibility to cleanup these rules when it starts again.

IPAMD Changes:
On init, IPAMD identifies stale allocations and prunes IP rules for these stale allocations. IPAMD cannot be sure if CNI was able to prune those rules before the pod was deleted. Also, IPAMD does an additional state file write after pruning entries.

Misc:
I also did some refactoring of network utilities to avoid code duplication.

N/A

Added unit test coverage for stale IPAM allocations. Manually created a cluster and verified behavior when IPAMD is unreachable, plus upgrade/downgrade scenarios.

**Automation added to e2e**:
N/A

N/A

No, Yes

No

Yes

```release-note
Update CNI to delete pod when IPAMD is unreachable.
```


pkg/networkutils/network.go
```diff
@@ -740,13 +746,34 @@ func (r iptablesRule) String() string {
...
+			if !containsNoSuchRule(err) {
+				return err
+			}
+			break
+		}
+	}
+	return nil
+}
+
+func ContainsNoSuchRule(err error) bool {
```
orsenthil, line 762: Are these introduced for backwards compatibility instead of refactoring the existing functions?

pkg/networkutils/network.go
```diff
@@ -740,13 +746,34 @@ func (r iptablesRule) String() string {
 	return fmt.Sprintf("%s/%s rule %s shouldExist %v rule %v", r.table, r.chain, r.name, r.shouldExist, r.rule)
 }
 
+// netLinkRuleDelAll deletes all matching route rules (instead of only first instance).
```
orsenthil, line 749: NetLinkRuleDelAll ( public)

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -325,13 +322,16 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 			args.ContainerID, err)
 
 		// When IPAMD is unreachable, try to teardown pod network using previous result. This action prevents rules from leaking while IPAMD is unreachable.
-		// Note that connection error is still returned to kubelet so that delete retries until IPAMD can be notified.
+		// Note that no error is returned to kubelet as there is no guarantee that kubelet will retry delete, and returning an error would prevent container runtime
+		// from cleaning up resources. When IPAMD again becomes responsive, it is responsible for reclaiming IP.
```
jayanthvn, line 326: Say if kubelet doesn't retry and if IPAMD is recovering from the state file then reclaiming IP is not possible right? Am is missing anything here?

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -325,13 +322,16 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 			args.ContainerID, err)
 
 		// When IPAMD is unreachable, try to teardown pod network using previous result. This action prevents rules from leaking while IPAMD is unreachable.
-		// Note that connection error is still returned to kubelet so that delete retries until IPAMD can be notified.
+		// Note that no error is returned to kubelet as there is no guarantee that kubelet will retry delete, and returning an error would prevent container runtime
```
orsenthil, line 325: > There is no guarantee that kubelet will retry delete

May be I am missing something, Why will not kubelet retry deletes? 

For the referenced issue,  https://github.com/aws/amazon-vpc-cni-k8s/issues/2331 was the case for the customer that the IPAM wasn't reachable, and if IPAMD was running these resource limitation won't be hit?

cmd/routed-eni-cni-plugin/cni.go
jayanthvn: Ideally I feel, IPAMD cleanup should happen at the end, i.e CNI should go ahead and cleanup resources (teardownnetwork) in the networking stack and then invoke IPAMD (delnetwork) to release IP.. that will also help with leaked rules..we can discuss offline more..

pkg/ipamd/datastore/data_store.go
```diff
@@ -438,6 +439,11 @@ func (ds *DataStore) ReadBackingStore(isv6Enabled bool) error {
 		}
 	}
 
+	// Some entries may have been purged during recovery, so write to backing store
+	if err := ds.writeBackingStoreUnsafe(); err != nil {
```
jayanthvn, line 443: Recovery is from backing store right? Do we need this again?

pkg/ipamd/datastore/data_store.go
```diff
@@ -1465,22 +1471,27 @@ func (ds *DataStore) CheckFreeableENIexists() bool {
...
+			ds.log.Warnf("stale IP allocation for ID(%v): IPv4(%v), IPv6(%v) due to %v", allocation.ContainerID, allocation.IPv4, allocation.IPv6, err)
+			staleAllocations = append(staleAllocations, allocation)
 		} else {
 			validatedAllocations = append(validatedAllocations, allocation)
 		}
 	}
 	checkpoint.Allocations = validatedAllocations
+	// Stale allocations may have dangling IP rules that need cleanup
+	if len(staleAllocations) > 0 {
+		ds.PruneStaleAllocations(staleAllocations)
```
jayanthvn, line 1493: We don't move the IPAMD/CNI delete logic, we need to cleanup the backing store too right?

cmd/routed-eni-cni-plugin/cni.go
jdn5126: Kubelet will not retry deletes if the resource has already been deleted at the API server, which can happen if a raw delete is issued or the pod had a termination grace period of 0 seconds.

#2331 is neither of those cases, though. It is the case where `aws-node` pod is not running on the node, and pods cannot be deleted on the node to make room for it due to IPAMD not running.

cmd/routed-eni-cni-plugin/cni.go
jdn5126: IPAMD will reclaim the IP from the ENI because it will detect that its entry in the state file is invalid. CNI does clean up all of the resources (really just the IP rules) if the pod was created by v1.12.1+. It is only if there is no PrevResult that the IP rules could leak and so IPAMD has to try to clear them

pkg/ipamd/datastore/data_store.go
jdn5126: We need to write here to purge any stale entries that were present. Otherwise, those stale entries would remain in the file, but not be loaded into memory (into datastore).

pkg/ipamd/datastore/data_store.go
```diff
@@ -1465,22 +1471,27 @@ func (ds *DataStore) CheckFreeableENIexists() bool {
...
 	if err != nil {
 		return CheckpointData{}, err
 	}
 	var validatedAllocations []CheckpointEntry
+	var staleAllocations []CheckpointEntry
 	for _, allocation := range checkpoint.Allocations {
 		if err := ds.validateAllocationByPodVethExistence(allocation, hostNSLinks); err != nil {
-			ds.log.Warnf("ignore IP allocation for %v:%v,%v due to %v", allocation.ContainerID, allocation.IPv4, allocation.IPv6, err)
+			ds.log.Warnf("stale IP allocation for ID(%v): IPv4(%v), IPv6(%v) due to %v", allocation.ContainerID, allocation.IPv4, allocation.IPv6, err)
+			staleAllocations = append(staleAllocations, allocation)
```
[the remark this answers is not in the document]
jdn5126: Right, but we can only do that if the pod was created by v1.12.1+. It would be great to remove this overhead in the future, but in the common case, there should never be any stale allocations to clean up

pkg/ipamd/datastore/data_store.go
jdn5126: The backing store is cleaned up just by writing from the in-memory datastore after init

pkg/networkutils/network.go
jdn5126: They are moved from the deleted `cmd/routed-eni-cni-plugin/driver/utils.go` so that they can be used by IPAMD and CNI logic without duplicating code

cmd/routed-eni-cni-plugin/cni.go
M00nF1sh: TBH, ideally i think we don't need this change, it should be kubelet's job to make sure always retry call cniDel even the Pod have been gone from APIServer. Kubelet should fix their bug. and this change is just our workaround for this kubelet flaw.

also, let's test the logic that IPAMD can reclaim the IP in this case.

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -325,13 +322,16 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
...
+		// from cleaning up resources. When IPAMD again becomes responsive, it is responsible for reclaiming IP.
 		if teardownPodNetworkWithPrevResult(driverClient, conf, k8sArgs, args.IfName, log) {
 			log.Infof("Handled pod teardown using prevResult: ContainerID(%s) Netns(%s) IfName(%s) PodNamespace(%s) PodName(%s)",
 				args.ContainerID, args.Netns, args.IfName, string(k8sArgs.K8S_POD_NAMESPACE), string(k8sArgs.K8S_POD_NAME))
+		} else {
+			log.Infof("Could not teardown pod using prevResult: ContainerID(%s) Netns(%s) IfName(%s) PodNamespace(%s) PodName(%s)",
```
M00nF1sh, line 331: this shall be a log.errorf or at least `warnf`

cmd/routed-eni-cni-plugin/cni.go
jdn5126: From kubelet perspective, this is working as expected: the pod has disappeared from the API server, so there is nothing to retry. In the case that motivated this PR, it is because the pod was force deleted, but I don't see that as a kubelet bug.

Outside of kubelet, though, the scheduling deadlock still exists, where the aws-node pod cannot be placed until pods are deleted, and pods cannot be deleted unless IPAMD is running.

jdn5126: I did confirm from testing that IPAMD reclaims the IP from the ENI and can reuse it later

cmd/routed-eni-cni-plugin/cni.go
jdn5126: The most common case where this happens is the pod already being deleted. Kubelet often sends the DEL request 2-3 times, even if the first one succeeded. I left this as "info" to be inline with the other info above, where IPAMD sees the container as "not found"

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -356,15 +356,18 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
...
 			return nil
 		}
-		log.Errorf("Error received from DelNetwork gRPC call for container %s: %v",
-			args.ContainerID, err)
+		log.Errorf("Error received from DelNetwork gRPC call for container %s: %v", args.ContainerID, err)
 
-		//  DelNetworkRequest may return a connection error, so try to delete using PrevResult whenever an error is returned.
+		// DelNetworkRequest may return a connection error, so try to delete using PrevResult whenever an error is returned. As with the case above, do
+		// not return error to kubelet, as there is no guarantee that delete is retried.
 		if teardownPodNetworkWithPrevResult(driverClient, conf, k8sArgs, args.IfName, log) {
```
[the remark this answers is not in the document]
jdn5126: That sounds reasonable. I think it's best to file an issue and have that as a future refactor rather than doing in this PR, as this PR has gotten quite large.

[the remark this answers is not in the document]
jdn5126: Filed https://github.com/aws/amazon-vpc-cni-k8s/issues/2374

M00nF1sh (approved):
/approve
approving given we'll do the mentioned refactor as follow up.

PR #2351, opened 2023-04-12
roll back event changes for now

haouc:
**What type of PR is this?**

**Automation added to e2e**:

Due to the changed plan, we need use another strategy to implement this channel. Roll back changes introduced for node events from commits.

```release-note

```


pkg/utils/eventrecorder/eventrecorder.go
```diff
@@ -108,21 +106,3 @@ func findMyPod(cachedK8SClient client.Client) (corev1.Pod, error) {
 	}
 	return pod, err
 }
-
-// SendNodeEvent sends an event regarding pod object and related to node object
-func (e *EventRecorder) SendNodeEvent(eventType, reason, action, message string) error {
```
jdn5126, line 113: I think `findMyNode` can be removed here as well, as only `SendNodeEvent` uses `e.hostNode`

jdn5126 (commented):
Change overall looks good to me. Can you confirm that `test/integration/pod-eni/` suite passes after this change?

pkg/utils/eventrecorder/eventrecorder.go
haouc: Thanks, Jeff. Let me check on the function. Also will confirm the tests.

haouc: Tests focused on pod_eni was passed.
```
[AfterSuite] PASSED [173.059 seconds]

Ran 1 of 6 Specs in 614.365 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 5 Skipped
PASS | FOCUSED

Ginkgo ran 1 suite in 10m23.396413883s
Test Suite Passed
```

PR #2354, opened 2023-04-19
Increase datastore pool at startup

jdn5126:
**What type of PR is this?**
bug

#1724 

This PR fixes two issues:
1. The container entrypoint was not properly validating IPAMD startup.
2. During node initialization, IPAMD used to only build datastore from already attached ENIs and available CIDRs. Now, if the datastore pool is too low (based on settings), it tries to increase the datastore pool by attaching ENIs. This improves the startup time for custom networking and branch ENIs. There is also a check for custom networking such that if the datastore is empty after the increase attempt, there must be a configuration error, so IPAMD returns an error, which will put the pod into a CrashLoopBackOff until the configuration error is resolved. In the future, we want to generate an event for this case.

N/A

Manually verified that when ENIConfigs are missing, node remains in "Not Ready" state as IPAMD returns error. Also verified that when ENIConfigs are present, ENIs are attached during node initialization.

Verified that CNI integration release tests pass. Scheduling GitHub runner as well.

**Automation added to e2e**:
N/A

No

No, Yes

No

Yes

```release-note
Enhance node initialization to attach ENIs if needed.
```

### DIFF
--- a/cmd/aws-vpc-cni/main.go
+++ b/cmd/aws-vpc-cni/main.go
@@ -36,7 +36,6 @@
 package main
 
 import (
-	"bytes"
 	"encoding/json"
 	"net"
 	"os"
@@ -171,13 +170,11 @@ type Range struct {
 	Gateway    net.IP      `json:"gateway,omitempty"`
 }
 
+// Wait for IPAMD health check to pass. Note that if IPAMD fails to start, wait happens indefinitely until liveness probe kills pod
 func waitForIPAM() bool {
 	for {
 		cmd := exec.Command("./grpc-health-probe", "-addr", "127.0.0.1:50051", ">", "/dev/null", "2>&1")
-		var outb bytes.Buffer
-		cmd.Stdout = &outb
-		cmd.Run()
-		if outb.String() == "" {
+		if err := cmd.Run(); err == nil {
 			return true
 		}
 	}
@@ -443,7 +440,7 @@ func _main() int {
 
 	err = cp.CopyFile(tmpAWSconflistFile, defaultHostCNIConfDirPath+awsConflistFile)
 	if err != nil {
-		log.WithError(err).Errorf("Failed to copy 10-awsconflist")
+		log.WithError(err).Errorf("Failed to copy %s", awsConflistFile)
 		return 1
 	}
 	log.Infof("Successfully copied CNI plugin binary and config file.")
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -412,7 +412,7 @@ func New(rawK8SClient client.Client, cachedK8SClient client.Client) (*IPAMContex
 
 	c.primaryIP = make(map[string]string)
 	c.reconcileCooldownCache.cache = make(map[string]time.Time)
-	//WARM and Min IP/Prefix targets are ignored in IPv6 mode
+	// WARM and Min IP/Prefix targets are ignored in IPv6 mode
 	c.warmENITarget = getWarmENITarget()
 	c.warmIPTarget = getWarmIPTarget()
 	c.minimumIPTarget = getMinimumIPTarget()
@@ -428,8 +428,7 @@ func New(rawK8SClient client.Client, cachedK8SClient client.Client) (*IPAMContex
 		return nil, err
 	}
 
-	//Let's validate if the configured combination of env variables is supported before we
-	//proceed any further
+	// Validate if the configured combination of env variables is supported before proceeding further
 	if !c.isConfigValid() {
 		return nil, fmt.Errorf("ipamd: failed to validate configuration")
 	}
@@ -446,7 +445,7 @@ func New(rawK8SClient client.Client, cachedK8SClient client.Client) (*IPAMContex
 
 	mac := c.awsClient.GetPrimaryENImac()
 
-	// retrieve security groups
+	// Retrieve security groups
 	if c.enableIPv4 && !c.disableENIProvisioning {
 		err = c.awsClient.RefreshSGIDs(mac)
 		if err != nil {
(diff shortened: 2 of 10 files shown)


pkg/ipamd/ipamd.go
```diff
@@ -601,21 +597,18 @@ func (c *IPAMContext) nodeInit() error {
...
-		} else if err != nil {
-			if containsInsufficientCIDRsOrSubnetIPs(err) {
-				log.Errorf("Unable to attach IPs/Prefixes for the ENI, subnet doesn't seem to have enough IPs/Prefixes. Consider using new subnet or carve a reserved range using create-subnet-cidr-reservation")
-				c.lastInsufficientCidrError = time.Now()
-				return nil
-			}
-			return err
+	// On node init, check if datastore pool needs to be increased. If so, attach CIDRs from existing ENIs and attach new ENIs.
+	if !c.disableENIProvisioning && c.isDatastorePoolTooLow() {
+		c.increaseDatastorePool(ctx)
```
jayanthvn, line 602: We will have to capture the `err`  from `c.increaseDatastorePool `.. similar to - 

```
else if err != nil {
 			if containsInsufficientCIDRsOrSubnetIPs(err) {
 				log.Errorf("Unable to attach IPs/Prefixes for the ENI, subnet doesn't seem to have enough IPs/Prefixes. Consider using new subnet or carve a reserved range using create-subnet-cidr-reservation")
 				c.lastInsufficientCidrError = time.Now()
 				return nil
 			}
 			return err
```

jayanthvn: Ideally any issues while increasing the cidr during nodeInit we used to return `err` so have to maintain the same behavior..

testdata/amazon-eks-nodegroup.yaml
```diff
@@ -74,7 +74,7 @@ Parameters:
 
   NodeImageIdSSMParam:
     Type: "AWS::SSM::Parameter::Value<AWS::EC2::Image::Id>"
-    Default: /aws/service/eks/optimized-ami/1.22/amazon-linux-2/recommended/image_id
+    Default: /aws/service/eks/optimized-ami/1.24/amazon-linux-2/recommended/image_id
```
jayanthvn, line 77: Nit: Consider adding a constant for `1.24`, I believe we have few places where we refer to K8s versions..

cmd/aws-vpc-cni/main.go
```diff
@@ -164,15 +163,18 @@ type Range struct {
...
 	for {
 		cmd := exec.Command("./grpc-health-probe", "-addr", "127.0.0.1:50051", ">", "/dev/null", "2>&1")
-		var outb bytes.Buffer
-		cmd.Stdout = &outb
-		cmd.Run()
-		if outb.String() == "" {
+		if err := cmd.Run(); err == nil {
 			return true
 		}
+		if time.Now().After(maxEnd) {
```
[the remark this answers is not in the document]
jdn5126: Yeah, I think we want this error so that the node does not get marked as ready. And while IPAMD will have the log, until we have an event for customers to look at, I think it is better for anyone running `kubectl log aws-node-xxxxx -n kube-system` to see this, which will point them to the IPAMD logs.

Having the `Run()` on one line is fine, I just split it up as I thought it was more readable

pkg/ipamd/ipamd.go
jdn5126: I considered that, but I want to be really careful in which cases we return error and crash IPAMD. For example, an EC2 API could fail temporarily, but then be resolved on the normal loop. While the pod would be restarted after the crash, it seems like we can allow a transient networking issue as long as we can make forward progress.

In the non-custom networking case, we can make forward progress since the primary ENI is already present. In the custom-networking case, we can make forward progress as long as at least one ENI was attached.

pkg/ipamd/ipamd.go
```diff
@@ -601,21 +597,18 @@ func (c *IPAMContext) nodeInit() error {
...
+		// If custom networking is enabled and the pool is empty, return an error, as there is a misconfiguration and
+		// the node should not become ready.
+		if c.useCustomNetworking && c.isDatastorePoolEmpty() {
+			podENIErrInc("nodeInit")
+			return errors.New("Failed to attach any ENIs for custom networking")
 		}
```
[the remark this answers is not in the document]
jdn5126: Commented above, still trying to figure out the right granularity for errors here.

testdata/amazon-eks-nodegroup.yaml
jdn5126: Good call, yeah this should be a parameter

pkg/ipamd/ipamd.go
jayanthvn: But this will be a behavior change, previously we used to err out nodeInit and now nodeInit might pass through..

cmd/aws-vpc-cni/main.go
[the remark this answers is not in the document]
jayanthvn: Yeah makes sense about `grpc-health-probe` but sorry I meant do we need the 1 min check?

pkg/ipamd/ipamd.go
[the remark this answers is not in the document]
jayanthvn: Sure, we can discuss offline about this too..

cmd/aws-vpc-cni/main.go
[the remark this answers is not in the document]
jdn5126: Since we don't want to wait indefinitely, I figured one minute was a reasonable upper limit for IPAMD to initialize and start gRPC server. We could also increase connect and rpc timeout args to `grpc-health-probe` instead of looping here, but I figured that was more prone to error as I wasn't sure if an errant TCP reset from gRPC server was possible

pkg/ipamd/ipamd.go
jdn5126: That's a good point, I can catch this error and maintain the original behavior

pkg/ipamd/ipamd.go
```diff
@@ -1353,8 +1355,11 @@ func podENIErrInc(fn string) {
 
 // nodeIPPoolReconcile reconcile ENI and IP info from metadata service and IP addresses in datastore
 func (c *IPAMContext) nodeIPPoolReconcile(ctx context.Context, interval time.Duration) {
+	// To reduce the number of EC2 API calls, skip reconciliation if IPs were recently added to the datastore.
 	timeSinceLast := time.Since(c.lastNodeIPPoolAction)
-	if timeSinceLast <= interval {
+	// Make an exception if node needs a trunk ENI and one is not currently attached.
```
jayanthvn, line 1360: One callout for this change is that, if trunk ENI fails to attach for any VPC RC issues then aws-node keeps running the reconciliation..maybe as an enhancement we should make it like a backoff retry in cases of SGPP..

jayanthvn (approved):
lgtm

Release testing should cover negative test cases while increasing the DS pool and trunk ENI failed to attach scenarios.

Issue #2319, opened 2023-03-16
Trunk ENI discovery were delayed due to IP pool increase


M00nF1sh:
**What happened**:
Our trunk eni discovery logic happens within [`nodeIPPoolReconcile`](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/ipamd.go#L1424), which skips itself when there were ip pool operations within past 1 minute.

However, when Node IP pool were increased(such as assign IPs), it [was counted as an "ip pool operation"](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/ipamd.go#L877), thus the trunk eni discovery logic can be skipped.

On an edge case cluster, Cx have a cluster that create/deletes health-check pods every 1 minute, and the trunk ENI discovery logic only runs after 10minute after it's been attached(in theory, it might never be discovered as well..)

**Attach logs**

**What you expected to happen**:
Once a new ENI is discovered in metadata, it's attributes should be inferred (at least for trunk ENI)

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- CNI Version
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):

PR #2356, opened 2023-04-20
IPAMD node ENI reconcile deadlock buster

jdn5126:
**What type of PR is this?**
bug

#2319 

This PR adds a deadlock punch-out for `nodeIPPoolReconcile`. In the future, we want to refactor this reconciliation to run without increasing EC2 API calls, perhaps by determining the subset of ENIs that we need to check tags for and having an expiry period for their cached tags. In the meantime, we want to fix this bug for the next release, so that's why the punch-out is done.

For #2319, where it took a long time to attach trunk ENIs, this will be helped by https://github.com/aws/amazon-vpc-cni-k8s/pull/2354 as well.

N/A

Manually verified that existing integration tests pass following this change.

**Automation added to e2e**:
N/A

N/A

No, Yes

No

Yes

```release-note
Add punch-out for node ENI reconcilation
```

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -257,6 +257,7 @@ type IPAMContext struct {
 	warmPrefixTarget          int
 	primaryIP                 map[string]string // primaryIP is a map from ENI ID to primary IP of that ENI
 	lastNodeIPPoolAction      time.Time
+	lastNodeIPPoolReconcile   time.Time
 	lastDecreaseIPPool        time.Time
 	// reconcileCooldownCache keeps timestamps of the last time an IP address was unassigned from an ENI,
 	// so that we don't reconcile and add it back too quickly if IMDS lags behind reality.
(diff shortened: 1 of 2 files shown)


haouc:
In terms of the bug, would it make sense to fire another goroutine to waitTillTrunkDiscovered() or timeout()? Since trunk interface attaching is one time deal, sounds no need to keep reconciling on it after requiring a trunk interface?

jdn5126:
> #2354

@haouc so https://github.com/aws/amazon-vpc-cni-k8s/pull/2354 alone should handle attaching the trunk ENI during startup. I think that mostly fixes https://github.com/aws/amazon-vpc-cni-k8s/issues/2319, but I wanted to also cover the theoretical case where we are never able to bust out of wait action. If this PR is not useful, it might be better to just focus on refactoring this after #2354 merges

haouc:
Thanks @jdn5126. If the other PR can mostly handle the issue, I would recommend we refactor the code as a follow-up instead of bringing another elapsed time variable as a patch. Please correct me if I misunderstood.

jdn5126:
Yeah, that makes sense. #2354 will try to attach a trunk ENI at init if one is needed, so it would resolve the initial starvation issue that #2319 saw. I'll close this PR and think about the refactoring more

PR #2358, opened 2023-04-21
chore: Add memory request for avoiding EC2 OOM

vazkarvishal:
**What type of PR is this?**

enhancement

OOM on EC2 instances.

When no memory request is specified for any pod running on the EC2's, the EC2 instance ends up going in to OOM and becoming non-responsive, leaving the kubelet to crash and leave the node in a not-ready state. Adding a minimum request for memory will allow VPC CNI to be always available and potentially trigger evictions for other application pods without CPU/Memory. 

The 200Mi has been set based on my observation across an average of 100 nodes used over two weeks.

- run pods without CPU and Memory limits on the EKS Cluster, then watch the node come up as healthy and then go into an unhealthy state due to the VPC CNI not having memory request and other pods using up all the resources on the EC2.

No

No

No

```release-note
Added minimum memory requests to avoid EC2 OOM cases causing kubelet to become unresponsive.

```

### DIFF
--- a/charts/aws-vpc-cni/values.yaml
+++ b/charts/aws-vpc-cni/values.yaml
@@ -118,6 +118,7 @@ readinessProbeTimeoutSeconds: 10
 resources:
   requests:
     cpu: 25m
+    memory: 100Mi
 
 updateStrategy:
   type: RollingUpdate


jdn5126 (commented):
@vazkarvishal thanks, this makes sense. I did not realize there was not a default value to begin with. 200Mi seems a bit high, as this is a daemonset pod and will trigger evictions like you mentioned. Can we use 100Mi as the default? Also, can you mark this PR as an "enhancement" instead of "bug"?

jdn5126 (approved):

achevuru:
@vazkarvishal VPC CNI pod is marked as `system-node-critical` (https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/master/aws-k8s-cni.yaml#L115) and will provide a shield against kubelet evicting it due to a random OOM on the node. By specifying a memory request, for a system critical daemonset pod by default in the spec - we will end up unnecessarily blocking memory on the node (have seen customers running 1-2 pods per node and 200M is rather high for those use cases). I think that should be something that an user sets based on their use case and should be avoided in the default manifest..

vazkarvishal:
> @vazkarvishal VPC CNI pod is marked as `system-node-critical` (https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/master/aws-k8s-cni.yaml#L115) and will provide a shield against kubelet evicting it due to a random OOM on the node. By specifying a memory request, for a system critical daemonset pod by default in the spec - we will end up unnecessarily blocking memory on the node (have seen customers running 1-2 pods per node and 200M is rather high for those use cases). I think that should be something that an user sets based on their use case and should be avoided in the default manifest..

@achevuru  - This is only true if the node is still able to respond and let the kubelet restart. By the time the OOM event is triggered, the EC2 becomes fully unresponsive, which means the kubelet has no chance to restart (We are unable to even SSM into the EC2) and EC2 System Checks start to fail.

Based on the request to reduce 200Mi to 100Mi, I have reflected the change. I also have an AWS support ticket I can share perhaps to give you some more insights into this if required. But it is a repeating behaviour and very likely to happen, which is why I thought having a bare minimum request would be a good shield against such behaviour.

achevuru:
@vazkarvishal  I'm not following the issue here. If a K8S worker node is completely down due to memory pressure, how can specifying a mem `request` prevent the OOM issue? or recover from it? Do you want to place an upper limit on how much `aws-node` pod can consume so that you can prevent it from using more than the set value? Are you looking for  memory `limits` instead? 

`kubelet` will evict the pods if the limits are breached.  Linux Kernel's OOM killer will only step if node is completely out of memory. 

https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/

VPC CNI's helm chart allows you to specify resource requests/limits that can match your use case. Consider a scenario, where we upgrade VPC CNI and there are few nodes under mem pressure and can't satisfy the mem request of the CNI pod but we need the CNI pod to be running on all the nodes as it essential for pod networking setup. Although daemonset controller adds a taint by default to allow it to be scheduled on nodes with mem pressure, we still need the node to satisfy the VPC CNI pod's mem req.



`kubelet`
https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/

vazkarvishal:
> @vazkarvishal I'm not following the issue here. If a K8S worker node is completely down due to memory pressure, how can specifying a mem `request` prevent the OOM issue? or recover from it? Do you want to place an upper limit on how much `aws-node` pod can consume so that you can prevent it from using more than the set value? Are you looking for memory `limits` instead?
> 
> `kubelet` will evict the pods if the limits are breached. Linux Kernel's OOM killer will only step if node is completely out of memory.
> 
> https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/
> 
> VPC CNI's helm chart allows you to specify resource requests/limits that can match your use case. Consider a scenario, where we upgrade VPC CNI and there are few nodes under mem pressure and can't satisfy the mem request of the CNI pod but we need the CNI pod to be running on all the nodes as it is essential for pod networking setup. Although daemonset controller adds a taint by default to allow it to be scheduled on nodes with mem pressure, we still need the node to satisfy the VPC CNI pod's mem req.


I have raised this PR based on our resolution working with AWS support, so thought of contributing it back to the source. Let me try to explain with an example using 5 microservices with no requests/limits set. 

1. A new node gets provisioned by Karpenter when the microservices are deployed
2. VPC CNI is provisioned first and registers the node as healthy once the **aws-node** pod is up and healthy
3. All microservice pods are provisioned
4. The node is still healthy and everything is running fine
5. two out of the 5 microservices start consuming a lot of memory, more than that available on the EC2
6. EC2 goes into high memory pressure/OOM state which causes VPC CNI health checks to fail
7. Kubernetes restarts the pod due to health check failures but now unable to stabilise as the node is already out of memory
8. EC2 dashbnoard instance health checks fail on systems checks
9. On diagnosis these nodes are marked unhealthy, and due to OOM we are unable to even login to diagnose the issue using SSM (As the system is unresponsive).
10. Since there are no memory requests, VPC CNI is not prioritised for memory over the other pods

Now, I agree with you that based on the theory of scheduling eviction, [node out of memory](https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/#node-out-of-memory-behavior) should be best effort, but that is not how it works (at least after trying it many times, we are able to replicate this issue and thus the PR to add a memory request by default).

Also as explained on the OOM behaviour docs:
_If the kubelet can't reclaim memory before a node experiences OOM, the oom_killer calculates an oom_score based on the percentage of memory it's using on the node, and then adds the oom_score_adj to get an effective oom_score for each container. It then kills the container with the highest score._

This does not seem to be happening as the VPC CNI Pods and the EC2 nodes are unable to stabilise at all, and we do not see any evictions.

jdn5126:
Closing PR as this will not be accepted

PR #2361, opened 2023-04-26
Add IPv6 egress support to eks IPv4 cluster

wanyufe:
**What type of PR is this?**
feature

#1925 

IPv6 egress support for IPv4 EKS cluster

Manual testing - unit testing has finished

```
bash-5.1# ip -6 a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 state UNKNOWN qlen 1000
    inet6 ::1/128 scope host 
5: v6if0@if130: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9001 state UP 
    inet6 fd00::34/8 scope global 
    inet6 fe80::e0ac:89ff:fe27:4bdb/64 scope link 
bash-5.1# ip -6 r
fd00::/8 dev v6if0 proto kernel metric 256 pref medium
fe80::/64 dev v6if0 proto kernel metric 256 pref medium
default via fe80::5c51:9ff:fe55:3ab2 dev v6if0 metric 1024 pref medium
bash-5.1# ping6 2600::
PING 2600::(2600::) 56 data bytes
64 bytes from 2600::: icmp_seq=1 ttl=43 time=37.6 ms
64 bytes from 2600::: icmp_seq=2 ttl=43 time=37.6 ms
64 bytes from 2600::: icmp_seq=3 ttl=43 time=37.6 ms
^C
--- 2600:: ping statistics ---
3 packets transmitted, 3 received, 0% packet loss, time 2003ms
rtt min/avg/max/mdev = 37.598/37.619/37.633/0.015 ms
```

**Automation added to e2e**:

No

No

no

```release-note

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,13 +105,23 @@ The following environment variables are available, and all of them are optional.
 
 ---
 
+#### `ENABLE_V6_EGRESS` (v1.13.0+)
+
+Type: Boolean as a String
+
+Default: `false`
+
+Specifies whether PODs in v4 cluster support IPv6 egress. If env is set to `true`, range `fd00::ac:00/118` is reserved for IPv6 egress.
+
+---
+
 #### `AWS_MANAGE_ENIS_NON_SCHEDULABLE` (v1.12.6+)
 
 Type: Boolean as a String
 
 Default: `false`
 
-Specifies whether IPAMD should allocate or deallocate ENIs on a non-schedulable node. 
+Specifies whether IPAMD should allocate or deallocate ENIs on a non-schedulable node.
 
 ---
 
--- a/cmd/aws-vpc-cni-init/main.go
+++ b/cmd/aws-vpc-cni-init/main.go
@@ -21,6 +21,7 @@ import (
 	"github.com/aws/amazon-vpc-cni-k8s/utils"
 	"github.com/aws/amazon-vpc-cni-k8s/utils/cp"
 	"github.com/aws/amazon-vpc-cni-k8s/utils/imds"
+
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
@@ -33,10 +34,12 @@ const (
 	metadataMAC                     = "mac"
 	defaultDisableIPv4TcpEarlyDemux = false
 	defaultEnableIPv6               = false
+	defaultEnableIPv6Egress         = false
 
 	envDisableIPv4TcpEarlyDemux = "DISABLE_TCP_EARLY_DEMUX"
 	envEnableIPv6               = "ENABLE_IPv6"
 	envHostCniBinPath           = "HOST_CNI_BIN_PATH"
+	envEgressV6                 = "ENABLE_V6_EGRESS"
 )
 
 func getNodePrimaryIF() (string, error) {
@@ -110,19 +113,22 @@ func configureIPv6Settings(procSys procsyswrapper.ProcSys, primaryIF string) err
 		}
 		val, _ := procSys.Get(entry)
 		log.Infof("Updated %s to %s", entry, val)
-
-		entry = "net/ipv6/conf/all/forwarding"
+	}
+	// Check if IPv6 egress supporting is enabled in IPv4 cluster
+	ipv6EgressEnabled := utils.GetBoolAsStringEnvVar(envEgressV6, defaultEnableIPv6Egress)
+	if enableIPv6 || ipv6EgressEnabled {
+		entry := "net/ipv6/conf/all/forwarding"
 		err = procSys.Set(entry, "1")
 		if err != nil {
-			return errors.Wrap(err, "Failed to enable ipv6 forwarding")
+			return errors.Wrap(err, "Failed to enable IPv6 forwarding")
 		}
-		val, _ = procSys.Get(entry)
+		val, _ := procSys.Get(entry)
 		log.Infof("Updated %s to %s", entry, val)
 
 		entry = "net/ipv6/conf/" + primaryIF + "/accept_ra"
 		err = procSys.Set(entry, "2")
 		if err != nil {
-			return errors.Wrap(err, "Failed to enable ipv6 accept_ra")
+			return errors.Wrap(err, "Failed to enable IPv6 accept_ra")
 		}
 		val, _ = procSys.Get(entry)
 		log.Infof("Updated %s to %s", entry, val)
--- a/cmd/aws-vpc-cni/main.go
+++ b/cmd/aws-vpc-cni/main.go
@@ -41,6 +41,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
+	"strconv"
 	"strings"
 	"time"
 
@@ -49,15 +50,18 @@ import (
 
 	"github.com/containernetworking/cni/pkg/types"
 
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/cniutils"
 	"github.com/aws/amazon-vpc-cni-k8s/utils"
 	"github.com/aws/amazon-vpc-cni-k8s/utils/cp"
-	"github.com/aws/amazon-vpc-cni-k8s/utils/imds"
 )
 
 const (
 	egressPluginIpamSubnetV4     = "169.254.172.0/22"
+	egressPluginIpamSubnetV6     = "fd00::ac:00/118"
 	egressPluginIpamDstV4        = "0.0.0.0/0"
+	egressPluginIpamDstV6        = "::/0"
 	egressPluginIpamDataDirV4    = "/run/cni/v6pd/egress-v4-ipam"
+	egressPluginIpamDataDirV6    = "/run/cni/v4pd/egress-v6-ipam"
 	defaultHostCNIBinPath        = "/host/opt/cni/bin"
 	defaultHostCNIConfDirPath    = "/host/etc/cni/net.d"
 	defaultAWSconflistFile       = "/app/10-aws.conflist"
@@ -68,8 +72,10 @@ const (
 	defaultPodSGEnforcingMode    = "strict"
 	defaultPluginLogFile         = "/var/log/aws-routed-eni/plugin.log"
 	defaultEgressV4PluginLogFile = "/var/log/aws-routed-eni/egress-v4-plugin.log"
+	defaultEgressV6PluginLogFile = "/var/log/aws-routed-eni/egress-v6-plugin.log"
 	defaultPluginLogLevel        = "Debug"
-	defaultEnableIPv6            = "false"
+	defaultEnableIPv6            = false
+	defaultEnableIPv6Egress      = false
 	defaultRandomizeSNAT         = "prng"
 	defaultEnableNftables        = false
 	awsConflistFile              = "/10-aws.conflist"
@@ -86,12 +92,14 @@ const (
 	envPluginLogFile         = "AWS_VPC_K8S_PLUGIN_LOG_FILE"
 	envPluginLogLevel        = "AWS_VPC_K8S_PLUGIN_LOG_LEVEL"
 	envEgressV4PluginLogFile = "AWS_VPC_K8S_EGRESS_V4_PLUGIN_LOG_FILE"
+	envEgressV6PluginLogFile = "AWS_VPC_K8S_EGRESS_V6_PLUGIN_LOG_FILE"
 	envEnPrefixDelegation    = "ENABLE_PREFIX_DELEGATION"
 	envWarmIPTarget          = "WARM_IP_TARGET"
 	envMinIPTarget           = "MINIMUM_IP_TARGET"
 	envWarmPrefixTarget      = "WARM_PREFIX_TARGET"
 	envEnBandwidthPlugin     = "ENABLE_BANDWIDTH_PLUGIN"
 	envEnIPv6                = "ENABLE_IPv6"
+	envEnIPv6Egress          = "ENABLE_V6_EGRESS"
 	envRandomizeSNAT         = "AWS_VPC_K8S_CNI_RANDOMIZESNAT"
 	envEnableNftables        = "ENABLE_NFTABLES"
 )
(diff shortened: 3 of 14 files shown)


cmd/aws-vpc-cni-init/main.go
```diff
@@ -33,10 +34,12 @@ const (
 	metadataMAC                     = "mac"
 	defaultDisableIPv4TcpEarlyDemux = false
 	defaultEnableIPv6               = false
+	defaultEnableIPv6Egress         = false
 
 	envDisableIPv4TcpEarlyDemux = "DISABLE_TCP_EARLY_DEMUX"
 	envEnableIPv6               = "ENABLE_IPv6"
 	envHostCniBinPath           = "HOST_CNI_BIN_PATH"
+	envEgress                   = "ENABLE_V6_EGRESS"
```
jdn5126, line 42: nit: `envEgressV6`

cmd/aws-vpc-cni/main.go
```diff
@@ -196,39 +204,84 @@ func waitForInit() error {
...
+
+	if ipv4 {
+		key = "local-ipv4"
+	} else {
+		key = "ipv6"
+	}
+	hostIP, err = cniutils.GetNodeMetadata(key)
+	if err != nil {
+		if ipv4 {
+			log.WithError(err).Fatalf("failed to rtrieve local-ipv4 address in imds metadata")
```
jdn5126, line 220: nit: "retrieve"

cmd/aws-vpc-cni/main.go
```diff
@@ -196,39 +204,84 @@ func waitForInit() error {
...
+		egressEnabled = utils.GetBoolAsStringEnvVar(envEnIPv6Egress, defaultEnableIPv6Egress)
+		egressPluginLogFile = utils.GetEnv(envEgressV6PluginLogFile, defaultEgressV6PluginLogFile)
+		if egressEnabled {
+			nodeIP, err = getPrimaryIP(false)
+			// Node may not have a IPv6 address if there is no need to support IPv6 egress in eks IPv4 cluster
+			if err != nil {
+				log.Errorf("To support IPv6 egress, node primary ENI must have a global IPv6 address, error: %v", err)
+				// Global unicast IPv6 nodeIP is mandatory to support IPv6 egress traffic in EKS IPv4 cluster
+				// no return err here so that aws-node daemonset pod could start success
+				// chained egress-cni-plugin will emit error message about empty nodeIP if IPv6 egress is enabled
```
jdn5126, line 276: I think we can return error here, as the combination of setting `ENABLE_V6_EGRESS` and not configuring an IPv6 node IP is an error

pkg/utils/cniutils/cni_utils.go
```diff
@@ -74,6 +75,22 @@ func WaitForAddressesToBeStable(netLink netlinkwrapper.NetLink, ifName string, t
 	}
 }
 
+// GetNodeMetadata calling node local imds metadata service using provided key
+// return either a non-empty value or an error
+func GetNodeMetadata(key string) (string, error) {
```
jdn5126, line 80: Why do we need another function here? Seems like callers can just call `imds.GetMetaData(key)` directly.

jdn5126 (commented):
Changes look good me overall, thanks @wanyufe ! Just a few nits

cmd/aws-vpc-cni-init/main.go
wanyufe: fixed

cmd/aws-vpc-cni/main.go
wanyufe: fixed

cmd/aws-vpc-cni/main.go
wanyufe: changed.

pkg/utils/cniutils/cni_utils.go
wanyufe: Whole lopic inside GetNodeMetadata has loop to deal with cases like no error but value still empty.
```
	for {
		value, err = imds.GetMetaData(key)
		if err != nil {
			return "", err
		}
		if value != "" {
			return value, nil
		}
	}
```

cmd/aws-vpc-cni-init/main.go
```diff
@@ -110,13 +113,16 @@ func configureIPv6Settings(procSys procsyswrapper.ProcSys, primaryIF string) err
...
-
-		entry = "net/ipv6/conf/all/forwarding"
+	}
+	// Check if IPv6 egress supporting is enabled in IPv4 cluster
+	ipv6EgressEnabled := utils.GetBoolAsStringEnvVar(envEgress, defaultEnableIPv6Egress)
+	if enableIPv6 || ipv6EgressEnabled {
+		entry := "net/ipv6/conf/all/forwarding"
 		err = procSys.Set(entry, "1")
 		if err != nil {
 			return errors.Wrap(err, "Failed to enable ipv6 forwarding")
```
pmotyka, line 123: nit: inconsistent casing of ipv6 vs IPv6 (https://github.com/aws/amazon-vpc-cni-k8s/blob/3d9e67decb355cf39cf5cbc5927484d9ecc8dda9/cmd/aws-vpc-cni-init/main.go#L175) in error log messages

cmd/egress-cni-plugin/egress-plugin.log
pmotyka, line 1: Should this log be part of the pull request?

cmd/aws-vpc-cni-init/main.go
wanyufe: Thanks, will use `IPv6` in logs for consistency.

cmd/egress-cni-plugin/egressContext.go
```diff
@@ -233,101 +239,426 @@ func (c *EgressContext) SetupHostVeth(vethName string) error {
...
+	return nil
+}
+
+// cmdAddEgressV6 exec necessary settings to support IPv6 egress traffic in EKS IPv4 cluster
+func (ec *egressContext) cmdAddEgressV6() (err error) {
+	// Per best practice, a new veth pair is created between container ns and node ns
+	// this newly created veth pair is used for container's egress IPv6 traffic
+	// NOTE:
+	//	1. link-local IPv6 addresses are automatically assigned to veth's both ends.
+	//	2. unique-local IPv6 address allocated from host-local IPAM plugin is assigned to veth's container end only
```
jdn5126, line 359: nit: alignment

cmd/egress-cni-plugin/egressContext.go
```diff
@@ -233,101 +239,426 @@ func (c *EgressContext) SetupHostVeth(vethName string) error {
...
+	hostInterface, containerInterface, err := ec.setupContainerVethV6()
+	if err != nil {
+		ec.Log.Errorf("veth created failed, ns: %s name: %s, mtu: %d, ipam-result: %+v err: %v",
+			ec.NsPath, ec.NetConf.IfName, ec.Mtu, *ec.TmpResult, err)
+		return err
+	}
+	ec.Log.Debugf("veth pair created for container IPv6 egress traffic, container interface: %s ,host interface: %s",
+		containerInterface.Name, hostInterface.Name)
+
+	containerIPv6, err := ec.getContainerIpv6GlobalAddrs(containerInterface.Name)
```
jdn5126, line 386: Sorry, I think I asked this before, but I can't find the conversation. This function finds the global IPv6 address assigned to the container veth intf. Isn't this just the address we got from host-ipam? i.e. shouldn't it be available in `ec.TmpResult`?

cmd/egress-cni-plugin/egressContext.go
```diff
@@ -233,101 +239,426 @@ func (c *EgressContext) SetupHostVeth(vethName string) error {
...
+			var containerLink netlink.Link
+			var _err error
+			ec.Log.Debugf("trying to get link by name %s", ec.NetConf.IfName)
+			containerLink, _err = ec.Link.LinkByName(ec.NetConf.IfName)
+			if _err != nil {
+				if !cniutils.IsLinkNotFoundError(_err) {
+					ec.Log.Debugf("failed to get link by name %s: %v", ec.NetConf.IfName, _err)
+				}
+				// must return when err, following code needs containerLink
+				return _err
```
jdn5126, line 455: I think we should return `nil` here, as this means the container veth was already deleted somehow, so there is nothing we can clean up

cmd/egress-cni-plugin/egressContext.go
```diff
@@ -233,101 +239,426 @@ func (c *EgressContext) SetupHostVeth(vethName string) error {
...
 			}
 
-			if err != nil && err == ip.ErrLinkNotFound {
-				c.Log.Debugf("DEL: Link Not Found, returning", err)
-				return nil
+			contIPAddrs, _err = ec.Link.AddrList(containerLink, netlink.FAMILY_V6)
```
jdn5126, line 458: Is the container IPv6 address available in PrevResult? I think we might be able to get it from `ec.NetConfig.PrevResult`

cmd/egress-cni-plugin/egressContext.go
```diff
@@ -233,101 +239,426 @@ func (c *EgressContext) SetupHostVeth(vethName string) error {
...
+			_err = ec.Link.LinkDel(containerLink)
+			if _err != nil {
+				if !cniutils.IsLinkNotFoundError(_err) {
+					ec.Log.Debugf("failed to delete veth %s in container: %v", containerLink.Attrs().Name, _err)
+				}
+			} else {
+				ec.Log.Debugf("Successfully deleted veth %s in container", containerLink.Attrs().Name)
+			}
+			return _err
 		})
```
jdn5126, line 474: We don't need to delete the veth in the container, right? Returning `nil` will cause the container runtime to delete it. I see that we do delete it in the IPv4 version as well, but I think we actually don't need either. Might be better to not do it for v6 to confirm, then we can always refactor v4 if there's time later

cmd/egress-cni-plugin/egressContext.go
```diff
@@ -233,101 +239,426 @@ func (c *EgressContext) SetupHostVeth(vethName string) error {
...
+}
+
+func (ec *egressContext) disableInterfaceIPv6(ifName string) error {
+	return ec.Ns.WithNetNSPath(ec.NsPath, func(hostNS ns.NetNS) error {
+		var entry = "net/ipv6/conf/" + ifName + "/disable_ipv6"
+		if content, err := ec.Procsys.Get(entry); err == nil {
+			if strings.TrimSpace(content) == "1" {
+				return nil
+			}
+		}
```
jdn5126, line 496: nit: since checking does a syscall as well, we may as well just blindly set it to 1 to decrease the number of syscalls

cmd/egress-cni-plugin/main_test.go
```diff
@@ -58,20 +62,20 @@ func TestCmdAddV4(t *testing.T) {
 							{"name":"eth0","sandbox":"/var/run/netns/cni-266298c1-b141-9c7f-f26b-97ff084f3fcc"},
 							{"name":"dummy36e5b0ee702","mac":"0","sandbox":"0"}],
 					"ips":
-						[{"version":"4","interface":1,"address":"192.168.13.226/32"}],
+						[{"version":"6","interface":1,"address":"2600:1f16:828:c404:af46:9f44:d2ea:4569/128"}],
```
jdn5126, line 65: Is this a typo? Function name is `TestCmdAddV4`, but this is a v6 address

cmd/egress-cni-plugin/snat/snat.go
```diff
@@ -91,31 +92,45 @@ func Add(ipt iptableswrapper.IPTablesIface, nodeIP, src net.IP, multicastRange,
...
-		return err
+func Del(ipt iptableswrapper.IPTablesIface, src net.IP, chain, comment string) (err error) {
+	var ipAddr string
+	if src == nil {
+		// since src ip is not provided, we list all nat POSTROUTING rules then filter by matching chain
+		// which is a hash of container id as part of its value, and it is unique for each container's nat POSTROUTING rule
+		// if no matching rules found, it means rule had already been removed
+		var rules []string
+		rules, err = ipt.List("nat", "POSTROUTING")
+		for _, rule := range rules {
```
jdn5126, line 103: Looping through all of the rules can be an expensive operation, depending on the size of the cluster. I think we should stick with hard-coding the deletes

pkg/utils/cniutils/cni_utils.go
jdn5126: Is the concern that the ipv6 address may not be available in IMDS yet, so we get an empty value and retry until it is? I think we can skip that, as if the address is not available, then aws-node pod will error and restart, at which case it will become available. My concern with this for loop is that it could run indefinitely if no IPv6 addr is configured.

jdn5126 (commented):
Did a pass focusing on optimizations, and I am realizing we may be able to get some optimizations on the delete path. Hoping so, at least.

pkg/utils/cniutils/cni_utils.go
wanyufe: if no ipv6 configured, it will return err which is some kind of 404, so it will not run indefinitely.
I am not the original author who added the for, I am guessing it is for imds service throttling.

cmd/egress-cni-plugin/egressContext.go
wanyufe: fixed.

cmd/egress-cni-plugin/egressContext.go
wanyufe: This is a direct translation of logic [here](https://github.com/containernetworking/plugins/blob/main/pkg/ip/ipforward_linux.go#L56-L61), the translation is to make it unit testable.

cmd/egress-cni-plugin/snat/snat.go
wanyufe: I did more test in eks cluster, and found out that `DEL` action was called 3 times in a normal pod removal, the first time with netns and preResult having value, the following 2 without preResult and netns value.

The currently production plugins worked the same.

Based on above observation, I changed the `del` code back to original version -- without loop through links to find src ip.

cmd/egress-cni-plugin/main_test.go
jdn5126: I see, thanks for explaining

pkg/utils/cniutils/cni_utils.go
jdn5126: I see, we can leave it as is then

cmd/egress-cni-plugin/egressContext.go
wanyufe: the updated version combined v4 del and v6 del, and removed v6 mergeResult, so the prevResult will not have ip address anymore. The v4 did not have ip stored in result in our original code.

cmd/egress-cni-plugin/main.go
```diff
@@ -45,11 +43,19 @@ func cmdAdd(args *skel.CmdArgs) error {
...
 	if err != nil {
 		return fmt.Errorf("failed to parse config: %v", err)
 	}
+
+	ec.Log.Debugf("Received an Add request: nsPath: %s conf=%+v", ec.NsPath, *ec.NetConf)
+	// uncomment following lines to debug inputs
+	//stdin := args.StdinData
+	//args.StdinData = nil
+	//ec.Log.Debugf("args: %+v, stdinData: %+v", *args, string(stdin))
+	//args.StdinData = stdin
```
jdn5126, line 57: nit: I think we can remove these

cmd/egress-cni-plugin/main.go
```diff
@@ -76,77 +80,75 @@ func add(args *skel.CmdArgs, ec *EgressContext) (err error) {
...
-
-	// IPv4 egress
-	ec.NetConf.IfName = EgressIPv4InterfaceName
-	// explicitly set err var so that above defer function can call ipam.ExecDel
-	err = ec.CmdAddEgressV4()
 	return err
 }
 
-func cmdDel(args *skel.CmdArgs) error {
+func cmdDel(args *skel.CmdArgs) (err error) {
```
jdn5126, line 111: nit: this should be fine with just returning `error`, right?

jdn5126 (dismissed):

jdn5126 (dismissed):

cmd/aws-vpc-cni/main.go
```diff
@@ -49,15 +50,18 @@ import (
...
 
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/cniutils"
 	"github.com/aws/amazon-vpc-cni-k8s/utils"
 	"github.com/aws/amazon-vpc-cni-k8s/utils/cp"
-	"github.com/aws/amazon-vpc-cni-k8s/utils/imds"
 )
 
 const (
 	egressPluginIpamSubnetV4     = "169.254.172.0/22"
+	egressPluginIpamSubnetV6     = "fd00::/8"
```
jdn5126, line 60: Let's use a smaller ULA range here, say `fd00::/118`

pkg/networkutils/network.go
```diff
@@ -224,41 +224,54 @@ func findPrimaryInterfaceName(primaryMAC string) (string, error) {
...
+}
+
+// Set up a rule to block traffic directed to v4/v6 egress interface of the Pod
+func (n *linuxNetwork) setupRuleToBlockNodeLocalAccess(protocol iptables.Protocol) error {
+	ipVersion := "v4"
+	localIpCidr := "169.254.172.0/22"
+	iptableCmd := "iptables"
+	if protocol == iptables.ProtocolIPv6 {
+		ipVersion = "v6"
+		localIpCidr = "fd00::/8"
```
jdn5126, line 248: Let's use a smaller ULA range here, say `fd00::/118`

cmd/aws-vpc-cni/main.go
wanyufe: I think narrow it down is a good idea, but not sure if fd00::/118 is a good choice.

`"CreateTransitGatewayConnectPeerRequest$InsideCidrBlocks": "The range of inside IP addresses that are used for BGP peering. You must specify a size /29 IPv4 CIDR block from the 169.254.0.0/16 range. The first address from the range must be configured on the appliance as the BGP IP address. You can also optionally specify a size /125 IPv6 CIDR block from the fd00::/8 range.`

README.md
```diff
@@ -105,13 +105,23 @@ The following environment variables are available, and all of them are optional.
 
 ---
 
+#### `ENABLE_V6_EGRESS` (v1.13.0+)
+
+Type: Boolean as a String
+
+Default: `false`
+
+Specifies whether PODs in v4 cluster support IPv6 egress. If env is set to `true`, range fd00::ac:00/118 is reserved for PODs IPv6 egress.
```
jdn5126, line 114: nit: let's surround `fd00::ac:00/118` in backticks, and change "reserved for PODs IPv6 egress" to "reserved for IPv6 egress"

jdn5126 (approved):

README.md
wanyufe: no problem.

jdn5126 (approved):

Issue #2363, opened 2023-04-28
IP Deallocation delay when multiple enis attached to host


bikashmishra100:
**What happened**:
The IP deallocation gets delayed in a particular case. When there are multiple enis attached to host and during the iteration to check freeable cidrs, if the first eni does not return anything, the entire loop exits it seems.

On this occasion, there are two enis attached to the host, one ending with `d6ac` and another ending with `86af`.

> {"level":"debug","ts":"2023-04-22T00:00:03.178Z","caller":"datastore/data_store.go:1104","msg":"ENI eni-***d6ac cannot be deleted because it is primary"}
{"level":"debug","ts":"2023-04-22T00:00:03.178Z","caller":"datastore/data_store.go:1104","msg":"ENI eni-***865f cannot be deleted because it has pods assigned"}

We are tracking two IP addresses which are allocated.

> {"level":"info","ts":"2023-04-22T00:00:50.702Z","caller":"ipamd/ipamd.go:1412","msg":"Trying to add 10.246.106.92"}
{"level":"info","ts":"2023-04-22T00:00:50.702Z","caller":"ipamd/ipamd.go:1536","msg":"Adding 10.246.106.92/32 to DS for eni-***865f"}
{"level":"info","ts":"2023-04-22T00:00:50.702Z","caller":"ipamd/ipamd.go:1536","msg":"IP already in DS"}
{"level":"info","ts":"2023-04-22T00:00:50.702Z","caller":"ipamd/ipamd.go:1412","msg":"Trying to add 10.246.105.43"}
{"level":"info","ts":"2023-04-22T00:00:50.702Z","caller":"ipamd/ipamd.go:1536","msg":"Adding 10.246.105.43/32 to DS for eni-***865f"}
{"level":"info","ts":"2023-04-22T00:00:50.702Z","caller":"ipamd/ipamd.go:1536","msg":"IP already in DS"}

We get a unassignment of IP `10.246.105.43`.

> {"level":"info","ts":"2023-04-22T00:03:32.592Z","caller":"rpc/rpc.pb.go:731","msg":"Send DelNetworkReply: IPv4Addr 10.246.105.43, DeviceNumber: 1, err: <nil>"}

When the decrease datastore is invoked, the `d6ac` eni comes first and the loop exits when the over count is > 1.

[...]

PR #2368, opened 2023-05-03
Deallocate IP address according to warm IP target when multiple enis are present

bikashmishra100:
**What type of PR is this?**
bug

https://github.com/aws/amazon-vpc-cni-k8s/issues/2363

Previously when multiple enis were present on the host, the deallocation loop used to exit when first eni did not have any available IP addresses to free. This PR fixes that. Also the PR deallocates exactly the required addresses below the target in case of multiple enis.

**Automation added to e2e**:

No

No

No

No

```release-note

```

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -771,7 +771,7 @@ func (c *IPAMContext) tryUnassignCidrsFromAll() {
 			cidrs := c.dataStore.FindFreeableCidrs(eniID)
 			if cidrs == nil {
 				log.Errorf("Error finding unassigned IPs for ENI %s", eniID)
-				return
+				continue
 			}
 
 			// Free the number of Cidrs `over` the warm IP target, unless `over` is greater than the number of available Cidrs on
@@ -799,6 +799,11 @@ func (c *IPAMContext) tryUnassignCidrsFromAll() {
 
 			// Deallocate Cidrs from the instance if they are not used by pods.
 			c.DeallocCidrs(eniID, deletedCidrs)
+
+			// reduce the deallocation target, if the deallocation target is achieved, we can exit
+			if over = over - len(deletedCidrs); over <= 0 {
+				break
+			}
 		}
 	}
 }
--- a/pkg/ipamd/ipamd_test.go
+++ b/pkg/ipamd/ipamd_test.go
@@ -22,6 +22,7 @@ import (
 	"reflect"
 	"sort"
 	"testing"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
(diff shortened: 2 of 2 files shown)


pkg/ipamd/ipamd.go
```diff
@@ -770,11 +770,16 @@ func (c *IPAMContext) tryUnassignCidrsFromAll() {
 	if over > 0 {
 		eniInfos := c.dataStore.GetENIInfos()
 		for eniID := range eniInfos.ENIs {
+			// if the deallocated target is achieved, we can exit
+			if over <= 0 {
+				break
+			}
```
jdn5126, line 776: nit: this check would be cleaner at the end of the loop, i.e. after we decrease `over`

pkg/ipamd/ipamd_test.go
```diff
@@ -745,6 +746,41 @@ func testIncreasePrefixPool(t *testing.T, useENIConfig bool) {
...
+	mockContext.dataStore.AddENI(secENIid, secDevice, true, false, false)
+	mockContext.dataStore.AddIPv4CidrToStore(secENIid, testAddr11, false)
+	mockContext.dataStore.AddIPv4CidrToStore(secENIid, testAddr12, false)
+	mockContext.dataStore.AssignPodIPv4Address(datastore.IPAMKey{ContainerID: "container2"}, datastore.IPAMMetadata{K8SPodName: "pod2"})
+
+	m.awsutils.EXPECT().DeallocPrefixAddresses(gomock.Any(), gomock.Any()).Times(1)
+	m.awsutils.EXPECT().DeallocIPAddresses(gomock.Any(), gomock.Any()).Times(1)
+
+	mockContext.decreaseDatastorePool(10 * time.Second)
+	mockContext.decreaseDatastorePool(10 * time.Second)
```
jdn5126, line 781: why are we calling this twice?

pkg/ipamd/ipamd_test.go
```diff
@@ -745,6 +746,41 @@ func testIncreasePrefixPool(t *testing.T, useENIConfig bool) {
 	mockContext.increaseDatastorePool(ctx)
 }
 
+func TestDecreaseIPPool(t *testing.T) {
```
jdn5126, line 749: Can you add some comments in this test case explaining what the goal is, what the setup is, etc.

jdn5126 (commented):
Thanks for the contribution @bikashmishra100 ! Just a few nits from me

pkg/ipamd/ipamd.go
bikashmishra100: sure, done

pkg/ipamd/ipamd_test.go
bikashmishra100: I have added some, please let me know if it helps

pkg/ipamd/ipamd_test.go
bikashmishra100: I added the call just to ensure that multiple deallocations back to back does not take the target below warm ip. If it is extraneous, i can drop it. I added a comment.

jdn5126: Ah, I see, that's fine with me. I think we should assert the datastore target state (calling `datastoreTargetState`) is fine, before the first call to `decreaseDatastorePool`, and then after each. That way, we can assert that before we do anything, we are two over, and after the first call, we are 0 over, and after the second call, we are still 0 over

pkg/ipamd/ipamd_test.go
jdn5126: That's perfect, thank you!

pkg/ipamd/ipamd_test.go
bikashmishra100: I updated the assertions

jdn5126 (approved):

jdn5126:
@bikashmishra100 sorry for the delay. I approved these changes, but it looks like this branch requires merge conflict resolution. Can you rebase and resolve conflicts? Also @jayanthvn can you review as well?

bikashmishra100:
No problem at all @jdn5126 . I have squashed all the commits and rebased it.

Issue #2360, opened 2023-04-25
Allow configuration of the leaked ENI reconciliation


alanty:
**What would you like to be added**:
As of v1.12.6 the CNI checks for "leaked" ENIs every hour: 
https://github.com/aws/amazon-vpc-cni-k8s/blob/v1.12.6/pkg/awsutils/awsutils.go#L440
```
	// Clean up leaked ENIs in the background
	if !disableENIProvisioning {
		go wait.Forever(cache.cleanUpLeakedENIs, time.Hour)
	}
```
rather than having the time.hour hardcoded could we parameterize that interval, or allow this process to only run once on node init? 

**Why is this needed**:

This check eventually issues a DescribeNetworkInterfaces command to get all available ENIs that were allocated by the AWS CNI plugin, but were not deleted.
https://github.com/aws/amazon-vpc-cni-k8s/blob/v1.12.6/pkg/awsutils/awsutils.go#L1748

In accounts with a large number of nodes or network interfaces this can cause a large number of Describe calls. By controlling the reconciliation we can continue to allow the leaked ENIs to be cleaned up, but reduce the unnecessary Describe calls for large accounts.

jayanthvn:
@alanty - We are thinking about moving this to the control plane (VPC RC) hence planning to add this knob - https://github.com/aws/amazon-vpc-cni-k8s/pull/1641 but that would need some additional testing and release so in the mean time we can think of making it configurable. It will be still suboptimal even if it is called once since all nodes will make a call once and if there is a node churn then we won't get much of the benefit.

jdn5126:
Closing as fixed by https://github.com/aws/amazon-vpc-cni-k8s/pull/2370

Issue #1223, opened 2020-09-18
Avoid detaching ENIs on nodes being drained


mogren:
**What would you like to be added**:
We should prevent ipamd from trying to free ENIs when a node is about to terminated. 

For spot instances we could do something similar to the [aws-node-termination-handler](https://github.com/aws/aws-node-termination-handler) and [check some metadata endpoints](https://github.com/aws/aws-node-termination-handler#metadata-endpoints).

For the case where a node is cordoned off before being terminated, meaning it is marked as "unschedulable", we should be able to check this node taints before trying to attach or detach any ENIs.

**Why is this needed**:
Since there is no EC2 API call to directly "delete" an ENI that is attached, instead they first have to be detached, which takes a few seconds, then deleted. If the instance gets terminated after the ENI has been detached, but before it has been deleted, it will be leaked. This leaked ENI can prevent Security Groups and VPCs from being deleted and require manual clean up. 

Related issues: #608 #69, #690

jayanthvn:
There are 2 parts to this - 1. There is an internal tracking ticket with Ec2 team to see if this can be handled. 2. IPAMD can check if the node is marked as unschedulable then prevent ENI deletion. But the cleaner approach is #1.

jayanthvn:
Pending POC on the IPAMD changes. Also following up with ec2 team if this can be handled internally.

hurriyetgiray-ping:
Do we have an ETA on this fix please? This is causing us a lot of issues, and substantial coding to workaround it.

hurriyetgiray-ping:
Can this be tagged as a bug please, as this is more than a feature request?

jayanthvn:
Hi @hurriyetgiray-ping,

What is the CNI version you are using and are you having short lived clusters in the account? currently we have a background thread - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/awsutils/awsutils.go#L382 which cleans up leaked ENIs in the account and I agree this would need atleast one aws-node pod running in the account.

hurriyetgiray-ping:
Hello @jayanthvn. Thank you for your response and also for tagging this issue as a bug. CNI version is V1.7.5 as per the below kubectl describe.  
```
bash-4.2$ kubectl describe daemonset aws-node --namespace kube-system | grep Image | cut -d "/" -f 2
amazon-k8s-cni-init:v1.7.5
amazon-k8s-cni:v1.7.5 
```
[...]

PR #2370, opened 2023-05-03
Introduce `DISABLE_LEAKED_ENI_CLEANUP` to disable leaked ENI cleanup task

jdn5126:
**What type of PR is this?**
feature

#2360 

This PR adds a new environment variable, `DISABLE_LEAKED_ENI_CLEANUP`, that disables the leaked ENI cleanup task that IPAMD runs every hour. It also disables the task for IPv6 clusters. The task runs on each node every hour and creates a burst of EC2 API calls that can lead to throttling in larger accounts. Eventually, the task of cleaning up leaked ENIs will move to the VPC Resource Controller. This solution is to help customers in the interim.

N/A

Manually verified that task does not run when environment variable is set and that existing integration tests pass.

**Automation added to e2e**:
N/A

N/A

No, Yes

No

Yes

```release-note
Introduce DISABLE_LEAKED_ENI_CLEANUP to disable leaked ENI cleanup task
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -658,6 +658,15 @@ Default: empty
 
 Specify the EC2 endpoint to use. This is useful if you are using a custom endpoint for EC2. For example, if you are using a proxy for EC2, you can set this to the proxy endpoint. Any kind of URL or IP address is valid such as `https://localhost:8080` or `http://ec2.us-west-2.customaws.com`. If this is not set, the default EC2 endpoint will be used.
 
+#### `DISABLE_LEAKED_ENI_CLEANUP` (v1.13.0+)
+
+Type: Boolean as a String
+
+Default: `false`
+
+On IPv4 clusters, IPAMD schedules an hourly background task per node that cleans up leaked ENIs. Setting this environment variable to `true` disables that job. The primary motivation to disable this task is to decrease the amount of EC2 API calls made from each node.
+Note that disabling this task should be considered carefully, as it requires users to manually cleanup ENIs leaked in their account. See [#1223](https://github.com/aws/amazon-vpc-cni-k8s/issues/1223) for a related discussion.
+
 ### VPC CNI Feature Matrix
 
 IP Mode | Secondary IP Mode | Prefix Delegation | Security Groups Per Pod | WARM & MIN IP/Prefix Targets | External SNAT
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -118,8 +118,11 @@ const (
 	// should not manage it in any form.
 	eniNoManageTagKey = "node.k8s.amazonaws.com/no_manage"
 
-	// disableENIProvisioning is used to specify that ENI doesn't need to be synced during initializing a pod.
+	// disableENIProvisioning is used to specify that ENIs do not need to be synced during initializing a pod.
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -658,6 +658,14 @@ Default: empty
 
 Specify the EC2 endpoint to use. This is useful if you are using a custom endpoint for EC2. For example, if you are using a proxy for EC2, you can set this to the proxy endpoint. Any kind of URL or IP address is valid such as `https://localhost:8080` or `http://ec2.us-west-2.customaws.com`. If this is not set, the default EC2 endpoint will be used.
 
+#### `DISABLE_LEAKED_ENI_CLEANUP` (v1.13.0+)
+
+Type: Boolean as a String
+
+Default: `false`
+
+On IPv4 clusters, IPAMD schedules an hourly background task per node that cleans up leaked ENIs. Setting this environment variable to `true` disables that job. The primary motivation to disable this task is to decrease the amount of EC2 API calls made from each node.
```
jayanthvn, line 667: Nit: Lets add a note that this has to be considered carefully since this might lead to leaked ENIs (https://github.com/aws/amazon-vpc-cni-k8s/issues/1223) and would need manual cleanup..

pkg/ipamd/ipamd.go
```diff
@@ -391,10 +393,10 @@ func New(rawK8SClient client.Client, cachedK8SClient client.Client) (*IPAMContex
 	c.enablePrefixDelegation = usePrefixDelegation()
 	c.enableIPv4 = isIPv4Enabled()
 	c.enableIPv6 = isIPv6Enabled()
+	c.disableENIProvisioning = disableENIProvisioning()
+	disableLeakedENICleanup := c.disableENIProvisioning || disableLeakedENICleanup()
```
jayanthvn, line 397: Nit: Maybe lets find another variable name - `disableLeakedENICleanup` internally now holds `disableENIProvisioning`...

jayanthvn (commented):
Minor nits..

README.md
jdn5126: added in latest revision

pkg/ipamd/ipamd.go
jdn5126: I decided to move the `disableENIProvisioning` check into the function, the same as the IPv6 check

pkg/ipamd/ipamd.go
```diff
@@ -1771,10 +1767,18 @@ func getMinimumIPTarget() int {
 	return noMinimumIPTarget
 }
 
-func disablingENIProvisioning() bool {
+func disableENIProvisioning() bool {
 	return getEnvBoolWithDefault(envDisableENIProvisioning, false)
 }
 
+func disableLeakedENICleanup() bool {
```
jayanthvn, line 1774: Can we rename this? I am not getting a good variable name but something like `disableENIOperations`

jdn5126: This is used exclusively for disabling the leaked ENI cleanup task, though. It doesn't disable other ENI operations. `disableENIProvisioning` is the flag we use for cases like not allocating or deallocating an ENI

jayanthvn: Will sync up offline on this..

jayanthvn (approved):
discussed offline, lgtm, we will consider refactoring later time.

PR #2375, opened 2023-05-08
Add parameters for cni upgrade downgrade test

jaydeokar:
**What type of PR is this?**
feature 

N/A

To pass parameters required for running cni-upgrade-downgrade integration test

N/A

Yes on 1.25 cluster

**Automation added to e2e**:
N/A

N/A

N/A

N/A

N/A


scripts/run-ginkgo-integration-suite.sh
```diff
@@ -24,6 +24,15 @@ function load_test_parameters(){
   if [[ ! -z $TARGET_ADDON_VERSION ]]; then
     EXTRA_OPTIONS+=" --target-addon-version $TARGET_ADDON_VERSION"
   fi
+
+  if [[ ! -z $INITIAL_MANIFEST_FILE ]]; then
+    EXTRA_OPTIONS+=" --initial-manifest-file $INITIAL_MANIFEST_FILE"
+  fi
+
+  if [[ ! -z $TARGET_MANIFEST_FILE ]]; then
```
jdn5126, line 32: Question: are these two parameters exclusive? Like can they both be set?

jaydeokar: Not really. I didn't want to complicate the logic here since we already have that check in the test. The test checks if we have both the parameters defined and not empty.

jdn5126 (approved):

PR #2377, opened 2023-05-09
Fix IPv6 test flakiness around updating veth prefix

jdn5126:
**What type of PR is this?**
bug

N/A

This PR resolves test flakiness in the CNI IPv6 integration test suite. Tests would occasionally fail as pods would be created with a stale veth prefix and MTU.

After updating the veth prefix and MTU in the daemonset, the tests would wait for the daemonset pods to get to "Ready" state before continuing. Even after the pods are "Ready", we need to wait for the conflist to get updated with the new values and for kubelet/container-runtime (depending on k8s version) to be ready to call CNI ADD with the new command line args.

The fix here is to add a short sleep (5 seconds) after the veth prefix and MTU are updated. This was chosen as there is no more granular way that I could find to make sure kubelet/container-runtime (depending on k8s version) would pass the latest values via the command line to CNI ADD.

I added the equivalent waits to the CNI IPv4 host networking test as this issue is theoretically possible there as well. Failures happened most frequently on k8s versions < 1.24, which is when kubelet called CNI add instead of container-runtime.

N/A

Manually reproduced issue and verified that these changes prevent it from reproducing again. Getting further coverage through integration test runners. 

**Automation added to e2e**:
N/A

No

No, Yes

No

No

```release-note
Fix IPv6 integration test flakiness
```

### DIFF
--- a/test/integration/cni/host_networking_test.go
+++ b/test/integration/cni/host_networking_test.go
@@ -57,6 +57,9 @@ var _ = Describe("test host networking", func() {
 				"AWS_VPC_ENI_MTU":            DEFAULT_MTU_VAL,
 				"AWS_VPC_K8S_CNI_VETHPREFIX": DEFAULT_VETH_PREFIX,
 			})
+			// After updating daemonset pod, we must wait until conflist is updated so that container-runtime calls CNI ADD with the latest VETH prefix and MTU.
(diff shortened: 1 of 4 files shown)


orsenthil (approved):
LGTM.

PR #2397, opened 2023-06-01
Added an environment variable to configure the pod IP cooldown - 2378

arturzeitler:
**What type of PR is this?**
feature

2378

Fixes an open issue.
Added new environment variable ADDRESS_COOLING_PERIOD in charts/aws-vpc-cni/values.yaml expecting an int as string.
pkg/ipamd/datastore/data_store.go defines two envAddressCoolingPeriod with env var name and defaultAddressCoolingPeriod as 30s time.Duration in const section. Then a new function getCoolingPeriod, returning a time.Duration, either returns the default cooling period of 30s or the custom one, if it is not smaller than 0. Also a new logger is defined to log in case the custom duration is used. There was no existing logger in the required scope. getCoolingPeriod is called in inCoolingPeriod with an AddressInfo receiver. Finally, in data_store_test.go, the two hardcoded instances of time.Sleep(30 * time.Second) are replaced by calls to getCoolingPeriod.

Only ran the docker unit-tests. have not yet tested it on a running cluster.

**Automation added to e2e**:
No

No

Very unlikely. Did not test it on running cluster.

No

A new environment variable is added, so kind of user facing.

```release-note

```


charts/aws-vpc-cni/values.yaml
```diff
@@ -57,6 +57,7 @@ env:
   DISABLE_NETWORK_RESOURCE_PROVISIONING: "false"
   ENABLE_IPv4: "true"
   ENABLE_IPv6: "false"
+  ADDRESS_COOLING_PERIOD: "30"
```
jdn5126, line 60: nit: how about `IP_COOLDOWN_PERIOD`?

charts/aws-vpc-cni/values.yaml
```diff
@@ -81,25 +82,26 @@ podLabels: {}
 securityContext:
   capabilities:
     add:
-    - "NET_ADMIN"
-    - "NET_RAW"
+      - "NET_ADMIN"
```
jdn5126, line 85: Please revert all of these unnecessary changes

pkg/ipamd/datastore/data_store.go
```diff
@@ -266,9 +268,27 @@ func (addr AddressInfo) Assigned() bool {
 	return !addr.IPAMKey.IsZero()
 }
 
+var log = logger.Get()
+
+// getCoolingPeriod returns the time duration in seconds configured by ADDRESS_COOLING_PERIOD env variable
+func getCoolingPeriod() time.Duration {
```
jdn5126, line 274: This can be a function of `DataStore`, and it can use `ds.log`

pkg/ipamd/datastore/data_store.go
```diff
@@ -266,9 +268,27 @@ func (addr AddressInfo) Assigned() bool {
...
+		log.Debugf("Using ADDRESS_COOLING_PERIOD %v", input)
+		return time.Duration(input) * time.Second
+	}
+	return defaultAddressCoolingPeriod
+}
+
 // InCoolingPeriod checks whether an addr is in addressCoolingPeriod
 func (addr AddressInfo) inCoolingPeriod() bool {
-	return time.Since(addr.UnassignedTime) <= addressCoolingPeriod
+	return time.Since(addr.UnassignedTime) <= getCoolingPeriod()
```
jdn5126, line 291: nit: Set it during DataStore construction and then you can access a member variable instead of having the overhead of a function call each time

jdn5126 (commented):
@arturzeitler thanks for the contribution! You will also need to add a README entry for this new environment variable. We should have very strict wording around this being a "use at your own risk" setting.

I also think we need to determine whether 0 is an acceptable value.

Also, we should parse this in `aws-vpc-cni/main.go` and refuse to start if it is not set to a valid integer

charts/aws-vpc-cni/values.yaml
arturzeitler: Yeah, that's clearer. Will change it.

pkg/ipamd/datastore/data_store.go
arturzeitler: I'm not sure about this one. If addressCoolingPeriod is set as a member variable of Datastore during NewDataStore, then it would not be available while calling (addr AddressInfo) inCoolingPeriod. Alternatively, addressCoolingPeriod could be a normal variable in the package scope.

README.md
```diff
@@ -669,6 +669,16 @@ Default: `false`
 On IPv4 clusters, IPAMD schedules an hourly background task per node that cleans up leaked ENIs. Setting this environment variable to `true` disables that job. The primary motivation to disable this task is to decrease the amount of EC2 API calls made from each node.
 Note that disabling this task should be considered carefully, as it requires users to manually cleanup ENIs leaked in their account. See [#1223](https://github.com/aws/amazon-vpc-cni-k8s/issues/1223) for a related discussion.
 
+#### `IP_COOLDOWN_PERIOD` (v1.13.0+)
```
jdn5126, line 672: nit: this will become `v1.13.1+` now

README.md
```diff
@@ -669,6 +669,16 @@ Default: `false`
...
+
+Type: Integer as a String
+
+Default: `30`
+
+Valid Values: Positive Integer
```
jdn5126, line 678: nit: I would specify that 0 is not a supported value

charts/aws-vpc-cni/values.yaml
```diff
@@ -57,6 +57,7 @@ env:
   DISABLE_NETWORK_RESOURCE_PROVISIONING: "false"
   ENABLE_IPv4: "true"
   ENABLE_IPv6: "false"
+  IP_COOLDOWN_PERIOD: "30"
```
jdn5126, line 60: No entry is needed here, as this is not a mandatory environment variable. Let's revert all changes in this file

cmd/aws-vpc-cni/main.go
```diff
@@ -101,6 +101,7 @@ const (
 	envEnIPv6Egress          = "ENABLE_V6_EGRESS"
 	envRandomizeSNAT         = "AWS_VPC_K8S_CNI_RANDOMIZESNAT"
 	envEnableNftables        = "ENABLE_NFTABLES"
+	envIPCoolingPeriod       = "IP_COOLING_PERIOD"
```
jdn5126, line 104: nit: let's use a default value as well, to follow the others

cmd/aws-vpc-cni/main.go
```diff
@@ -354,6 +355,17 @@ func validateEnvVars() bool {
 	warmPrefixTarget := utils.GetEnv(envWarmPrefixTarget, "0")
 	minimumIPTarget := utils.GetEnv(envMinIPTarget, "0")
 
+	// Validate that IP_COOLING_PERIOD is a valid integer
+	ipCoolingPeriod, err := utils.GetIntAsStringEnvVar(envIPCoolingPeriod, 30)
+	if ipCoolingPeriod < 1 {
+		log.Errorf("IP_COOLING_PERIOD cannot be smaller than 1")
+		return false
+	}
+	if err != nil {
```
jdn5126, line 364: nit: let's switch the order of these checks and first validate that it is an integer before doing integer compare

pkg/ipamd/datastore/data_store.go
```diff
@@ -78,6 +79,8 @@ const backfillNetworkIface = "unknown"
 // ErrUnknownPod is an error when there is no pod in data store matching pod name, namespace, sandbox id
 var ErrUnknownPod = errors.New("datastore: unknown pod")
 
+var ipCoolingPeriod time.Duration
```
jdn5126, line 82: nit: let's add a comment here explaining this value

README.md
```diff
@@ -669,6 +669,16 @@ Default: `false`
...
+
+Specify the number of seconds an IP address is in cooldown after a pod is deleted. The cooldown is used to give kube-proxy time to update node's IP table rules it is running on. Lower values should be used if the pod is not connected to any service.
```
jdn5126, line 680: On this wording, I think it should be something like:
```
Specifies the number of seconds an IP address is in cooldown after pod deletion. The cooldown period gives kube-proxy time to update node iptables rules when the IP was registered as a valid endpoint, such as for a service. Modify this value with caution, as kube-proxy update time scales with the number of nodes and services.
```

jdn5126 (changes requested):
Changes overall are looking good to me. I just have some nits

jdn5126:
@arturzeitler it looks like there are some conflicts here that need to be resolved, then this can build and run tests

cmd/aws-vpc-cni/main.go
```diff
@@ -99,6 +100,8 @@ const (
 	envEnIPv6                = "ENABLE_IPv6"
 	envEnIPv6Egress          = "ENABLE_V6_EGRESS"
 	envRandomizeSNAT         = "AWS_VPC_K8S_CNI_RANDOMIZESNAT"
+	envEnableNftables        = "ENABLE_NFTABLES"
```
jdn5126, line 103: looks like this was added back unintentionally after merge conflict resolution

pkg/ipamd/datastore/data_store.go
```diff
@@ -35,9 +36,9 @@ const (
 	// minENILifeTime is the shortest time before we consider deleting a newly created ENI
 	minENILifeTime = 1 * time.Minute
 
-	// addressCoolingPeriod is used to ensure an IP not get assigned to a Pod if this IP is used by a different Pod
-	// in addressCoolingPeriod
-	addressCoolingPeriod = 30 * time.Second
+	// envIPCoolingPeriod is used to ensure an IP not get assigned to a Pod if this IP is used by a different Pod
+	// in envIPCoolingPeriod
```
jdn5126, line 40: nit: can we update this comment to: "envIPCoolingPeriod controls the time that a deleted pod IP address is held before being eligible to be assigned to a new pod. This holding time gives components that depended on the deleted IP a chance to react to the update."

pkg/ipamd/datastore/data_store_test.go
```diff
@@ -925,8 +925,8 @@ func TestGetIPStatsV4(t *testing.T) {
 		*ds.GetIPStats("4"),
 	)
 
-	// wait 30s (cooldown period)
-	time.Sleep(30 * time.Second)
+	// wait number of seconds configured in IP_COOLING_PERIOD (cooldown period)
```
jdn5126, line 928: Can we add a test case where we actually set `IP_COOLING_PERIOD`?

pkg/ipamd/datastore/data_store.go
```diff
@@ -266,9 +270,25 @@ func (addr AddressInfo) Assigned() bool {
 	return !addr.IPAMKey.IsZero()
 }
 
-// InCoolingPeriod checks whether an addr is in addressCoolingPeriod
+// getCoolingPeriod returns the time duration in seconds configured by the IP_COOLING_PERIOD env variable
+func (ds *DataStore) getCoolingPeriod() time.Duration {
+	inputStr, found := os.LookupEnv(envIPCoolingPeriod)
```
jdn5126, line 275: nit: why not call the `GetIntAsStringEnvVar` function that was added?

cmd/aws-vpc-cni/main.go
```diff
@@ -99,6 +100,8 @@ const (
...
+	envIPCoolingPeriod       = "IP_COOLING_PERIOD"
```
jdn5126, line 104: This should be `IP_COOLDOWN_PERIOD`, as that's what the README calls it. Let's also use `envIPCooldownPeriod`

cmd/aws-vpc-cni/main.go
```diff
@@ -352,6 +355,17 @@ func validateEnvVars() bool {
 	warmPrefixTarget := utils.GetEnv(envWarmPrefixTarget, "0")
 	minimumIPTarget := utils.GetEnv(envMinIPTarget, "0")
 
+	// Validate that IP_COOLING_PERIOD is a valid integer
+	ipCoolingPeriod, err := utils.GetIntAsStringEnvVar(envIPCoolingPeriod, defaultIPCoolingPeriod)
+	if err != nil {
+		log.Errorf("IP_COOLING_PERIOD has to be a valid integer")
```
jdn5126, line 361: `IP_COOLDOWN_PERIOD`

cmd/aws-vpc-cni/main.go
```diff
@@ -352,6 +355,17 @@ func validateEnvVars() bool {
...
+		return false
+	}
+	if ipCoolingPeriod < 1 {
+		log.Errorf("IP_COOLING_PERIOD cannot be smaller than 1")
```
jdn5126, line 365: `IP_COOLDOWN_PERIOD`

jdn5126 (changes requested):
Sorry for the delay @arturzeitler . A few more comments and then I think this is good to go

pkg/ipamd/datastore/data_store_test.go
arturzeitler: yes, it would help though if you could point me to a similar test case.

jdn5126:
@arturzeitler are you still able to drive this? I am good to merge this after we rebase, fix a few more nits, and pass a run of integration tests

jdn5126:
Closing as https://github.com/aws/amazon-vpc-cni-k8s/pull/2492 is taking this over

PR #2398, opened 2023-06-02
refactor canary test to access images from AWS registries

oliviassss:
**What type of PR is this?**

Refactor test

**Which issue does this PR fix**: NA

The PR refactors the canary tests:
1. Adds a new ENV ```TEST_IMAGE_REGISTRY``` in test script to map the AWS registries for canary test images. The value will be passed from internal CI jobs, which differentiates the default region (us-west-2) vs. isolated regions.
2. Skips installing latest addon version if the test is running in ADC regions.
3. Adds a new framework option TestImageRegistry so that in canary test it will pull the images from AWS registries, instead of from internet.

No

ran the canary test in local

**Automation added to e2e**:

No

No

No

NA

No
```release-note

```

### DIFF
--- a/scripts/run-canary-test.sh
+++ b/scripts/run-canary-test.sh
@@ -7,6 +7,9 @@ set -e
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 GINKGO_TEST_BUILD="$SCRIPT_DIR/../test/build"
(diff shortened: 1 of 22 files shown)


jdn5126 (commented):
Change LGTM, and I can't think of any better way to do this than to pass the test image registry when we create pods. Luckily this is just a one-time change. I will approve once the tests pass

jdn5126 (approved):

Issue #2388, opened 2023-05-20
Auto-detect iptables mode


hakman:
**What would you like to be added**:

https://github.com/aws/amazon-vpc-cni-k8s/pull/2155 adds support for setting the iptables mode using the `ENABLE_NFTABLES` env var.

AWS VPC CNI should default to the same logic as `kube-proxy` to auto-detect the `iptables` mode of the host:
https://github.com/kubernetes/release/blob/31530bf54df0322e8f112affdb15cb65e6f9e2e3/images/build/debian-iptables/bullseye/iptables-wrapper

The `iptables-wrapper` logic is pretty simple and already used by other CNI plugins:
https://github.com/projectcalico/calico/blob/d433a29710de512716479870a3937c3a730431f2/felix/environment/feature_detect_linux.go#L326

**Why is this needed**:

The most obvious use case is that operators are migrating from older distro versions to newer ones and getting the new default, which is `iptables-nft` (for example ubuntu-20.04 to ubuntu-22.04).

There are also cases of (more rare) clusters with mixed distros.

jdn5126:
@hakman the reason we did not use the `iptables-wrapper` script initially is that all the implementation did was count which mode (`iptables-nft` vs `iptables-legacy`) had more entries present. This would cause issues as customers upgraded or dealt with other components still relying on `iptables-legacy`, so we introduced `ENABLE_NFTABLES` to be more explicit.

Newer versions of `iptables-wrapper` look for rules explicitly installed by kubelet, and are implemented in Golang, so it does seem like it would be good to move to `iptables-wrapper` going forward and deprecate `ENABLE_NFTABLES`.

jdn5126:
Closing as this is added in v1.13.2 release

PR #2402, opened 2023-06-05
Set iptables mode automatically and deprecate ENABLE_NFTABLES

jdn5126:
**What type of PR is this?**
enhancement

#2388 

This PR sets the iptables mode for `aws-node` pod automatically. Rather than using `ENABLE_NFTABLES` to decide when to run `update-alternatives`, we now rely on the `iptables-wrapper` script to set the mode based on which mode kubelet is using (kubelet iptables hint).

This PR also deprecates `ENABLE_NFTABLES` environment variable as it is no longer needed.

N/A

Manually verified that `iptables-wrapper` script is run.

**Automation added to e2e**:
N/A

No

No, Yes

No

Yes

```release-note
Set iptables mode automatically and deprecate ENABLE_NFTABLES
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -620,7 +620,7 @@ configured to operate in IPv6 mode. Prefix delegation is only supported on nitro
 
 ---
 
-#### `ENABLE_NFTABLES` (v1.12.1+)
+#### `ENABLE_NFTABLES` (introduced in v1.12.1, deprecated in v1.13.1+)
 
 Type: Boolean as a String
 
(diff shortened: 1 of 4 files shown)


jdn5126:
> /lgtm Nice!! Have we verified if it handles the cases correct? i.e setting the right mode based on what's being used by the node?

Yep, I verified that the script runs as part of container init and it will switch to `iptables-nft` when there is a "kubelet hint" in the table for `iptables-nft`. I also verified that `update-alternatives --set", "iptables /usr/sbin/iptables-wrapper` can be run manually using `kubectl exec`.

Issue #2426, opened 2023-06-15
Custom Networking ENIs fail to attach in v1.13.0 when no security groups are specified


jdn5126:
**What happened**:
In `v1.13.0`, https://github.com/aws/amazon-vpc-cni-k8s/pull/2354 was introduced, which improved custom networking startup time by attaching custom ENIs during node initialization. This introduced an issue, though, when no security group was associated with the ENIConfig object. In this case, we should fall back to the security group assigned to the node's primary ENI.

The security group assigned to the node's primary ENI comes from the EC2 metadata cache, though, and this value is not synced until after node initialization: https://github.com/aws/amazon-vpc-cni-k8s/blob/v1.13.0/pkg/ipamd/ipamd.go#L450

Our retry strategy in `v1.13.0` is to let the `aws-node` pod crash so that the node does not become ready for pods to be scheduled. When the `aws-node` pod restarts, we hit the same issue, as the cache never syncs the primary ENI's security group ID until after node initialization.

**Attach logs**
```
{"level":"info","ts":"2023-06-15T21:55:31.095Z","caller":"ipamd/ipamd.go:903","msg":"Found ENI Config Name: us-west-2b"}
{"level":"info","ts":"2023-06-15T21:55:31.196Z","caller":"ipamd/ipamd.go:873","msg":"ipamd: using custom network config: [], subnet-07f48a842d5d33cee"}
{"level":"info","ts":"2023-06-15T21:55:31.196Z","caller":"awsutils/awsutils.go:763","msg":"Using a custom network config for the new ENI"}
{"level":"warn","ts":"2023-06-15T21:55:31.196Z","caller":"awsutils/awsutils.go:763","msg":"No custom networking security group found, will use the node's primary ENI's SG: []"}
{"level":"info","ts":"2023-06-15T21:55:31.196Z","caller":"awsutils/awsutils.go:763","msg":"Creating ENI with security groups: [] in subnet: subnet-07f48a842d5d33cee"}
{"level":"error","ts":"2023-06-15T21:55:31.299Z","caller":"awsutils/awsutils.go:763","msg":"Failed to CreateNetworkInterface InvalidParameterValue: user [REDACTED] does not own a resource\n\tstatus code: 400, request id: e12e797e-19f5-47fa-8d8e-4a5de5a677b7"}
```

**What you expected to happen**:
Custom ENIs should be attached during node initialization, and there should be a better retry strategy.

[...]

PR #2427, opened 2023-06-16
Sync node security groups to cache before node initialization

jdn5126:
**What type of PR is this?**
bug

https://github.com/aws/amazon-vpc-cni-k8s/issues/2426

This PR fixes a bug in v1.13.0 where custom ENIs could not be created when an `ENIConfig` definition did not contain a security group ID. The expected behavior is that we fall back to the security group ID assigned to the primary ENI. The issue is that we were trying to create the ENI before the security group assigned to the primary ENI was synced to the EC2 metadata cache.

N/A

Manually verified that `ENIConfig`s with no security group work following this change. Integration test coverage is pending, as this PR is being fast-tracked to make `v1.13.1` release. The manual test cases that were run are being added to custom networking test suite.

**Automation added to e2e**:
Pending, will be added in a follow-up PR

No

No, Yes

No

Yes

```release-note
Fix custom networking issue in v1.13.0 when ENIConfig did not include any security group IDs
```

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -439,14 +439,8 @@ func New(rawK8SClient client.Client, cachedK8SClient client.Client) (*IPAMContex
 	checkpointer := datastore.NewJSONFile(dsBackingStorePath())
 	c.dataStore = datastore.NewDataStore(log, checkpointer, c.enablePrefixDelegation)
 
-	err = c.nodeInit()
-	if err != nil {
-		return nil, err
-	}
-
-	mac := c.awsClient.GetPrimaryENImac()
-
 	// Retrieve security groups
+	mac := c.awsClient.GetPrimaryENImac()
 	if c.enableIPv4 && !c.disableENIProvisioning {
 		err = c.awsClient.RefreshSGIDs(mac)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/ipamd/ipamd.go
```diff
@@ -439,14 +439,8 @@ func New(rawK8SClient client.Client, cachedK8SClient client.Client) (*IPAMContex
...
 
-	err = c.nodeInit()
-	if err != nil {
-		return nil, err
-	}
-
-	mac := c.awsClient.GetPrimaryENImac()
-
 	// Retrieve security groups
+	mac := c.awsClient.GetPrimaryENImac()
```
orsenthil, line 443: I looked at this code in CNI 1.12.1
https://github.com/aws/amazon-vpc-cni-k8s/blob/v1.12.1/pkg/ipamd/ipamd.go#L414

It hasn't changed in that. 

- So, I wasn't entirely certain if this is the cause of regression here.
- Secondly, will `c.awsClient.GetPrimaryENImac` return a success if the there a chance of `c.nodeInit` failure?

jdn5126: What changed is that node initialization will now attach new ENIs: https://github.com/aws/amazon-vpc-cni-k8s/blob/v1.13.0/pkg/ipamd/ipamd.go#L612 instead of just reconciling already attached CNIs: https://github.com/aws/amazon-vpc-cni-k8s/blob/v1.12.6/pkg/ipamd/ipamd.go#L589

So the change is that we are attempting ENI allocations before the security groups from the primary ENI are synced, which leads to this error.

For `GetPrimaryENImac`, this will always be populated during container init from IMDS, and the `aws-node` will not be able to start if we cannot find the primary ENI MAC associated with the instance. So there is no behavior change there. It is still possible for `nodeInit` to fail for many reasons, but this failure was a "false positive".

orsenthil: Got it. Thanks for the context and explanation.

orsenthil (approved):
LGTM.

PR #2439, opened 2023-06-23
Scope the cached k8s client to the node

alam0rt:
**What type of PR is this?** bug (regression)

**Which issue does this PR fix**: https://github.com/aws/amazon-vpc-cni-k8s/issues/2436

**What does this PR do / Why do we need it**: CNI daemonsets should use as minimal memory as possible since they are run on most nodes.

**Testing done on this change**:

Deploy to a medium sized staging cluster (3000+ pods).

Before change
```
aws-cni-xw4ts                            5m           432Mi
aws-cni-zt5wr                            6m           446Mi
aws-cni-zxch8                            8m           453Mi
```

After scoping caching client
```
aws-cni-ppx92                            9m           72Mi
aws-cni-ppzml                            7m           39Mi
aws-cni-q27wl                            6m           42Mi
aws-cni-qg6ww                            5m           48Mi
aws-cni-r2lwc                            8m           77Mi
aws-cni-r5nr5                            9m           49Mi
aws-cni-rjxjj                            7m           76Mi
aws-cni-rmhzf                            2m           68Mi
aws-cni-rx7wq                            4m           53Mi
aws-cni-scpc6                            7m           44Mi
aws-cni-sfbkf                            6m           42Mi
```

**Automation added to e2e**:

**Will this PR introduce any new dependencies?**: reliance on `MY_NODE_NAME` being set via downward API.

**Will this break upgrades or downgrades. Has updating a running cluster been tested?**: This has been tested on a running cluster

**Does this change require updates to the CNI daemonset config files to work?**: No

**Does this PR introduce any user-facing change?**: Substantially less memory usage (still more than before the caching client was used though).

```release-note

```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #2160, opened 2022-12-13
Reduce API server calls


jayanthvn:
**What would you like to be added**:
Today we do a list call on aws-node init and regular watch calls. This is not needed for customers who are not using custom networking and/or SGPP.

**Why is this needed**:
Reduce unnecessary API server calls and moreover cross AZ calls when not needed.

jayanthvn:
This optimization will be done post merge of https://github.com/aws/amazon-vpc-cni-k8s/pull/2149.

PR #2442, opened 2023-06-26
add support to VPC resource controller's CNINode CRD

haouc:
**What type of PR is this?**
We want to replace node labels when enabling SGP feature through VPC CNI. VPC resource controller will create the new CRD objects with one to one mapping to k8s nodes. VPC CNI is responsible to patch the CR to add features.
The resource controller's PR is https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/270

The tests have been done with resource controller change. 

We need to switch to the new CRD based implementation.

- new install
- upgrade from older CNI 1.11.x and 1.12.x
- downgrade to older CNI version
- scale tests with 300 nodes
- 
**Automation added to e2e**:

Will need wait for the VPC resource controller release

Yes, depending on the new CRD

It will not.

no

The node's label for trunk interface will no longer be added.
```release-note

```

### DIFF
--- a/charts/aws-vpc-cni/templates/clusterrole.yaml
+++ b/charts/aws-vpc-cni/templates/clusterrole.yaml
@@ -33,3 +33,8 @@ rules:
     resources:
       - events
     verbs: ["create", "patch", "list"]
+  - apiGroups:
+      - vpcresources.k8s.aws
+    resources:
+      - cninodes
+    verbs: ["get", "list", "patch"]
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.20
 require (
 	github.com/apparentlymart/go-cidr v1.0.1
 	github.com/aws/amazon-vpc-cni-k8s/test/agent v0.0.0-20230614172555-5788cde362fa
-	github.com/aws/amazon-vpc-resource-controller-k8s v1.1.5
+	github.com/aws/amazon-vpc-resource-controller-k8s v1.1.8
 	github.com/aws/aws-sdk-go v1.43.29
 	github.com/containernetworking/cni v1.1.1
 	github.com/containernetworking/plugins v1.1.1
@@ -123,6 +123,7 @@ require (
 	github.com/rubenv/sql-migrate v1.2.0 // indirect
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
 	github.com/safchain/ethtool v0.0.0-20210803160452-9aa261dae9b1 // indirect
+	github.com/samber/lo v1.38.1
 	github.com/shopspring/decimal v1.2.0 // indirect
 	github.com/spf13/cast v1.4.1 // indirect
 	github.com/spf13/cobra v1.6.1 // indirect
@@ -135,13 +136,14 @@ require (
 	go.uber.org/atomic v1.7.0 // indirect
 	go.uber.org/multierr v1.6.0 // indirect
 	golang.org/x/crypto v0.5.0 // indirect
+	golang.org/x/exp v0.0.0-20230522175609-2e198f4a06a1 // indirect
 	golang.org/x/oauth2 v0.4.0 // indirect
 	golang.org/x/sync v0.1.0 // indirect
 	golang.org/x/term v0.6.0 // indirect
 	golang.org/x/text v0.8.0 // indirect
 	golang.org/x/time v0.3.0 // indirect
 	golang.org/x/tools v0.6.0 // indirect
-	gomodules.xyz/jsonpatch/v2 v2.2.0 // indirect
+	gomodules.xyz/jsonpatch/v2 v2.3.0 // indirect
 	google.golang.org/appengine v1.6.7 // indirect
 	google.golang.org/genproto v0.0.0-20220502173005-c8bf987b8c21 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -31,6 +31,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/samber/lo"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -44,6 +45,7 @@ import (
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/networkutils"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/eventrecorder"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
+	rcv1alpha1 "github.com/aws/amazon-vpc-resource-controller-k8s/apis/vpcresources/v1alpha1"
 )
 
 // The package ipamd is a long running daemon which manages a warm pool of available IP addresses.
@@ -134,12 +136,6 @@ const (
 
 	// envNodeName will be used to store Node name
 	envNodeName = "MY_NODE_NAME"
-
-	// vpcENIConfigLabel is used by the VPC resource controller to pick the right ENI config.
-	vpcENIConfigLabel = "vpc.amazonaws.com/eniConfig"
-
-	// trunkInterfaceLabel is used by CNI to check if correct label was added by the VPC resource controller
-	trunkInterfaceLabel = "vpc.amazonaws.com/has-trunk-attached"
 
 	//envEnableIpv4PrefixDelegation is used to allocate /28 prefix instead of secondary IP for an ENI.
 	envEnableIpv4PrefixDelegation = "ENABLE_PREFIX_DELEGATION"
@@ -575,34 +571,19 @@ func (c *IPAMContext) nodeInit() error {
 
 	eniConfigName, err := eniconfig.GetNodeSpecificENIConfigName(node)
 	if err == nil && c.useCustomNetworking && eniConfigName != "default" {
-		// Signal to VPC Resource Controller that the node is using custom networking
-		err := c.SetNodeLabel(ctx, vpcENIConfigLabel, eniConfigName)
+		// Add the feature name to CNINode of this node
+		err := c.AddFeatureToCNINode(ctx, rcv1alpha1.CustomNetworking, eniConfigName)
 		if err != nil {
-			log.Errorf("Failed to set eniConfig node label", err)
+			log.Errorf("Failed to add feature custom networking into CNINode", err)
 			podENIErrInc("nodeInit")
 			return err
 		}
-	} else {
-		// Remove the custom networking label
-		err := c.SetNodeLabel(ctx, vpcENIConfigLabel, "")
-		if err != nil {
-			log.Errorf("Failed to delete eniConfig node label", err)
-			podENIErrInc("nodeInit")
-			return err
-		}
+		log.Infof("Enabled feature %s in CNINode for node %s if not existing", rcv1alpha1.CustomNetworking, c.myNodeName)
 	}
 
-	if metadataResult.TrunkENI != "" {
-		// Signal to VPC Resource Controller that the node has a trunk already
-		err := c.SetNodeLabel(ctx, trunkInterfaceLabel, "true")
-		if err != nil {
-			log.Errorf("Failed to set node label", err)
-			podENIErrInc("nodeInit")
-			// If this fails, we probably can't talk to the API server. Let the pod restart
-			return err
-		}
-	} else {
-		// Check if we want to label node for trunk interface during node init
+	if c.enablePodENI {
+		// Check if we want to add feature into CNINode for trunk interface during node init
+		// we don't check if the node already has trunk added during initialization
 		c.askForTrunkENIIfNeeded(ctx)
 	}
 
@@ -701,7 +682,9 @@ func (c *IPAMContext) StartNodeIPPoolManager() {
 }
 
 func (c *IPAMContext) updateIPPoolIfRequired(ctx context.Context) {
-	c.askForTrunkENIIfNeeded(ctx)
+	if c.enablePodENI && c.dataStore.GetTrunkENI() == "" {
+		c.askForTrunkENIIfNeeded(ctx)
+	}
 
 	if c.isDatastorePoolTooLow() {
 		c.increaseDatastorePool(ctx)
@@ -1283,27 +1266,18 @@ func (c *IPAMContext) logPoolStats(dataStoreStats *datastore.DataStoreStats) {
 }
 
 func (c *IPAMContext) askForTrunkENIIfNeeded(ctx context.Context) {
-	if c.enablePodENI && c.dataStore.GetTrunkENI() == "" {
-		// Check that there is room for a trunk ENI to be attached:
-		if c.dataStore.GetENIs() >= (c.maxENI - c.unmanagedENI) {
-			log.Debug("No slot available for a trunk ENI to be attached. Not labeling the node")
-			return
-		}
-		// We need to signal that VPC Resource Controller needs to attach a trunk ENI
-		err := c.SetNodeLabel(ctx, trunkInterfaceLabel, "false")
-		if err != nil {
-			podENIErrInc("askForTrunkENIIfNeeded")
-			log.Errorf("Failed to set node label", err)
-		}
+	// Check that there is room for a trunk ENI to be attached:
+	if c.dataStore.GetENIs() >= (c.maxENI - c.unmanagedENI) {
+		log.Error("No slot available for a trunk ENI to be attached. Not init the node")
+		return
 	}
-}
-
-func (c *IPAMContext) getNodeLabels(ctx context.Context) (map[string]string, error) {
-	var node corev1.Node
-	if err := c.cachedK8SClient.Get(ctx, types.NamespacedName{Name: c.myNodeName}, &node); err != nil {
-		return nil, err
+	// We need to signal that VPC Resource Controller needs to attach a trunk ENI
+	err := c.AddFeatureToCNINode(ctx, rcv1alpha1.SecurityGroupsForPods, "")
+	if err != nil {
+		podENIErrInc("askForTrunkENIIfNeeded")
+		log.Errorf("Failed to add SGP feature into the CNINode", err)
 	} else {
-		return node.GetLabels(), err
+		log.Infof("Successfully added feature %s to CNINode if not existing", rcv1alpha1.SecurityGroupsForPods)
 	}
 }
 
@@ -1417,13 +1391,6 @@ func (c *IPAMContext) nodeIPPoolReconcile(ctx context.Context, interval time.Dur
 
 		if c.enablePodENI && metadataResult.TrunkENI != "" {
 			log.Debugf("Trunk interface (%s) has been added to the node already.", metadataResult.TrunkENI)
-			// Label the node that we have a trunk
-			err = c.SetNodeLabel(ctx, trunkInterfaceLabel, "true")
-			if err != nil {
-				podENIErrInc("askForTrunkENIIfNeeded")
-				log.Errorf("Failed to set node label for trunk. Aborting reconcile", err)
-				return
-			}
 		}
 		// Update trunk ENI
 		trunkENI = metadataResult.TrunkENI
@@ -1994,54 +1961,6 @@ func (c *IPAMContext) getTrunkLinkIndex() (int, error) {
 	return -1, errors.New("no trunk found")
 }
 
-// SetNodeLabel sets or deletes a node label
-func (c *IPAMContext) SetNodeLabel(ctx context.Context, key, value string) error {
-	request := types.NamespacedName{
-		Name: c.myNodeName,
-	}
-
-	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		node := &corev1.Node{}
-		// Find my node
-		err := c.cachedK8SClient.Get(ctx, request, node)
-		if err != nil {
-			log.Errorf("Failed to get node: %v", err)
-			return err
-		}
-		log.Debugf("Node found %q - no of labels - %d", node.Name, len(node.Labels))
-
-		if labelValue, ok := node.Labels[key]; ok && labelValue == value {
-			log.Debugf("Node label %q is already %q", key, labelValue)
-			return nil
-		}
-
-		// Make deep copy for modification
-		updateNode := node.DeepCopy()
-
-		// in case for some reason node object doesn't have label map
-		if updateNode.Labels == nil {
-			updateNode.Labels = make(map[string]string)
-		}
-
-		// Set node label
-		if value != "" {
-			updateNode.Labels[key] = value
-		} else {
-			// Empty value, delete the label
-			log.Debugf("Deleting label %q", key)
-			delete(updateNode.Labels, key)
-		}
-
-		if err = c.cachedK8SClient.Update(ctx, updateNode); err != nil {
-			log.Errorf("Failed to patch node %s with label %q: %q, error: %v", c.myNodeName, key, value, err)
-			return err
-		}
-		log.Debugf("Updated node %s with label %q: %q", c.myNodeName, key, value)
-		return nil
-	})
-	return err
-}
-
 // GetPod returns the pod matching the name and namespace
 func (c *IPAMContext) GetPod(podName, namespace string) (*corev1.Pod, error) {
 	ctx := context.TODO()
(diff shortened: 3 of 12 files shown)


haouc:
**Note**: Since this is waiting for a new release tag of VPC resource controller for an update on the CNINode CRD, it is referring to my fork repo tag by `replacing`. We need remove the `replacing` before merging this PR.

pkg/k8sapi/k8sutils.go
```diff
@@ -68,6 +70,7 @@ func CreateCachedKubeClient(rawK8SClient client.Client, mapper meta.RESTMapper)
 	vpcCniScheme := runtime.NewScheme()
 	clientgoscheme.AddToScheme(vpcCniScheme)
 	eniconfigscheme.AddToScheme(vpcCniScheme)
+	rcscheme.AddToScheme(vpcCniScheme)
```
jdn5126, line 73: Similar to https://github.com/aws/amazon-vpc-cni-k8s/pull/2439 , do we need to scope down caching for this CRD by adding an object selector for this node only? Alternatively, we may not want to cache this object since there will always only be one per node, and we only do patch operations

pkg/sgpp/utils.go
```diff
@@ -30,3 +38,68 @@ func LoadEnforcingModeFromEnv() EnforcingMode {
...
+	newCNINode := cniNode.DeepCopy()
+	newFeature := rc.Feature{
+		Name:  featureName,
+		Value: featureValue,
+	}
+	newCNINode.Spec.Features = append(newCNINode.Spec.Features, newFeature)
+	return k8sClient.Patch(ctx, newCNINode, client.MergeFromWithOptions(cniNode, client.MergeFromWithOptimisticLock{}))
+}
+
+func RemoveFeatureFromCNINode(ctx context.Context, k8sClient client.Client, nodeName string, features []rc.FeatureName) error {
```
jdn5126, line 70: nit: as above, I think we can remove this function if we do not plan to call it

pkg/sgpp/utils.go
```diff
@@ -30,3 +38,68 @@ func LoadEnforcingModeFromEnv() EnforcingMode {
 		return DefaultEnforcingMode
 	}
 }
+
+func AddFeatureToCNINode(ctx context.Context, k8sClient client.Client, nodeName string, featureName rc.FeatureName, featureValue string) error {
```
jdn5126, line 42: Since this package is for `sgpp`, and this feature is going to be more general, I am wondering if this belongs in a different package. Maybe `k8sapi` or `utils`?

jdn5126 (commented):
Change overall LGTM, no major concerns. I added a few questions and comments. For the manifest changes, can we make the `aws-vpc-cni` chart changes as well? Since these are only additions, they are fine to add in this PR

pkg/ipamd/ipamd.go
```diff
@@ -575,35 +574,28 @@ func (c *IPAMContext) nodeInit() error {
 
 	eniConfigName, err := eniconfig.GetNodeSpecificENIConfigName(node)
 	if err == nil && c.useCustomNetworking && eniConfigName != "default" {
-		// Signal to VPC Resource Controller that the node is using custom networking
-		err := c.SetNodeLabel(ctx, vpcENIConfigLabel, eniConfigName)
+		// Add the feature name to CNINode of this node
+		err := sgpp.AddFeatureToCNINode(ctx, c.cachedK8SClient, c.myNodeName, rc.CustomNetworking, eniConfigName)
```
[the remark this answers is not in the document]
ellistarn: Do we really want to crash the CNI? Can't we simply retry and emit a log each time? Crashing will subject us to crash loop backoff durations.

pkg/ipamd/ipamd.go
```diff
@@ -575,35 +574,28 @@ func (c *IPAMContext) nodeInit() error {
...
 		if err != nil {
-			log.Errorf("Failed to set eniConfig node label", err)
+			log.Errorf("Failed to add feature custom networking into CNINode", err)
 			podENIErrInc("nodeInit")
 			return err
 		}
+		log.Infof("Successfully added feature %s to CNINode if not existing", rc.CustomNetworking)
```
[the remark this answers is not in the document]
ellistarn: Another idea for logging. Consider avoiding words like "successfully" and "failed" in INFO and ERROR logs respectively. Both INFO and ERROR implicitly convey these concepts, but nature of their log level. 

e.g. `enabled feature %q for node %q`


Also -- should we be using a structured logger, so this is more easily parsable?

log.With("node", node.Name, "feature", rc.CustomNetworking).Info("enabled feature")

pkg/sgpp/utils.go
ellistarn, line 42: This function is not attached to any struct and is "dangling". Is it possible to attach this to a greater context, so we can avoid the complexity of passing dependencies around through method arguments (e.g. k8sClient). 

Another thought -- does it make sense to decouple these two concepts:
1. Constructs of the CNINode
2. Application of the CNINode

Something like:
```
node := v1.Node{}
_, err := client.Get(ctx, &node) { ... }
cniNode := CNINodeFor(node)
client.Patch(ctx, &cniNode) { ... }

File: pkg/sgpp/utils.go [hunk 6]
```diff
@@ -1,6 +1,14 @@
 package sgpp
```
ellistarn, line 1: As a general pattern, I'd avoid `utils` packages as much as possible, and encapsulate the logic where its owned wherever possible. Of course, some things do need to be shared, but often there is a better named abstraction that encapsulates the concept. `utils` packages have a tendency to become a dumping ground for code, and become more complex and less maintainable over time.

pkg/ipamd/ipamd.go
```diff
@@ -1283,19 +1277,28 @@ func (c *IPAMContext) logPoolStats(dataStoreStats *datastore.DataStoreStats) {
...
-		}
-		// We need to signal that VPC Resource Controller needs to attach a trunk ENI
-		err := c.SetNodeLabel(ctx, trunkInterfaceLabel, "false")
-		if err != nil {
-			podENIErrInc("askForTrunkENIIfNeeded")
-			log.Errorf("Failed to set node label", err)
-		}
+	// Check that there is room for a trunk ENI to be attached:
+	if c.dataStore.GetENIs() >= (c.maxENI - c.unmanagedENI) {
+		log.Debug("No slot available for a trunk ENI to be attached. Not labeling the node")
```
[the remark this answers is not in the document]
ellistarn: Should this merely emit a debug log? It seems like if we need a trunk eni, but can't get one, that should be considered as an error case.

pkg/sgpp/utils.go
haouc: This can be added to ipamd.go instead of from another package. I prefer to create the CNINode objects from VPC resource controller side since in general VPC resource controller is persisting and tracking/managing the life time of the node object from its point of view. VPC CNI will restart for many reasons and seems no need to worry about creation of this CRD object, like it doesn't care the creation of node object.

haouc: According to Ellis' suggestion, I think it makes sense to have this function associated to IPAM.

pkg/sgpp/utils.go
haouc: Makes sense. I was hesitating the same and decided keeping it here to save the option to support disable the feature without terminating nodes. I can remove them.

pkg/k8sapi/k8sutils.go
haouc: Do you mean always use raw client for this operation?

pkg/ipamd/ipamd.go
```diff
@@ -1283,19 +1277,28 @@ func (c *IPAMContext) logPoolStats(dataStoreStats *datastore.DataStoreStats) {
...
+	err := sgpp.AddFeatureToCNINode(ctx, c.cachedK8SClient, c.myNodeName, rc.SecurityGroupsForPods, "")
+	if err != nil {
+		podENIErrInc("askForTrunkENIIfNeeded")
+		log.Errorf("Failed to add SGP feature into the CNINode", err)
+	} else {
+		log.Infof("Successfully added feature %s to CNINode if not existing", rc.SecurityGroupsForPods)
+	}
+}
+
+func (c *IPAMContext) cleanupCNINode(ctx context.Context, features []rc.FeatureName) error {
```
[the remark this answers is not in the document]
haouc: Sure, will remove it.

pkg/ipamd/ipamd.go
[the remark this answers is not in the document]
haouc: Makes sense.

pkg/ipamd/ipamd.go
```diff
@@ -1283,19 +1277,28 @@ func (c *IPAMContext) logPoolStats(dataStoreStats *datastore.DataStoreStats) {
...
-			podENIErrInc("askForTrunkENIIfNeeded")
-			log.Errorf("Failed to set node label", err)
-		}
+	// Check that there is room for a trunk ENI to be attached:
+	if c.dataStore.GetENIs() >= (c.maxENI - c.unmanagedENI) {
+		log.Debug("No slot available for a trunk ENI to be attached. Not labeling the node")
+		return
+	}
+	// We need to signal that VPC Resource Controller needs to attach a trunk ENI
+	err := sgpp.AddFeatureToCNINode(ctx, c.cachedK8SClient, c.myNodeName, rc.SecurityGroupsForPods, "")
```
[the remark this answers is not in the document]
haouc: Yes, it is just log the error and continue. This is ok for nodeinit and the reconciling will try it again later.

pkg/ipamd/ipamd.go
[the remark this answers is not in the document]
haouc: If Get returns error of NotFound, we just log the error and move on. The reconciler will try again.

pkg/ipamd/ipamd.go
```diff
@@ -575,35 +574,28 @@ func (c *IPAMContext) nodeInit() error {
...
 	} else {
-		// Remove the custom networking label
-		err := c.SetNodeLabel(ctx, vpcENIConfigLabel, "")
-		if err != nil {
-			log.Errorf("Failed to delete eniConfig node label", err)
-			podENIErrInc("nodeInit")
-			return err
-		}
+		// Warning: Enble this here will cause VPC RC resync resources from different subnets
+		// TODO: If we decide to support remove SGP support without termintaing node, we can enable this
```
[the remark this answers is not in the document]
haouc: OK. Since we don't current plan to support remove networking without terminating the node, I will update the comment and remove the code. Thanks

pkg/sgpp/utils.go
ellistarn: > I prefer to create the CNINode objects from VPC resource controller side since in general VPC resource controller is persisting and tracking/managing the life time of the node object from its point of view. 

Huge +1. I could see taking this a step further, and taking an assumption that the CRD exists for the given node before taking any actions as a CNI. WDYT?

pkg/ipamd/ipamd.go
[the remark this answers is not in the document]
haouc: I agree. I think this should be qualified for an error even it is induced by customers.

pkg/sgpp/utils.go
haouc: That would be nice. I am trying to find out if there is a k8s mechanism which is like a "generator" (the opposite side of finalizer) with Ownership. I think this is no such native controller to do this? In this case, we are relying on the VPC resource controller to generate the CRD object upon its owner node object's creation. Not sure if this is what you were referring to?

pkg/sgpp/utils.go
haouc: added algo/algo.go to utils folder. Seems ok practice referring to https://github.com/kubernetes/kubernetes/tree/master/pkg/util
Probably should use `util` instead of `utils` in VPC CNI?

pkg/ipamd/ipamd.go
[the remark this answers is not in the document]
jdn5126: In this implementation, for custom networking, it looks like we only patch the CRD during node init. So if we cannot patch during node init, it would never be patched, and if the k8s client returns an error, `aws-node` pod will crash.

It looks like only SGPP uses a reconciler approach

pkg/sgpp/utils.go
ellistarn: Yeah there's no "generator" concept. I think of the CNINode is being owned by the CNI-agent, as it is the source of truth for the CNINode.spec. I'm not sure I see the benefit of creating the object from VPCRC, regardless of the restart strategy of VPCRC or VPCCNI.

What do you think about the following:
* VPCCNI wakes up, and applies (create/update) the CNINode to be what it expects it to be.
* VPCRC watches for the CNINode objects and takes action accordingly (i.e., attaching trunk eni)

pkg/utils/algo/algos.go
```diff
@@ -0,0 +1,13 @@
+package algo
+
+import "golang.org/x/exp/constraints"
+
+// Include checks if a string existing in a string slice and returns true or false.
+func Include[T constraints.Ordered](target T, values []T) bool {
```
ellistarn, line 6: Consider using https://github.com/samber/lo#contains

pkg/k8sapi/k8sutils.go
```diff
@@ -68,9 +72,22 @@ func CreateCachedKubeClient(rawK8SClient client.Client, mapper meta.RESTMapper)
...
 
 	stopChan := ctrl.SetupSignalHandler()
-	cache, err := cache.New(restCfg, cache.Options{Scheme: vpcCniScheme, Mapper: mapper})
+
+	opts := cache.Options{Scheme: vpcCniScheme, Mapper: mapper}
+	if nodeName := os.Getenv("MY_NODE_NAME"); nodeName != "" {
```
ellistarn, line 80: Is this env var already defined?

pkg/ipamd/ipamd.go
```diff
@@ -2371,3 +2302,31 @@ func (c *IPAMContext) isConfigValid() bool {
...
+		Name:      c.myNodeName,
+		Namespace: "default",
+	}
+	err := c.cachedK8SClient.Get(ctx, key, cniNode)
+	if err != nil {
+		return err
+	}
+
+	var names []rcv1alpha1.FeatureName
+	for _, feature := range cniNode.Spec.Features {
```
ellistarn, line 2318: Check out: https://github.com/samber/lo#containsby

```
if lo.ContainsBy(cniNode.Spec.Features, func(feature rcv1alpha1.Feature) bool {
  return feature == rcv1alpha1.Feature{ Name: featureName, Value: featureValue }
}) {
  return nil
}
```

pkg/utils/algo/algos.go
ellistarn, line 6: Consider using github.com/samber/lo for functional interfaces like this.

pkg/sgpp/utils.go
haouc: >the CNINode is being owned by the CNI-agent
I am not sure if this is correct since the agent will change for any reason but the CNINode should be living with the node. My preference is
- VPC RC (sort of) manage "nodes" and their owning CNINode objects
- VPC CNI doesn't have to worry about the CNINode creation since it should be taken care by VPC RC
- VPC CNI shouldn't have create permission on this resource
- VPC RC managing these objects will benefit future improvements (we can sync offline)

pkg/ipamd/ipamd.go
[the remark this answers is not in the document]
haouc: I don't think we support reconciling on the eniconfig label currently. Can you double check? I checked the code in master branch and seems we only do that in init.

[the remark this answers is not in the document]
jdn5126: Right, we do not currently support reconciling on ENIConfig label, so this is not a change in behavior, I am just pointing out that our restart strategy here is to crash the `aws-node` pod when the CRD does not exist. Which may not be the best idea as @ellistarn pointed out, but it's the least complex

pkg/k8sapi/k8sutils.go
haouc: It is defined in CNI config yaml.
https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/master/aws-k8s-cni.yaml#L204

pkg/sgpp/utils.go
ellistarn: Worth a discussion, I think.

pkg/ipamd/ipamd.go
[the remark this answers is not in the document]
ellistarn: Isn't it straightforward for us to write a controller-runtime reconciler that's included with the ipamd server?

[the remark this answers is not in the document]
jdn5126: If you mean to reconcile on the CRD, then yeah, but I think that overhead is not worth it. We have a push to decrease the amount of API server calls from IPAMD: https://github.com/aws/amazon-vpc-cni-k8s/issues/2160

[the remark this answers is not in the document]
ellistarn: ack, this is a good idea for daemons. As an alternative we could backoff-poll to retrieve the CRD instead of crashing the process.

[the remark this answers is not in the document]
jdn5126: For reference, when we refer to the "reconciliation loop", it refers to this: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/ipamd.go#L685

This is the "main loop" for IPAMD, where we continuously check if we can reclaim IPs, need to attach new ENIs, etc.

pkg/ipamd/ipamd.go
haouc: nice lib. Will use this. Thanks

pkg/ipamd/ipamd.go
```diff
@@ -2371,3 +2302,31 @@ func (c *IPAMContext) isConfigValid() bool {
...
 	return true
 }
+
+func (c *IPAMContext) AddFeatureToCNINode(ctx context.Context, featureName rcv1alpha1.FeatureName, featureValue string) error {
+	cniNode := &rcv1alpha1.CNINode{}
+	key := types.NamespacedName{
+		Name:      c.myNodeName,
+		Namespace: "default",
+	}
+	err := c.cachedK8SClient.Get(ctx, key, cniNode)
```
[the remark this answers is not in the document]
haouc: Sure, let me make them inline.

pkg/utils/algo/algos.go
haouc: Changed to the powerful lib.

go.mod
```diff
@@ -171,3 +173,5 @@ replace golang.org/x/crypto => golang.org/x/crypto v0.1.0
 replace github.com/containernetworking/cni => github.com/containernetworking/cni v0.8.1
 
 replace github.com/containernetworking/plugins => github.com/containernetworking/plugins v0.9.1
+
+replace github.com/aws/amazon-vpc-resource-controller-k8s => github.com/haouc/amazon-vpc-resource-controller-k8s v1.2.3
```
ellistarn, line 177: Check out go.work files, which you can add to .gitignore and not need to edit the module: https://go.dev/doc/tutorial/workspaces

haouc: I will check the doc. My understanding is the multi modules workspace in this case will keep the change local and this PR will fail in GitHub actions? Which is fine since this `replace` change will be removed after the VPC RC has a new tag for CRD update. But I think I get your point.

ellistarn (commented):
LGTM.

charts/aws-vpc-cni/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: aws-vpc-cni
-version: 1.13.2
-appVersion: "v1.13.2"
+version: 1.13.3
+appVersion: "v1.13.3"
```
jayanthvn, line 4: Why is the appVersion updated? We generally update this during release..

charts/aws-vpc-cni/templates/clusterrole.yaml
```diff
@@ -33,3 +33,8 @@ rules:
     resources:
       - events
     verbs: ["create", "patch", "list"]
+  - apiGroups:
+      - vpcresources.k8s.aws
+    resources:
+      - cninodes
+    verbs: ["get", "list", "patch"]
```
jayanthvn, line 40: Let's discuss on adding "patch" permission offline.

go.mod
jayanthvn: @haouc lets please make sure PR is merged after tag is released, or other option is cut a RC release on VPC RC and use that tag here. Since it will break master builds.

pkg/ipamd/ipamd.go
```diff
@@ -138,9 +140,6 @@ const (
 	// vpcENIConfigLabel is used by the VPC resource controller to pick the right ENI config.
 	vpcENIConfigLabel = "vpc.amazonaws.com/eniConfig"
```
jayanthvn, line 141: Even this can be cleaned up right?

pkg/ipamd/ipamd.go
```diff
@@ -575,34 +574,19 @@ func (c *IPAMContext) nodeInit() error {
 
 	eniConfigName, err := eniconfig.GetNodeSpecificENIConfigName(node)
 	if err == nil && c.useCustomNetworking && eniConfigName != "default" {
-		// Signal to VPC Resource Controller that the node is using custom networking
-		err := c.SetNodeLabel(ctx, vpcENIConfigLabel, eniConfigName)
+		// Add the feature name to CNINode of this node
+		err := c.AddFeatureToCNINode(ctx, rcv1alpha1.CustomNetworking, eniConfigName)
```
jayanthvn, line 578: nit -> something like `AddFeatureToCNINodeResource` or anything else maybe?

pkg/ipamd/ipamd.go
```diff
@@ -1283,18 +1269,18 @@ func (c *IPAMContext) logPoolStats(dataStoreStats *datastore.DataStoreStats) {
...
-		}
-		// We need to signal that VPC Resource Controller needs to attach a trunk ENI
-		err := c.SetNodeLabel(ctx, trunkInterfaceLabel, "false")
-		if err != nil {
-			podENIErrInc("askForTrunkENIIfNeeded")
-			log.Errorf("Failed to set node label", err)
-		}
+	// Check that there is room for a trunk ENI to be attached:
+	if c.dataStore.GetENIs() >= (c.maxENI - c.unmanagedENI) {
+		log.Error("No slot available for a trunk ENI to be attached. Not labeling the node")
```
jayanthvn, line 1274: Shouldn't we add these 2 lines here -

```
podENIErrInc("askForTrunkENIIfNeeded")
log.Errorf("Failed to add SGP feature into the CNINode", err)
```

charts/aws-vpc-cni/templates/clusterrole.yaml
jayanthvn: Also we need to remove the `update` permission from node resource right?

pkg/ipamd/ipamd.go
```diff
@@ -2371,3 +2302,43 @@ func (c *IPAMContext) isConfigValid() bool {
 
 	return true
 }
+
+func (c *IPAMContext) AddFeatureToCNINode(ctx context.Context, featureName rcv1alpha1.FeatureName, featureValue string) error {
+	cniNode := &rcv1alpha1.CNINode{}
+
+	if err := c.cachedK8SClient.Get(ctx, types.NamespacedName{Name: c.myNodeName}, cniNode); err != nil {
+		return err
```
jayanthvn, line 2310: I feel, we need to add some new metrics to track CNINode updates/deletes..you can refer here - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/ipamd.go#L187

pkg/k8sapi/k8sutils.go
```diff
@@ -47,6 +49,7 @@ func CreateKubeClient(mapper meta.RESTMapper) (client.Client, error) {
 	vpcCniScheme := runtime.NewScheme()
 	clientgoscheme.AddToScheme(vpcCniScheme)
 	eniconfigscheme.AddToScheme(vpcCniScheme)
+	rcscheme.AddToScheme(vpcCniScheme)
```
jayanthvn, line 52: Have we measured the number of API server calls increase with this change?

pkg/ipamd/ipamd.go
ellistarn: Client go should give us these for free

pkg/ipamd/ipamd.go
haouc: We don't unset the features in CNINode. CNiNode itself has owner reference to the node object. When node is deleted, the CNINode will be deleted.

pkg/k8sapi/k8sutils.go
haouc: We are changing to use rawclient.

charts/aws-vpc-cni/Chart.yaml
haouc: ack

charts/aws-vpc-cni/templates/clusterrole.yaml
haouc: We need remove node update permission. The problem is we can merge the deletion to master. I have synced with Jeff offline. We may use a feature branch instead.

jdn5126 (approved):

go.mod
haouc: Yeah, @jayanthvn. We can discuss offline. This PR has been changed to target release branch as a possible RC release.

pkg/ipamd/ipamd.go
haouc: Cleaning work has been done in the latest update.

pkg/ipamd/ipamd.go
haouc: The error message has been updated in the updates.

PR #2443, opened 2023-06-27
Remove canary runs against default cluster version from nightly/weekly integration tests

jdn5126:
**What type of PR is this?**
cleanup

N/A

This PR removes a test section from `scripts/run-integration-tests.sh` where we run the canary tests against the default VPC CNI version for the cluster.

This coverage is unnecessary, as the goal for our GitHub Actions here is to run regression and performance tests against the latest code/version. Also, we get this coverage from other nightly runs. Also, we are moving towards unifying the default and latest addon versions on cluster creation. 

N/A

N/A

**Automation added to e2e**:
Test case removed

No

N/A

No

No

```release-note

```

### DIFF
--- a/scripts/run-integration-tests.sh
+++ b/scripts/run-integration-tests.sh
@@ -27,7 +27,6 @@ ARCH=$(go env GOARCH)
 : "${RUN_CONFORMANCE:=false}"
 : "${RUN_TESTER_LB_ADDONS:=false}"
 : "${RUN_KOPS_TEST:=false}"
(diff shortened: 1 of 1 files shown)


ellistarn:
How do we catch any regressions against older kubernetes versions?

jdn5126:
> How do we catch any regressions against older kubernetes versions?

That comes from our internal prow testing. The general strategy is that GitHub Actions covers new (unreleased) code and prow covers released versions against all k8s versions, but there is a lot of overlap.

ellistarn (approved):
Nice!

PR #2445, opened 2023-06-27
Add custom-networking integration test covering ENIConfig objects with no security groups

jdn5126:
**What type of PR is this?**
testing

N/A

This PR adds integration test coverage for `custom-networking` when ENIConfig objects are created without any security groups. The test does the following:
1. Deletes existing ENIConfigs and recreates them without security groups
2. Terminates nodes and waits for them to be recycled
3. Creates a deployment and verifies that it succeeds

More connectivity cases may be added in the future, but pod connectivity is skipped from this new case as it is trivial and covered in the above cases. It also significantly increases the runtime.

N/A

Stress tested case to make sure that it is not flaky

**Automation added to e2e**:
Added test case

No

No, Yes

No

No

```release-note

```


jaydeokar (approved):
lgtm

Issue #2437, opened 2023-06-22
Add memory usage tests for daily and release runs


jayanthvn:
**What would you like to be added**:
Capture aws-node memory usage during startup, pod churns, long lived clusters/pods and so on

**Why is this needed**:
Identify memory leaks

jdn5126:
Closing as resolved by #2455

PR #2455, opened 2023-07-03
Refactor performance tests: capture memory usage, add README

jdn5126:
**What type of PR is this?**
enhancement, testing

https://github.com/aws/amazon-vpc-cni-k8s/issues/2437

This PR refactors and cleans up `scripts/run-integration-tests.sh`, primarily for performance testing. It also captures memory usage of the `aws-node` pod during performance tests and uploads it to an S3 bucket. It also adds a README for running `run-integration-tests.sh` and cleans up unused variables.

The usage of the memory statistics may be extended in the future, i.e. to compare to previous runs, but for now we only flag when memory usage exceeds our recommended value: 200Mi.

N/A

Manually running integration tests and evaluating performance statistics.

**Automation added to e2e**:
Added capturing of memory usage by `aws-node` pods to performance tests.

No

No, Yes

No

No

```release-note

```

### DIFF
--- a/.github/workflows/weekly-cron-tests.yaml
+++ b/.github/workflows/weekly-cron-tests.yaml
@@ -23,7 +23,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: "1.19"
+          go-version: "1.20"
       - name: Set up tools
         run: |
           # Install ginkgo version from go.mod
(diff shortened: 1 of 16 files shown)


.github/workflows/weekly-cron-tests.yaml
```diff
@@ -39,55 +39,47 @@ jobs:
...
           RUN_BOTTLEROCKET_TEST: true
           RUN_TESTER_LB_ADDONS: true
-          RUN_INTEGRATION_DEFAULT_CNI: false
         run: |
           ./scripts/run-integration-tests.sh
         if: always()
       - name: Run calico tests
         env:
           DISABLE_PROMPT: true
-          S3_BUCKET_CREATE: false
```
orsenthil, line 83: What's the motivation of these changes, especially the removal of S3 bucket creates and setting RUN_INTEGRATION_DEFAULT_CNI to false in the weekly cron?

orsenthil (approved):
LGTM. Could you please paste an output of the test run after this refactor. It will be good to verify what what to expect, and see a successful run. 


Thank you.

.github/workflows/weekly-cron-tests.yaml
jdn5126: `S3_BUCKET_CREATE` and `S3_BUCKET_NAME` were unused variables that appear to have been missed in a previous cleanup, so I removed them here.

For `RUN_INTEGRATION_DEFAULT_CNI`, we were running the CNI integration tests with every invocation of `run-integration-tests`. For the weekly tests, that means we were running them 4 times with no real benefit. Since they already run nightly as part of the `nightly-cron-tests`, I concluded they should be removed from the weekly tests. This should drop our weekly test runtime by about 1.5 hours, as well.

jdn5126:
For test results, I kicked off https://github.com/aws/amazon-vpc-cni-k8s/actions/runs/5488008758 to show the results from GitHub

jdn5126:
> For test results, I kicked off https://github.com/aws/amazon-vpc-cni-k8s/actions/runs/5488008758 to show the results from GitHub

Oh, actually that won't work, as the run cannot happen with the new workflow file until this merges. For example output, I can show the latest S3 bucket results from my manual run:
```
Iteration, scaleUpTime (s), scaleDownTime (s), scaleUpMem (Mi), scaleDownMem (Mi)
1, 13, 25, 57, 59
2, 13, 26, 59, 59
3, 13, 25, 58, 59
```

Issue #2436, opened 2023-06-22
10x memory usage in `1.13.2` compared to `1.12.x`


alam0rt:
Deploying 1.13.2 revealed that the memory usage at both start up + after running for some time has drastically increased.

At start up, we saw memory usage climb to 450 or so Mi and settle down to about 400Mi.

This appears to increase with the number of nodes in the cluster.

![image](https://github.com/aws/amazon-vpc-cni-k8s/assets/35046326/40fe43f5-0545-4fa8-b7a1-327447b67248)
The red bar is OOM kills, pre-spike is running 1.12.0, post is 1.13.2.

`1.12.0`
```
aws-cni-pqd9g                           6m           63Mi
aws-cni-q6x2x                           7m           40Mi
aws-cni-q7sbw                           8m           63Mi
aws-cni-q8msl                           10m          57Mi
aws-cni-rfrll                           5m           54Mi
aws-cni-s2qwn                           7m           55Mi
aws-cni-s9ndl                           6m           64Mi
aws-cni-s9pvj                           6m           54Mi
aws-cni-sg9g7                           6m           61Mi
aws-cni-srvvz                           6m           40Mi
aws-cni-szqg8                           6m           52Mi
aws-cni-v8gvc                           7m           64Mi
aws-cni-vjhff                           6m           40Mi
aws-cni-w56kk                           6m           52Mi
aws-cni-w9mb8                           6m           53Mi
aws-cni-wmxwf                           8m           40Mi
aws-cni-wtpgg                           6m           53Mi
aws-cni-wxrtp                           6m           54Mi
```

`1.13.2`
```
aws-cni-wbk75                            7m           413Mi
aws-cni-wg857                            8m           434Mi
aws-cni-wgkrj                            8m           422Mi
aws-cni-wkqnm                            11m          433Mi
aws-cni-wlzlw                            8m           428Mi
aws-cni-x22hb                            7m           424Mi
aws-cni-x8r4n                            6m           447Mi
aws-cni-xk2hd                            7m           426Mi
aws-cni-xw4ts                            5m           432Mi
aws-cni-zt5wr                            6m           446Mi
aws-cni-zxch8                            8m           453Mi
```

**What happened**:
Updated to 1.13.2

**Attach logs**

**What you expected to happen**:

For aws-cni not to use 10 times the memory.

**How to reproduce it (as minimally and precisely as possible)**:

Deploy 1.13.2

**Anything else we need to know?**:

**Environment**:
[...]

PR #2463, opened 2023-07-11
Decrease memory usage by K8S Clients

jdn5126:
**What type of PR is this?**
bug

https://github.com/aws/amazon-vpc-cni-k8s/issues/2436

This PR modifies the schemas used by the raw and cached Kubernetes clients in the IPAMD and Metrics Helper agents. These clients only need to add the `corev1` scheme: https://github.com/kubernetes/api/blob/v0.26.5/core/v1/register.go#L45 , as the clients only make GET and LIST calls for objects added by this scheme.

Decreasing the number of objects loaded decreases the memory consumed by these clients. For IPAMD, a selector is added for pods such that only pods on the deployed node are cached. Metrics Helper does not add this selector, as it needs to query `aws-node` pods on all nodes. Metrics Helper already has a label selector in its pod watcher.

Running the performance tests yielded the following memory decrease:
`v1.13.2`:
* Base memory usage of aws-node pod: 44Mi
* After deploying 5000 pods: 107Mi
* After scaling down to 0 pods: 70Mi

`this PR`:
* Base memory usage of aws-node pod: 30Mi
* After deploying 5000 pods: 45Mi
* After scaling down to 0 pods: 45Mi

I also validated that all integration tests pass. I will schedule a manual run against this PR as well.

**Automation added to e2e**:
N/A

No

No, Yes

No

Yes

```release-note
Decrease `aws-node` and `cni-metrics-helper` memory usage.
```

### DIFF
--- a/cmd/aws-k8s-agent/main.go
+++ b/cmd/aws-k8s-agent/main.go
@@ -52,7 +52,7 @@ func _main() int {
 		return 1
 	}
 
-	cacheK8SClient, err := k8sapi.CreateCachedKubeClient(rawK8SClient, mapper)
+	cacheK8SClient, err := k8sapi.CreateCachedKubeClient(rawK8SClient, mapper, true)
 	if err != nil {
 		log.Errorf("Failed to create cached kube client: %s", err)
 		return 1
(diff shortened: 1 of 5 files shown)


jdn5126:
Integration test run: https://github.com/aws/amazon-vpc-cni-k8s/actions/runs/5522624871

haouc (approved):
lgtm

Issue #2466, opened 2023-07-13
Panic in event recorder with missing IAM permission - 1.13.2


jayanthvn:
**What happened**:

If there is a node IAM permission issue then on aws-node startup, refreshSGID which is called as part of awsutils init will fail to update the ENI and tries to add a pod event. But the event recorder is initialized after awsutils init.

We are passing `c.eventRecorder` - > https://github.com/aws/amazon-vpc-cni-k8s/blob/v1.13.2/pkg/ipamd/ipamd.go#L399 

But the value is set here - https://github.com/aws/amazon-vpc-cni-k8s/blob/v1.13.2/pkg/ipamd/ipamd.go#L412

Needs code reordering.

**Attach logs**

**What you expected to happen**:
No panic, just log event.

**How to reproduce it (as minimally and precisely as possible)**:
Remove IAM permission and restart aws-node.

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- CNI Version - 1.13.2
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):

jdn5126:
Closing as fixed by #2467

Issue #2448, opened 2023-06-29
Panic on v1.13.2


jose-ledesma:
**What happened**:
VPC CNI panic

**Attach logs**
```
$ kubectl logs aws-node-4pts8 -p
Installed /host/opt/cni/bin/aws-cni
Installed /host/opt/cni/bin/egress-cni
time="2023-06-29T15:12:21Z" level=info msg="Starting IPAM daemon... "
time="2023-06-29T15:12:21Z" level=info msg="Checking for IPAM connectivity... "
time="2023-06-29T15:12:23Z" level=info msg="Copying config file... "
time="2023-06-29T15:12:23Z" level=info msg="Successfully copied CNI plugin binary and config file."
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x560ff5791b21]

goroutine 398 [running]:
github.com/aws/amazon-vpc-cni-k8s/pkg/awsutils.(*EC2InstanceMetadataCache).DescribeAllENIs(0xc0009508c0)
	/go/src/github.com/aws/amazon-vpc-cni-k8s/pkg/awsutils/awsutils.go:1239 +0xd01
github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd.(*IPAMContext).nodeIPPoolReconcile(0xc000950780, {0x560ff6815970, 0xc000134020}, 0xdf8475800)
	/go/src/github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd/ipamd.go:1412 +0x5e9
github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd.(*IPAMContext).StartNodeIPPoolManager(0xc000950780)
	/go/src/github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd/ipamd.go:699 +0x66
created by main._main
	/go/src/github.com/aws/amazon-vpc-cni-k8s/cmd/aws-k8s-agent/main.go:68 +0x42b
time="2023-06-29T15:12:28Z" level=error msg="Failed to wait for IPAM daemon to complete" error="exit status 2"
```

**What you expected to happen**:
`aws-node` pods don't panic ^^

**How to reproduce it (as minimally and precisely as possible)**:
unsure

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
```
[...]

PR #2467, opened 2023-07-13
Fix event recorder initialization and add check to log

jdn5126:
**What type of PR is this?**
bug

https://github.com/aws/amazon-vpc-cni-k8s/issues/2466
https://github.com/aws/amazon-vpc-cni-k8s/issues/2448

For #2466 , this PR fixes an issue with creating events when IAM permissions are missing. The EventRecorder object was not being properly initialized. Also, the `EventRecorder` package is now more modular, so dependent packages can simply call `eventrecorder.Get()` to get the object. Initialization is left to `aws-k8s-agent` on startup.

For #2448 , this PR fixes a logging issue where a malformed EC2 response (missing `Attachment`) was trying to access the missing field in order to log.

N/A

Manually verified that integration tests pass. Scheduling runner against this PR.

**Automation added to e2e**:
N/A

No

No, Yes

No

Yes

```release-note
Fix issue where missing IAM permissions did not generate Event; handle malformed EC2 response
```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -80,9 +80,6 @@ var (
 
 var log = logger.Get()
 
-// event recorder to raise aws-node pod events eg. if EC2 API calls fail due to UnauthorizedOperation error
-var eventRecorder *eventrecorder.EventRecorder
-
 var (
 	awsAPILatency = prometheus.NewSummaryVec(
 		prometheus.SummaryOpts{
@@ -248,8 +245,6 @@ type EC2InstanceMetadataCache struct {
 
 	imds   TypedIMDS
 	ec2SVC ec2wrapper.EC2
-
-	eventRecorder *eventrecorder.EventRecorder
 }
 
 // ENIMetadata contains information about an ENI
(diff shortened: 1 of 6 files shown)


jdn5126:
https://github.com/aws/amazon-vpc-cni-k8s/actions/runs/5548274272

pkg/awsutils/awsutils.go
```diff
@@ -419,28 +415,33 @@ func New(useCustomNetworking, disableLeakedENICleanup, v4Enabled, v6Enabled bool
...
 	}
 
+	// Setup Event Recorder, which creates a Kubernetes event when IAM permissions are missing
+	eventRecorder, err := eventrecorder.New(cachedK8SClient)
+	if err != nil {
+		log.Errorf("Failed to create event recorder: %v", err)
+		return nil, errors.Wrapf(err, "failed to create event recorder")
+	}
+	eventRecorder.HostID = cache.instanceID
+	cache.eventRecorder = eventRecorder
```
jayanthvn, line 439: nit: we are adding `eventRecorder` to `EC2InstanceMetadataCache` which is not a metadata property. It would have been good if we had something like - `var log = logger.Get()` for eventRecorder...

jdn5126: Yeah, `EC2InstanceMetadataCache` is the only user today, but it would be better long-term if `EventRecorder` created itself and others could just call `Get`. I'll try that out tomorrow

jdn5126: This has been refactored in the latest revision

jayanthvn (approved):
lgtm

jdn5126:
New integration test run: https://github.com/aws/amazon-vpc-cni-k8s/actions/runs/5557806856

pkg/utils/eventrecorder/eventrecorder.go
```diff
@@ -89,3 +94,16 @@ func findMyPod(cachedK8SClient client.Client) (corev1.Pod, error) {
 	}
 	return pod, err
 }
+
+// Functions used for mocking package
+func InitMockEventRecorder() *events.FakeRecorder {
```
orsenthil, line 99: Post merge review; a mock recorder could be test files as, I hope, we don't use this in real code.

PR #2471, opened 2023-07-18
Enable Cloudwatch insights for performance tests

jaydeokar:
**What type of PR is this?**
automation

N/A

This collects pod memory, CPU and other metrics through cloudwatchagent and pushes it to cloudwatch insights. It will collect metrics every 30s which gives us granular information about resource usage during performance tests

N/A

Yes

**Automation added to e2e**:
N/A

N/A

N/A

N/A

N/A

### DIFF
--- a/scripts/test/config/perf-cluster.yml
+++ b/scripts/test/config/perf-cluster.yml
@@ -2,6 +2,8 @@
 ---
 apiVersion: eksctl.io/v1alpha5
 kind: ClusterConfig
+iam:
+  withOIDC: true
 
 metadata:
   name: CLUSTER_NAME_PLACEHOLDER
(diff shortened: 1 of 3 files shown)


jdn5126 (commented):
Change LGTM. Do we need the metrics-server if we have CloudWatch logs?

jaydeokar:
> Change LGTM. Do we need the metrics-server if we have CloudWatch logs?

No we don't. I kept it still as we don't want to remove the existing metrics collection flow (s3 bucket). We can clean that up once we have the right dashboards and metrics in place with CW

scripts/test/config/perf-cluster.yml
```diff
@@ -2,6 +2,8 @@
 ---
 apiVersion: eksctl.io/v1alpha5
 kind: ClusterConfig
+iam:
```
orsenthil, line 5: Is this more for security / improvement or is it required?

jaydeokar: I wanted to use IRSA for creating the SA/permissions for cloudwatch agent. Unless we enable this, IRSA doesn't work

jdn5126 (approved):

Issue #2474, opened 2023-07-19
Expand integration test coverage for unmanaged ENIs


jdn5126:
**What would you like to be added**:
An integration test dedicated to unmanaged ENIs

**Why is this needed**:
Recent refactoring around node initialization led to a regression related to unmanaged ENIs: https://github.com/aws/amazon-vpc-cni-k8s/issues/2470. We need better integration test coverage for unmanaged ENIs to prevent future regressions.

Issue #2470, opened 2023-07-18
IPAMD changes security group of un-managed ENIs


constant-mihai:
**What happened**:

We are attaching extra ENIs to the EKS nodes.
We assign dedicated security groups to these ENIs.
These ENIs are tagged with "node.k8s.amazonaws.com/no_manage: true".
We upgraded the VPC CNI add-on from "v1.13.0-eksbuild.1" -> "v1.13.2-eksbuild.1".
The security group of the extra ENIs changed from the dedicated one to the security group of the node.

**Attach logs**

```
{"level":"info","ts":"2023-07-17T15:12:05.081Z","caller":"aws-k8s-agent/main.go:27","msg":"Starting L-IPAMD   ..."}
{"level":"info","ts":"2023-07-17T15:12:05.081Z","caller":"aws-k8s-agent/main.go:38","msg":"Testing communication with server"}
{"level":"info","ts":"2023-07-17T15:12:05.095Z","caller":"wait/wait.go:222","msg":"Successful communication with the Cluster! Cluster Version is: v1.27+. git version: v1.27.2-eks-c12679a. git tree state: clean. commit: 002c6bc4e142b1f48b9405140e559194a094dcff. platform: linux/amd64"}
{"level":"warn","ts":"2023-07-17T15:12:05.105Z","caller":"awssession/session.go:65","msg":"HTTP_TIMEOUT env is not set or set to less than 10 seconds, defaulting to httpTimeout to 10sec"}
{"level":"debug","ts":"2023-07-17T15:12:05.107Z","caller":"ipamd/ipamd.go:399","msg":"Discovered region: us-east-1"}
{"level":"info","ts":"2023-07-17T15:12:05.107Z","caller":"ipamd/ipamd.go:399","msg":"Custom networking enabled false"}
{"level":"debug","ts":"2023-07-17T15:12:05.107Z","caller":"awsutils/awsutils.go:434","msg":"Found availability zone: us-east-1c "}
{"level":"debug","ts":"2023-07-17T15:12:05.108Z","caller":"awsutils/awsutils.go:434","msg":"Discovered the instance primary IPv4 address: 10.70.116.114"}
{"level":"debug","ts":"2023-07-17T15:12:05.108Z","caller":"awsutils/awsutils.go:434","msg":"Found instance-id: i-0f7052bb03aa5c778 "}
[...]

PR #2475, opened 2023-07-19
RefreshSecurityGroups must be called after unmanaged ENIs are set

jdn5126:
**What type of PR is this?**
bug

https://github.com/aws/amazon-vpc-cni-k8s/issues/2470

This PR moves the call to `RefreshSGIDs` to `nodeInit` after unmanaged ENIs are set. This is required to fix a regression where security groups on unmanaged ENIs were getting modified during node init.

N/A

This is caused by https://github.com/aws/amazon-vpc-cni-k8s/pull/2427, which fixed https://github.com/aws/amazon-vpc-cni-k8s/issues/2426. I manually verified that https://github.com/aws/amazon-vpc-cni-k8s/issues/2426 is still fixed.

I also manually verified the dependencies on the ordering here. All integration tests pass, and I am kicking off a GitHub runner for this as well.

I filed https://github.com/aws/amazon-vpc-cni-k8s/issues/2474 to add more integration test coverage for unmanaged ENIs.

**Automation added to e2e**:
N/A

No

No, Yes

No

Yes

```release-note
Fix regression where unmanaged ENIs had their security groups modified during init.
```

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -426,21 +426,7 @@ func New(rawK8SClient client.Client, cachedK8SClient client.Client) (*IPAMContex
 	checkpointer := datastore.NewJSONFile(dsBackingStorePath())
 	c.dataStore = datastore.NewDataStore(log, checkpointer, c.enablePrefixDelegation)
 
-	// Retrieve security groups
-	mac := c.awsClient.GetPrimaryENImac()
-	if c.enableIPv4 && !c.disableENIProvisioning {
-		err = c.awsClient.RefreshSGIDs(mac)
-		if err != nil {
-			return nil, err
-		}
-
-		// Refresh security groups and VPC CIDR blocks in the background
-		// Ignoring errors since we will retry in 30s
-		go wait.Forever(func() { _ = c.awsClient.RefreshSGIDs(mac) }, 30*time.Second)
-	}
-
-	err = c.nodeInit()
-	if err != nil {
+	if err := c.nodeInit(); err != nil {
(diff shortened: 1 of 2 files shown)


jdn5126:
https://github.com/aws/amazon-vpc-cni-k8s/actions/runs/5601184868

pkg/ipamd/ipamd.go
```diff
@@ -552,6 +538,19 @@ func (c *IPAMContext) nodeInit() error {
 		vpcV4CIDRs = c.updateCIDRsRulesOnChange(vpcV4CIDRs)
 	}, 30*time.Second)
 
+	// Retrieve security groups
```
jayanthvn, line 541: We could have moved the logic after `nodeInit` is successful but should be fine since if nodeInit fails then aws-node will be in crash loop..Can you update the comment since it has a tight dependency on unmanaged ENIs and ENIs on multicard interfaces?

jayanthvn (commented):
minor nit..rest lgtm.

pkg/ipamd/ipamd.go
jdn5126: We can't move it after `nodeInit` is successful as the first call to `RefreshSGIDs` populates the ENI cache, and that cache must be populated before we attach any new ENIs

jdn5126: I'll update the comment to mention the dependency

jdn5126: Updated in latest commit

jayanthvn (approved):
lgtm

Issue #2378, opened 2023-05-10
Configurability for IP cooldown period


jdn5126:
**What would you like to be added**:
An environment variable to configure the pod IP cooldown period: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/datastore/data_store.go#L40. Configuring it to 0 seconds should be allowed. Currently, the period is hard-coded to 30 seconds.

**Why is this needed**:
Customers with bursts of pod churn can have pods sparsely populated across ENIs, which leads to more VPC IP consumption than desired. While IPs are in cooldown, new ENIs are allocated to service pod ADDs.

Note that changing the default IP cooldown period can cause issues with kube-proxy, as the main motivation for the cooldown period is to give kube-proxy a chance to cleanup endpoints. If you do not use services, setting the value to 0 is a valid option. If you want to take advantage of kube-proxy performance improvements over time, tuning the setting to a lower value is a valid option.

jdn5126:
Closing as #2492 has merged

PR #2492, opened 2023-08-09
IP_COOLDOWN_PERIOD environment variable for ip cooldown period configuration

jchen6585:
**What type of PR is this?**
Continuing PR #2397 

Issue #2378 

Continuing PR #2397: Added new environment variable IP_COOLDOWN_PERIOD.  cmd/aws-vpc-cni/main.go defines two new variables in the const section, envIPCooldownPeriod with IP_COOLDOWN_PERIOD and defaultIPCooldownPeriod as 30s time.Duration. pkg/ipamd/datastore/data_store.go creates a new function getCooldownPeriod(), returns a time.Duration being either the default cooldown period of 30s or the custom one if it is not smaller than 0. getCooldownPeriod() is called when creating a new DataStore, storing the cooldown period value in a variable ipCooldownPeriod which is used in inCoolingPeriod(). pkg/ipamd/datastore/data_store_test.go creates a new test case TestGetIPStatsV4WithCustomIPCooldown(). util/utils.go adds a new function GetIntAsStringEnvVar() to parse the environment string into an integer.

Ran against cni and ipamd integration tests with IP_COOLDOWN_PERIOD values set as: unset, 0, 10, 20
Screenshot 2023-08-09 at 2 25 25 PM
Screenshot 2023-08-09 at 2 26 11 PM

**Automation added to e2e**:

No

No

Have not updated a running cluster

No

Yes

```release-note
#### `IP_COOLDOWN_PERIOD` (v1.13.4+)

Type: Integer as a String

Default: `30`

Specifies the number of seconds an IP address is in cooldown after pod deletion. The cooldown period gives kube-proxy time to update node iptables rules when the IP was registered as a valid endpoint, such as for a service. Modify this value with caution, as kube-proxy update time scales with the number of nodes and services.

**Note:** 0 is a supported value, however it is highly discouraged.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -680,6 +680,19 @@ This environment variable must be set for both the `aws-vpc-cni-init` and `aws-n
 
 Note that enabling/disabling this feature only affects whether newly created pods have an IPv6 interface created. Therefore, it is recommended that you reboot existing nodes after enabling/disabling this feature. Also note that if you are using this feature in conjunction with `ENABLE_POD_ENI` (Security Groups for Pods), the security group rules will NOT be applied to egressing IPv6 traffic.
 
+----
+
+#### `IP_COOLDOWN_PERIOD` (v1.14.1+)
+
+Type: Integer as a String
+
+Default: `30`
+
+Specifies the number of seconds an IP address is in cooldown after pod deletion. The cooldown period gives network proxies, such as kube-proxy, time to update node iptables rules when the IP was registered as a valid endpoint, such as for a service. Modify this value with caution, as kube-proxy update time scales with the number of nodes and services.
+
+**Note:** 0 is a supported value, however it is highly discouraged.
+**Note:** Higher cooldown periods may lead to a higher number of EC2 API calls as IPs are in cooldown cache.
+
 ### VPC CNI Feature Matrix
 
 IP Mode | Secondary IP Mode | Prefix Delegation | Security Groups Per Pod | WARM & MIN IP/Prefix Targets | External SNAT
--- a/cmd/aws-vpc-cni/main.go
+++ b/cmd/aws-vpc-cni/main.go
@@ -80,6 +80,7 @@ const (
 	vpcCniInitDonePath           = "/vpc-cni-init/done"
 	defaultEnBandwidthPlugin     = false
 	defaultEnPrefixDelegation    = false
+	defaultIPCooldownPeriod      = 30
 
 	envHostCniBinPath        = "HOST_CNI_BIN_PATH"
 	envHostCniConfDirPath    = "HOST_CNI_CONFDIR_PATH"
@@ -99,6 +100,7 @@ const (
 	envEnIPv6                = "ENABLE_IPv6"
 	envEnIPv6Egress          = "ENABLE_V6_EGRESS"
 	envRandomizeSNAT         = "AWS_VPC_K8S_CNI_RANDOMIZESNAT"
+	envIPCooldownPeriod      = "IP_COOLDOWN_PERIOD"
 )
 
 // NetConfList describes an ordered list of networks.
@@ -345,6 +347,13 @@ func validateEnvVars() bool {
 			log.Errorf("%s must be set to either 'strict' or 'standard'", envPodSGEnforcingMode)
 			return false
 		}
+	}
+
+	// Validate that IP_COOLDOWN_PERIOD is a valid integer
+	ipCooldownPeriod, err, input := utils.GetIntFromStringEnvVar(envIPCooldownPeriod, defaultIPCooldownPeriod)
+	if err != nil || ipCooldownPeriod < 0 {
+		log.Errorf("IP_COOLDOWN_PERIOD MUST be a valid positive integer. %s is invalid", input)
+		return false
 	}
 
 	prefixDelegationEn := utils.GetBoolAsStringEnvVar(envEnPrefixDelegation, defaultEnPrefixDelegation)
--- a/pkg/ipamd/datastore/data_store.go
+++ b/pkg/ipamd/datastore/data_store.go
@@ -27,6 +27,7 @@ import (
 	"golang.org/x/sys/unix"
 
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
+	"github.com/aws/amazon-vpc-cni-k8s/utils"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 )
@@ -35,9 +36,8 @@ const (
 	// minENILifeTime is the shortest time before we consider deleting a newly created ENI
 	minENILifeTime = 1 * time.Minute
 
-	// addressCoolingPeriod is used to ensure an IP not get assigned to a Pod if this IP is used by a different Pod
-	// in addressCoolingPeriod
-	addressCoolingPeriod = 30 * time.Second
+	// envIPCooldownPeriod (default 30 seconds) specifies the time after pod deletion before an IP can be assigned to a new pod
+	envIPCooldownPeriod = "IP_COOLDOWN_PERIOD"
 
 	// DuplicatedENIError is an error when caller tries to add an duplicate ENI to data store
 	DuplicatedENIError = "data store: duplicate ENI"
@@ -249,12 +249,12 @@ type CidrStats struct {
 }
 
 // Gets number of assigned IPs and the IPs in cooldown from a given CIDR
-func (cidr *CidrInfo) GetIPStatsFromCidr() CidrStats {
+func (cidr *CidrInfo) GetIPStatsFromCidr(ipCooldownPeriod time.Duration) CidrStats {
 	stats := CidrStats{}
 	for _, addr := range cidr.IPAddresses {
 		if addr.Assigned() {
 			stats.AssignedIPs++
-		} else if addr.inCoolingPeriod() {
+		} else if addr.inCoolingPeriod(ipCooldownPeriod) {
 			stats.CooldownIPs++
 		}
 	}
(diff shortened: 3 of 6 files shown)


README.md
```diff
@@ -680,6 +680,18 @@ This environment variable must be set for both the `aws-vpc-cni-init` and `aws-n
 
 Note that enabling/disabling this feature only affects whether newly created pods have an IPv6 interface created. Therefore, it is recommended that you reboot existing nodes after enabling/disabling this feature. Also note that if you are using this feature in conjunction with `ENABLE_POD_ENI` (Security Groups for Pods), the security group rules will NOT be applied to egressing IPv6 traffic.
 
+----
+
+#### `IP_COOLDOWN_PERIOD` (v1.13.4+)
```
jdn5126, line 685: This will end up being `v1.14.1+`, as we will have to hold off on merging this until the `v1.14.0` release is complete

README.md
```diff
@@ -680,6 +680,18 @@ This environment variable must be set for both the `aws-vpc-cni-init` and `aws-n
...
+
+Type: Integer as a String
+
+Default: `30`
+
+Specifies the number of seconds an IP address is in cooldown after pod deletion. The cooldown period gives kube-proxy time to update node iptables rules when the IP was registered as a valid endpoint, such as for a service. Modify this value with caution, as kube-proxy update time scales with the number of nodes and services.
```
jdn5126, line 691: nit: let's change "gives kube-proxy" to "gives network proxies, such as kube-proxy,"

cmd/aws-vpc-cni/main.go
```diff
@@ -352,6 +354,17 @@ func validateEnvVars() bool {
 	warmPrefixTarget := utils.GetEnv(envWarmPrefixTarget, "0")
 	minimumIPTarget := utils.GetEnv(envMinIPTarget, "0")
 
+	// Validate that IP_COOLDOWN_PERIOD is a valid integer
```
jdn5126, line 357: nit: let's move this above the variables defined above. Otherwise it makes that logic harder to read

utils/utils.go
```diff
@@ -21,6 +21,20 @@ func GetBoolAsStringEnvVar(env string, defaultVal bool) bool {
 	return defaultVal
 }
 
+// Parse environment variable and return integer representation of string, or default value if environment variable is unset
+func GetIntAsStringEnvVar(env string, defaultVal int) (int, error) {
```
jdn5126, line 25: nit: maybe `GetIntFromStringEnvVar`?

cmd/aws-vpc-cni/main.go
```diff
@@ -352,6 +354,17 @@ func validateEnvVars() bool {
...
+	ipCooldownPeriod, err := utils.GetIntAsStringEnvVar(envIPCooldownPeriod, defaultIPCooldownPeriod)
+	if err != nil {
+		log.Errorf("IP_COOLDOWN_PERIOD has to be a valid integer")
```
jdn5126, line 360: For both error cases, let's print the value that the user passed

pkg/ipamd/datastore/data_store.go
```diff
@@ -266,9 +274,19 @@ func (addr AddressInfo) Assigned() bool {
...
 }
 
+// getCooldownPeriod returns the time duration in seconds configured by the IP_COOLING_PERIOD env variable
+func (ds *DataStore) getCooldownPeriod() time.Duration {
+	cooldownVal, err := utils.GetIntAsStringEnvVar(envIPCooldownPeriod, 30)
+	if err != nil {
+		ds.log.Debugf("Error parsing cooldown period, using default cooldown period: 30")
+		return 30 * time.Second
+	}
+	return time.Duration(cooldownVal) * time.Second
```
jdn5126, line 284: should be no need for `time.Duration` here. It is just an integer times `time.Second`

pkg/ipamd/datastore/data_store.go
```diff
@@ -341,13 +359,15 @@ func prometheusRegister() {
...
 	prometheusRegister()
-	return &DataStore{
+	ds := &DataStore{
 		eniPool:      make(ENIPool),
 		log:          log,
 		backingStore: backingStore,
 		netLink:      netlinkwrapper.NewNetLink(),
 		isPDEnabled:  isPDEnabled,
 	}
+	ipCooldownPeriod = ds.getCooldownPeriod()
```
jdn5126, line 369: I still think we should just make this as a member variable of `DataStore`

jdn5126 (changes requested):

pkg/ipamd/datastore/data_store.go
jchen6585: Removing time.Duration results in this error `invalid operation: cooldownVal * time.Second (mismatched types int and time.Duration)`

pkg/ipamd/datastore/data_store.go
```diff
@@ -35,9 +36,9 @@ const (
 	// minENILifeTime is the shortest time before we consider deleting a newly created ENI
 	minENILifeTime = 1 * time.Minute
 
-	// addressCoolingPeriod is used to ensure an IP not get assigned to a Pod if this IP is used by a different Pod
-	// in addressCoolingPeriod
-	addressCoolingPeriod = 30 * time.Second
+	// envIPCooldownPeriod is used to ensure an IP not get assigned to a Pod if this IP is used by a different Pod
+	// in envIPCooldownPeriod
```
jdn5126, line 40: nit: let's clean up this comment to say: "envIPCooldownPeriod (default 30 seconds) specifies the time after pod deletion before an IP can be assigned to a new pod"

pkg/ipamd/datastore/data_store_test.go
```diff
@@ -925,7 +925,6 @@ func TestGetIPStatsV4(t *testing.T) {
 		*ds.GetIPStats("4"),
 	)
 
-	// wait 30s (cooldown period)
```
jdn5126, line 928: nit: this comment should remain, as it still applies

pkg/ipamd/datastore/data_store_test.go
```diff
@@ -976,7 +975,6 @@ func TestGetIPStatsV4WithPD(t *testing.T) {
...
```
jdn5126, line 979: this comment should remain, as it still applies

pkg/ipamd/datastore/data_store_test.go
```diff
@@ -925,7 +925,6 @@ func TestGetIPStatsV4(t *testing.T) {
...
 	time.Sleep(30 * time.Second)
```
jdn5126, line 928: We should take this opportunity to fix this test, as sleeping for 30 seconds is not ideal. Maybe we can set the cooldown period lower to start? Or change the time

jdn5126 (commented):
A few comments, this is almost there

utils/utils.go
```diff
@@ -21,6 +21,20 @@ func GetBoolAsStringEnvVar(env string, defaultVal bool) bool {
...
 
+// Parse environment variable and return integer representation of string, or default value if environment variable is unset
+func GetIntFromStringEnvVar(env string, defaultVal int) (int, error, string) {
+	if val, ok := os.LookupEnv(env); ok {
+		parsedVal, err := strconv.Atoi(val)
+		if err == nil {
+			return parsedVal, nil, val
+		}
+		log.Errorf("Failed to parse variable %s with value %s as integer", env, val)
+		return 0, err, val
```
jdn5126, line 32: nit: let's return -1 in the failure case

cmd/aws-vpc-cni/main.go
```diff
@@ -347,6 +349,17 @@ func validateEnvVars() bool {
 		}
 	}
 
+	// Validate that IP_COOLDOWN_PERIOD is a valid integer
+	ipCooldownPeriod, err, input := utils.GetIntFromStringEnvVar(envIPCooldownPeriod, defaultIPCooldownPeriod)
+	if err != nil {
+		log.Errorf("IP_COOLDOWN_PERIOD has to be a valid integer, user inputted: %v", input)
```
jdn5126, line 355: nit: Let's combine this case and the case below, i.e. `if err != nil || ipCooldownPeriod < 0`.

Since this is a container log, let's change this to "IP_COOLDOWN_PERIOD MUST be a valid positive integer. %s is invalid"

jdn5126 (changes requested):
Sorry for the delay, these are the last nits, then this is good to go! Let's also update the title and clean up the description a bit for future readers.

utils/utils.go
jdn5126: Let's also add a small unit test for this in `utils_test.go`

jayanthvn:
Can you add tests here - https://github.com/aws/amazon-vpc-cni-k8s/tree/master/test/integration/ipamd before merging the PR..Also callout increasing the cooldown timer will cause higher EC2 calls since IPs will be in cache for more time.

jdn5126:
> Can you add tests here - https://github.com/aws/amazon-vpc-cni-k8s/tree/master/test/integration/ipamd before merging the PR..Also callout increasing the cooldown timer will cause higher EC2 calls since IPs will be in cache for more time.

@jayanthvn I am not sure that an integration test case adds much value here, as the number itself is really arbitrary. We have only been using 30 seconds as a default because that's what upstream Kubernetes recommended early on. If we run the service connectivity test with 10 second cooldown and it passes consistently, that isn't going to change our default, so  this overhead doesn't seem worth it.

@jchen6585 the callout for the README is a good point. More IPs in cooldown means more IPs need to be allocated, which means more EC2 API calls

jdn5126 (approved):

jayanthvn:
But IMO, I feel we need to ensure DS functionality is fine i.e, no of IPs/total cooldown or at least the IP/ENIs allocated are expected with the setting..also if some change goes in breaking this functionality we should be able to capture it which UTs will not do..

jchen6585:
> But IMO, I feel we need to ensure DS functionality is fine i.e, no of IPs/total cooldown or at least the IP/ENIs allocated are expected with the setting..also if some change goes in breaking this functionality we should be able to capture it which UTs will not do..
> 
> @jchen6585 - Do add our recommendation and reason of 30s in EKS best practices..

@jayanthvn is this what you are referring to? https://aws.github.io/aws-eks-best-practices/networking/index/
if so, how do i get access to edit it?

jayanthvn:
> > But IMO, I feel we need to ensure DS functionality is fine i.e, no of IPs/total cooldown or at least the IP/ENIs allocated are expected with the setting..also if some change goes in breaking this functionality we should be able to capture it which UTs will not do..
> > @jchen6585 - Do add our recommendation and reason of 30s in EKS best practices..
> 
> @jayanthvn is this what you are referring to? https://aws.github.io/aws-eks-best-practices/networking/index/ if so, how do i get access to edit it?

You can open a request here - https://github.com/aws/aws-eks-best-practices

jdn5126 (approved):

Issue #2483, opened 2023-08-01
[Containerd] Disabling IPv6 


xuoja-acc:
**What happened**: 
After upgrade to kubernetes 1.24 and containerd pods resolve localhost on the ipv6 address ::1.
The upgrade documentation warn us about this behavior but the documentation doesn't provides the way to disable IPv6 at **the instance level**. 

We tried to disable the IPv6 through the bootstrap + sysctl but containers are still created with IPv6 (reboot is **not** an option but we tried with no effects)
```
cat <<'EOF' > /etc/sysctl.d/10-disable-ipv6.conf
net.ipv6.conf.all.disable_ipv6 = 1
net.ipv6.conf.default.disable_ipv6 = 1
net.ipv6.conf.lo.disable_ipv6 = 1
EOF
```

**What you expected**:
We shoul'd be able to disable the ipv6 by configuring CNI with the[ **cni tuning** plugin ](https://www.cni.dev/plugins/current/meta/tuning/) or any other way.

**NB** : Adding the tuning part to nerdctl-bridge.conflist is perfectly working

```
{
  "cniVersion": "0.4.0",
  "name": "aws-cni",
  "disableCheck": true,
  "plugins": [
    {
      "name": "aws-cni",
      "type": "aws-cni",
      "vethPrefix": "eni",
      "mtu": "9001",
      "podSGEnforcingMode": "strict",
      "pluginLogFile": "/var/log/aws-routed-eni/plugin.log",
      "pluginLogLevel": "DEBUG"
    },
    {
      "name": "egress-cni",
      "type": "egress-cni",
      "mtu": "9001",
      "enabled": "false",
      "randomizeSNAT": "prng",
      "nodeIP": "",
      "ipam": {
         "type": "host-local",
         "ranges": [[{"subnet": "fd00::ac:00/118"}]],
         "routes": [{"dst": "::/0"}],
         "dataDir": "/run/cni/v4pd/egress-v6-ipam"
      },
      "pluginLogFile": "/var/log/aws-routed-eni/egress-v6-plugin.log",
      "pluginLogLevel": "DEBUG"
    },
    {
      "type": "portmap",
      "capabilities": {"portMappings": true},
      "snat": true
    },
    {
      "type": "tuning",
      "sysctl": {
          "net.ipv6.conf.all.disable_ipv6": "1",
          "net.ipv6.conf.default.disable_ipv6": "1",
[...]

PR #2499, opened 2023-08-17
Add DISABLE_CONTAINER_V6 to disable IPv6 networking in container network namespaces

jdn5126:
**What type of PR is this?**
feature

https://github.com/aws/amazon-vpc-cni-k8s/issues/2483

This PR introduces a new environment variable: `DISABLE_CONTAINER_V6`. When this environment variable is set, the `tuning` plugin will be chained in the AWS conflist (`/etc/cni/net.d/10-aws.conflist`) and configured to disable IPv6 networking in newly created container network namespaces. This is done to provide a solution to https://github.com/aws/amazon-vpc-cni-k8s/issues/2483.

1. Create an IPv4 cluster with containerd as the container-runtime (EKS 1.24+)
2. Create a pod
3. See that IPv6 is enabled on the loopback interface within the container network namespace regardless of host sysctls.

Manually verified that conflist is rendered correctly and IPv6 is disabled in newly created container network namespaces. Verified that all integration tests pass following this change.

**Automation added to e2e**:
N/A

No

No, Yes

No

Yes

```release-note
Introduce DISABLE_CONTAINER_V6 to disable IPv6 networking in container namespaces.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -682,7 +682,7 @@ Note that enabling/disabling this feature only affects whether newly created pod
 
 ----
 
-#### `IP_COOLDOWN_PERIOD` (v1.14.1+)
+#### `IP_COOLDOWN_PERIOD` (v1.15.0+)
 
 Type: Integer as a String
 
@@ -692,6 +692,19 @@ Specifies the number of seconds an IP address is in cooldown after pod deletion.
 
 **Note:** 0 is a supported value, however it is highly discouraged.
 **Note:** Higher cooldown periods may lead to a higher number of EC2 API calls as IPs are in cooldown cache.
+
+----
+
+#### `DISABLE_POD_V6` (v1.15.0+)
+
+Type: Boolean as a String
+
+Default: `false`
+
+When `DISABLE_POD_V6` is set, the [tuning plugin](https://www.cni.dev/plugins/current/meta/tuning/) is chained and configured to disable IPv6 networking in each newly created pod network namespace. Set this variable when you have an IPv4 cluster and containerized applications that cannot tolerate IPv6 being enabled.
+Container runtimes such as `containerd` will enable IPv6 in newly created container network namespaces regardless of host settings.
+
+Note that if you set this while using Multus, you must ensure that any chained plugins do not depend on IPv6 networking. You must also ensure that chained plugins do not also modify these sysctls.
 
 ### VPC CNI Feature Matrix
 
--- a/cmd/aws-vpc-cni/main.go
+++ b/cmd/aws-vpc-cni/main.go
@@ -81,6 +81,7 @@ const (
 	defaultEnBandwidthPlugin     = false
 	defaultEnPrefixDelegation    = false
 	defaultIPCooldownPeriod      = 30
+	defaultDisablePodV6          = false
 
 	envHostCniBinPath        = "HOST_CNI_BIN_PATH"
 	envHostCniConfDirPath    = "HOST_CNI_CONFDIR_PATH"
@@ -101,6 +102,7 @@ const (
 	envEnIPv6Egress          = "ENABLE_V6_EGRESS"
 	envRandomizeSNAT         = "AWS_VPC_K8S_CNI_RANDOMIZESNAT"
 	envIPCooldownPeriod      = "IP_COOLDOWN_PERIOD"
+	envDisablePodV6          = "DISABLE_POD_V6"
 )
 
 // NetConfList describes an ordered list of networks.
@@ -116,11 +118,12 @@ type NetConfList struct {
 type NetConf struct {
 	CNIVersion string `json:"cniVersion,omitempty"`
 
-	Name         string          `json:"name,omitempty"`
-	Type         string          `json:"type,omitempty"`
-	Capabilities map[string]bool `json:"capabilities,omitempty"`
-	IPAM         *IPAMConfig     `json:"ipam,omitempty"`
-	DNS          *types.DNS      `json:"dns,omitempty"`
+	Name         string            `json:"name,omitempty"`
+	Type         string            `json:"type,omitempty"`
+	Capabilities map[string]bool   `json:"capabilities,omitempty"`
+	IPAM         *IPAMConfig       `json:"ipam,omitempty"`
+	DNS          *types.DNS        `json:"dns,omitempty"`
+	Sysctl       map[string]string `json:"sysctl,omitempty"`
 
 	RawPrevResult map[string]interface{} `json:"prevResult,omitempty"`
 	PrevResult    types.Result           `json:"-"`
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -693,6 +693,17 @@ Specifies the number of seconds an IP address is in cooldown after pod deletion.
 **Note:** 0 is a supported value, however it is highly discouraged.
 **Note:** Higher cooldown periods may lead to a higher number of EC2 API calls as IPs are in cooldown cache.
 
+----
+
+#### `DISABLE_CONTAINER_V6` (v1.14.x+)
```
achevuru, line 698: This will not be part of v1.14.0, as we already have the build tag for it. Let's instead make it v1.14.1+

README.md
```diff
@@ -693,6 +693,17 @@ Specifies the number of seconds an IP address is in cooldown after pod deletion.
...
+----
+
+#### `DISABLE_CONTAINER_V6` (v1.15.0+)
+
+Type: Boolean as a String
+
+Default: `false`
+
+When `DISABLE_CONTAINER_V6` is set, the [tuning plugin](https://www.cni.dev/plugins/current/meta/tuning/) is chained and configured to disable IPv6 networking in each newly created container network namespace. Set this variable when you have an IPv4 cluster and containerized applications that cannot tolerate IPv6 being enabled.
+Container runtimes such as `containerd` will enable IPv6 in newly created container network namespaces regardless of host settings.
```
achevuru, line 705: Let's add a blurb around Multus as well (i.e.,) user needs to make sure that any chained plugins are not dependent on this setting and/or not modifying these settings.

README.md
```diff
@@ -693,6 +693,17 @@ Specifies the number of seconds an IP address is in cooldown after pod deletion.
 **Note:** 0 is a supported value, however it is highly discouraged.
 **Note:** Higher cooldown periods may lead to a higher number of EC2 API calls as IPs are in cooldown cache.
 
+----
+
+#### `DISABLE_CONTAINER_V6` (v1.15.0+)
```
achevuru, line 698: nit: `DISABLE_CONTAINER_V6` --> `DISABLE_POD_V6`. We're disabling it at Pod level and so for all the containers in the Pod. Some might misinterpret it.

cmd/aws-vpc-cni/main.go
```diff
@@ -292,23 +295,41 @@ func generateJSON(jsonFile string, outFile string, getPrimaryIP func(ipv4 bool)
...
 			Type:         "bandwidth",
 			Capabilities: map[string]bool{"bandwidth": true},
 		}
 		data.Plugins = append(data.Plugins, &bwPlugin)
-		byteValue, err = json.MarshalIndent(data, "", "  ")
-		if err != nil {
-			return err
+	}
+
+	// Chain the tuning plugin (configured to disable IPv6 in container network namespace) when requested
```
achevuru, line 315: `container network namespace` --> `pod network namespace`

README.md
orsenthil, line 705: I am not sure how to understand this statement. What is intention for mentioning this information:

> Container runtimes such as `containerd` will enable IPv6 in newly created container network namespaces regardless of host settings.

- Even if we set `DISABLE_CONTAINER_V6=true, the container runtime can enable IPV6 in the newly created namespace?

cmd/aws-vpc-cni/main.go
```diff
@@ -292,23 +295,41 @@ func generateJSON(jsonFile string, outFile string, getPrimaryIP func(ipv4 bool)
...
+	disableContainerV6 := utils.GetBoolAsStringEnvVar(envDisableContainerV6, defaultDisableContainerV6)
+	if disableContainerV6 {
+		tuningPlugin := NetConf{
+			Type: "tuning",
+			Sysctl: map[string]string{
+				"net.ipv6.conf.all.disable_ipv6":     "1",
+				"net.ipv6.conf.default.disable_ipv6": "1",
+				"net.ipv6.conf.lo.disable_ipv6":      "1",
```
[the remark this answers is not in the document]
jdn5126: Not in any of my testing, but since the loopback interface is already created and assigned an IPv6 address by the time this chained plugin runs, I figured it was best to be safe and not make any assumptions on how the container kernel will handle cleanup.

cmd/aws-vpc-cni/main.go
```diff
@@ -292,23 +295,41 @@ func generateJSON(jsonFile string, outFile string, getPrimaryIP func(ipv4 bool)
 
 	byteValue = []byte(netconf)
 
+	// Unmarshall current conflist into data in order to chain any requested plugins
+	data := NetConfList{}
```
achevuru, line 299: While, I'm Ok with this flow. How about we do this only if either of the cases is true. We're anyways deriving both the env vars and we can use them to decide if we even need to do this..

```
enBandwidthPlugin := utils.GetBoolAsStringEnvVar(envEnBandwidthPlugin, defaultEnBandwidthPlugin)
disableContainerV6 := utils.GetBoolAsStringEnvVar(envDisableContainerV6, defaultDisableContainerV6)

if enBandwidthPlugin || disableContainerV6{
    //Unmarshall the conflist file
}
```

I understand that you might need another set of `if` checks inside to decide which one to chain but considering these env vars are probably used by <1% of Cx, I'm thinking it's alright.

README.md
jdn5126: The container runtime, in this case containerd, will create the network namespace and create the loopback interface with an IPv6 address assigned before any CNI plugins run. This statement is to clarify that containerd will not abide by the host settings (host networking sysctls) when it creates the network namespace.

achevuru: VPC CNI is down the chain. So, we're undoing what the CRI did

cmd/aws-vpc-cni/main.go
jdn5126: Yeah, I figured one extra JSON unmarshall and marshall was fine, but I can avoid this

cmd/aws-vpc-cni/main.go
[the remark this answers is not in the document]
achevuru: Yeah, we wouldn't need this. All the containers share the Kernel with the host machine/OS and we're only doing this per network ns. But there is no harm doing this, so we can leave it.

achevuru (approved):

README.md
```diff
@@ -693,6 +693,19 @@ Specifies the number of seconds an IP address is in cooldown after pod deletion.
 **Note:** 0 is a supported value, however it is highly discouraged.
 **Note:** Higher cooldown periods may lead to a higher number of EC2 API calls as IPs are in cooldown cache.
 
+----
+
+#### `DISABLE_POD_V6` (v1.15.0+)
+
+Type: Boolean as a String
+
+Default: `false`
```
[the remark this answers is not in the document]
jdn5126: I don't think so, as we don't want to change default behavior. There could be some customers that rely on IPv6 being enabled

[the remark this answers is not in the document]
jayanthvn: But lets add a line, we shouldn't set this for v6 clusters..

jayanthvn (commented):
lgtm..

PR #2517, opened 2023-08-29
Merge release 1.14 to master

jaydeokar:
**What type of PR is this?**
merge from release 1.14.0

N/A

N/A

N/A

**Automation added to e2e**:
N/A

N/A

N/A

N/A

N/A

### DIFF
--- a/charts/aws-vpc-cni/templates/clusterrole.yaml
+++ b/charts/aws-vpc-cni/templates/clusterrole.yaml
@@ -24,7 +24,7 @@ rules:
     resources:
       - pods
     verbs: ["list", "watch", "get"]
-{{- end }}        
+{{- end }}
   - apiGroups: [""]
     resources:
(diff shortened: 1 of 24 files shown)


charts/aws-vpc-cni/templates/clusterrole.yaml
```diff
@@ -38,3 +38,11 @@ rules:
     resources:
       - cninodes
     verbs: ["get", "list", "patch"]
+  - apiGroups: ["networking.k8s.aws"]
```
jdn5126, line 41: small nit: can we move these new entries to before `CNINode`? Then they will match up with the release order

jaydeokar: Updated

jdn5126 (approved):

PR #2522, opened 2023-08-31
Fixed the image incorrect image version

bhargaje:
**What type of PR is this?**
bug

The yaml are still pointing to older 1.13.4 instead of 1.14.0

NetworkPolicy support which was launched with VPC CNI 1.14.0 doesn't work because of this incorrect image version of aws-node and aws-init containers.

If you do kubectl apply -f with the existing CNI file you will see aws-node running with both vpc-cni and aws-node but when you create policy and it doesn't have any effect.

Were able to deploy and see network security policy taking effect.

**Automation added to e2e**:
NA

NA

NA

NA

NA

```release-note

```

### DIFF
--- a/config/master/aws-k8s-cni-us-gov-east-1.yaml
+++ b/config/master/aws-k8s-cni-us-gov-east-1.yaml
@@ -266,7 +266,22 @@ metadata:
     app.kubernetes.io/name: aws-node
     app.kubernetes.io/instance: aws-vpc-cni
     k8s-app: aws-node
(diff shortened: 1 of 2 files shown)


jdn5126:
Thanks for catching this @bhargaje! It looks like the job used for https://github.com/aws/amazon-vpc-cni-k8s/pull/2517 had some errors. We want to generate these files automatically, so we will take up fixing this

jdn5126:
Closing in favor of https://github.com/aws/amazon-vpc-cni-k8s/pull/2525

PR #2528, opened 2023-08-31
Instance limits api pkg

jchen6585:
**What type of PR is this?**
feature

N/A

Creates an API pkg that users can import and make calls rather than importing the entire instance limits map. All instance limit related resources all moved into the `pkg/vpc`. Modifies code that previously directly did operations on the instance limit map.

Passed all test cases against CNI and IPAMD integration suite.

**Automation added to e2e**:

No

No

No, No

No

No

```release-note

```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -31,6 +31,7 @@ import (
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/eventrecorder"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/retry"
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/vpc"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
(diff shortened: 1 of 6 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -1399,13 +1383,13 @@ func (cache *EC2InstanceMetadataCache) AllocIPAddress(eniID string) error {
...
 func (cache *EC2InstanceMetadataCache) FetchInstanceTypeLimits() error {
-	_, ok := InstanceNetworkingLimits[cache.instanceType]
+	_, ok := vpc.GetInstance(cache.instanceType)
 	if ok {
 		return nil
 	}
 
 	log.Debugf("Instance type limits are missing from vpc_ip_limits.go hence making an EC2 call to fetch the limits")
-	var eniLimits InstanceTypeLimits
+	var eniLimits vpc.InstanceTypeLimits
```
jdn5126, line 1392: I don't have a great suggestion here, but it feels to me like we shouldn't be creating a `vpc.InstanceTypeLimits` object and passing to `vpc.SetInstance`. It seems like we should be passing the relevant values and the `vpc` package should create and add the struct. Then it is more following a constructor package. Maybe `vpc` package needs a `New` function that takes all of the fields as arguments

jdn5126 (commented):
Changes overall LGTM, just one comment on a constructor

jdn5126 (approved):
Change LGTM

PR #2536, opened 2023-09-01
Update kops tests for 1.28 and fix generate-cni-yaml script

jaydeokar:
**What type of PR is this?**
Test scripts updates

This updates the kops weekly tests to run on 1.28 cluster and also pins the host instance type to ubuntu 20.04. 

Weekly integration tests to run on upstream kops cluster

N/A

N/A

**Automation added to e2e**:
N/A

N/A

N/A

N/A

N/A

```release-note

```

### DIFF
--- a/scripts/lib/cluster.sh
+++ b/scripts/lib/cluster.sh
@@ -76,6 +76,7 @@ function up-kops-cluster {
     mv kops-linux-amd64 $KOPS_BIN
     CLUSTER_NAME=kops-cni-test-cluster-${TEST_ID}.k8s.local
     export KOPS_STATE_STORE=s3://${KOPS_S3_BUCKET}
+    HOST_IMAGE_SSM_PARAMETER="ssm:/aws/service/canonical/ubuntu/server/20.04/stable/current/amd64/hvm/ebs-gp2/ami-id"
 
     SSH_KEYS=~/.ssh/devopsinuse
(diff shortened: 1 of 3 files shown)


scripts/lib/cluster.sh
```diff
@@ -85,6 +85,7 @@ function up-kops-cluster {
     else
         echo -e "\nSSH keys are already in place!"
     fi
+# Using Ubuntu 20.04 because of https://github.com/aws/amazon-vpc-cni-k8s/issues/2103
```
jdn5126, line 88: nit: alignment if this is needed

scripts/lib/cluster.sh
```diff
@@ -95,6 +96,7 @@ function up-kops-cluster {
     --node-size c5.xlarge \
     --ssh-public-key=~/.ssh/devopsinuse.pub \
     --kubernetes-version ${K8S_VERSION} \
+    --image "ssm:/aws/service/canonical/ubuntu/server/20.04/stable/current/amd64/hvm/ebs-gp2/ami-id" \
```
jdn5126, line 99: nit: can this be a global variable?

jaydeokar: We are not reusing it anywhere else, but I can update it in a variable

jdn5126 (approved):

PR #2547, opened 2023-09-06
Integration test cleanup: Security Groups for Pods

jdn5126:
**What type of PR is this?**
cleanup

N/A

This PR refactors the `pod-eni` integration test suite. The suite was previously failing due to scheduling issues, i.e. the test made assumptions about when pods should be scheduled on the self-managed node group that it created vs the existing managed node group. This PR removes the self-managed node group and runs the suite against the existing managed node group. This fixes the issues, decreases the complexity of the suite, and decreases the suite runtime.

N/A

Manually verified that the `pod-eni` suite passes following this change.

**Automation added to e2e**:
N/A

No

No, Yes

No

No

```release-note

```

### DIFF
--- a/test/framework/resources/aws/utils/nodegroup.go
+++ b/test/framework/resources/aws/utils/nodegroup.go
@@ -18,6 +18,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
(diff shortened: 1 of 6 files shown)


jchen6585 (approved):
LGTM

PR #2557, opened 2023-09-11
Only metrics

lnhanks:
**What type of PR is this?**

n/a

This PR adds to additional metrics to better visualize IP allocation. No available addresses error counter metric is currently a logged error but turning it into a metric will help visualize how often an IP allocation fails due to there being no addresses available. The second gauge metric ENI utilization expands on the existing ENIs allocated metric by partitioning data by ENI id and counting how many IP addresses are in use on each ENI. There are also two additional log statements created when an IP address is allocated or deallocated.

**Automation added to e2e**:

No

No

Yes

No

```release-note
Adds to additional metrics to better visualize IP allocation. No available addresses error counter metric is currently a logged error but turning it into a metric will help visualize how often an IP allocation fails due to there being no addresses available. The second gauge metric ENI utilization expands on the existing ENIs allocated metric by partitioning data by ENI id and counting how many IP addresses are in use on each ENI. 
```


pkg/ipamd/datastore/data_store.go
```diff
@@ -122,6 +122,19 @@ var (
 		},
 		[]string{"cidr"},
 	)
+	noAvailableAddrs = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "awscni_err_no_avail_addrs",
+			Help: "The number of IP/Prefix assignments that fail due to no available addresses at the ENI level",
```
jdn5126, line 128: nit on wording, can you change this to:
"The number of pod IP assignments that fail due to no available IP addresses"

pkg/ipamd/datastore/data_store.go
```diff
@@ -797,6 +815,7 @@ func (ds *DataStore) assignPodIPAddressUnsafe(addr *AddressInfo, ipamKey IPAMKey
 	addr.IPAMMetadata = ipamMetadata
 	addr.AssignedTime = assignedTime
 
+	ds.log.Debugf("IP allocation request")
```
jdn5126, line 818: nit: I don't think this log is needed, so let's remove it

pkg/ipamd/datastore/data_store.go
```diff
@@ -813,6 +832,7 @@ func (ds *DataStore) unassignPodIPAddressUnsafe(addr *AddressInfo) {
 	addr.IPAMKey = IPAMKey{} // unassign the addr
 	addr.IPAMMetadata = IPAMMetadata{}
 	ds.assigned--
+	ds.log.Debugf("IP deallocation request")
```
jdn5126, line 835: nit: same here, I don't think we need this log

pkg/ipamd/datastore/data_store.go
```diff
@@ -122,6 +122,19 @@ var (
 		},
 		[]string{"cidr"},
 	)
+	noAvailableAddrs = prometheus.NewCounter(
```
jdn5126, line 125: nit: can we rename this to `noAvailableIPAddrs`?

jdn5126 (changes requested):

pkg/ipamd/datastore/data_store.go
```diff
@@ -122,6 +122,19 @@ var (
...
+			Name: "awscni_err_no_avail_addrs",
+			Help: "The number of pod IP assignments that fail due to no available IP addresses",
+		},
+	)
+	eniUtilization = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "awscni_eni_util",
+			Help: "The number of allocated ips partitioned by eni",
+		},
+		[]string{"fn"},
```
jdn5126, line 136: nit: the string key should be renamed from "fn", which is short for "function", to "eni"

pkg/ipamd/datastore/data_store.go
```diff
@@ -122,6 +122,19 @@ var (
 		},
 		[]string{"cidr"},
 	)
+	noAvailableIPAddrs = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "awscni_err_no_avail_addrs",
+			Help: "The number of pod IP assignments that fail due to no available IP addresses",
+		},
+	)
+	eniUtilization = prometheus.NewGaugeVec(
```
jdn5126, line 131: sorry, I should have pointed this out earlier. Can we rename this to something like `eniIPsInUse`?

pkg/ipamd/datastore/data_store.go
```diff
@@ -711,9 +728,12 @@ func (ds *DataStore) AssignPodIPv6Address(ipamKey IPAMKey, ipamMetadata IPAMMeta
 				delete(V6Cidr.IPAddresses, addr.Address)
 				return "", -1, err
 			}
+			// Increment ENI Utilization gauge
```
jdn5126, line 731: nit: "Increment ENI IP usage on pod IPv6 allocation"

pkg/ipamd/datastore/data_store.go
```diff
@@ -711,9 +728,12 @@ func (ds *DataStore) AssignPodIPv6Address(ipamKey IPAMKey, ipamMetadata IPAMMeta
...
+			eniUtilization.WithLabelValues(eni.ID)
```
jdn5126, line 732: This is missing an `Inc()`

pkg/ipamd/datastore/data_store.go
```diff
@@ -776,11 +796,14 @@ func (ds *DataStore) AssignPodIPv4Address(ipamKey IPAMKey, ipamMetadata IPAMMeta
 				ipsPerCidr.With(prometheus.Labels{"cidr": availableCidr.Cidr.String()}).Dec()
 				return "", -1, err
 			}
+			// Increment ENI Utilization gauge
```
jdn5126, line 799: nit: "Increment ENI IP usage on pod IPv4 allocation"

pkg/ipamd/datastore/data_store.go
```diff
@@ -1166,6 +1193,8 @@ func (ds *DataStore) UnassignPodIPAddress(ipamKey IPAMKey) (e *ENI, ip string, d
 	ipsPerCidr.With(prometheus.Labels{"cidr": availableCidr.Cidr.String()}).Dec()
 	ds.log.Infof("UnassignPodIPAddress: sandbox %s's ipAddr %s, DeviceNumber %d",
 		ipamKey, addr.Address, eni.DeviceNumber)
+	// Decrement ENI Utilization gauge when pod IP is removed
+	eniUtilization.WithLabelValues(eni.ID)
```
jdn5126, line 1197: This is missing a `Dec()`

pkg/ipamd/datastore/data_store.go
```diff
@@ -1166,6 +1193,8 @@ func (ds *DataStore) UnassignPodIPAddress(ipamKey IPAMKey) (e *ENI, ip string, d
 	ipsPerCidr.With(prometheus.Labels{"cidr": availableCidr.Cidr.String()}).Dec()
 	ds.log.Infof("UnassignPodIPAddress: sandbox %s's ipAddr %s, DeviceNumber %d",
 		ipamKey, addr.Address, eni.DeviceNumber)
+	// Decrement ENI Utilization gauge when pod IP is removed
```
jdn5126, line 1196: nit: "Decrement ENI Utilization gauge when ENI is removed"

jdn5126 (changes requested):

pkg/ipamd/datastore/data_store.go
lnhanks: This is when a pod is removed, right? Would I want something like Decrement ENI IP usage on pod deallocation

jdn5126: Oh sorry, I was looking at the wrong code, you are right. Please ignore this comment

jdn5126 (dismissed):
Changes LGTM, thanks @lnhanks !

jdn5126 (approved):

pkg/ipamd/datastore/data_store.go
```diff
@@ -122,6 +122,19 @@ var (
...
 	)
+	noAvailableIPAddrs = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "awscni_err_no_avail_addrs",
+			Help: "The number of pod IP assignments that fail due to no available IP addresses",
+		},
+	)
+	eniIPsInUse = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "awscni_eni_util",
```
M00nF1sh, line 133: i think this metric should be named `awscni_assigned_ip_per_eni`, since it serves similar purpose of `awscni_assigned_ip_per_cidr` metric(expect the label value is eni instead of cidr).
Also, `util` is not a good abbreviation for "utilization"( i assume you mean utilization here :D)

pkg/ipamd/datastore/data_store.go
```diff
@@ -122,6 +122,19 @@ var (
 		},
 		[]string{"cidr"},
 	)
+	noAvailableIPAddrs = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "awscni_err_no_avail_addrs",
```
M00nF1sh, line 127: maybe change to `awscni_no_available_ip_addresses`, which
1. aligns with other metrics's use of "ip_addresses". 
2. and we don't need to have this "err" in metrics name, since it's a expected behavior instead of err.
3. remove nonstandard use of abbreviation(avail is not a standard abbreviation for available iirc)

PR #2559, opened 2023-09-12
Remove Calico test suite from weekly integration tests

jdn5126:
**What type of PR is this?**
cleanup

N/A

This PR removes the Calico integration test suite from weekly integration test runs. Now that AWS VPC CNI supports Kubernetes Network Policy natively (via https://github.com/aws/aws-network-policy-agent), we do not need to test against Calico's Network Policy implementation as frequently. We will still test against Calico's Network Policy implementation before every release and treat any failure as a release gate.

Removing `RUN_CALICO_TEST` decreases the complexity of `run-integration-tests` and decreases the overall runtime of our weekly integration test runner. 

N/A

N/A

**Automation added to e2e**:
None

No

No, No

No

No

```release-note

```

### DIFF
--- a/.github/workflows/weekly-cron-tests.yaml
+++ b/.github/workflows/weekly-cron-tests.yaml
@@ -71,15 +71,3 @@ jobs:
         run: |
           ./scripts/run-integration-tests.sh
         if: always()
-      - name: Run calico tests
-        env:
-          DISABLE_PROMPT: true
-          ROLE_CREATE: false
-          ROLE_ARN: ${{ secrets.EKS_CLUSTER_ROLE_ARN }}
-          RUN_CNI_INTEGRATION_TESTS: false
-          RUN_CALICO_TEST: true
-          RUN_LATEST_CALICO_VERSION: true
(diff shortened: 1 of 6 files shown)


orsenthil:
Hi @jdn5126 , 

Is it a maintenance burden to keep the CALICO test in weekly runs?

If the suite has been healthy, don't see a harm in keeping the calico suite running in addition to the NP controller suite. I will defer the judgement to the maintainer(s) tough.

jdn5126:
> 

The Calico tests are currently causing the weekly integration test runs to fail: https://github.com/aws/amazon-vpc-cni-k8s/actions/runs/6099567307/job/16552025168, and that was part of the motivation for this PR. The root cause seems to be from the helm upgrade, but that is still being investigated.

The other motivation, and where I think the maintenance burden comes from, is that Calico leaves the cluster in a bad state after the tests finish. We do not have a way to uninstall Calico, so it always has to run last and requires the cluster to be deleted right after running.

jayanthvn (approved):
lgtm

Issue #2518, opened 2023-08-30
No version number reported in CloudTrail history.


guessi:
**What happened**:

While troubleshooting VPC CNI related event, it is better to have version number included.

But the fact is that there's no version number reported in CloudTrail history, hard to tell which version was applied.

Version defined [HERE](https://github.com/aws/amazon-vpc-cni-k8s/blob/v1.14.0/pkg/awsutils/awssession/session.go#L38-L39) and [HERE](https://github.com/aws/amazon-vpc-cni-k8s/blob/v1.14.0/pkg/awsutils/awssession/session.go#L90-L98) seems like not defined, but I'm not sure how to fix it correctly, so I create the bug report here.

![image](https://github.com/aws/amazon-vpc-cni-k8s/assets/984415/fdab3abb-dc48-4dc7-b929-0dbc59577405)

**Attach logs**

n/a

**What you expected to happen**:

Version info should be included in CloudTrail events.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

It's a long lasting bug, version is not that important.

jdn5126:
This is a good call out @guessi , thank you for the report! Yeah, that `version` variable should be set based on the current VPC CNI version. We should do the same for our Kubernetes client userAgent: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/k8sapi/k8sutils.go#L174

I marked this as a good first issue, and will try to find someone to work on this.

jdn5126:
Closing as this merged in #2566 and will ship in v1.15.1

PR #2566, opened 2023-09-13
Add cni version to userAgent

jchen6585:
**What type of PR is this?**

enhancement

#2518 

Includes current CNI version in userAgent.

Manually tested through looking at cloudwatch logs.

**Automation added to e2e**:

No

No

No

No

```release-note

```

### DIFF
--- a/charts/aws-vpc-cni/values.yaml
+++ b/charts/aws-vpc-cni/values.yaml
@@ -78,6 +78,7 @@ env:
   DISABLE_NETWORK_RESOURCE_PROVISIONING: "false"
   ENABLE_IPv4: "true"
   ENABLE_IPv6: "false"
+  VPC_CNI_VERSION: "v1.15.0"
 
 # this flag enables you to use the match label that was present in the original daemonset deployed by EKS
 # You can then annotate and label the original aws-node resources and 'adopt' them into a helm release
--- a/pkg/awsutils/awssession/session.go
+++ b/pkg/awsutils/awssession/session.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/logger"
+	"github.com/aws/amazon-vpc-cni-k8s/utils"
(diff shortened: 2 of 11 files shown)


charts/aws-vpc-cni/values.yaml
```diff
@@ -78,6 +78,7 @@ env:
   DISABLE_NETWORK_RESOURCE_PROVISIONING: "false"
   ENABLE_IPv4: "true"
   ENABLE_IPv6: "false"
+  VPC_CNI_VERSION: "v1.15.0"
```
jchen6585, line 81: Based on what I saw from the implementation of other variables, it seems the other env variable values are eventually hardcoded within the code. This means when upgrading versions we'll have to manually update the default value when calling `GetEnv()` in `session.go`, which defeats the purpose of having this environment variable in the first place.

jdn5126: If the environment variable is not defined, we can stick with the behavior as it is today, i.e. not including a version in the user agent string

pkg/awsutils/awssession/session.go
```diff
@@ -89,6 +91,7 @@ func New() *session.Session {
 
 // injectUserAgent will inject app specific user-agent into awsSDK
 func injectUserAgent(handlers *request.Handlers) {
+	version = utils.GetEnv(envVpcCniVersion, "v1.15.0")
```
jdn5126, line 94: Let's remove `version` as a global variable and make the default be the empty string, which matches the current behavior.

jdn5126 (commented):
Let's also modify https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/generate-cni-yaml.sh#L67 to pass `$VPC_CNI_VERSION`, i.e. `--set env.VPC_CNI_VERSION=$VPC_CNI_VERSION`

jdn5126:
Manual integration test run: https://github.com/aws/amazon-vpc-cni-k8s/actions/runs/6191596936

jdn5126 (approved):

PR #2567, opened 2023-09-14
fix clusterrole permissions:  watch cninodes

marcincuber:
**What type of PR is this?**
Fix permissions issue

VPC CNI errors:
```
E0914 13:07:56.908969      12 reflector.go:148] pkg/mod/k8s.io/client-go@v0.27.3/tools/cache/reflector.go:231: Failed to watch *v1alpha1.CNINode: unknown (get cninodes.vpcresources.k8s.aws)
```

### DIFF
--- a/charts/aws-vpc-cni/templates/clusterrole.yaml
+++ b/charts/aws-vpc-cni/templates/clusterrole.yaml
@@ -45,4 +45,4 @@ rules:
       - vpcresources.k8s.aws
     resources:
       - cninodes
-    verbs: ["get", "list", "patch"]
+    verbs: ["get", "list", "patch", "watch"]
--- a/config/master/aws-k8s-cni-cn.yaml
+++ b/config/master/aws-k8s-cni-cn.yaml
@@ -327,7 +327,7 @@ rules:
       - vpcresources.k8s.aws
     resources:
       - cninodes
-    verbs: ["get", "list", "patch"]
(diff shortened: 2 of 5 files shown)


m477r1x (approved):

marcincuber:
> @marcincuber in what context are you seeing this error? The VPC CNI only does a GET and PATCH on `CNINode` (https://github.com/aws/amazon-vpc-cni-k8s/blob/v1.15.0/pkg/ipamd/ipamd.go#L2283), so WATCH should not be required

Not sure what you mean by context. But we upgraded to 1.15.0 today and saw the error immediately being thrown every 50 seconds or so. Adding `watch` perms fixed it. I believe it was happening on the main container `aws-node`.

jdn5126:
This is just an error log in the IPAMD logs, right? In other words, the container itself is not crashing and restarting, right? What I assume is happening is that the k8s client is trying to cache `CNINode`, so it issues a WATCH and logs an error since it lacks permission. We do not need or want to cache this resource locally, though, so this log is harmless. We should still fix it in a future version, though.

jdn5126:
Thanks for finding this @marcincuber! I want to do some more digging on a solution that does not involve caching `CNINode` resources, as they scale with the number of nodes in the cluster. Adding `watch` permission is a great workaround in the meantime.

jdn5126:
Closing in favor of #2570

PR #2570, opened 2023-09-15
Pass CNINode scheme to k8s client only

jdn5126:
**What type of PR is this?**
cleanup

https://github.com/aws/amazon-vpc-cni-k8s/pull/2567

The `aws-node` ClusterRole does not have WATCH permission for `CNINode` resources as IPAMD has no need to cache these objects, and it consumes excess memory to cache them. By passing the `CNINode` scheme to the cache, `controller-runtime` was issuing a WATCH, which resulted in the following log:
```
E0914 13:07:56.908969      12 reflector.go:148] pkg/mod/k8s.io/client-go@v0.27.3/tools/cache/reflector.go:231: Failed to watch *v1alpha1.CNINode: unknown (get cninodes.vpcresources.k8s.aws)
```

The log is harmless, but we want to avoid it, so this PR does not pass the `CNINode` scheme to the cache.

N/A

Manually verified that the `pod-eni` integration test suite passes and that the log is no longer present.

**Automation added to e2e**:
N/A

No

No, Yes

No

Yes

```release-note
Stop trying to cache CNINode resources.
```


jayanthvn (approved):
lgtm

orsenthil (commented):
LGTM too.

PR #2572, opened 2023-09-16
Bandwidth plugin with NP is currently unsupported

jayanthvn:
**What type of PR is this?**
documentation

N/A

Bandwidth plugin uses TC filter and attaches TBF(token bucket filter) queuing discipline (qdisc) on both ingress and egress. But we use NP enforcement with TC filter "clsact" which will collide with B/W plugin. Ref : https://github.com/aws/aws-network-policy-agent/issues/68

N/A

N/A

**Automation added to e2e**:

N/A

N/A

N/A

N/A

N/A

```release-note

```


README.md
```diff
@@ -572,6 +572,8 @@ Default: `false`
 
 Setting `ENABLE_BANDWIDTH_PLUGIN` to `true` will update `10-aws.conflist` to include upstream [bandwidth plugin](https://www.cni.dev/plugins/current/meta/bandwidth/) as a chained plugin.
 
+NOTE: Network Policy is supported from CNI version v1.14.0 onwards. We use TC(traffic classifier) system to enforce network policy for the pods ingress and egress traffic. If the bandwidth plugin and network policy is enabled and the network policy enforced pods have "ingress-bandwidth" and "egress-bandwidth" configured then network policy enforcement will fail since bandwidth plugin uses TC which will collide with aws-eks-nodeagent TC configuration. We will provide updates [here](https://github.com/aws/aws-network-policy-agent/issues/68)
```
jdn5126, line 575: nit: for the note, I think it is better to just say that the `bandwidth` plugin is not currently compatible with Network Policy. The explanation feels like it just leads to more confusion.

jayanthvn: This is for readme and IMO we need to give a proper reason. Once we fix it, this will be removed.

achevuru (approved):

PR #2573, opened 2023-09-18
Update example table 'Pod per Prefixes' value

rlaisqls:
**What type of PR is this?**

documentation

In the example table, the value of 'Pod per ENIs' for t3.small instance with 58 pods was changed from '48, 2' to '48, 10'.

Since the total number of pods is 58, the sum of the values in that column must also be 58. The value of 'Pod per Prefixes' can also be inferred that the value of the corresponding column will be 10. I think this notation error can cause confusion for people who want to grasp the concept through Docs.

**Automation added to e2e**:

```release-note

```


jdn5126:
Nice catch, thanks @rlaisqls!

jdn5126 (approved):

PR #2577, opened 2023-09-21
Add ENABLE_V4_EGRESS env var to control IPv4 egress in IPv6 clusters

jdn5126:
**What type of PR is this?**
feature

https://github.com/aws/aws-network-policy-agent/issues/55

This PR adds a new environment variable, `ENABLE_V4_EGRESS`, which controls whether the `egress-cni` chained plugin is enabled for IPv6 clusters. The value is enabled by default, preserving the existing IPv6 default. When disabled, an IPv4 interface will not be created in IPv6 pods, blocking those pods from reaching IPv4 endpoints.

The primary motivation for this PR is to address https://github.com/aws/aws-network-policy-agent/issues/55, where the concern is network policies in IPv6 clusters not applying to IPv4 traffic.

N/A

Manually verified that chained plugin is enabled or disabled based on env var.

**Automation added to e2e**:
N/A

No

No, Yes

No

Yes

```release-note
Introduce ENABLE_V4_EGRESS, which controls whether egress-cni plugin is enabled in IPv6 clusters.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -670,7 +670,17 @@ Specifies whether PODs in an IPv4 cluster support IPv6 egress. If env is set to 
 
 This environment variable must be set for both the `aws-vpc-cni-init` and `aws-node` containers in order for this feature to work properly. This feature also requires that the node has an IPv6 address assigned to its primary ENI, as this address is used for SNAT to IPv6 endpoints outside of the cluster. If the configuration prerequisites are not met, the `egress-cni` plugin is not enabled and an error log is printed in the `aws-node` container.
 
-Note that enabling/disabling this feature only affects whether newly created pods have an IPv6 interface created. Therefore, it is recommended that you reboot existing nodes after enabling/disabling this feature. Also note that if you are using this feature in conjunction with `ENABLE_POD_ENI` (Security Groups for Pods), the security group rules will NOT be applied to egressing IPv6 traffic.
+Note that enabling/disabling this feature only affects whether newly created pods have an IPv6 interface created. Therefore, it is recommended that you reboot existing nodes after enabling/disabling this feature.
+
+#### `ENABLE_V4_EGRESS` (v1.15.1+)
+
+Type: Boolean as a String
+
+Default: `true`
+
+Specifies whether PODs in an IPv6 cluster support IPv4 egress. If env is set to `true`, range `169.254.172.0/22` is reserved for IPv4 egress. When enabled, traffic egressing an IPv6 pod destined to an IPv4 endpoint will be SNAT'ed via the node IPv4 address.
+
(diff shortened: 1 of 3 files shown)


jchen6585 (approved):
lgtm

README.md
```diff
@@ -672,6 +672,16 @@ This environment variable must be set for both the `aws-vpc-cni-init` and `aws-n
...
 
+#### `ENABLE_V4_EGRESS` (v1.15.1+)
+
+Type: Boolean as a String
+
+Default: `true`
+
+Specifies whether PODs in an IPv6 cluster support IPv4 egress. If env is set to `true`, range `169.254.172.0/22` is reserved for IPv4 egress. When enabled, traffic egressing an IPv6 pod destined to an IPv4 endpoint will be SNAT'ed via the node IPv4 address.
+
+Note that enabling/disabling this feature only affects whether newly created pods have an IPv4 interface created. Therefore, it is recommended that you reboot existing nodes after enabling/disabling this feature. Also note that if you are using this feature in conjunction with `ENABLE_POD_ENI` (Security Groups for Pods), the security group rules will NOT be applied to egressing IPv4 traffic.
```
achevuru, line 683: I don't think we need to add the part about SGPP as we don't support SGPP in IPv6 mode yet.

jdn5126: I figured it was fine to keep that there for when future support arrives

orsenthil: Should we have the env variable name as ``ENABLE_V4_EGRESS_IN_V6` ? 

because the context matters.

If we specify the context, then statement of SGPP can be omitted. Don't feel strongly about mentioning it here.

jdn5126: Yeah, I chose this name because we have `ENABLE_V6_EGRESS` for IPv4 clusters, so I wanted to keep the naming the same

orsenthil: > ENABLE_V6_EGRESS for IPv4 clusters

For some reason, the IPV4 clusters are considered 'default' and that made immediate sense to me.   When I read `ENABLE_V4_EGRESS` I was thinking what I am missing, and then realized the context.

Keeping it consistent sounds good.

achevuru: It will not be an issue even when we support it because the SG is applied against a branch ENI in SGPP mode (and not pod's interfaces) and no matter which interface a packet takes to exit out of the pod it will have to go out of the branch ENI and so will be evaluated against the SG the customer provides and is configured on branch ENIs.

jdn5126: Ah that makes sense, I am not sure why I originally had that line for `ENABLE_V6_EGRESS` then. Maybe because I was trying to point out that IPv4 policies would not be applied to IPv6 traffic and so you need IPv6-specific policies. Anyway, updated in the latest revision

achevuru (approved):

PR #2587, opened 2023-09-26
Increase time for service readiness in integration test

jdn5126:
**What type of PR is this?**
testing

N/A

The `cni` service connectivity integration test was occasionally failing as 20 seconds is not enough time for a Classic Load Balancer to become ready. This PR increases the time to one minute as a temporary solution to resolve the flakiness. Long-term, we need to make an EC2 API call to ensure that the load balancer is ready.

N/A

None

**Automation added to e2e**:

None

No

No, No

No

No

```release-note

```


orsenthil (approved):
LGTM.

PR #2588, opened 2023-09-26
ipamd: add deleted ip to cooldown cache in tryUnassignIPFromENI

zhouhaibing089:
There are two places where `awsClient.DeallocIPAddresses` is called:

1. tryUnassignIPFromENI
2. DeallocCidrs

However only `DeallocCidrs` tracks the deleted IP in cooldown cache, and it appears to be missing in `tryUnassignIPFromENI`.

We have found several cases in our production environment where IP(s) were added back in `nodeIPPoolReconcile` given that they were just freed up several seconds ago.

This makes the cooldown behavior more consistent.

**What type of PR is this?**

bug

**Automation added to e2e**:

```release-note

```

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -2044,6 +2044,7 @@ func (c *IPAMContext) tryUnassignIPFromENI(eniID string) {
 			continue
 		} else {
 			deletedIPs = append(deletedIPs, toDelete.IP.String())
+			c.reconcileCooldownCache.Add(toDelete.IP.String())
 		}
 	}
 


jdn5126:
@zhouhaibing089 sorry, I am not following the change here. When an ENI is attached, all of the secondary IPs are added to the datastore. When the pod is deleted, that IP is available in the datastore again. Could you add more detail here?

zhouhaibing089:
@jdn5126: Sorry for my late response. I am no longer working on this issue anymore. My previous team member may continue offer inputs.

As I still recall it, aws-cni plugin also frees IP(s) when the number of allocated IP(s) while not being used exceeds certain threshold. When that happens, the deletion API call is eventual consistency, and the next list call will get the IP again (even though it is being deleted) and assign this IP to some new pod - which is apparently going to be useless as the IP is eventually going to be de-registered.

PR #2590, opened 2023-09-28
Remove self-managed node group from custom-networking suite

jdn5126:
**What type of PR is this?**
testing

N/A

This PR removes the creation of a self-managed node group for the `custom-networking` integration suite. This removal makes the test more flexible and decreases its complexity. It also removes the flakiness that was present in the test, as not all jobs had the proper node selectors.

It is unfortunate that this test suite has so much waiting time. The time spent waiting for instance termination cannot be avoided, nor can the ENI cooldown time. It would be nice to eliminate some of the other waits, though.

N/A

Manually verified that integration suite passes following change

**Automation added to e2e**:

N/A

No

No, Yes

No

No

```release-note

```


jchen6585 (approved):
LGTM

Issue #2584, opened 2023-09-26
Can't use aws-node without VPC Resource Controller?


mattburgess:
**What happened**:

I tried upgrading our v1.12.6 aws-node daemonset to v1.15.0 but the pods fail to start up. They log the following:

```text
{"level":"info","ts":"2023-09-26T14:17:02.870Z","caller":"ipamd/ipamd.go:550","msg":"Get Node Info for: ip-10-50-97-227.eu-west-1.compute.internal"}
{"level":"error","ts":"2023-09-26T14:17:02.974Z","caller":"ipamd/ipamd.go:423","msg":"Failed to add feature custom networking into CNINode%!(EXTRA *fmt.wrapError=failed to get API group resources: unable to retrieve the complete list of server APIs: vpcresources.k8s.aws/v1alpha1: the server could not find the requested resource)"}
{"level":"error","ts":"2023-09-26T14:17:02.974Z","caller":"aws-k8s-agent/main.go:32","msg":"Initialization failure: failed to get API group resources: unable to retrieve the complete list of server APIs: vpcresources.k8s.aws/v1alpha1: the server could not find the requested resource"}
```

That seems to be due to #2503. I took a look at the various env vars but couldn't see anything there, or in the code, that makes this feature optional. Having a hard dependency on a controller that looks specifically designed for EKS means it looks like we can't upgrade to this release on a non-EKS-but-still-hosted-in-AWS cluster. Have I understood things correctly or have I missed something in the docs?

**Attach logs**

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): 1.24.17
- CNI Version: 1.15.0
- OS (e.g: `cat /etc/os-release`): Ubuntu 20.04
- Kernel (e.g. `uname -a`): 5.19.0-1029-aws

jdn5126:
@mattburgess how did you upgrade from v1.12.6 to v1.15.0? Did you install the full manifest or helm chart?

The VPC CNI has always had a dependency on the VPC Resource Controller for certain features, and a new CRD was introduced in v1.15.0. It sounds like that CRD is not installed in your cluster

mattburgess:
[...]

PR #2591, opened 2023-09-28
Do not patch CNINode for custom networking unless SGPP is enabled

jdn5126:
**What type of PR is this?**
bug

https://github.com/aws/amazon-vpc-cni-k8s/issues/2584

When Custom Networking is enabled, IPAMD patches the `CNINode` resource for this node by default, when it only needs to patch it when Security Groups for Pods is also enabled. Patching by default leads to an issue if the VPC Resource Controller is not running, as that `CNINode` resource will not exist.

This PR avoids patching the `CNINode` resource when unless a VPC Resource Controller feature is enabled. This is done so that customers can run the VPC CNI and configure certain features without needing the controller.

N/A

Manually verified the fix.

**Automation added to e2e**:

N/A

N/A

No, Yes

No

Yes

```release-note
Patch CNINode only when Security Groups for Pods is enabled
```

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -560,14 +560,16 @@ func (c *IPAMContext) nodeInit() error {
 
 		eniConfigName, err := eniconfig.GetNodeSpecificENIConfigName(node)
 		if err == nil && eniConfigName != "default" {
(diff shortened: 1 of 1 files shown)


haouc (approved):
Glynn. Thanks.

PR #2595, opened 2023-09-30
Quote images in chart

### LINKED ISSUE #2594: Error when setting image.account in chart
**What happened**:

 
The aws-vpc-cni chart allows specifying `image.account` and `init.image.account`. The following works fine in chart versions before https://github.com/aws/amazon-vpc-cni-k8s/commit/41f8d8c9ac4ad3b56ddbc5165c5f2bda8e3d7af9, but now generates an error.

To reproduce, run the following on `master`:
`helm template --debug . --set 'image.account=602401143452'`

The resulting YAML has the following:

> image: %!s(int64=602401143452).dkr.ecr.us-west-2.amazonaws.com/amazon-k8s-cni:v1.15.0

The same issue exists for `init.image.account`.

You can use `--set-string` as a workaround, but that wasn't necessary in prior versions of this chart.

**Attach logs**

> Error: YAML parse error on aws-vpc-cni/templates/daemonset.yaml: error converting YAML to JSON: yaml: line 55: found character that cannot start any token

**What you expected to happen**: Valid YAML

**How to reproduce it (as minimally and precisely as possible)**: `helm template . --set 'image.account=602401143452'`

I've opened https://github.com/aws/amazon-vpc-cni-k8s/pull/2595 as a fix for this.

**Environment**:
- Kubernetes version (use `kubectl version`): N/A
- CNI Version: N/A
- OS (e.g: `cat /etc/os-release`): N/A
- Kernel (e.g. `uname -a`): N/A

gabegorelick:
This is necessary to ensure that YAML doesn't confuse the leading account as a number. That won't happen with the default account from values.yaml, since that's set as a string, but it can happen when users pass in their own values.

Fixes #2594

**What type of PR is this?**
bug

#2594

Fixes a regression in the chart that broke existing users.

**If an issue # is not available please add repro steps and logs from IPAMD/CNI showing the issue**: n/a

✅ `helm template . --set 'init.image.account=602401143452,image.account=602401143452,nodeAgent.account=602401143452'` generates correct YAML. 

✅ `helm template .` generates correct YAML

**Automation added to e2e**:

no

**Will this break upgrades or downgrades. Has updating a running cluster been tested?**: n/a

**Does this change require updates to the CNI daemonset config files to work?**: no

```release-note

```

### DIFF
--- a/charts/aws-vpc-cni/templates/daemonset.yaml
+++ b/charts/aws-vpc-cni/templates/daemonset.yaml
@@ -40,7 +40,7 @@ spec:
       hostNetwork: true
       initContainers:
       - name: aws-vpc-cni-init
-        image: {{ include "aws-vpc-cni.initImage" . }}
+        image: "{{ include "aws-vpc-cni.initImage" . }}"
         env:
 {{- range $key, $value := .Values.init.env }}
           - name: {{ $key }}
@@ -66,7 +66,7 @@ spec:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
       containers:
         - name: aws-node
-          image: {{ include "aws-vpc-cni.image" . }}
+          image: "{{ include "aws-vpc-cni.image" . }}"
           ports:
             - containerPort: 61678
               name: metrics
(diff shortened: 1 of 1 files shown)


jdn5126:
@gabegorelick I am going to close this pull request as I think `--set-string` is the right way to truly handle these cases

PR #2617, opened 2023-10-15
Stricter dependency/security review

dims:
- Add new GH action for dependency review
- Add new GH action for govulncheck

action.yaml copied from here (untouched):
https://github.com/golang/govulncheck-action/blob/master/action.yml

under BSD 3-Clause "New" or "Revised" License:
https://github.com/golang/govulncheck-action/blob/master/LICENSE

Need to copy the file because github actions in github orgs with enterprise license cannot refer to things as shown by this error:
image

### DIFF
--- a/.github/actions/govulncheck/action.yaml
+++ b/.github/actions/govulncheck/action.yaml
@@ -1,0 +1,46 @@
+name: 'golang-govulncheck-action'
+description: 'Run govulncheck'
+inputs:
+  go-version-input:  # version of Go to use for govulncheck
+    description: 'Version of Go to use for govulncheck'
+    required: false
+  check-latest:
+    description: 'Set this option to true if you want the action to always check for the latest available Go version that satisfies the version spec'
+    required: false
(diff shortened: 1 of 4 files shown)


jdn5126:
Thanks for adding this @dims! I am wondering if `govulncheck` should be run as a periodic job rather than a PR gate. The reason being that when new vulns pop up, unrelated PRs will be blocked until `govulncheck` is addressed. What do you think?

dims:
@jdn5126 of course! feel free to use the pattern that works for you. you can land this and modify or just borrow code from this PR.

Here's the tough part ... someone to watch the periodic job and take remediation steps. having this block landing PR(s) forces everyone to deal with it immediately as it is annoying and gets in the way :)

jdn5126 (approved):

orsenthil:
LGTM. Thank you!. 

Periodic job will be better, provided a way to alert and act on the periodic job failure.

PR #2623, opened 2023-10-19
feat(chart): Made node agent optional

### LINKED ISSUE #2524: Node agent resources not configurable in the Helm chart
**What would you like to be added**:
The Helm chart uses the same resource configuration for all of the containers; when there was only a single init and regular container this didn't matter as the resource usage would be exactly as was configured. Now that there is an additional node agent container which uses the same resource configuration the DaemonSet will always use twice as many resources as is configured.

**Why is this needed**:
The resource configuration should be container specific and be clear to end-users.

stevehipwell:
**What type of PR is this?**

feature

- Fixes #2524

This PR updates the Helm chart to makes the node agent container optional (defaults to present) and allows for custom resource specification for the init container and the node agent. These changes should be a no-op unless someone makes use of the new chart inputs.

n/a

I've templated the Helm chart and these changes are a no-op. The new chart values work as expected.

**Automation added to e2e**:
n/a

No

No

 No

Yes

```release-note

```

### DIFF
--- a/charts/aws-vpc-cni/templates/daemonset.yaml
+++ b/charts/aws-vpc-cni/templates/daemonset.yaml
@@ -48,8 +48,10 @@ spec:
 {{- end }}
         securityContext:
           {{- toYaml .Values.init.securityContext | nindent 12 }}
+        {{- with default .Values.resources .Values.init.resources }}
         resources:
-          {{- toYaml .Values.resources | nindent 12 }}
+          {{- toYaml . | nindent 12 }}
+        {{- end }}
         volumeMounts:
           - mountPath: /host/opt/cni/bin
             name: cni-bin-dir
@@ -91,8 +93,10 @@ spec:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: metadata.name
+          {{- with .Values.resources }}
           resources:
-            {{- toYaml .Values.resources | nindent 12 }}
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
           volumeMounts:
@@ -116,6 +120,7 @@ spec:
           {{- with .Values.extraVolumeMounts  }}
           {{- toYaml .| nindent 10 }}
           {{- end }}
+        {{- if .Values.nodeAgent.enabled }}
         - name: aws-eks-nodeagent
           image: {{ include "aws-vpc-cni.nodeAgentImage" . }}
           env:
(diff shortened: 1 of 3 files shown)


stevehipwell:
CC @jdn5126

charts/aws-vpc-cni/values.yaml
```diff
@@ -22,8 +22,10 @@ init:
     ENABLE_IPv6: "false"
   securityContext:
     privileged: true
+  resources:
```
jdn5126, line 25: I'm not sure that specifying specific init container specific resources is worth it, or a good idea, as the idea is that the node needs to reserve enough resources for the actual containers. I am not sure how the node handles when the init container requires more or less resources than the actual containers that will run, so let's leave this as is.

charts/aws-vpc-cni/values.yaml
```diff
@@ -44,6 +46,7 @@ nodeAgent:
   enableIpv6: "false"
   metricsBindAddr: "8162"
   healthProbeBindAddr: "8163"
+  resources:
```
jdn5126, line 49: nit: let's use `resources: nil` and add a comment that when not specified, node agent container resources are inherited from `Values.resources`

jdn5126 (changes requested):
Thanks for working on this @stevehipwell! Changes overall LGTM, just a few nits. I'll also give others a chance to review

charts/aws-vpc-cni/values.yaml
stevehipwell: Kubernetes scheduling deals with this in a [well defined way](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#resource-sharing-within-containers). The current implementation is misleading to the user and although I don't think it actually matters for most people I think it should be customisable as I've configured here (if I wasn't aiming for a no-op I'd have made it fully explicit instead of using `default`).

charts/aws-vpc-cni/values.yaml
stevehipwell: Sorry, I'm not sure what you mean by `resources: nil`? The idiomatic pattern for an empty value in a Helm chart is either `key:` or `key: {}`, not defining the key is bad practice as it breaks tooling. The documentation for how to use this is in the README.

charts/aws-vpc-cni/templates/daemonset.yaml
```diff
@@ -144,6 +151,7 @@ spec:
             name: log-dir
           - mountPath: /var/run/aws-node
             name: run-dir
+      {{- end }}
       volumes:
       - name: bpf-pin-path
```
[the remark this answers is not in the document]
stevehipwell: Fixed

charts/aws-vpc-cni/values.yaml
jdn5126: Got it. Having the extra config value is fine, let's just use `resources: {}` as below

charts/aws-vpc-cni/values.yaml
jdn5126: Let's use `resources: {}` to make it cleaner and more explicit

jdn5126 (changes requested):

stevehipwell:
@jdn5126 I've made the changes.

jdn5126 (approved):

PR #2632, opened 2023-10-25
Upgrade CNI spec from 0.4.0 to 1.0.0

jdn5126:
**What type of PR is this?**
spec upgrade

N/A

This PR updates the container networking spec used by the VPC CNI from 0.4.0 to 1.0.0. This is needed to stay up to date with the latest `containternetworking` package.

Note that `containernetworking/cni v1.1.2` provides libraries such that spec 0.4.0 and 1.0.0 are fully forward and backward compatible. There are co changes in this PR related to the new spec other than the CNI version.

Also note that this PR will require a minor version update, i.e. `v1.16.x`. This PR also introduces a requirement that the EKS version be at least 1.24.

N/A

Manually verified that pods created with image using 0.4.0 spec can be properly loaded and deleted by image using 1.0.0 spec (and vice versa).

Manually verified that all integration tests pass.

**Automation added to e2e**:

N/A

Yes, this PR makes VPC CNI compatible only with EKS 1.24+.

No, Yes

No

Yes

```release-note
Upgrade CNI spec from 0.4.0 to 1.0.0
```


jaydeokar (approved):
lgtm

PR #2639, opened 2023-10-31
IPv6 Security Groups for Pods Support

jdn5126:
**What type of PR is this?**

feature

N/A

This PR adds VPC CNI support for IPv6 Security Groups for Pods. It is dependent on VPC Resource Controller support added in https://github.com/aws/amazon-vpc-resource-controller-k8s/pull/309.

This PR is unfortunately quite large, so to help reviewers, I will call out major areas to focus on:
*Initialization*
- `cmd/aws-vpc-cni-init/main.go` - default sysctls for IPv6 have been modified
- `pkg/ipamd/ipamd.go` - configuration of IPv6 Security Groups for Pods needs to be allowed

*ENI Management*
- `pkg/awsutils/awsutils.go` - when attaching an IPv6 trunk ENI, the metadata associated with the ENI needs to incorporate IPv6.
- `pkg/ipamd/ipamd.go` - when IPv6 Security Groups for Pods is enabled, IPAMD must wait for a trunk ENI to be attached
- `pkg/networkutils/network.go` - support for attaching an IPv6 trunk ENI, which includes getting the IPv6 gateway address and configuring relevant routes and rules

*Pod Interface Management*
- `cmd/routed-eni-cni-plugin/driver/driver.go` - sysctls for branch interfaces and IPv6 veth interfaces are set
- `pkg/ipamd/rpc_handler.go` - support for deriving IPv6 branch ENI information from annotation

*Caching*
- We do not want to constantly derive the IPv6 gateway address on an instance. When the IPv6 trunk ENI is configured, we cache the gateway address in IPAMD for use by CNI ADD.

*Testing*
- `test/integration/pod-eni` has been made compatible for IPv6

*Other*
I realized that `make unit-test` was not invoking unit tests defined in `cmd/...`. Some of these unit tests were failing, so I made sure all were passing in this PR.

N/A

Manually verified that `test/integration/pod-eni` passes. Also manually verified that all existing integration tests pass.

Further validation is in progress for upgrade/downgrade, scale, performance, etc.

No

No, Yes

No

Yes

```release-note
IPv6 Security Groups for Pods support
```


test/integration/pod-eni/security_group_per_pod_test.go
```diff
@@ -100,15 +100,17 @@ var _ = Describe("Security Group for Pods Test", func() {
 				TrafficServerDeploymentBuilder: serverDeploymentBuilder,
 				ServerPort:                     openPort,
 				ServerProtocol:                 "tcp",
-				ClientCount:                    20,
+				ClientCount:                    6,
```
jdn5126, line 103: this needs to be reverted to 20

cmd/aws-vpc-cni-init/main.go
```diff
@@ -125,10 +129,33 @@ func configureIPv6Settings(procSys procsyswrapper.ProcSys, primaryIF string) err
...
+
+		entry = "net/ipv6/conf/default/accept_redirects"
+		err = procSys.Set(entry, "1")
+		if err != nil {
+			return errors.Wrap(err, "Failed to enable IPv6 accept redirects")
+		}
+		val, _ = procSys.Get(entry)
+		log.Infof("Updated %s to %s", entry, val)
+
+		// For the primary ENI in IPv6, sysctls are set to:
```
achevuru, line 151: Do we even need this block now? We're now configuring the same set of values for all interfaces on the node, so I don't see a need to repeat these again for primary IF.

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -328,9 +331,13 @@ func (n *linuxNetwork) TeardownBranchENIPodNetwork(containerAddr *net.IPNet, vla
 		return errors.Wrapf(err, "TeardownBranchENIPodNetwork: failed to teardown vlan")
 	}
 
+	family := unix.AF_INET
```
achevuru, line 334: nit: `family` -> `ipFamily`

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -360,20 +367,29 @@ func (n *linuxNetwork) setupVeth(hostVethName string, contVethName string, netns
...
+	// 3. forwarding=0
 	if err := n.procSys.Set(fmt.Sprintf("net/ipv6/conf/%s/accept_ra", hostVethName), "0"); err != nil {
 		if !os.IsNotExist(err) {
 			return nil, errors.Wrapf(err, "failed to disable IPv6 router advertisements")
 		}
 		log.Debugf("Ignoring '%v' writing to accept_ra: Assuming kernel lacks IPv6 support", err)
 	}
-
-	if err := n.procSys.Set(fmt.Sprintf("net/ipv6/conf/%s/accept_redirects", hostVethName), "0"); err != nil {
+	if err := n.procSys.Set(fmt.Sprintf("net/ipv6/conf/%s/accept_redirects", hostVethName), "1"); err != nil {
```
achevuru, line 380: Why do we need to enable this?

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -360,20 +367,29 @@ func (n *linuxNetwork) setupVeth(hostVethName string, contVethName string, netns
...
 		if !os.IsNotExist(err) {
 			return nil, errors.Wrapf(err, "failed to disable IPv6 ICMP redirects")
 		}
 		log.Debugf("Ignoring '%v' writing to accept_redirects: Assuming kernel lacks IPv6 support", err)
 	}
-	log.Debugf("Successfully disabled IPv6 RA and ICMP redirects on hostVeth %s", hostVethName)
+	if err := n.procSys.Set(fmt.Sprintf("net/ipv6/conf/%s/forwarding", hostVethName), "0"); err != nil {
```
achevuru, line 386: Same Q as above - Why is this even needed for hostVeth?

jdn5126: Just to mirror `ec2-net-utils`, really. Also once forwarding is disabled, `accept_redirects` is implicitly 1: https://sysctl-explorer.net/net/ipv6/accept_redirects/

cmd/aws-vpc-cni-init/main.go
```diff
@@ -125,10 +129,33 @@ func configureIPv6Settings(procSys procsyswrapper.ProcSys, primaryIF string) err
...
+		entry = "net/ipv6/conf/default/accept_ra"
+		err = procSys.Set(entry, "2")
+		if err != nil {
+			return errors.Wrap(err, "Failed to set IPv6 accept Router Advertisements to 2")
+		}
+		val, _ = procSys.Get(entry)
+		log.Infof("Updated %s to %s", entry, val)
+
+		entry = "net/ipv6/conf/default/accept_redirects"
+		err = procSys.Set(entry, "1")
```
jayanthvn, line 144: There are multiple instances of procSys.Set...Maybe we can move this to a single function/package and input is the "entry" above..

pkg/ipamd/ipamd.go
```diff
@@ -618,10 +626,6 @@ func (c *IPAMContext) StartNodeIPPoolManager() {
 }
 
 func (c *IPAMContext) updateIPPoolIfRequired(ctx context.Context) {
-	if c.enablePodENI && c.dataStore.GetTrunkENI() == "" {
```
achevuru, line 621: As discussed, let's restore the current flow as we're trading IMDS calls with EC2 calls and might contribute to API throttling issues we run in to in VPC RC.

jdn5126: restored in latest revision

pkg/ipamd/ipamd.go
```diff
@@ -1060,12 +1073,23 @@ func (c *IPAMContext) setupENI(eni string, eniMetadata awsutils.ENIMetadata, isT
...
-		// Either case add the IPs and prefixes to datastore.
-		c.addENIsecondaryIPsToDataStore(eniMetadata.IPv4Addresses, eni)
-		c.addENIv4prefixesToDataStore(eniMetadata.IPv4Prefixes, eni)
+		if !v6TrunkEni {
+			log.Infof("Found ENIs having %d secondary IPs and %d Prefixes", len(eniMetadata.IPv4Addresses), len(eniMetadata.IPv4Prefixes))
+			// Either case add the IPs and prefixes to datastore.
+			c.addENIsecondaryIPsToDataStore(eniMetadata.IPv4Addresses, eni)
+			c.addENIv4prefixesToDataStore(eniMetadata.IPv4Prefixes, eni)
+		} else {
+			// Cache IPv6 gateway address to speed up CNI ADD
```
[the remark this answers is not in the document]
jdn5126: added in latest revision

pkg/ipamd/ipamd.go
```diff
@@ -1060,12 +1078,24 @@ func (c *IPAMContext) setupENI(eni string, eniMetadata awsutils.ENIMetadata, isT
 				return errors.Wrapf(err, "failed to set up ENI %s network", eni)
 			}
 		}
-		log.Infof("Found ENIs having %d secondary IPs and %d Prefixes", len(eniMetadata.IPv4Addresses), len(eniMetadata.IPv4Prefixes))
-		// Either case add the IPs and prefixes to datastore.
-		c.addENIsecondaryIPsToDataStore(eniMetadata.IPv4Addresses, eni)
-		c.addENIv4prefixesToDataStore(eniMetadata.IPv4Prefixes, eni)
+		if !v6TrunkEni {
```
achevuru, line 1081: `v6TrunkEni` var sounds a bit odd to me for this check here. Why can't this just be `c.enableIPv4` (or) `!c.enableIPv6` instead?

jdn5126: It can be, I was trying to have a graceful solution to the case where a customer manually attaches a secondary ENI in IPv6 mode and does not mark it as unmanaged. I can change the `v6TrunkEni` checks here to just `c.enableIPv6` and comment that it currently only applies to v6 trunk ENIs

achevuru (approved):

PR #2640, opened 2023-10-31
Refactor AllocENI

jchen6585:
**What type of PR is this?**
enhancement

N/A

Reduces the number of API calls made when creating an ENI and allocating IP addresses. Combines AllocENI and AllocIpAddresses into one API call.

Manually tested against custom networking, cni, and ipv6 integration suites.
Screenshot 2023-11-06 at 1 24 56 PM
Screenshot 2023-11-06 at 11 50 41 AM
Screenshot 2023-11-06 at 12 19 36 PM

No

No

No

```release-note

```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -123,7 +123,7 @@ var (
 // APIs defines interfaces calls for adding/getting/deleting ENIs/secondary IPs. The APIs are not thread-safe.
 type APIs interface {
 	// AllocENI creates an ENI and attaches it to the instance
-	AllocENI(useCustomCfg bool, sg []*string, subnet string) (eni string, err error)
+	AllocENI(useCustomCfg bool, sg []*string, subnet string, numIPs int) (eni string, err error)
 
 	// FreeENI detaches ENI interface and deletes it
 	FreeENI(eniName string) error
@@ -740,8 +740,8 @@ func (cache *EC2InstanceMetadataCache) awsGetFreeDeviceNumber() (int, error) {
 
 // AllocENI creates an ENI and attaches it to the instance
 // returns: newly created ENI ID
-func (cache *EC2InstanceMetadataCache) AllocENI(useCustomCfg bool, sg []*string, subnet string) (string, error) {
-	eniID, err := cache.createENI(useCustomCfg, sg, subnet)
+func (cache *EC2InstanceMetadataCache) AllocENI(useCustomCfg bool, sg []*string, subnet string, numIPs int) (string, error) {
+	eniID, err := cache.createENI(useCustomCfg, sg, subnet, numIPs)
 	if err != nil {
 		return "", errors.Wrap(err, "AllocENI: failed to create ENI")
 	}
@@ -812,7 +812,7 @@ func (cache *EC2InstanceMetadataCache) attachENI(eniID string) (string, error) {
 }
 
 // return ENI id, error
-func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string, subnet string) (string, error) {
+func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string, subnet string, numIPs int) (string, error) {
 	eniDescription := eniDescriptionPrefix + cache.instanceID
 	tags := map[string]string{
 		eniCreatedAtTagKey: time.Now().Format(time.RFC3339),
@@ -826,14 +826,34 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
 			Tags:         convertTagsToSDKTags(tags),
 		},
 	}
+	var needIPs = numIPs
 
-	input := &ec2.CreateNetworkInterfaceInput{
-		Description:       aws.String(eniDescription),
-		Groups:            aws.StringSlice(cache.securityGroups.SortedList()),
-		SubnetId:          aws.String(cache.subnetID),
-		TagSpecifications: tagSpec,
+	ipLimit := cache.GetENIIPv4Limit()
+	if ipLimit < needIPs {
+		needIPs = ipLimit
 	}
 
+	log.Infof("Trying to allocate %d IP addresses on new ENI", needIPs)
+	log.Debugf("PD enabled - %t", cache.enablePrefixDelegation)
+	input := &ec2.CreateNetworkInterfaceInput{}
+
+	if cache.enablePrefixDelegation {
+		input = &ec2.CreateNetworkInterfaceInput{
+			Description:       aws.String(eniDescription),
+			Groups:            aws.StringSlice(cache.securityGroups.SortedList()),
+			SubnetId:          aws.String(cache.subnetID),
+			TagSpecifications: tagSpec,
+			Ipv4PrefixCount:   aws.Int64(int64(needIPs)),
+		}
+	} else {
+		input = &ec2.CreateNetworkInterfaceInput{
+			Description:                    aws.String(eniDescription),
+			Groups:                         aws.StringSlice(cache.securityGroups.SortedList()),
+			SubnetId:                       aws.String(cache.subnetID),
+			TagSpecifications:              tagSpec,
+			SecondaryPrivateIpAddressCount: aws.Int64(int64(needIPs)),
+		}
+	}
 	if useCustomCfg {
 		log.Info("Using a custom network config for the new ENI")
 		if len(sg) != 0 {
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -884,41 +884,39 @@ func (c *IPAMContext) tryAllocateENI(ctx context.Context) error {
 		subnet = eniCfg.Subnet
 	}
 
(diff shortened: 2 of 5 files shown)


pkg/ipamd/ipamd.go
```diff
@@ -884,26 +884,14 @@ func (c *IPAMContext) tryAllocateENI(ctx context.Context) error {
...
 		return err
 	}
 
-	resourcesToAllocate := c.GetENIResourcesToAllocate()
-	_, err = c.awsClient.AllocIPAddresses(eni, resourcesToAllocate)
-	if err != nil {
-		log.Warnf("Failed to allocate %d IP addresses on an ENI: %v", resourcesToAllocate, err)
-		// Continue to process the allocated IP addresses
-		ipamdErrInc("increaseIPPoolAllocIPAddressesFailed")
-		if containsInsufficientCIDRsOrSubnetIPs(err) {
```
jdn5126, line 900: I think we still want to be able to log an error like this when the subnet is exhausted. Can you test what we will log after this change when the subnet does not have enough IPs? You can do this by creating small subnets and passing them to eksctl: https://eksctl.io/usage/creating-and-managing-clusters/

pkg/awsutils/awsutils.go
```diff
@@ -827,11 +827,39 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
+
+	ipLimit := cache.GetENIIPv4Limit()
+
+	if ipLimit < needIPs {
+		needIPs = ipLimit
+	}
+
+	if needIPs < 1 {
+		return "", nil
+	}
```
jdn5126, line 840: nit: let's add a comment here explaining that we cannot allocate more IPs for an ENI than it can handle

pkg/awsutils/awsutils.go
```diff
@@ -827,11 +827,39 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
+
+	log.Infof("Trying to allocate %d IP addresses on new ENI", needIPs)
+	log.Debugf("PD enabled - %t", cache.enablePrefixDelegation)
+	input := &ec2.CreateNetworkInterfaceInput{}
+
+	if cache.enablePrefixDelegation {
+		needPrefixes := needIPs
```
jdn5126, line 847: nit: no need to create a new variable here

jdn5126 (commented):
Refactor overall looks good to me, but yeah we will need unit test coverage for this

pkg/ipamd/ipamd.go
```diff
@@ -884,24 +884,22 @@ func (c *IPAMContext) tryAllocateENI(ctx context.Context) error {
...
 		if containsInsufficientCIDRsOrSubnetIPs(err) {
+			ipamdErrInc("increaseIPPoolAllocIPAddressesFailed")
 			log.Errorf("Unable to attach IPs/Prefixes for the ENI, subnet doesn't seem to have enough IPs/Prefixes. Consider using new subnet or carve a reserved range using create-subnet-cidr-reservation")
 			c.lastInsufficientCidrError = time.Now()
-			return err
 		}
+
+		return err
+	}
+	if eni == "" {
```
jdn5126, line 901: What is this case? Should we log or do something here?

pkg/awsutils/awsutils.go
```diff
@@ -740,11 +740,14 @@ func (cache *EC2InstanceMetadataCache) awsGetFreeDeviceNumber() (int, error) {
...
-func (cache *EC2InstanceMetadataCache) AllocENI(useCustomCfg bool, sg []*string, subnet string) (string, error) {
-	eniID, err := cache.createENI(useCustomCfg, sg, subnet)
+func (cache *EC2InstanceMetadataCache) AllocENI(useCustomCfg bool, sg []*string, subnet string, numIPs int) (string, error) {
+	eniID, err := cache.createENI(useCustomCfg, sg, subnet, numIPs)
 	if err != nil {
 		return "", errors.Wrap(err, "AllocENI: failed to create ENI")
 	}
+	if eniID == "" {
+		return "", nil
+	}
```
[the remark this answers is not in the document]
jchen6585: This is for the case that `needIPs < 1`. createENI would return `"", nil` since no ENI is created. I'm not sure if combining with above case makes sense since it technically isn't an error.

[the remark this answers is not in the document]
jdn5126: I see, let's at least log and/or comment then

jdn5126 (approved):

pkg/awsutils/awsutils.go
```diff
@@ -740,11 +740,16 @@ func (cache *EC2InstanceMetadataCache) awsGetFreeDeviceNumber() (int, error) {
...
 // returns: newly created ENI ID
-func (cache *EC2InstanceMetadataCache) AllocENI(useCustomCfg bool, sg []*string, subnet string) (string, error) {
-	eniID, err := cache.createENI(useCustomCfg, sg, subnet)
+func (cache *EC2InstanceMetadataCache) AllocENI(useCustomCfg bool, sg []*string, subnet string, numIPs int) (string, error) {
+	eniID, err := cache.createENI(useCustomCfg, sg, subnet, numIPs)
 	if err != nil {
 		return "", errors.Wrap(err, "AllocENI: failed to create ENI")
 	}
+	// This is to handle the case where we need 0 IPs, so ENI will not be created
+	if eniID == "" {
```
jayanthvn, line 749: Why are we even entering AllocENI the num of IPs is 0?

pkg/ipamd/ipamd.go
```diff
@@ -884,24 +884,24 @@ func (c *IPAMContext) tryAllocateENI(ctx context.Context) error {
...
+			ipamdErrInc("increaseIPPoolAllocIPAddressesFailed")
 			log.Errorf("Unable to attach IPs/Prefixes for the ENI, subnet doesn't seem to have enough IPs/Prefixes. Consider using new subnet or carve a reserved range using create-subnet-cidr-reservation")
 			c.lastInsufficientCidrError = time.Now()
-			return err
 		}
+
+		return err
+	}
+	// This case is to handle an ENI not being created due to 0 IPs being needed.
+	if eni == "" {
```
jayanthvn, line 902: IMO we shouldn't be doing these checks, if we don't need IPs then we shouldn't even call Alloc.

But trying to understand more about the change, why are we merging the functions? Still we will end up with 2 EC2 calls isn't it?

jayanthvn: Nvm, we are passing the count to createNetworkInterfaceWithContext..but we should refactor the eni check..

pkg/awsutils/awsutils.go
jchen6585: Refactoring based on out offline discussion (Don't call AllocENI if we need 0 IPs)

jchen6585:
> Can you please run Ipv6, V4- PD and custom networking tests..

all suites passed

jdn5126 (approved):

pkg/awsutils/awsutils.go
```diff
@@ -826,14 +826,38 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
-		Description:       aws.String(eniDescription),
-		Groups:            aws.StringSlice(cache.securityGroups.SortedList()),
-		SubnetId:          aws.String(cache.subnetID),
-		TagSpecifications: tagSpec,
+	ipLimit := cache.GetENIIPv4Limit()
+	if ipLimit < needIPs {
+		needIPs = ipLimit
 	}
+	// If we don't need any more IPs, don't create ENI. It should never be the case.
+	if needIPs < 1 {
```
jayanthvn, line 836: This is safe to remove now? since we have the below check and this function is called from AllocENI...

```
if resourcesToAllocate > 0 {
		eni, err := c.awsClient.AllocENI(c.useCustomNetworking, securityGroups, subnet, resourcesToAllocate)
....
```

pkg/ipamd/ipamd.go
```diff
@@ -884,41 +884,37 @@ func (c *IPAMContext) tryAllocateENI(ctx context.Context) error {
...
 	resourcesToAllocate := c.GetENIResourcesToAllocate()
-	_, err = c.awsClient.AllocIPAddresses(eni, resourcesToAllocate)
-	if err != nil {
-		log.Warnf("Failed to allocate %d IP addresses on an ENI: %v", resourcesToAllocate, err)
-		// Continue to process the allocated IP addresses
-		ipamdErrInc("increaseIPPoolAllocIPAddressesFailed")
-		if containsInsufficientCIDRsOrSubnetIPs(err) {
-			log.Errorf("Unable to attach IPs/Prefixes for the ENI, subnet doesn't seem to have enough IPs/Prefixes. Consider using new subnet or carve a reserved range using create-subnet-cidr-reservation")
-			c.lastInsufficientCidrError = time.Now()
+	if resourcesToAllocate > 0 {
```
jayanthvn, line 888: TBH we don't need this check because the func is `tryAllocateENI` and we come here only if we need to increase the DS pool..Maybe for now we can leave it but we should revisit it later. Also can you add a debug line if we ever land here and resourceToAllocate is not greater than 0...

PR #2655, opened 2023-11-07
Make test binaries for run-ginkgo-integration-suite by default

jdn5126:
**What type of PR is this?**

testing

N/A

This PR changes the default for `SKIP_MAKE_TEST_BINARIES` in `scripts/run-ginkgo-integration-suite.sh` so that test suites are compiled by default. This is required for calling `scripts/run-ginkgo-integration-suite.sh` from automated environments.

N/A

N/A

No

N/A

N/A

No

```release-note

```

### DIFF
--- a/scripts/run-ginkgo-integration-suite.sh
+++ b/scripts/run-ginkgo-integration-suite.sh
@@ -9,7 +9,7 @@ set -e
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 GINKGO_TEST_BUILD="$SCRIPT_DIR/../test/build"
-SKIP_MAKE_TEST_BINARIES=1
+: "${SKIP_MAKE_TEST_BINARIES:=}"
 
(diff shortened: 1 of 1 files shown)


scripts/run-ginkgo-integration-suite.sh
```diff
@@ -9,7 +9,7 @@ set -e
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 GINKGO_TEST_BUILD="$SCRIPT_DIR/../test/build"
-SKIP_MAKE_TEST_BINARIES=1
+: "${SKIP_MAKE_TEST_BINARIES:=}"
```
orsenthil, line 12: What does `:` in the front do? Does it nullify the variable `SKIP_MAKE_TEST_BINARIES` ?

jdn5126: That defines the env variable if it is not already defined (not already passed to the script) and lets you pass a default value on the other side of `:=`, which in this case is `nil`.

orsenthil (approved):
LGTM

Issue #2650, opened 2023-11-01
CNI Failed to assign IPs while using WARM_IP_TARGET & MINIMUM_IP_TARGET


sidewinder12s:
**What happened**:

We have run into IP allocation issues in many of our clusters. Due to some technical constraints, we have a cluster that is constrained most of its nodes to a single subnet. We realized we had a few hundred IPs wasted on these nodes due to the workload/scheduling constraints (largely 1 to 1 mapping of node to workload pod), so we tried modifying our IP allocation settings to reduce that waste.

Original:

```yaml
    WARM_ENI_TARGET: "null"
    WARM_IP_TARGET: "5"
    MINIMUM_IP_TARGET: "10"
```

New:

```yaml
    WARM_ENI_TARGET: "null"
    WARM_IP_TARGET: "0"
    MINIMUM_IP_TARGET: "11"
```

After applying these settings, the majority of workloads/nodes appeared to function as expected with no IP allocation issues. We however realized we a few longer lived nodes had run out of IPs and were failing to assign more IPs, even though the subnet had plenty of slack capacity.

We largely had issues on nodes with more than 11 pods that required IPs. It seems like there might be some unhandled condition if you keep the warm IP pool empty?

Once I raised the `WARM_IP_TARGET` to 1, all pods that were failing to assign IPs appeared to recover.

**Attach logs**

Logs sent in support case: 14209321561

**What you expected to happen**:

We would be able to assign IPs

**How to reproduce it (as minimally and precisely as possible)**:

I'd assume just take our original settings:

- apply them
- Launch pods
- Apply new settings

I think what might trigger it is if you have a minimum IP target that does not meet all pod requirements and then have another pod get scheduled.

**Anything else we need to know?**:

We are aware this configuration is not recommended and have EC2 API Metrics enabled to observe for throttling.

**Environment**:
- Kubernetes version (use kubectl version): 1.25
- CNI Version: v1.13.3
- OS (e.g: cat /etc/os-release): EKS AMI v20230825
- Kernel (e.g. uname -a): 5.10.186-179.751.amzn2

jdn5126:
@sidewinder12s I see the problem here, and we really need better documentation around what `WARM_IP_TARGET=0` or `WARM_ENI_TARGET=0` mean.

Today, the VPC CNI only supports "static" allocation of IPs. At startup, we allocate the `MINIMUM` targets, then we continuously maintain the `WARM` thresholds as IPs are used for pods.

[...]

PR #2659, opened 2023-11-08
Add configuration check for env WARM_IP_TARGET and MINIMUM_IP_TARGET

oliviassss:
**What type of PR is this?**

bug fix

https://github.com/aws/amazon-vpc-cni-k8s/issues/2650
This PR adds a configuration check on env, if `MINIMUM_IP_TARGET` is set but `WARM_IP_TARGET` is not set, it will error out. Also update docs on this change.

```release-note

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,7 +333,7 @@ elasticity, but uses roughly half as many IPs as using WARM_IP_TARGET alone (32 
 This also improves the reliability of the EKS cluster by reducing the number of calls necessary to allocate or deallocate
 private IPs, which may be throttled, especially at scaling-related times.
 
-Setting both `WARM_IP_TARGET` and `MINIMUM_IP_TARGET` will override `WARM_PREFIX_TARGET`. For a detailed explanation, see
+Setting both `WARM_IP_TARGET` and `MINIMUM_IP_TARGET` will override `WARM_PREFIX_TARGET` and `WARM_ENI_TARGET`. For a detailed explanation, see
 [`WARM_PREFIX_TARGET`, `WARM_IP_TARGET` and `MINIMUM_IP_TARGET`](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/docs/prefix-and-ip-target.md).
 
 #### `MAX_ENI`
--- a/README.md
+++ b/README.md
@@ -290,7 +290,7 @@ Type: Integer
 
 Default: None
 
-Specifies the number of free IP addresses that the `ipamd` daemon should attempt to keep available for pod assignment on the node. Setting this to a non-positive value is the same as setting this to 0.
+Specifies the number of free IP addresses that the `ipamd` daemon should attempt to keep available for pod assignment on the node. Setting this to a non-positive value is the same as setting this to 0 or not setting the variable.
 With `ENABLE_PREFIX_DELEGATION` set to `true` then `ipamd` daemon will check if the existing (/28) prefixes are enough to maintain the
 `WARM_IP_TARGET` if it is not sufficient then more prefixes will be attached.
 
--- a/README.md
+++ b/README.md
@@ -288,9 +288,9 @@ If `WARM_IP_TARGET` is set, then this environment variable is ignored and the `W
 
 Type: Integer
 
-Default: None
+Default: `1`
 
-Specifies the number of free IP addresses that the `ipamd` daemon should attempt to keep available for pod assignment on the node. Setting this to a non-positive value is the same as setting this to 0 or not setting the variable.
+Specifies the number of free IP addresses that the `ipamd` daemon should attempt to keep available for pod assignment on the node. Setting this to a non-positive value is the same as setting this to 0.
 With `ENABLE_PREFIX_DELEGATION` set to `true` then `ipamd` daemon will check if the existing (/28) prefixes are enough to maintain the
 `WARM_IP_TARGET` if it is not sufficient then more prefixes will be attached.
 
@@ -323,8 +323,8 @@ Specifies the number of total IP addresses that the `ipamd` daemon should attemp
 addresses to keep available at all times, it sets a target number for a floor on how many total IP addresses are allocated. Setting to a
 non-positive value is same as setting this to 0 or not setting the variable.
 
-`MINIMUM_IP_TARGET` is for pre-scaling, `WARM_IP_TARGET` is for dynamic scaling. For example, suppose a cluster has an
-expected pod density of approximately 30 pods per node. If `WARM_IP_TARGET` is set to 30 to ensure there are enough IPs
+`MINIMUM_IP_TARGET` is for pre-scaling, `WARM_IP_TARGET` is for dynamic scaling, and when `MINIMUM_IP_TARGET` is set, it is required that `WARM_IP_TARGET` is also set to positive value. 
+For example, suppose a cluster has an expected pod density of approximately 30 pods per node. If `WARM_IP_TARGET` is set to 30 to ensure there are enough IPs
 allocated up front by the CNI, then 30 pods are deployed to the node, the CNI will allocate an additional 30 IPs, for
 a total of 60, accelerating IP exhaustion in the relevant subnets. If instead `MINIMUM_IP_TARGET` is set to 30 and
 `WARM_IP_TARGET` to 2, after the 30 pods are deployed the CNI would allocate an additional 2 IPs. This still provides
@@ -332,6 +332,9 @@ elasticity, but uses roughly half as many IPs as using WARM_IP_TARGET alone (32 
 
 This also improves the reliability of the EKS cluster by reducing the number of calls necessary to allocate or deallocate
 private IPs, which may be throttled, especially at scaling-related times.
+
+Setting both `WARM_IP_TARGET` and `MINIMUM_IP_TARGET` will override `WARM_PREFIX_TARGET`. For a detailed explanation, see
+[`WARM_PREFIX_TARGET`, `WARM_IP_TARGET` and `MINIMUM_IP_TARGET`](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/docs/prefix-and-ip-target.md).
 
 #### `MAX_ENI`
 
--- a/README.md
+++ b/README.md
@@ -288,7 +288,7 @@ If `WARM_IP_TARGET` is set, then this environment variable is ignored and the `W
 
 Type: Integer
 
-Default: `1`
+Default: None
 
 Specifies the number of free IP addresses that the `ipamd` daemon should attempt to keep available for pod assignment on the node. Setting this to a non-positive value is the same as setting this to 0.
 With `ENABLE_PREFIX_DELEGATION` set to `true` then `ipamd` daemon will check if the existing (/28) prefixes are enough to maintain the
--- a/cmd/aws-vpc-cni/main.go
+++ b/cmd/aws-vpc-cni/main.go
@@ -83,7 +83,7 @@ const (
 	defaultEnPrefixDelegation    = false
 	defaultIPCooldownPeriod      = 30
 	defaultDisablePodV6          = false
-	defaultWarmIPTarget          = 1
+	defaultWarmIPTarget          = 0
 	defaultMinIPTarget           = 0
 	defaultWarmPrefixTarget      = 0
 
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -288,9 +288,9 @@ If `WARM_IP_TARGET` is set, then this environment variable is ignored and the `W
 
 Type: Integer
 
-Default: None
+Default: `1`
```
jdn5126, line 291: So the default here is still `None`, as when set, the WARM/MINIMUM IP targets take precedence over the WARM_ENI_TARGET. The default when `MINIMUM_IP_TARGET` is set is 1

cmd/aws-vpc-cni/main.go
```diff
@@ -83,6 +83,9 @@ const (
 	defaultEnPrefixDelegation    = false
 	defaultIPCooldownPeriod      = 30
 	defaultDisablePodV6          = false
+	defaultWarmIPTarget          = "1"
+	defaultMinIPTarget           = "0"
+	defaultWarmPrefixTarget      = "0"
```
jdn5126, line 88: nit: we should use the actual integer value here, i.e. `1` or `0`

cmd/aws-vpc-cni/main.go
```diff
@@ -384,14 +387,17 @@ func validateEnvVars() bool {
 	}
 
 	prefixDelegationEn := utils.GetBoolAsStringEnvVar(envEnPrefixDelegation, defaultEnPrefixDelegation)
-	warmIPTarget := utils.GetEnv(envWarmIPTarget, "0")
-	warmPrefixTarget := utils.GetEnv(envWarmPrefixTarget, "0")
-	minimumIPTarget := utils.GetEnv(envMinIPTarget, "0")
+	warmIPTarget := utils.GetEnv(envWarmIPTarget, defaultWarmIPTarget)
+	warmPrefixTarget := utils.GetEnv(envWarmPrefixTarget, defaultWarmPrefixTarget)
+	minimumIPTarget := utils.GetEnv(envMinIPTarget, defaultMinIPTarget)
```
jdn5126, line 392: With the integer values, we can use `utils.GetIntFromStringEnvVar`

cmd/aws-vpc-cni/main.go
```diff
@@ -384,14 +387,17 @@ func validateEnvVars() bool {
...
 
 	// Note that these string values should probably be cast to integers, but the comparison for values greater than 0 works either way
 	if prefixDelegationEn && (warmIPTarget <= "0" && warmPrefixTarget <= "0" && minimumIPTarget <= "0") {
 		log.Errorf("Setting WARM_PREFIX_TARGET = 0 is not supported while WARM_IP_TARGET/MINIMUM_IP_TARGET is not set. Please configure either one of the WARM_{PREFIX/IP}_TARGET or MINIMUM_IP_TARGET env variables")
 		return false
+	} else if warmIPTarget <= "0" && minimumIPTarget > "0" {
```
jdn5126, line 398: nit: with the int values above, we can change this from string checks to integer checks. Also, we should allow `WARM_IP_TARGET=0` when `MINIMUM_IP_TARGET` is set. We just want to default `WARM_IP_TARGET` to 1 in this case (env var is unset)

jdn5126 (changes requested):

oliviassss:
As discussed internally, we decided not to change the default value of `WARM_IP_TARGET`, as it may cause breaking change to existing users by overriding the env `WARM_ENI_TARGET`. Instead, we add a configuration check, when `MINIMUM_IP_TARGET` is defined but `WARM_IP_TARGET` is not defined, return an error.

cmd/aws-vpc-cni/main.go
oliviassss: updated

cmd/aws-vpc-cni/main.go
```diff
@@ -384,14 +387,29 @@ func validateEnvVars() bool {
...
+		log.Errorf("error when trying to get env MINIMUM_IP_TARGET: %s; input is %v", err, input)
+		return false
+	}
 
 	// Note that these string values should probably be cast to integers, but the comparison for values greater than 0 works either way
-	if prefixDelegationEn && (warmIPTarget <= "0" && warmPrefixTarget <= "0" && minimumIPTarget <= "0") {
+	if prefixDelegationEn && (warmIPTarget <= 0 && warmPrefixTarget <= 0 && minimumIPTarget <= 0) {
 		log.Errorf("Setting WARM_PREFIX_TARGET = 0 is not supported while WARM_IP_TARGET/MINIMUM_IP_TARGET is not set. Please configure either one of the WARM_{PREFIX/IP}_TARGET or MINIMUM_IP_TARGET env variables")
 		return false
+	} else if warmIPTarget <= 0 && minimumIPTarget > 0 {
```
jdn5126, line 410: nit: I think this check should be: `if (minimumIPTarget is set and warmIPTarget is not set) { ... }`
And the way we can check if they are set is by using the `input` value returned from `GetIntFromStringEnvVar`

oliviassss: Updated, thanks

cmd/aws-vpc-cni/main.go
```diff
@@ -384,14 +387,29 @@ func validateEnvVars() bool {
...
+		return false
+	}
 
 	// Note that these string values should probably be cast to integers, but the comparison for values greater than 0 works either way
-	if prefixDelegationEn && (warmIPTarget <= "0" && warmPrefixTarget <= "0" && minimumIPTarget <= "0") {
+	if prefixDelegationEn && (warmIPTarget <= 0 && warmPrefixTarget <= 0 && minimumIPTarget <= 0) {
 		log.Errorf("Setting WARM_PREFIX_TARGET = 0 is not supported while WARM_IP_TARGET/MINIMUM_IP_TARGET is not set. Please configure either one of the WARM_{PREFIX/IP}_TARGET or MINIMUM_IP_TARGET env variables")
 		return false
+	} else if inputWarmIP == "" && inputMinIP != "" {
+		log.Errorf("WARM_IP_TARGET should be set while MINIMUM_IP_TARGET is set, in order to avoid no new ENIs being allocated after minimum threshold is meet, please configure properly")
```
jdn5126, line 411: last nit: Can we change the error message to - "WARM_IP_TARGET MUST be set when MINIMUM_IP_TARGET is set. Please configure both environment variables."

cmd/aws-vpc-cni/main.go
```diff
@@ -384,14 +387,29 @@ func validateEnvVars() bool {
...
+		log.Errorf("error when trying to get env WARM_PREFIX_TARGET: %s; input is %v", err, inputWarmPrefix)
+		return false
+	}
+	minimumIPTarget, err, inputMinIP := utils.GetIntFromStringEnvVar(envMinIPTarget, defaultMinIPTarget)
+	if err != nil {
+		log.Errorf("error when trying to get env MINIMUM_IP_TARGET: %s; input is %v", err, inputMinIP)
+		return false
+	}
 
 	// Note that these string values should probably be cast to integers, but the comparison for values greater than 0 works either way
```
jdn5126, line 406: nit: this comment can be deleted now

cmd/aws-vpc-cni/main.go
```diff
@@ -384,14 +387,29 @@ func validateEnvVars() bool {
...
-	if prefixDelegationEn && (warmIPTarget <= "0" && warmPrefixTarget <= "0" && minimumIPTarget <= "0") {
+	if prefixDelegationEn && (warmIPTarget <= 0 && warmPrefixTarget <= 0 && minimumIPTarget <= 0) {
 		log.Errorf("Setting WARM_PREFIX_TARGET = 0 is not supported while WARM_IP_TARGET/MINIMUM_IP_TARGET is not set. Please configure either one of the WARM_{PREFIX/IP}_TARGET or MINIMUM_IP_TARGET env variables")
 		return false
```
jdn5126, line 409: nit: Since we already check for error conditions above, we can change this if condition to:
```
if prefixDelegationEn && (warmPrefixTarget == 0 && warmIPTarget == 0 && minimumIPTarget == 0) {
    log.Errorf("Setting WARM_PREFIX_TARGET=0 is not supported while WARM_IP_TARGET/MINIMUM_IP_TARGET is not set. Please configure either one of the WARM_{PREFIX/IP}_TARGET or MINIMUM_IP_TARGET env variables")
    return false
}
```

jdn5126 (changes requested):
Last nits, and then this is good to merge

cmd/aws-vpc-cni/main.go
oliviassss: updated

cmd/aws-vpc-cni/main.go
oliviassss: comment deleted

cmd/aws-vpc-cni/main.go
oliviassss: updated, thanks

jdn5126 (approved):

cmd/aws-vpc-cni/main.go
```diff
@@ -384,14 +387,28 @@ func validateEnvVars() bool {
 	}
 
 	prefixDelegationEn := utils.GetBoolAsStringEnvVar(envEnPrefixDelegation, defaultEnPrefixDelegation)
-	warmIPTarget := utils.GetEnv(envWarmIPTarget, "0")
-	warmPrefixTarget := utils.GetEnv(envWarmPrefixTarget, "0")
-	minimumIPTarget := utils.GetEnv(envMinIPTarget, "0")
+	warmIPTarget, err, inputWarmIP := utils.GetIntFromStringEnvVar(envWarmIPTarget, defaultWarmIPTarget)
```
jayanthvn, line 390: Minor nit - `inputWarmIP` -> `WarmIPTargetStr`

Similar change for below sections. Ideally we should also change the func name `GetIntFromStringEnvVar` to `GetEnvVar` (maybe?) since it returns both int and string..we can do that as fast followup if there are lot of places to change.

cmd/aws-vpc-cni/main.go
```diff
@@ -384,14 +387,28 @@ func validateEnvVars() bool {
...
+		log.Errorf("error when trying to get env MINIMUM_IP_TARGET: %s; input is %v", err, inputMinIP)
+		return false
+	}
 
-	// Note that these string values should probably be cast to integers, but the comparison for values greater than 0 works either way
-	if prefixDelegationEn && (warmIPTarget <= "0" && warmPrefixTarget <= "0" && minimumIPTarget <= "0") {
+	if prefixDelegationEn && (warmIPTarget == 0 && warmPrefixTarget == 0 && minimumIPTarget == 0) {
 		log.Errorf("Setting WARM_PREFIX_TARGET = 0 is not supported while WARM_IP_TARGET/MINIMUM_IP_TARGET is not set. Please configure either one of the WARM_{PREFIX/IP}_TARGET or MINIMUM_IP_TARGET env variables")
 		return false
+	} else if inputWarmIP == "" && inputMinIP != "" {
```
jayanthvn, line 409: We need to call it out in our docs about this change since if anyone is using this might start seeing aws-node startup failures on upgrades.

cmd/aws-vpc-cni/main.go
oliviassss: Renamed the `GetIntFromStringEnvVar` to `GetEnvVar`, and also renamed the output var. Thanks

README.md
```diff
@@ -333,6 +333,9 @@ elasticity, but uses roughly half as many IPs as using WARM_IP_TARGET alone (32
 This also improves the reliability of the EKS cluster by reducing the number of calls necessary to allocate or deallocate
 private IPs, which may be throttled, especially at scaling-related times.
 
+Setting both `WARM_IP_TARGET` and `MINIMUM_IP_TARGET` will override `WARM_PREFIX_TARGET`. For a detailed explanation, see
```
jayanthvn, line 336: will override WARM_PREFIX and WARM_ENI

README.md
```diff
@@ -323,8 +323,8 @@ Specifies the number of total IP addresses that the `ipamd` daemon should attemp
 addresses to keep available at all times, it sets a target number for a floor on how many total IP addresses are allocated. Setting to a
 non-positive value is same as setting this to 0 or not setting the variable.
 
-`MINIMUM_IP_TARGET` is for pre-scaling, `WARM_IP_TARGET` is for dynamic scaling. For example, suppose a cluster has an
-expected pod density of approximately 30 pods per node. If `WARM_IP_TARGET` is set to 30 to ensure there are enough IPs
+`MINIMUM_IP_TARGET` is for pre-scaling, `WARM_IP_TARGET` is for dynamic scaling, and when `MINIMUM_IP_TARGET` is set, it is required that `WARM_IP_TARGET` is also set to positive value.
```
jayanthvn, line 326: Does this mean we will support WARM_IP_TARGET = 0? Just in case anyone wants to use just the min ip and that will be max number of pods on that node needing IPs from CNI....

cmd/aws-vpc-cni/main.go
```diff
@@ -377,21 +380,35 @@ func validateEnvVars() bool {
...
+		log.Errorf("error when trying to get env MINIMUM_IP_TARGET: %s; input is %v", err, minimumIPTargetStr)
+		return false
+	}
 
-	// Note that these string values should probably be cast to integers, but the comparison for values greater than 0 works either way
-	if prefixDelegationEn && (warmIPTarget <= "0" && warmPrefixTarget <= "0" && minimumIPTarget <= "0") {
+	if prefixDelegationEn && (warmIPTarget == 0 && warmPrefixTarget == 0 && minimumIPTarget == 0) {
 		log.Errorf("Setting WARM_PREFIX_TARGET = 0 is not supported while WARM_IP_TARGET/MINIMUM_IP_TARGET is not set. Please configure either one of the WARM_{PREFIX/IP}_TARGET or MINIMUM_IP_TARGET env variables")
 		return false
+	} else if warmIPTargetStr == "" && minimumIPTargetStr != "" {
```
jayanthvn, line 409: We can have warm eni target (which is the default config) and not warm ip/min ip right?

jayanthvn (changes requested):
We can just have WARM ENI and not have WARM IP/MIN IP..might need to make some changes to accommodate that..

README.md
jdn5126: Yes, we will continue to allow setting `WARM_IP_TARGET=0`, which customers can use in combination with `MINIMUM_IP_TARGET` and max pods to prevent all future ENI allocation

cmd/aws-vpc-cni/main.go
jdn5126: Right. In that case, `minimumIPTargetStr` would be empty as well

jdn5126:
> We can just have WARM ENI and not have WARM IP/MIN IP..might need to make some changes to accommodate that..

That is covered. When MIN IP is set, WARM ENI is ignored, though

README.md
jdn5126: will fix after this merges

jdn5126 (approved):

jdn5126:
Closing this PR as we have chosen to update the README to call this behavior out instead: https://github.com/aws/amazon-vpc-cni-k8s#minimum_ip_target-v160

PR #2663, opened 2023-11-14
Merge master into release-1.15 for v1.15.4 release

jdn5126:
**What type of PR is this?**

release

N/A

This PR merges master into `release-1.15` for the upcoming v1.15.4 release. The only difference between `v1.15.4` and the current `master` will be that the CNI spec will be held at `0.4.0` for backward compatibility with EKS 1.23.

N/A

Manually verified that all integration tests pass (used EKS 1.23 and 1.27)

No

No, Yes

N/A

N/A

```release-note

```


jayanthvn (approved):
Certain test files have spec v1.0.0 but those are for egress plugin and doesn't actually test anything. So lgtm..

PR #2667, opened 2023-11-15
Manifest for Multus 4.0.2 thick plugin support

jdn5126:
**What type of PR is this?**

dependency

N/A

This PRs adds a new manifest for Multus CNI v4.0.2 (https://github.com/k8snetworkplumbingwg/multus-cni/releases/tag/v4.0.2). The AWS-built image in this manifest has been validated against the latest VPC CNI. Note that we are only publishing the `thick` plugin currently. Additionally publishing the `thin` plugin is future work.

N/A

Manually verified that this manifest can be applied to an existing cluster. Multus pods come up and pod IPs are assigned. The existing `multus` integration test checks no longer apply now that the multus architecture has changed. The existing log checks are no longer present.

No, Yes

No

Yes

```release-note
Provide manifest for compatibility with Multus v4.0.2
```

### DIFF
--- a/config/multus/v4.0.2-eksbuild.1/multus-daemonset-thick.yml
+++ b/config/multus/v4.0.2-eksbuild.1/multus-daemonset-thick.yml
@@ -1,0 +1,239 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: network-attachment-definitions.k8s.cni.cncf.io
+spec:
+  group: k8s.cni.cncf.io
+  scope: Namespaced
+  names:
+    plural: network-attachment-definitions
+    singular: network-attachment-definition
+    kind: NetworkAttachmentDefinition
+    shortNames:
+      - net-attach-def
+  versions:
+    - name: v1
+      served: true
+      storage: true
+      schema:
+        openAPIV3Schema:
+          description: 'NetworkAttachmentDefinition is a CRD schema specified by the Network Plumbing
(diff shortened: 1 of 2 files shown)


haouc (approved):
lgtm

emmanuel:
Thanks for publishing an updated manifest.

Where can I find the Dockerfile(s) from which the Multus image(s) is (are) built?

jdn5126:
@emmanuel we build using the upstream Multus Dockerfile, so for the v4.0.2 thick plugin, that is https://github.com/k8snetworkplumbingwg/multus-cni/blob/v4.0.2/images/Dockerfile.thick

We also construct the manifest from the upstream Multus manifest, so for the v4.0.2 thick plugin, that is https://github.com/k8snetworkplumbingwg/multus-cni/blob/v4.0.2/deployments/multus-daemonset-thick.yml

emmanuel:
@jdn5126 thank you. I wonder then, where does the (AWS-specific) `entrypoint.sh` which is the entrypoint of the image distributed here come from?

jdn5126:
> @jdn5126 thank you. I wonder then, where does the (AWS-specific) `entrypoint.sh` which is the entrypoint of the image distributed here come from?

There's no `entrypoint.sh` in v4.0.2, but in past releases, that also came from Multus upstream: https://github.com/k8snetworkplumbingwg/multus-cni/blob/master/images/README.md#entrypointsh-parameters

The only thing AWS-specific here is that the image is hosted in private ECRs. When needed, we sometimes build Multus against an EKS Distro minimal base image and carry CVE patches. So we want to maintain that flexibility.

PR #2603, opened 2023-10-04
Prometheus metrics scraping from CNI metrics helper

jayanthvn:
**What type of PR is this?**
Feature

**Which issue does this PR fix**:
Fixes #2071 

**What does this PR do / Why do we need it**:
Reopening #2192 and some additional refactoring.

CNI metrics helper merges all the metrics from aws-nodes and aggregates the metrics to send for CW. With this PR the aggregated data is sent to Prometheus when "USE_PROMETHEUS" is set. As part of initial enhancement only Prometheus type GAUGE is streamed for prometheus. Future enhancements will be made for other types of metrics.

This will include -

```
                "awscni_eni_max":                   EnisMax,
		"awscni_ip_max":                    IpMax,
		"awscni_add_ip_req_count":          AddIPCnt,
		"awscni_del_ip_req_count":          DelIPCnt,
		"awscni_eni_allocated":             Enis,
		"awscni_total_ip_addresses":        TotalIPs,
		"awscni_assigned_ip_addresses":     AssignedIPs,
		"awscni_force_removed_enis":        ForceRemovedENIs,
		"awscni_force_removed_ips":         ForceRemovedIPs,
		"awscni_total_ipv4_prefixes":       TotalPrefixes,
		"awscni_no_available_ip_addresses": NoAvailableIPAddrs,
```

**If an issue # is not available please add repro steps and logs from IPAMD/CNI showing the issue**:

**Testing done on this change**:

```
kube-system   cni-metrics-helper-5d4d84b8d4-6pxmn                    1/1     Running   0          3m3s   192.168.38.249   
```

```
curl http://192.168.38.249:61681/metrics
# HELP awscni_add_ip_req_count The number of add IP address requests
# TYPE awscni_add_ip_req_count gauge
awscni_add_ip_req_count 0
# HELP awscni_assigned_ip_addresses The number of IP addresses assigned to pods
# TYPE awscni_assigned_ip_addresses gauge
awscni_assigned_ip_addresses 14
# HELP awscni_eni_allocated The number of ENIs allocated
# TYPE awscni_eni_allocated gauge
awscni_eni_allocated 4
# HELP awscni_eni_max The maximum number of ENIs that can be attached to the instance, accounting for unmanaged ENIs
# TYPE awscni_eni_max gauge
awscni_eni_max 8
# HELP awscni_force_removed_enis The number of ENIs force removed while they had assigned pods
# TYPE awscni_force_removed_enis gauge
awscni_force_removed_enis 0
[...]

PR #2670, opened 2023-11-16
fix: Add parameters for tuning revisionHistory and securityContext

bodgit:
**What type of PR is this?**

feature

n/a

This PR adds support for overriding the `revisionHistoryLimit` setting on the Deployment as well as the setting the `securityContext` at both the Pod and container level.

Defaults values are set to the implicit Kubernetes defaults.

```release-note

```


jdn5126:
@bodgit two questions here:
1. What is the value of the `revisionHistoryLimit`?
2. Why would you want a different `securityContext` for the pod spec and container spec where there is only one container in this pod?

bodgit:
> 1. What is the value of the `revisionHistoryLimit`?

10 is the implicit default in current Kubernetes clusters if it's not set, so that's used as the default. My intent with the change is to be able to set this to 0 to not keep any prior revisions. Unfortunately Helm considers 0 a false value so trying to wrap the line with `{{ if .Values.revisionHistoryLimit }}...{{ end }}` results in no line being added for that use case.

> 2. Why would you want a different `securityContext` for the pod spec and container spec where there is only one container in this pod?

Force of habit on my part, however there's a common subset in both places but some settings can only be set in either place, (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#securitycontext-v1-core vs https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#podsecuritycontext-v1-core)

charts/cni-metrics-helper/values.yaml
```diff
@@ -28,3 +28,9 @@ serviceAccount:
     # eks.amazonaws.com/role-arn: arn:aws:iam::AWS_ACCOUNT_ID:role/IAM_ROLE_NAME  
 
 resources: {}
+
+revisionHistoryLimit: 10
+
+podSecurityContext: {}
+
+securityContext: {}
```
jdn5126, line 36: nit: maybe we should call this `containerSecurityContext` to differentiate it?

jdn5126:
> > 1. What is the value of the `revisionHistoryLimit`?
> 
> 10 is the implicit default in current Kubernetes clusters if it's not set, so that's used as the default. My intent with the change is to be able to set this to 0 to not keep any prior revisions. Unfortunately Helm considers 0 a false value so trying to wrap the line with `{{ if .Values.revisionHistoryLimit }}...{{ end }}` results in no line being added for that use case.
> 
> > 2. Why would you want a different `securityContext` for the pod spec and container spec where there is only one container in this pod?
> 
> Force of habit on my part, however there's a common subset in both places but some settings can only be set in either place, (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#securitycontext-v1-core vs https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#podsecuritycontext-v1-core)

Got it, thank you for explaining, and thank you for taking this up! I am fine with both of these changes. For the YAML comment, you check if the value is greater than or equal to 0, but it's not a big deal

charts/cni-metrics-helper/values.yaml
bodgit: Some charts use `securityContext` & `containerSecurityContext`, others use `podSecurityContext` and `securityContext` so just a case of settling on a convention. I can update the name.

bodgit:
Parameter names tweaked and rebased.

jdn5126 (approved):

jdn5126:
Thanks @bodgit! Since you are using CNI Metrics Helper, you may also be interested in https://github.com/aws/amazon-vpc-cni-k8s/pull/2603, which takes the aggregated metrics published to CloudWatch and makes some of them available via Prometheus

bodgit:
> Thanks @bodgit! Since you are using CNI Metrics Helper, you may also be interested in #2603, which takes the aggregated metrics published to CloudWatch and makes some of them available via Prometheus

Yes, that does look useful. Thanks.

Issue #2373, opened 2023-05-08
AWS CNI failed to add chain rule for each CIDR in VPC with nf_tables mode


HenryXie1:
**What happened**:
After adopting Red Hat Enterprise Linux 8.7 (Ootpa) 4.18.0-425.19.2.el8_7.x86_64 AMI (08d4770d37a0b97a1) which enabled nftable, all AWS CNI pods are pending with:

time="2023-04-17T03:17:29Z" level=info msg="Updating iptables mode to nft" 
Installed /host/opt/cni/bin/aws-cni 
Installed /host/opt/cni/bin/egress-v4-cni 
time="2023-04-17T03:17:29Z" level=info msg="Starting IPAM daemon... " 
time="2023-04-17T03:17:29Z" level=info msg="Checking for IPAM connectivity... " 
time="2023-04-17T03:17:30Z" level=info msg="Copying config file... " 
time="2023-04-17T03:17:30Z" level=info msg="Successfully copied CNI plugin binary and config file." 
time="2023-04-17T03:17:30Z" level=error msg="Failed to wait for IPAM daemon to complete" error="exit status 1"

Node's ipamd log has error: 

{"level":"error","ts":"2023-04-27T08:27:59.957Z","caller":"networkutils/network.go:368",
"msg":"host network setup: failed to add nat/AWS-SNAT-CHAIN-14 rule [14] AWS-SNAT-CHAIN shouldExist true rule [
! -d 100.64.0.0/16 -m comment --comment AWS SNAT CHAIN -j AWS-SNAT-CHAIN-15], running [/usr/sbin/iptables -t nat -A AWS-SNAT-CHAIN-14 ! -d 100.64.0.0/16 -m comment --comment AWS SNAT CHAIN -j AWS-SNAT-CHAIN-15 --wait]: exit status 4: iptables v1.8.4 (nf_tables):  RULE_APPEND failed (Too many links): rule in chain AWS-SNAT-CHAIN-14\n"} 

AWS CNI adds chain rule for each CIDR in VPC, and nftable has hard limitation NFT_JUMP_STACK_SIZE = 16  defined in the code (e.g https://codebrowser.dev/linux/linux/include/net/netfilter/nf_tables.h.html#22). So for VPCs which has 15 or above CIDRs, CNI can't attach all rules successfully in nftable which caused error above.

**Attach logs**

**What you expected to happen**:
[...]

PR #2697, opened 2023-12-01
Refactor IPTable Rules

jchen6585:
**What type of PR is this?**
bug

#2373 

nf_tables mode restricts the jump stack size to at most 16. We currently have (one jump per VPC CIDR/Excluded CIDR) + 1, so having 15+ CIDRS will result in a jump stack size greater than 16, causing failure. This pr refactors our iptable rules to retain current behavior while keeping the jump stack size to a size of 2.

Manually checked iptable rules for 16 CIDRS, aws-node is up and running
Screenshot 2023-12-01 at 11 31 45 AM

Manually tested adding exclusion cidrs using `AWS_VPC_K8S_CNI_EXCLUDE_SNAT_CIDRS`

Also manually checked up to VPC Limit (50 CIDRS).
Manually checked packet using tcpdump
- Leaving VPC goes out eth0
- Routing within VPC goes out correct eth

Integration Tests:
- CNI: Passing
Screenshot 2023-12-01 at 10 55 52 AM

- IPAMD: Passing
Screenshot 2023-12-04 at 11 20 26 AM

- SNAT: Passing
Screenshot 2023-12-05 at 2 21 07 PM

Ensured traffic to the internet aren't dropped while rules are being updated (deleted and added 10+ cidrs while ping in a pod was running)
Screenshot 2023-12-08 at 4 34 32 PM

Scale Testing
130 Pod
Screenshot 2023-12-01 at 4 25 34 PM

730 pod
Screenshot 2023-12-01 at 4 25 46 PM

5000 pod
Screenshot 2023-12-01 at 4 25 55 PM

No

No, Yes

No

No

```release-note

```

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -395,6 +395,11 @@ func (c *IPAMContext) nodeInit() error {
 	if err != nil {
 		return errors.Wrap(err, "ipamd init: failed to set up host network")
 	}
+	err = c.networkClient.CleanUpStaleAWSChains(c.enableIPv4, c.enableIPv6)
+	if err != nil {
+		// We should not error if clean up fails since these chains don't affect the rules
+		log.Debugf("Failed to clean up stale AWS chains: %v", err)
+	}
 
 	metadataResult, err := c.awsClient.DescribeAllENIs()
 	if err != nil {
--- a/pkg/networkutils/network.go
+++ b/pkg/networkutils/network.go
@@ -152,6 +152,7 @@ type NetworkAPIs interface {
 	SetupENINetwork(eniIP string, mac string, deviceNumber int, subnetCIDR string) error
 	// UpdateHostIptablesRules updates the nat table iptables rules on the host
 	UpdateHostIptablesRules(vpcCIDRs []string, primaryMAC string, primaryAddr *net.IP, v4Enabled bool, v6Enabled bool) error
+	CleanUpStaleAWSChains(v4Enabled, v6Enabled bool) error
 	UseExternalSNAT() bool
 	GetExcludeSNATCIDRs() []string
 	GetExternalServiceCIDRs() []string
@@ -375,6 +376,51 @@ func (n *linuxNetwork) UpdateHostIptablesRules(vpcCIDRs []string, primaryMAC str
 	return n.updateHostIptablesRules(vpcCIDRs, primaryMAC, primaryAddr, v4Enabled, v6Enabled)
 }
 
+func (n *linuxNetwork) CleanUpStaleAWSChains(v4Enabled, v6Enabled bool) error {
+	ipProtocol := iptables.ProtocolIPv4
+	if v6Enabled {
+		ipProtocol = iptables.ProtocolIPv6
+	}
+
+	ipt, err := n.newIptables(ipProtocol)
+	if err != nil {
+		return errors.Wrap(err, "stale chain cleanup: failed to create iptables")
+	}
+
+	exists, err := ipt.ChainExists("nat", "AWS-SNAT-CHAIN-1")
+	if err != nil {
+		return errors.Wrap(err, "stale chain cleanup: failed to check if AWS-SNAT-CHAIN-1 exists")
+	}
+
+	if exists {
+		existingChains, err := ipt.ListChains("nat")
+		if err != nil {
+			return errors.Wrap(err, "stale chain cleanup: failed to list iptables nat chains")
+		}
+
+		for _, chain := range existingChains {
+			if !strings.HasPrefix(chain, "AWS-CONNMARK-CHAIN") && !strings.HasPrefix(chain, "AWS-SNAT-CHAIN") {
+				continue
+			}
+			parsedChain := strings.Split(chain, "-")
+			chainNum, err := strconv.Atoi(parsedChain[len(parsedChain)-1])
+			if err != nil {
+				return errors.Wrap(err, "stale chain cleanup: failed to convert string to int")
+			}
+			// Chains 1 --> x (0 indexed) will be stale
+			if chainNum > 0 {
+				// No need to clear the chain since computeStaleIptablesRules cleans up all rules already
+				log.Infof("Deleting stale chain: %s", chain)
+				err := ipt.DeleteChain("nat", chain)
+				if err != nil {
+					return errors.Wrapf(err, "stale chain cleanup: failed to delete chain %s", chain)
+				}
+			}
+		}
+	}
+	return nil
+}
+
 func (n *linuxNetwork) updateHostIptablesRules(vpcCIDRs []string, primaryMAC string, primaryAddr *net.IP, v4Enabled bool,
 	v6Enabled bool) error {
 	primaryIntf, err := findPrimaryInterfaceName(primaryMAC)
@@ -434,15 +480,13 @@ func (n *linuxNetwork) buildIptablesSNATRules(vpcCIDRs []string, primaryAddr *ne
 	log.Debugf("Total CIDRs to program - %d", len(allCIDRs))
 	// build IPTABLES chain for SNAT of non-VPC outbound traffic and excluded CIDRs
 	var chains []string
-	for i := 0; i <= len(allCIDRs); i++ {
-		chain := fmt.Sprintf("AWS-SNAT-CHAIN-%d", i)
-		log.Debugf("Setup Host Network: iptables -N %s -t nat", chain)
-		if err := ipt.NewChain("nat", chain); err != nil && !containChainExistErr(err) {
-			log.Errorf("ipt.NewChain error for chain [%s]: %v", chain, err)
-			return []iptablesRule{}, errors.Wrapf(err, "host network setup: failed to add chain")
-		}
-		chains = append(chains, chain)
+	chain := "AWS-SNAT-CHAIN-0"
+	log.Debugf("Setup Host Network: iptables -N %s -t nat", chain)
+	if err := ipt.NewChain("nat", chain); err != nil && !containChainExistErr(err) {
+		log.Errorf("ipt.NewChain error for chain [%s]: %v", chain, err)
+		return []iptablesRule{}, errors.Wrapf(err, "host network setup: failed to add chain")
 	}
+	chains = append(chains, chain)
 
 	// build SNAT rules for outbound non-VPC traffic
 	var iptableRules []iptablesRule
@@ -456,23 +500,20 @@ func (n *linuxNetwork) buildIptablesSNATRules(vpcCIDRs []string, primaryAddr *ne
 			"-m", "comment", "--comment", "AWS SNAT CHAIN", "-j", "AWS-SNAT-CHAIN-0",
 		}})
 
-	for i, cidr := range allCIDRs {
-		curChain := chains[i]
-		curName := fmt.Sprintf("[%d] AWS-SNAT-CHAIN", i)
-		nextChain := chains[i+1]
+	for _, cidr := range allCIDRs {
 		comment := "AWS SNAT CHAIN"
 		if cidr.isExclusion {
 			comment += " EXCLUSION"
 		}
-		log.Debugf("Setup Host Network: iptables -A %s ! -d %s -t nat -j %s", curChain, cidr, nextChain)
+		log.Debugf("Setup Host Network: iptables -A %s -d %s -t nat -j %s", chain, cidr, "RETURN")
 
 		iptableRules = append(iptableRules, iptablesRule{
-			name:        curName,
+			name:        chain,
 			shouldExist: !n.useExternalSNAT,
 			table:       "nat",
-			chain:       curChain,
+			chain:       chain,
 			rule: []string{
-				"!", "-d", cidr.cidr, "-m", "comment", "--comment", comment, "-j", nextChain,
+				"-d", cidr.cidr, "-m", "comment", "--comment", comment, "-j", "RETURN",
 			}})
 	}
 
@@ -494,21 +535,20 @@ func (n *linuxNetwork) buildIptablesSNATRules(vpcCIDRs []string, primaryAddr *ne
 		}
 	}
 
-	lastChain := chains[len(chains)-1]
-	iptableRules = append(iptableRules, iptablesRule{
-		name:        "last SNAT rule for non-VPC outbound traffic",
-		shouldExist: !n.useExternalSNAT,
-		table:       "nat",
-		chain:       lastChain,
-		rule:        snatRule,
-	})
-
 	snatStaleRules, err := computeStaleIptablesRules(ipt, "nat", "AWS-SNAT-CHAIN", iptableRules, chains)
 	if err != nil {
 		return []iptablesRule{}, err
 	}
 
 	iptableRules = append(iptableRules, snatStaleRules...)
+
+	iptableRules = append(iptableRules, iptablesRule{
+		name:        "last SNAT rule for non-VPC outbound traffic",
+		shouldExist: !n.useExternalSNAT,
+		table:       "nat",
+		chain:       chain,
+		rule:        snatRule,
+	})
 
 	iptableRules = append(iptableRules, iptablesRule{
 		name:        "connmark for primary ENI",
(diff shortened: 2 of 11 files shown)


pkg/networkutils/network.go
```diff
@@ -451,23 +453,20 @@ func (n *linuxNetwork) buildIptablesSNATRules(vpcCIDRs []string, primaryAddr *ne
...
 		comment := "AWS SNAT CHAIN"
 		if cidr.isExclusion {
 			comment += " EXCLUSION"
 		}
-		log.Debugf("Setup Host Network: iptables -A %s ! -d %s -t nat -j %s", curChain, cidr, nextChain)
+		log.Debugf("Setup Host Network: iptables -A %s -d %s -t nat -j %s", "AWS-SNAT-CHAIN-0", cidr, "RETURN")
 
 		iptableRules = append(iptableRules, iptablesRule{
-			name:        curName,
+			name:        "AWS-SNAT-CHAIN-0",
```
jdn5126, line 464: nit: let's use a variable name here in case we have to change it in the future

pkg/networkutils/network.go
```diff
@@ -489,22 +488,31 @@ func (n *linuxNetwork) buildIptablesSNATRules(vpcCIDRs []string, primaryAddr *ne
...
-	})
-
 	snatStaleRules, err := computeStaleIptablesRules(ipt, "nat", "AWS-SNAT-CHAIN", iptableRules, chains)
 	if err != nil {
 		return []iptablesRule{}, err
 	}
 
 	iptableRules = append(iptableRules, snatStaleRules...)
 
+	// Force delete the SNAT rule as we want this to be the last rule of the AWS-SNAT-CHAIN-0 chain
```
jdn5126, line 498: Could you expand on this? Why do we have to delete this rule? Is this from the old chain (before this change)?

pkg/networkutils/network.go
```diff
@@ -585,37 +592,23 @@ func (n *linuxNetwork) buildIptablesConnmarkRules(vpcCIDRs []string, ipt iptable
...
 		comment := "AWS CONNMARK CHAIN, VPC CIDR"
 		if excludeCIDRs.Has(cidr) {
 			comment = "AWS CONNMARK CHAIN, EXCLUDED CIDR"
 		}
-		log.Debugf("Setup Host Network: iptables -A %s ! -d %s -t nat -j %s", curChain, cidr, nextChain)
+		log.Debugf("Setup Host Network: iptables -A %s -d %s -t nat -j %s", "AWS-CONNMARK-CHAIN-0", cidr, "RETURN")
 
 		iptableRules = append(iptableRules, iptablesRule{
-			name:        curName,
+			name:        "AWS-CONNMARK-CHAIN-0",
```
jdn5126, line 603: nit: let's use variable names here

pkg/networkutils/network.go
```diff
@@ -714,14 +730,38 @@ func listCurrentIptablesRules(ipt iptableswrapper.IPTablesIface, table, chainPre
 	return toClear, nil
 }
 
+func (n *linuxNetwork) cleanOldNatChains(ipt iptableswrapper.IPTablesIface) error {
```
jdn5126, line 733: Won't this clean up the chains that are already in place? Can we do the following:
1. Clean up any rules with `AWS-SNAT-CHAIN-%d` if they exist
2. Use `AWS-SNAT-CHAIN` for the new rules only and make sure stale rules are cleaned up

jchen6585: This would only clean up chains` AWS-SNAT-CHAIN-1` to `AWS-SNAT-CHAIN-<x>` and `AWS-CONNMARK-CHAIN-1` to `AWS-CONNMARK-CHAIN<x>`, which is wanted behavior since the new rules all rely on `-0`. We skip the clean up for a chain if it ends with `-0` and has the SNAT or CONNMARK prefix.

pkg/networkutils/network.go
jchen6585: Say we had the rules:
-d <CIDR1> -j RETURN
-d <CIDR2> -j RETURN
SNAT Rule

If we were to add a CIDR, we want to ensure the SNAT Rule is after that CIDR rule. If we didn't force delete this SNAT rule and readd it, it could potentially become:
-d <CIDR1> -j RETURN
-d <CIDR2> -j RETURN
SNAT Rule
-d <CIDR3> -j RETURN
This is unwanted behavior as the CIDR3 rule won't even be checked.

pkg/networkutils/network.go
```diff
@@ -407,6 +407,10 @@ func (n *linuxNetwork) updateHostIptablesRules(vpcCIDRs []string, primaryMAC str
 		if err := n.updateIptablesRules(iptablesConnmarkRules, ipt); err != nil {
 			return err
 		}
+		err = n.cleanOldNatChains(ipt)
```
jdn5126, line 410: nit: can you add a comment that this cleans up chains from the previous implementation that may have been missed by the reconciliation logic in `updateIptablesRules`?

pkg/networkutils/network.go
```diff
@@ -451,23 +453,22 @@ func (n *linuxNetwork) buildIptablesSNATRules(vpcCIDRs []string, primaryAddr *ne
 			"-m", "comment", "--comment", "AWS SNAT CHAIN", "-j", "AWS-SNAT-CHAIN-0",
 		}})
 
-	for i, cidr := range allCIDRs {
-		curChain := chains[i]
-		curName := fmt.Sprintf("[%d] AWS-SNAT-CHAIN", i)
-		nextChain := chains[i+1]
+	for _, cidr := range allCIDRs {
+		curChain := "AWS-SNAT-CHAIN-0"
+		curName := "AWS-SNAT-CHAIN-0"
```
jdn5126, line 458: nit: these should be defined outside the for loop since they are the same for each iteration

pkg/networkutils/network.go
```diff
@@ -451,23 +453,22 @@ func (n *linuxNetwork) buildIptablesSNATRules(vpcCIDRs []string, primaryAddr *ne
...
 		comment := "AWS SNAT CHAIN"
 		if cidr.isExclusion {
 			comment += " EXCLUSION"
 		}
-		log.Debugf("Setup Host Network: iptables -A %s ! -d %s -t nat -j %s", curChain, cidr, nextChain)
+		log.Debugf("Setup Host Network: iptables -A %s -d %s -t nat -j %s", "AWS-SNAT-CHAIN-0", cidr, "RETURN")
```
jdn5126, line 463: nit: use variable

pkg/networkutils/network.go
jdn5126: Got it, this is order within the chain, thanks for explaining. I suppose this can have an impact to existing traffic (while the rules are being updated), so we should think if there is any way to delete the rule last

pkg/networkutils/network.go
```diff
@@ -489,22 +490,31 @@ func (n *linuxNetwork) buildIptablesSNATRules(vpcCIDRs []string, primaryAddr *ne
...
 	snatStaleRules, err := computeStaleIptablesRules(ipt, "nat", "AWS-SNAT-CHAIN", iptableRules, chains)
 	if err != nil {
 		return []iptablesRule{}, err
 	}
 
 	iptableRules = append(iptableRules, snatStaleRules...)
 
+	// Force delete the SNAT rule as we want this to be the last rule of the AWS-SNAT-CHAIN-0 chain
+	iptableRules = append(iptableRules, iptablesRule{
+		name:        "last SNAT rule for non-VPC outbound traffic",
```
jdn5126, line 502: nit: I think the name does not matter, so we can add a "(force delete)" here so we can distinguish it from the rule added below?

pkg/networkutils/network.go
```diff
@@ -585,37 +594,25 @@ func (n *linuxNetwork) buildIptablesConnmarkRules(vpcCIDRs []string, ipt iptable
 			"-j", "AWS-CONNMARK-CHAIN-0",
 		}})
 
-	for i, cidr := range allCIDRs {
-		curChain := chains[i]
-		curName := fmt.Sprintf("[%d] AWS-SNAT-CHAIN", i)
-		nextChain := chains[i+1]
+	for _, cidr := range allCIDRs {
+		curChain := "AWS-CONNMARK-CHAIN-0"
+		curName := "AWS-CONNMARK-CHAIN-0"
```
jdn5126, line 599: nit: define variables outside of for loop

pkg/networkutils/network.go
```diff
@@ -585,37 +594,25 @@ func (n *linuxNetwork) buildIptablesConnmarkRules(vpcCIDRs []string, ipt iptable
...
 		comment := "AWS CONNMARK CHAIN, VPC CIDR"
 		if excludeCIDRs.Has(cidr) {
 			comment = "AWS CONNMARK CHAIN, EXCLUDED CIDR"
 		}
-		log.Debugf("Setup Host Network: iptables -A %s ! -d %s -t nat -j %s", curChain, cidr, nextChain)
+		log.Debugf("Setup Host Network: iptables -A %s -d %s -t nat -j %s", "AWS-CONNMARK-CHAIN-0", cidr, "RETURN")
```
jdn5126, line 604: nit: use variable

pkg/networkutils/network.go
```diff
@@ -647,14 +644,37 @@ func (n *linuxNetwork) buildIptablesConnmarkRules(vpcCIDRs []string, ipt iptable
 	}
 	iptableRules = append(iptableRules, connmarkStaleRules...)
 
+	// Force delete the CONNMARK rule as we want this to be the last rule of the AWS-CONNMARK-CHAIN-0 chain
+	iptableRules = append(iptableRules, iptablesRule{
+		name:        "connmark rule for external outbound traffic",
```
jdn5126, line 649: nit: same here, let's add "(force delete)" to the name for traceability

pkg/networkutils/network.go
jdn5126: Got it, let's add that as a comment above this function. It cleans up rules from previous implementations of this chain

jdn5126 (approved):

pkg/networkutils/network.go
jchen6585: We can potentially skip this chain cleanup as the rules are all cleaned up properly, so these existing chains won't affect any iptable rules. Upgrading/downgrading without chain cleanup was tested and there was no breakage. These existing unused chains will be auto-cleaned up during node restart.

test/framework/resources/k8s/manifest/container.go
```diff
@@ -52,7 +52,7 @@ func NewCurlContainer() *Container {
 func NewTestHelperContainer(testImageRegistry string) *Container {
 	return &Container{
 		name:            "test-helper",
-		image:           utils.GetTestImage(testImageRegistry, utils.TestAgentImage),
+		image:           "public.ecr.aws/q1l2n4k8/vpc-cni-temp:agent3",
```
jchen6585, line 55: This is hardcoded for now, if test/agent code changes look good will revert this and push a new image to ecr

test/agent/cmd/snat-utils/main.go
```diff
@@ -76,30 +76,40 @@ func validateIPTableRules(randomizedSNATValue string, numOfCidrs int) error {
...
+			}
+			if !foundNextChain {
+				return fmt.Errorf("failed: AWS-SNAT chain broken for %s", currChain)
 			}
 		}
-		if foundNextChain == false {
-			return fmt.Errorf("failed: AWS-SNAT chain broken for %s", currChain)
-		}
+	} else {
+		lastChain = "AWS-SNAT-CHAIN-0"
```
jdn5126, line 108: The old logic counted the number of `AWS-SNAT-CHAIN-%d` chains based on the number of CIDRs in the VPC. We should still do that check, but by counting the number of entries in the `AWS-SNAT-CHAIN-0` chain. That check seems like it can just be added below

jayanthvn:
This will lead to non-zero chains being in iptables after upgrades even though we are not using them. I think it is better to delete the AWS chains during nodeInit and we anyway rebuild the chains so should be safe... Let's discuss it offline.

pkg/ipamd/ipamd.go
```diff
@@ -391,10 +391,15 @@ func (c *IPAMContext) nodeInit() error {
 	}
 
 	primaryENIMac := c.awsClient.GetPrimaryENImac()
-	err = c.networkClient.SetupHostNetwork(vpcV4CIDRs, primaryENIMac, &primaryV4IP, c.enablePodENI, c.enableIPv4, c.enableIPv6)
+	ipt, err := c.networkClient.SetupHostNetwork(vpcV4CIDRs, primaryENIMac, &primaryV4IP, c.enablePodENI, c.enableIPv4, c.enableIPv6)
 	if err != nil {
 		return errors.Wrap(err, "ipamd init: failed to set up host network")
 	}
+	err = c.networkClient.CleanUpStaleAWSChains(ipt)
```
jdn5126, line 398: I see all of the other `networkClient` functions that modify iptables rules calling `n.newIptables(ipProtocol)`. Why do we pass it as an argument here?

jdn5126 (commented):
One question, otherwise all changes look good to me

pkg/ipamd/ipamd.go
jchen6585: I thought it would be better to pass down that variable. If it is unnecessary I can also call `n.newIptables(ipProtocol)` in the clean up function.

pkg/networkutils/network.go
```diff
@@ -375,6 +376,44 @@ func (n *linuxNetwork) UpdateHostIptablesRules(vpcCIDRs []string, primaryMAC str
...
+		}
+		parsedChain := strings.Split(chain, "-")
+		chainNum, err := strconv.Atoi(parsedChain[len(parsedChain)-1])
+		if err != nil {
+			return errors.Wrap(err, "stale chain cleanup: failed to convert string to int")
+		}
+		// Chains 1 --> x (0 indexed) will be stale
+		if chainNum > 0 {
+			// No need to clear the chain since computeStaleIptablesRules cleans up all rules already
+			err := ipt.DeleteChain("nat", chain)
```
jdn5126, line 407: nit: let's log the name of the chain being deleted

pkg/networkutils/network.go
```diff
@@ -375,6 +376,44 @@ func (n *linuxNetwork) UpdateHostIptablesRules(vpcCIDRs []string, primaryMAC str
...
+			if err != nil {
+				return errors.Wrapf(err, "stale chain cleanup: failed to delete chain %s", chain)
+			}
+		}
+	}
+	return nil
+
```
jdn5126, line 414: nit: empty line can be removed

pkg/networkutils/network.go
```diff
@@ -494,22 +530,21 @@ func (n *linuxNetwork) buildIptablesSNATRules(vpcCIDRs []string, primaryAddr *ne
...
 		return []iptablesRule{}, err
 	}
 
 	iptableRules = append(iptableRules, snatStaleRules...)
 
+	iptableRules = append(iptableRules, iptablesRule{
+		name:        "last SNAT rule for non-VPC outbound traffic",
+		shouldExist: !n.useExternalSNAT,
+		table:       "nat",
+		chain:       "AWS-SNAT-CHAIN-0",
```
jdn5126, line 544: nit: we should use the variable for chain name here

pkg/networkutils/network.go
```diff
@@ -456,23 +493,22 @@ func (n *linuxNetwork) buildIptablesSNATRules(vpcCIDRs []string, primaryAddr *ne
 			"-m", "comment", "--comment", "AWS SNAT CHAIN", "-j", "AWS-SNAT-CHAIN-0",
 		}})
 
-	for i, cidr := range allCIDRs {
-		curChain := chains[i]
-		curName := fmt.Sprintf("[%d] AWS-SNAT-CHAIN", i)
-		nextChain := chains[i+1]
+	curChain := "AWS-SNAT-CHAIN-0"
+	curName := "AWS-SNAT-CHAIN-0"
```
jdn5126, line 497: nit: these variables are unnecessary when we can just use `chain` defined above. You can rename it to `chainName` if it is easier

pkg/networkutils/network.go
```diff
@@ -590,37 +624,25 @@ func (n *linuxNetwork) buildIptablesConnmarkRules(vpcCIDRs []string, ipt iptable
 			"-j", "AWS-CONNMARK-CHAIN-0",
 		}})
 
-	for i, cidr := range allCIDRs {
-		curChain := chains[i]
-		curName := fmt.Sprintf("[%d] AWS-SNAT-CHAIN", i)
-		nextChain := chains[i+1]
+	curChain := "AWS-CONNMARK-CHAIN-0"
+	curName := "AWS-CONNMARK-CHAIN-0"
```
jdn5126, line 628: Same here, these variables are unnecessary with `chain` defined above

pkg/networkutils/network.go
```diff
@@ -652,14 +674,24 @@ func (n *linuxNetwork) buildIptablesConnmarkRules(vpcCIDRs []string, ipt iptable
 	}
 	iptableRules = append(iptableRules, connmarkStaleRules...)
 
+	iptableRules = append(iptableRules, iptablesRule{
+		name:        "connmark rule for external outbound traffic",
+		shouldExist: !n.useExternalSNAT,
+		table:       "nat",
+		chain:       "AWS-CONNMARK-CHAIN-0",
```
jdn5126, line 681: nit: use `chain` variable

jdn5126 (commented):
Few nits. I think we are good to update the test agent image in this PR now, assuming your testing shows everything passing

jdn5126 (dismissed):

pkg/networkutils/network.go
```diff
@@ -375,6 +376,44 @@ func (n *linuxNetwork) UpdateHostIptablesRules(vpcCIDRs []string, primaryMAC str
...
+	if v6Enabled {
+		ipProtocol = iptables.ProtocolIPv6
+	}
+
+	ipt, err := n.newIptables(ipProtocol)
+	if err != nil {
+		return errors.Wrap(err, "stale chain cleanup: failed to create iptables")
+	}
+
+	existingChains, err := ipt.ListChains("nat")
```
jayanthvn, line 390: This would get all chains and we walk thru all and skip the ones without AWS-* right..I was thinking if we can first check https://github.com/coreos/go-iptables/blob/8c6a487c56db75caf04fc043d562ea09b17315be/iptables/iptables.go#L319 if AWS-SNAT-CHAIN1 exists and then do ahead of ListChains and deleting..So we can avoid list when there is only CHAIN0..

```
iptables -t nat -S AWS-SNAT-CHAIN-1 1
-A AWS-SNAT-CHAIN-1 ! -o vlan+ -m comment --comment "AWS, SNAT" -m addrtype ! --dst-type LOCAL -j SNAT --to-source 192.168.95.7 --random-fully
```

jchen6585: Ack

jayanthvn (approved):
lgtm

PR #2700, opened 2023-12-04
Simplify IPv6 subnet gateway calculation

anguslees:
**What type of PR is this?**
cleanup

On AWS VPC, the IPv6 gateway address is [now always available at FE80:EC2::1](https://aws.amazon.com/about-aws/whats-new/2022/11/ipv6-subnet-default-gateway-router-multiple-addresses/).  This PR replaces the previous RouterAdvertisement-based approach with this hardcoded link-local address.  Simpler, avoids delay for RA, no chance of runtime failures, and removes the previous restriction that all IPv6 ENIs had to be on the same subnet.

None 😄 

Requires VPC functionality launched in Nov 2022: https://aws.amazon.com/about-aws/whats-new/2022/11/ipv6-subnet-default-gateway-router-multiple-addresses/

no/not-tested

no

no

### DIFF
--- a/pkg/networkutils/network.go
+++ b/pkg/networkutils/network.go
@@ -954,23 +954,10 @@ func linkByMac(mac string, netLink netlinkwrapper.NetLink, retryInterval time.Du
 	}
 }
 
-// For IPv6, derive gateway address from primary ENI's RA route. Note that in IPv6, all ENIs must be in the same subnet.
-func GetIPv6Gateway() (net.IP, error) {
-	primaryEni, err := netlink.LinkByName("eth0")
-	if err != nil {
-		return nil, errors.Wrapf(err, "GetIPv6Gateway failed to get primary ENI")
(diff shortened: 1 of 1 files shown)


anguslees:
Hey!  No urgency at all.  I was just wondering if that VPC feature ever launched, discovered it had, and so figured I would update one of the places I knew could be simplified.  If anyone on EKS team is bored, there's a similar case in EKS control plane apiserver ENI setup (iirc).

Feels good to close the loop, since EKS IPv6 was one of the customer experiences that motivated this VPC gateway simplification :)

dims:
@anguslees GUS!!!

jdn5126:
This is awesome, thanks for this @anguslees! I created https://github.com/aws/amazon-vpc-cni-k8s/pull/2703 to simplify the IPv6 gateway usage for Security Groups for Pods

PR #2703, opened 2023-12-05
Simplify IPv6 Gateway Calculation

jdn5126:
**What type of PR is this?**
cleanup

N/A

This PR takes the finding in https://github.com/aws/amazon-vpc-cni-k8s/pull/2700 and applies it to simplify IPv6 gateway calculation for IPv6 Security Groups for Pods support. The IPv6 gateway for a subnet can always be reached from the node via `fe80:ec2::1`.

N/A

Manually verified that all integration tests pass against an IPv6 cluster, including the `pod-eni` test suite.

No

No, Yes

No

No

```release-note

```

### DIFF
--- a/pkg/networkutils/network.go
+++ b/pkg/networkutils/network.go
@@ -149,7 +149,7 @@ type NetworkAPIs interface {
 	SetupHostNetwork(vpcCIDRs []string, primaryMAC string, primaryAddr *net.IP, enablePodENI bool,
 		v4Enabled bool, v6Enabled bool) error
 	// SetupENINetwork performs ENI level network configuration. Not needed on the primary ENI
-	SetupENINetwork(eniIP string, mac string, deviceNumber int, subnetCIDR string) (net.IP, error)
+	SetupENINetwork(eniIP string, mac string, deviceNumber int, subnetCIDR string) error
 	// UpdateHostIptablesRules updates the nat table iptables rules on the host
 	UpdateHostIptablesRules(vpcCIDRs []string, primaryMAC string, primaryAddr *net.IP, v4Enabled bool, v6Enabled bool) error
 	UseExternalSNAT() bool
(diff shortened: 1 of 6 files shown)


achevuru (approved):
lgtm

pkg/networkutils/network.go
```diff
@@ -1121,30 +1112,30 @@ func setupENINetwork(eniIP string, eniMAC string, deviceNumber int, eniSubnetCID
...
-	return gw, nil
+	return nil
 }
 
 // For IPv6 strict mode, ICMPv6 packets from the gateway must lookup in the local routing table so that branch interfaces can resolve their gateway.
-func (n *linuxNetwork) UpdateIPv6GatewayRule(gw *net.IP) error {
+func (n *linuxNetwork) createIPv6GatewayRule() error {
 	gatewayRule := n.netLink.NewRule()
-	gatewayRule.Src = &net.IPNet{IP: *gw, Mask: net.CIDRMask(128, 128)}
+	gatewayRule.Src = &net.IPNet{IP: GetIPv6Gateway(), Mask: net.CIDRMask(128, 128)}
```
anguslees, line 1124: I'm curious: Was this sufficient, during your testing?

If I remember my AWS-internal discussion with the VPC folks leading up to this FE80:EC2::1 launch, the idea was that the gateway would _respond_ to these additional addresses, but still use the original SLAAC address for traffic _initiated_ from the gateway (eg router advertisements).  This was an important detail, to avoid changing any user-visible behaviour.

I'm not sure what "can resolve their gateway" includes exactly.  I _think_ what you've done in this PR is break router-initiated RAs, but (continue to) allow explicit host-initiated NDs.  If so, I think one of the following is true:
1. sgpp needs ICMPv6, and this breaks it, and we need to fix/revert this line. 👎 
2. sgpp needs ICMPv6, and works with this change.  Best answer if true, but we might need to expand this rule to cover all the other variants of gateway address ($subnet::1, original SLAAC address). 👍
3. sgpp no longer needs ICMPv6 (RAs?) now that we're hardcoding fe80:ec2::1 elsewhere, and ICMPv6 to/from gateway is broken but we don't care.

If it's this third case, I think we should remove this `createIPv6GatewayRule()` function entirely.

PR #2708, opened 2023-12-11
No need to set accept_ra or accept_redirects for non-primary interfaces

jdn5126:
**What type of PR is this?**
cleanup

N/A

This PR removes the IPv6 sysctl setup done in the VPC CNI init container for `accept_ra` and `accept_redirects`. Our IPv6 implementation does not rely on RA for non-primary interfaces, so these sysctls do not need to be set.

N/A

Verified that the `pod-eni` integration test suite passes with and without these sysctls. Also verified that the RA route getting installed for the trunk ENI has no impact.

No

No, Yes

No

No

```release-note

```


dims (approved):
LGTM

PR #2714, opened 2023-12-15
Enhance subnet selection

jchen6585:
**What type of PR is this?**

feature

N/A

IP exhaustion in IPv4 is a prevalent issue. This PR helps with this by creating ENIs using all subnets in the VPC with the correct tags.

**Integration**
- CNI: Passing
Screenshot 2024-03-01 at 11 43 18 AM

- IPAMD: Passing
Screenshot 2023-12-14 at 2 42 01 PM

- Custom Networking: Passing
Screenshot 2023-12-14 at 3 10 57 PM

- IPv6: Passing
Screenshot 2023-12-15 at 12 17 10 PM

- ENI Subnet Selection (New Suite): Passing
Screenshot 2024-03-13 at 4 10 02 PM

**Upgrade/Downgrade**

- Upgrading and downgrading won't break `aws-node` pod
- Older cni version can delete ENIs created from the newer version

**Performance Test**
- 130 Pods
enhanced-subnet-selection-130

- 730 pods
enhanced-subnet-selection-730

- 5000 pods
enhanced-subnet-selection-5000

Yes

No, Yes

No

Yes

```release-note
All subnets tagged with `kubernetes.io*` or `*cluster-name` will be considered for eni allocation.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -545,6 +545,15 @@ You can use the below command to enable `DISABLE_TCP_EARLY_DEMUX` to `true` -
 kubectl patch daemonset aws-node -n kube-system -p '{"spec": {"template": {"spec": {"initContainers": [{"env":[{"name":"DISABLE_TCP_EARLY_DEMUX","value":"true"}],"name":"aws-vpc-cni-init"}]}}}}'
 ```
 
+#### `ENABLE_SUBNET_DISCOVERY` (v1.18.0+)
+
+Type: Boolean as a String
+
+Default: `true`
+
+Subnet discovery is enabled by default. VPC-CNI will pick the subnet with the most number of free IPs from the nodes' VPC/AZ to create the secondary ENIs. The subnets considered are the subnet the node is created in and subnets tagged with `kubernetes.io/role/cni`.
+If `ENABLE_SUBNET_DISCOVERY` is set to `false` or if DescribeSubnets fails due to IAM permissions, all secondary ENIs will be created in the subnet the node is created in.
+
 #### `ENABLE_PREFIX_DELEGATION` (v1.9.0+)
 
 Type: Boolean as a String
@@ -733,6 +742,7 @@ Note that if you set this while using Multus, you must ensure that any chained p
 This plugin interacts with the following tags on ENIs:
 
 * `cluster.k8s.amazonaws.com/name`
+* `kubernetes.io/role/cni`
 * `node.k8s.amazonaws.com/instance_id`
 * `node.k8s.amazonaws.com/no_manage`
 
@@ -740,6 +750,17 @@ This plugin interacts with the following tags on ENIs:
 
 The tag `cluster.k8s.amazonaws.com/name` will be set to the cluster name of the
 aws-node daemonset which created the ENI.
+
+#### CNI role tag
+
+The tag `kubernetes.io/role/cni` is read by the aws-node daemonset to determine
+if a secondary subnet can be used for creating secondary ENIs.
+
+This tag is not set by the cni plugin itself, but rather must be set by a user
+to indicate that a subnet can be used for secondary ENIs. Secondary subnets
+to be used must have this tag. The primary subnet (node's subnet) is not
+required to be tagged.
+
 
 #### Instance ID tag
 
--- a/charts/aws-vpc-cni/values.yaml
+++ b/charts/aws-vpc-cni/values.yaml
@@ -84,6 +84,7 @@ env:
   ENABLE_IPv4: "true"
   ENABLE_IPv6: "false"
   VPC_CNI_VERSION: "v1.16.4"
+  ENABLE_SUBNET_DISCOVERY: "true"
 
 # this flag enables you to use the match label that was present in the original daemonset deployed by EKS
 # You can then annotate and label the original aws-node resources and 'adopt' them into a helm release
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -22,6 +22,7 @@ import (
 	"net"
 	"os"
 	"regexp"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -58,6 +59,7 @@ const (
 	eniClusterTagKey        = "cluster.k8s.amazonaws.com/name"
 	additionalEniTagsEnvVar = "ADDITIONAL_ENI_TAGS"
 	reservedTagKeyPrefix    = "k8s.amazonaws.com"
+	subnetDiscoveryTagKey   = "kubernetes.io/role/cni"
 	// UnknownInstanceType indicates that the instance type is not yet supported
 	UnknownInstanceType = "vpc ip resource(eni ip limit): unknown instance type"
 
@@ -85,7 +87,7 @@ var log = logger.Get()
 // APIs defines interfaces calls for adding/getting/deleting ENIs/secondary IPs. The APIs are not thread-safe.
 type APIs interface {
 	// AllocENI creates an ENI and attaches it to the instance
-	AllocENI(useCustomCfg bool, sg []*string, subnet string, numIPs int) (eni string, err error)
+	AllocENI(useCustomCfg bool, sg []*string, eniCfgSubnet string, numIPs int) (eni string, err error)
 
 	// FreeENI detaches ENI interface and deletes it
 	FreeENI(eniName string) error
@@ -200,10 +202,12 @@ type EC2InstanceMetadataCache struct {
 	primaryENImac    string
 	availabilityZone string
 	region           string
+	vpcID            string
 
 	unmanagedENIs          StringSet
 	useCustomNetworking    bool
 	multiCardENIs          StringSet
+	useSubnetDiscovery     bool
 	enablePrefixDelegation bool
 
 	clusterName       string
@@ -347,7 +351,7 @@ func (i instrumentedIMDS) GetMetadataWithContext(ctx context.Context, p string) 
 }
 
 // New creates an EC2InstanceMetadataCache
-func New(useCustomNetworking, disableLeakedENICleanup, v4Enabled, v6Enabled bool) (*EC2InstanceMetadataCache, error) {
+func New(useSubnetDiscovery, useCustomNetworking, disableLeakedENICleanup, v4Enabled, v6Enabled bool) (*EC2InstanceMetadataCache, error) {
 	// ctx is passed to initWithEC2Metadata func to cancel spawned go-routines when tests are run
 	ctx := context.Background()
 
@@ -367,6 +371,8 @@ func New(useCustomNetworking, disableLeakedENICleanup, v4Enabled, v6Enabled bool
 	log.Debugf("Discovered region: %s", cache.region)
 	cache.useCustomNetworking = useCustomNetworking
 	log.Infof("Custom networking enabled %v", cache.useCustomNetworking)
+	cache.useSubnetDiscovery = useSubnetDiscovery
+	log.Infof("Subnet discovery enabled %v", cache.useSubnetDiscovery)
 	cache.v4Enabled = v4Enabled
 	cache.v6Enabled = v6Enabled
 
@@ -442,13 +448,21 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context) 
 	}
 	log.Debugf("%s is the primary ENI of this instance", cache.primaryENI)
 
-	// retrieve sub-id
+	// retrieve subnet-id
 	cache.subnetID, err = cache.imds.GetSubnetID(ctx, mac)
 	if err != nil {
 		awsAPIErrInc("GetSubnetID", err)
 		return err
 	}
 	log.Debugf("Found subnet-id: %s ", cache.subnetID)
+
+	// retrieve vpc-id
+	cache.vpcID, err = cache.imds.GetVpcID(ctx, mac)
+	if err != nil {
+		awsAPIErrInc("GetVpcID", err)
+		return err
+	}
+	log.Debugf("Found vpc-id: %s ", cache.vpcID)
 
 	// We use the ctx here for testing, since we spawn go-routines above which will run forever.
 	select {
@@ -730,8 +744,8 @@ func (cache *EC2InstanceMetadataCache) awsGetFreeDeviceNumber() (int, error) {
 
 // AllocENI creates an ENI and attaches it to the instance
 // returns: newly created ENI ID
-func (cache *EC2InstanceMetadataCache) AllocENI(useCustomCfg bool, sg []*string, subnet string, numIPs int) (string, error) {
-	eniID, err := cache.createENI(useCustomCfg, sg, subnet, numIPs)
+func (cache *EC2InstanceMetadataCache) AllocENI(useCustomCfg bool, sg []*string, eniCfgSubnet string, numIPs int) (string, error) {
+	eniID, err := cache.createENI(useCustomCfg, sg, eniCfgSubnet, numIPs)
 	if err != nil {
 		return "", errors.Wrap(err, "AllocENI: failed to create ENI")
 	}
@@ -803,7 +817,7 @@ func (cache *EC2InstanceMetadataCache) attachENI(eniID string) (string, error) {
 }
 
 // return ENI id, error
-func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string, subnet string, numIPs int) (string, error) {
+func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string, eniCfgSubnet string, numIPs int) (string, error) {
 	eniDescription := eniDescriptionPrefix + cache.instanceID
 	tags := map[string]string{
 		eniCreatedAtTagKey: time.Now().Format(time.RFC3339),
@@ -826,6 +840,7 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
 
 	log.Infof("Trying to allocate %d IP addresses on new ENI", needIPs)
 	log.Debugf("PD enabled - %t", cache.enablePrefixDelegation)
+
 	input := &ec2.CreateNetworkInterfaceInput{}
 
 	if cache.enablePrefixDelegation {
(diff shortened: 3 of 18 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -439,13 +441,13 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
 	}
 	log.Debugf("%s is the primary ENI of this instance", cache.primaryENI)
 
-	// retrieve sub-id
-	cache.subnetID, err = cache.imds.GetSubnetID(ctx, mac)
```
jdn5126, line 443: Do we no longer need to cache `subnetID`? If so, can we remove it from the data structure?

pkg/awsutils/awsutils.go
```diff
@@ -819,52 +821,98 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
-	input := &ec2.CreateNetworkInterfaceInput{}
 
-	if cache.enablePrefixDelegation {
-		input = &ec2.CreateNetworkInterfaceInput{
-			Description:       aws.String(eniDescription),
-			Groups:            aws.StringSlice(cache.securityGroups.SortedList()),
-			SubnetId:          aws.String(cache.subnetID),
-			TagSpecifications: tagSpec,
-			Ipv4PrefixCount:   aws.Int64(int64(needIPs)),
+	describeSubnetInput := &ec2.DescribeSubnetsInput{
```
jdn5126, line 825: We should skip the whole describe subnet block when we are using custom networking. Splitting this into its own function will make that easier

pkg/awsutils/awsutils.go
```diff
@@ -819,52 +821,98 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
+	prometheusmetrics.AwsAPILatency.WithLabelValues("DescribeSubnets", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
+	if err != nil {
+		checkAPIErrorAndBroadcastEvent(err, "ec2:DescribeSubnets")
+		awsAPIErrInc("DescribeSubnets", err)
+		prometheusmetrics.Ec2ApiErr.WithLabelValues("DescribeSubnets").Inc()
+		return "", errors.Wrap(err, "AllocENI: unable to describe subnets")
+	}
+
+	// Sort the subnet by available IP address counter (desc order) before determining subnet to use
+	sort.SliceStable(subnetResult.Subnets, func(i, j int) bool {
```
jdn5126, line 850: I'm not sure if there is a way, but we should think about whether we can cache the subnets from previous runs and avoid a describe call every time. Or at least not redo the sort every time. Anything to reduce the number of EC2 API calls

pkg/awsutils/awsutils.go
```diff
@@ -819,52 +821,98 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
-	prometheusmetrics.AwsAPILatency.WithLabelValues("CreateNetworkInterface", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
-	if err != nil {
+		start = time.Now()
+		result, err = cache.ec2SVC.CreateNetworkInterfaceWithContext(context.Background(), input)
+		prometheusmetrics.Ec2ApiReq.WithLabelValues("CreateNetworkInterface").Inc()
+		prometheusmetrics.AwsAPILatency.WithLabelValues("CreateNetworkInterface", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
+		if err == nil {
+			log.Infof("Created a new ENI: %s", aws.StringValue(result.NetworkInterface.NetworkInterfaceId))
+			return aws.StringValue(result.NetworkInterface.NetworkInterfaceId), nil
+		}
```
jdn5126, line 909: This block can just use whatever `ec2.CreateNetworkInterfaceInput` is returned by the custom networking function or subnet search function. That will make this cleaner

test/integration/eni-subnet-selection/eni_subnet_selection_test.go
```diff
@@ -0,0 +1,150 @@
...
+				expectedCount = 0
+			})
+			It(fmt.Sprintf("should not have ip addresses in CIDR range %s", cidrRangeString), func() {
+				CheckNumberOfIps(expectedCount, podList)
+			})
+		})
+	})
+})
+
+func CheckNumberOfIps(expectedCount int, podList coreV1.PodList) {
```
jdn5126, line 139: nit: instead of counting IPs, we can just make an EC2 API call to get ENIs attached to the instance and validate their subnets. That's the specific check we want for this test suite

pkg/awsutils/awsutils.go
jchen6585: Yes we no longer need it

pkg/awsutils/awsutils.go
```diff
@@ -838,33 +852,111 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
+			return "", errors.Wrap(vpcErr, "failed to get subnets in vpc")
+		}
+
+		for _, subnet := range subnetResult {
+			if *subnet.SubnetId != cache.subnetID {
+				if !validTag(subnet) {
+					continue
+				}
+			}
+
```
jdn5126, line 885: big nit: let's remove this newline to make it easier to read

pkg/awsutils/awsutils.go
```diff
@@ -838,33 +852,111 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
+			log.Infof("Creating ENI with security groups: %v in subnet: %s", aws.StringValueSlice(input.Groups), aws.StringValue(input.SubnetId))
+
+			input.SubnetId = subnet.SubnetId
+
```
jdn5126, line 889: big nit: let's remove newline to make it easier to read

pkg/awsutils/awsutils.go
```diff
@@ -838,33 +852,111 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
+				Values: []*string{aws.String(cache.availabilityZone)},
+			},
+		},
+	}
 
 	start := time.Now()
-	result, err := cache.ec2SVC.CreateNetworkInterfaceWithContext(context.Background(), input)
-	prometheusmetrics.Ec2ApiReq.WithLabelValues("CreateNetworkInterface").Inc()
-	prometheusmetrics.AwsAPILatency.WithLabelValues("CreateNetworkInterface", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
+	subnetResult, err := cache.ec2SVC.DescribeSubnetsWithContext(context.Background(), describeSubnetInput)
```
jdn5126, line 922: nit: we should investigate whether a paginated call would be better looks. Looks like this is unpaginated by default

pkg/awsutils/awsutils.go
```diff
@@ -838,33 +852,111 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
+	sort.SliceStable(subnetResult.Subnets, func(i, j int) bool {
+		return *subnetResult.Subnets[j].AvailableIpAddressCount < *subnetResult.Subnets[i].AvailableIpAddressCount
+	})
+
+	return subnetResult.Subnets, nil
+}
+
+func validTag(subnet *ec2.Subnet) bool {
+	for _, tag := range subnet.Tags {
+		if strings.HasSuffix(*tag.Key, "cluster-name") || strings.HasPrefix(*tag.Key, "kubernetes.io") {
```
jdn5126, line 942: big nit: let's switch the order here since `kubernetes.io` is more likely to be true

test/integration/eni-subnet-selection/eni_subnet_selection_test.go
```diff
@@ -0,0 +1,168 @@
...
+			container := manifest.NewNetCatAlpineContainer(f.Options.TestImageRegistry).
+				Command([]string{"sleep"}).
+				Args([]string{"3600"}).
+				Build()
+
+			deploymentBuilder := manifest.NewBusyBoxDeploymentBuilder(f.Options.TestImageRegistry).
+				Container(container).
+				Replicas(50).
+				PodLabel(podLabelKey, podLabelVal).
+				NodeSelector("topology.kubernetes.io/zone", *primaryInstance.Placement.AvailabilityZone).
```
jdn5126, line 59: shouldn't we target the node directly, as there could be multiple nodes in this availability zone

jdn5126 (commented):
Changes look good overall, just nits. Will do one more pass on the test code next

pkg/awsutils/awsutils.go
```diff
@@ -838,33 +852,109 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
+			return *result.NetworkInterface.NetworkInterfaceId, nil
 		}
-		input.SubnetId = aws.String(subnet)
+		checkAPIErrorAndBroadcastEvent(err, "ec2:CreateNetworkInterface")
+		awsAPIErrInc("CreateNetworkInterface", err)
+		prometheusmetrics.Ec2ApiErr.WithLabelValues("CreateNetworkInterface").Inc()
+		log.Errorf("Failed to CreateNetworkInterface %v for subnet %s", err, customSubnet)
 	} else {
-		log.Info("Using same config as the primary interface for the new ENI")
+		subnetResult, vpcErr := cache.getVpcSubnets()
```
jayanthvn, line 874: Won't this increase the EC2 call volume during ENI scale + node churn if done by default? Have we evaluated the percentage increase especially on XL clusters?

pkg/awsutils/awsutils.go
```diff
@@ -85,7 +86,7 @@ var log = logger.Get()
 // APIs defines interfaces calls for adding/getting/deleting ENIs/secondary IPs. The APIs are not thread-safe.
 type APIs interface {
 	// AllocENI creates an ENI and attaches it to the instance
-	AllocENI(useCustomCfg bool, sg []*string, subnet string, numIPs int) (eni string, err error)
+	AllocENI(useCustomCfg bool, sg []*string, customSubnet string, numIPs int) (eni string, err error)
```
jayanthvn, line 89: Nit: customSubnet doesn't look right to read..

pkg/awsutils/awsutils.go
```diff
@@ -838,33 +852,109 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
-	if useCustomCfg {
-		log.Info("Using a custom network config for the new ENI")
-		if len(sg) != 0 {
-			input.Groups = sg
-		} else {
-			log.Warnf("No custom networking security group found, will use the node's primary ENI's SG: %v", aws.StringValueSlice(input.Groups))
+
+	var err error
+	if cache.useCustomNetworking {
+		input = createEniUsingCustomCfg(sg, customSubnet, input)
```
jayanthvn, line 858: Nit: "Eni" -> "ENI"

pkg/awsutils/awsutils.go
```diff
@@ -838,33 +852,109 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
 	} else {
-		log.Info("Using same config as the primary interface for the new ENI")
+		subnetResult, vpcErr := cache.getVpcSubnets()
+		if vpcErr != nil {
+			return "", errors.Wrap(vpcErr, "failed to get subnets in vpc")
+		}
+
+		for _, subnet := range subnetResult {
+			if *subnet.SubnetId != cache.subnetID {
+				if !validTag(subnet) {
```
jayanthvn, line 881: What happens if we don't have the validTag at all on all subnets? Maybe someone removed it..

pkg/awsutils/awsutils.go
```diff
@@ -838,33 +852,109 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
-		} else {
-			log.Warnf("No custom networking security group found, will use the node's primary ENI's SG: %v", aws.StringValueSlice(input.Groups))
+
+	var err error
+	if cache.useCustomNetworking {
+		input = createEniUsingCustomCfg(sg, customSubnet, input)
+		log.Infof("Creating ENI with security groups: %v in subnet: %s", aws.StringValueSlice(input.Groups), aws.StringValue(input.SubnetId))
+
+		start := time.Now()
+		result, err = cache.ec2SVC.CreateNetworkInterfaceWithContext(context.Background(), input)
```
jayanthvn, line 862: we can move the CreateNetworkInterfaceWithContext and metrics to a sub function since we duplicate later..

pkg/awsutils/awsutils.go
jchen6585: Assuming the default subnet is out of IPs, it won't create an ENI and will try again later. Same as previous logic

pkg/awsutils/awsutils.go
jchen6585: Messaged offline about this

pkg/awsutils/awsutils.go
```diff
@@ -838,33 +852,109 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
-		return "", errors.Wrap(err, "failed to create network interface")
+		checkAPIErrorAndBroadcastEvent(err, "ec2:DescribeSubnets")
+		awsAPIErrInc("DescribeSubnets", err)
+		prometheusmetrics.Ec2ApiErr.WithLabelValues("DescribeSubnets").Inc()
+		return nil, errors.Wrap(err, "AllocENI: unable to describe subnets")
+	}
+
+	// Sort the subnet by available IP address counter (desc order) before determining subnet to use
+	sort.SliceStable(subnetResult.Subnets, func(i, j int) bool {
+		return *subnetResult.Subnets[j].AvailableIpAddressCount < *subnetResult.Subnets[i].AvailableIpAddressCount
```
jayanthvn, line 932: This might not work for PD completely..maybe a subnet has more available IPs but not /28 blocks i.e, more fragments..ideally for PD we should pick a subnet which has more /28 blocks i.e, contiguous 16IPs..

pkg/awsutils/awsutils.go
```diff
@@ -838,33 +852,109 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
+	})
+
+	return subnetResult.Subnets, nil
+}
+
+func validTag(subnet *ec2.Subnet) bool {
+	for _, tag := range subnet.Tags {
+		if strings.HasPrefix(*tag.Key, "kubernetes.io") || strings.HasSuffix(*tag.Key, "cluster-name") {
```
jayanthvn, line 940: Consider making these constants such as `eniNoManageTagKey = "node.k8s.amazonaws.com/no_manage"`

pkg/awsutils/awsutils.go
```diff
@@ -838,33 +852,109 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
+		if err == nil {
+			log.Infof("Created a new ENI: %s", aws.StringValue(result.NetworkInterface.NetworkInterfaceId))
+			return *result.NetworkInterface.NetworkInterfaceId, nil
 		}
-		input.SubnetId = aws.String(subnet)
+		checkAPIErrorAndBroadcastEvent(err, "ec2:CreateNetworkInterface")
+		awsAPIErrInc("CreateNetworkInterface", err)
+		prometheusmetrics.Ec2ApiErr.WithLabelValues("CreateNetworkInterface").Inc()
+		log.Errorf("Failed to CreateNetworkInterface %v for subnet %s", err, customSubnet)
 	} else {
```
jayanthvn, line 873: This is guarded a knob isn't it? I.e this should be an opt-in

pkg/awsutils/awsutils.go
ellistarn, line 940: I'm a bit concerned about the breadth of tagging here. Customers may be using kubernetes.io or cluster-name for other purposes (e.g. CPI subnets). Is there a reason we're not using a specific tag vs selecting broadly?

jchen6585: We had an internal discussion on tuesday and decided to use a specific tag. Just haven't gotten a chance to update the PR yet

pkg/awsutils/awsutils.go
```diff
@@ -838,33 +852,109 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
-		log.Info("Using same config as the primary interface for the new ENI")
+		subnetResult, vpcErr := cache.getVpcSubnets()
+		if vpcErr != nil {
+			return "", errors.Wrap(vpcErr, "failed to get subnets in vpc")
+		}
+
+		for _, subnet := range subnetResult {
```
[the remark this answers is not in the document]
jchen6585: We sort the subnets by most available IP addresses in `getVpcSubnets`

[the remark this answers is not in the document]
ellistarn: Nice!

pkg/awsutils/awsutils.go
jayanthvn: I am thinking if we can retry on remaining subnets if we hit InsufficientCidr error...that should work for both PD and secondary IP mode..

charts/aws-vpc-cni/values.yaml
```diff
@@ -84,6 +84,7 @@ env:
   ENABLE_IPv4: "true"
   ENABLE_IPv6: "false"
   VPC_CNI_VERSION: "v1.16.2"
+  ENABLE_ADV_ENI_CFG: "true"
```
jdn5126, line 87: I think we need a name other than `ENABLE_ADV_ENI_CFG` here. `ENIConfig` is too closely related to Custom Networking, so this could get confusing.

Since at the end of the day we are expanding the pool of subnets we choose from, I think `ENABLE_ADV_SUBNET_SEL` is the best option. The pitch to customers is that we will automatically start using subnets they tag, so this seems the clearest variable to denote that.

pkg/awsutils/awsutils.go
jdn5126: Agreed, that's a good plan

charts/aws-vpc-cni/values.yaml
jayanthvn: Agree, why do we call it *_ENI_CFG? Since we aren't using eniconfig right? Let's discuss internally on the naming..

pkg/awsutils/awsutils.go
```diff
@@ -367,6 +371,8 @@ func New(useCustomNetworking, disableLeakedENICleanup, v4Enabled, v6Enabled bool
 	log.Debugf("Discovered region: %s", cache.region)
 	cache.useCustomNetworking = useCustomNetworking
 	log.Infof("Custom networking enabled %v", cache.useCustomNetworking)
+	cache.useAdvancedENIConfig = useAdvancedENIConfig
+	log.Infof("Advanced ENI Config enabled %v", cache.useAdvancedENIConfig)
```
jayanthvn, line 375: Same as above, it is bit confusing when we call it advanced ENI config...

pkg/awsutils/awsutils.go
```diff
@@ -826,6 +840,8 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
 
 	log.Infof("Trying to allocate %d IP addresses on new ENI", needIPs)
 	log.Debugf("PD enabled - %t", cache.enablePrefixDelegation)
+
+	log.Info("Using same config as the primary interface for the new ENI")
```
jayanthvn, line 844: Lets clarify what config or remove this line...

pkg/awsutils/awsutils.go
```diff
@@ -845,33 +861,117 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
+			return networkInterfaceID, nil
+		}
+	} else {
+		if cache.useAdvancedENIConfig {
+			subnetResult, vpcErr := cache.getVpcSubnets()
+			if vpcErr != nil {
+				return "", errors.Wrap(vpcErr, "failed to get subnets in vpc")
+			}
+
+			for _, subnet := range subnetResult {
```
jayanthvn, line 882: This can be simplified, something like -

if (enhanced_subnet_selection) {
    1. get a subnet
    2. Set input.SubnetID
}

Then call ->
networkInterfaceID, err = cache.tryCreateNetworkInterface(input)

pkg/awsutils/awsutils.go
jayanthvn: Are we checking `InsufficientCidr` ?

README.md
```diff
@@ -545,6 +545,15 @@ You can use the below command to enable `DISABLE_TCP_EARLY_DEMUX` to `true` -
 kubectl patch daemonset aws-node -n kube-system -p '{"spec": {"template": {"spec": {"initContainers": [{"env":[{"name":"DISABLE_TCP_EARLY_DEMUX","value":"true"}],"name":"aws-vpc-cni-init"}]}}}}'
 ```
 
+#### `ENABLE_SUBNET_DISCOVERY` (v1.18.0+)
```
jayanthvn, line 548: Let's also add here - https://github.com/aws/amazon-vpc-cni-k8s?tab=readme-ov-file#eni-tags-related-to-allocation

pkg/awsutils/awsutils.go
```diff
@@ -58,6 +59,7 @@ const (
 	eniClusterTagKey        = "cluster.k8s.amazonaws.com/name"
 	additionalEniTagsEnvVar = "ADDITIONAL_ENI_TAGS"
 	reservedTagKeyPrefix    = "k8s.amazonaws.com"
+	subnetSelectionTagKey   = "kubernetes.io/role/cni"
```
jayanthvn, line 62: Nit: `subnetDiscoveryTagKey`

pkg/awsutils/awsutils.go
```diff
@@ -845,33 +860,118 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
+			},
+			{
+				Name:   aws.String("availability-zone"),
+				Values: []*string{aws.String(cache.availabilityZone)},
+			},
+		},
+	}
+
+	start := time.Now()
+	subnetResult, err := cache.ec2SVC.DescribeSubnetsWithContext(context.Background(), describeSubnetInput)
```
jchen6585, line 921: If error is due to missing permissions fall back to default subnet and provide a warning

README.md
```diff
@@ -545,6 +545,15 @@ You can use the below command to enable `DISABLE_TCP_EARLY_DEMUX` to `true` -
...
+
+Type: Boolean as a String
+
+Default: `true`
+
+Subnet discovery is enabled by default. VPC-CNI will pick the subnet with the most number of free IPs from the nodes' VPC/AZ to create the secondary ENIs. The subnets considered are the subnet the node is created in and subnets tagged with `kubernetes.io/role/cni`.
+If `ENABLE_SUBNET_DISCOVERY` is set to `false`, all secondary ENIs will be created in the subnet the node is created in.
```
jayanthvn, line 555: If `ENABLE_SUBNET_DISCOVERY` is set to `false` or if DescribeSubnets fails due to IAM permissions..

README.md
```diff
@@ -741,6 +751,17 @@ This plugin interacts with the following tags on ENIs:
 The tag `cluster.k8s.amazonaws.com/name` will be set to the cluster name of the
 aws-node daemonset which created the ENI.
 
+#### CNI role tag
+
+The tag `kubernetes.io/role/cni` is read by the aws-node daemonset to determine
+if a secondary subnet can be used for creating secondary ENIs.
+
+This tag is not set by the cni plugin itself, but rather may be set by a user
```
jayanthvn, line 759: rather may -> rather must

pkg/awsutils/awsutils.go
```diff
@@ -845,33 +860,123 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
+		log.Infof("Creating ENI with security groups: %v in subnet: %s", aws.StringValueSlice(input.Groups), aws.StringValue(input.SubnetId))
+
+		networkInterfaceID, err = cache.tryCreateNetworkInterface(input)
+		if err == nil {
+			return networkInterfaceID, nil
+		}
+	} else {
+		if cache.useSubnetDiscovery {
+			subnetResult, vpcErr := cache.getVpcSubnets()
+			if vpcErr != nil {
```
jayanthvn, line 877: Maybe as a fast follow up lets restructure the code..

1. Get subnet -> Either primary ENI, ENIConfig or pick a subnet
2. Get SG -> Either primary ENI SG or ENIConfig SG
3. Modify the "input" with the new subnet ID and SG ID
4. Call tryCreateNetworkInterface

test/integration/README.md
```diff
@@ -81,6 +81,13 @@ Custom networking tests validate use of the `AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG`
 Test info:
   - Pass `custom-networking-cidr-range` flag with *allowed* VPC CIDR that does not conflict with an existing one. So if existing VPC CIDR is `192.168.0.0/16`, you can use `custom-networking-cidr-range=100.64.0.0/16`. You can go to your cluster VPC to check existing/allowed CIDRs.
 
+### ENI Subnet Selection (eni_subnet_selection)
```
jayanthvn, line 84: Need to update this for subnet discovery..

achevuru (approved):

test/integration/eni-subnet-discovery/eni_subnet_discovery_test.go
```diff
@@ -0,0 +1,331 @@
...
+			})
+			Context("using a subnet tagged with kubernetes.io/role/cn", func() {
+				BeforeEach(func() {
+					By("Tagging kubernetes.io/role/cn to subnet")
+					_, err = f.CloudServices.EC2().
+						CreateTags(
+							[]string{createdSubnet},
+							[]*ec2.Tag{
+								{
+									Key:   aws.String("kubernetes.io/role/cn"),
```
alex-zige, line 199: `kubernetes.io/role/cni`instead of `cn`?

orsenthil: I think, it is alright. The tagging and un-tagging was consistent in this test.

Issue #2711, opened 2023-12-14
Refactor iptable mock


jchen6585:
**What would you like to be added**:
Issue to keep track of refactoring iptable mock

**Why is this needed**:
Currently, our mock has some empty functions and only map the chain to rules that start with `-A`. We specifically should add chain creation rules to the map as well `-N`. This leaves some gaps in our test coverage such as testing https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/networkutils/network.go#L379

jdn5126:
Closing as fixed by #2721

PR #2721, opened 2023-12-20
Iptables mock

jchen6585:
**What type of PR is this?**

enhancement

https://github.com/aws/amazon-vpc-cni-k8s/issues/2711

Refactors our iptables mock to have more coverage and mimic code behavior (e.g. differentiating append and insert)

`make unit-test`

No

No, No

No

No

```release-note

```

### DIFF
--- a/pkg/iptableswrapper/mocks/iptables_maps.go
+++ b/pkg/iptableswrapper/mocks/iptables_maps.go
@@ -16,6 +16,7 @@ package mock_iptableswrapper
 import (
 	"fmt"
 	"reflect"
+	"slices"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -44,7 +45,12 @@ func (ipt *MockIptables) Insert(table, chain string, pos int, rulespec ...string
 	if ipt.DataplaneState[table] == nil {
 		ipt.DataplaneState[table] = map[string][][]string{}
 	}
-	ipt.DataplaneState[table][chain] = append(ipt.DataplaneState[table][chain], rulespec)
+	if len(ipt.DataplaneState[table][chain]) == pos-1 {
+		ipt.DataplaneState[table][chain] = append(ipt.DataplaneState[table][chain], rulespec)
+	} else {
+		ipt.DataplaneState[table][chain] = append(ipt.DataplaneState[table][chain][:pos], ipt.DataplaneState[table][chain][pos-1:]...)
+		ipt.DataplaneState[table][chain][pos] = rulespec
+	}
 	return nil
 }
 
@@ -91,6 +97,10 @@ func (ipt *MockIptables) List(table, chain string) ([]string, error) {
 	var chains []string
 	chainContents := ipt.DataplaneState[table][chain]
 	for _, ruleSpec := range chainContents {
+		if slices.Contains(ruleSpec, "-N") {
+			chains = append(chains, strings.Join(ruleSpec, " "))
+			continue
+		}
 		sanitizedRuleSpec := []string{"-A", chain}
 		for _, item := range ruleSpec {
 			if strings.Contains(item, " ") {
@@ -101,10 +111,15 @@ func (ipt *MockIptables) List(table, chain string) ([]string, error) {
 		chains = append(chains, strings.Join(sanitizedRuleSpec, " "))
 	}
 	return chains, nil
-
 }
 
 func (ipt *MockIptables) NewChain(table, chain string) error {
+	exists, _ := ipt.ChainExists(table, chain)
+	if exists {
+		return errors.New("Chain already exists")
+	}
+	// Creating a new chain adds a -N chain rule to iptables
+	ipt.Append(table, chain, "-N", chain)
 	return nil
 }
 
@@ -113,6 +128,12 @@ func (ipt *MockIptables) ClearChain(table, chain string) error {
 }
 
 func (ipt *MockIptables) DeleteChain(table, chain string) error {
+	// More than just the create chain rule
+	if len(ipt.DataplaneState[table][chain]) > 1 {
+		err := fmt.Sprintf("Chain %s is not empty", chain)
+		return errors.New(err)
+	}
+	delete(ipt.DataplaneState[table], chain)
 	return nil
 }
 
--- a/pkg/networkutils/network_test.go
+++ b/pkg/networkutils/network_test.go
@@ -363,6 +363,10 @@ func TestSetupHostNetworkNodePortEnabledAndSNATDisabled(t *testing.T) {
 				},
 			},
 		},
+		"nat": {
+			"AWS-SNAT-CHAIN-0":     [][]string{{"-N", "AWS-SNAT-CHAIN-0"}},
+			"AWS-CONNMARK-CHAIN-0": [][]string{{"-N", "AWS-CONNMARK-CHAIN-0"}},
+		},
 	}, mockIptables.(*mock_iptables.MockIptables).DataplaneState)
 }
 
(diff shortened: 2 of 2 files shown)


pkg/iptableswrapper/mocks/iptables_maps.go
```diff
@@ -41,10 +41,16 @@ func (ipt *MockIptables) Exists(table, chainName string, rulespec ...string) (bo
 }
 
 func (ipt *MockIptables) Insert(table, chain string, pos int, rulespec ...string) error {
+	// We only have -N chainName in the map for chains we manually create
+	if !strings.HasPrefix(chain, "AWS-SNAT-CHAIN") && !strings.HasPrefix(chain, "AWS-CONNMARK-CHAIN") {
+		pos = pos - 1
```
jdn5126, line 46: This looks strange. It feels like it should be the caller's responsibility to pass the right position and we should not have any logic based on the string name of the chain

pkg/iptableswrapper/mocks/iptables_maps.go
```diff
@@ -41,10 +41,16 @@ func (ipt *MockIptables) Exists(table, chainName string, rulespec ...string) (bo
...
+	}
 	if ipt.DataplaneState[table] == nil {
 		ipt.DataplaneState[table] = map[string][][]string{}
 	}
-	ipt.DataplaneState[table][chain] = append(ipt.DataplaneState[table][chain], rulespec)
+	ipt.DataplaneState[table][chain] = append(ipt.DataplaneState[table][chain], []string{})
+	copy(ipt.DataplaneState[table][chain][pos+1:], ipt.DataplaneState[table][chain][pos:])
+	ipt.DataplaneState[table][chain][pos] = rulespec
```
jdn5126, line 53: The `copy` is a bit confusing to read. A simple `append` should work here. Can we follow the logic of something like https://stackoverflow.com/questions/46128016/insert-a-value-in-a-slice-at-a-given-index

pkg/iptableswrapper/mocks/iptables_maps.go
```diff
@@ -91,6 +97,10 @@ func (ipt *MockIptables) List(table, chain string) ([]string, error) {
 	var chains []string
 	chainContents := ipt.DataplaneState[table][chain]
 	for _, ruleSpec := range chainContents {
+		if len(ruleSpec) == 1 {
```
jdn5126, line 100: Can you add a comment explaining why this is needed?

pkg/iptableswrapper/mocks/iptables_maps.go
```diff
@@ -113,6 +127,12 @@ func (ipt *MockIptables) ClearChain(table, chain string) error {
 }
 
 func (ipt *MockIptables) DeleteChain(table, chain string) error {
+	// More than just the create chain rule
+	if len(ipt.DataplaneState[table][chain]) > 1 {
+		err := fmt.Sprintf("Chain %s is not empty", chain)
```
jdn5126, line 132: Why wouldn't we delete the chain if it has rules in it?

pkg/networkutils/network_test.go
```diff
@@ -363,6 +363,10 @@ func TestSetupHostNetworkNodePortEnabledAndSNATDisabled(t *testing.T) {
 				},
 			},
 		},
+		"nat": {
+			"AWS-SNAT-CHAIN-0":     [][]string{{"AWS-SNAT-CHAIN-0"}},
```
jdn5126, line 367: Is this first entry with the chain name supposed to represent the `-N` entry?

pkg/networkutils/network_test.go
```diff
@@ -838,6 +868,82 @@ func TestUpdateHostIptablesRules(t *testing.T) {
...
+
+		netLink: mockNetLink,
+		ns:      mockNS,
+		newIptables: func(iptables.Protocol) (iptableswrapper.IPTablesIface, error) {
+			return mockIptables, nil
+		},
+	}
+	setupNetLinkMocks(ctrl, mockNetLink)
+
+	_ = mockIptables.NewChain("nat", "AWS-SNAT-CHAIN-1")
```
jdn5126, line 892: nit: no need to even try to capture the return value with `_ =` if it won't be used. You can just call the function with `mockIptables.NewChain("nat", "AWS-SNAT-CHAIN-1")`

pkg/iptableswrapper/mocks/iptables_maps.go
jchen6585: This is because we can only insert chain creation rule (`-N chain`) into the map for chains we manually create (e.g. `AWS-SNAT-CHAIN`). This function will add all the `-d CIDR` rules to the position right after the chain creation rule. We insert a rule to the `FORWARD` chain in the `filter` table at https://github.com/aws/amazon-vpc-cni-k8s/blob/3be57994609107d9330a15c6a42e42ab1df9f107/pkg/networkutils/network.go#L284. Since this chain is an existing chain rather than one we explicitly create, we do not capture the `-N` rule in the map, hence adding to `pos=1` will result in an out of bounds exception.

pkg/iptableswrapper/mocks/iptables_maps.go
jchen6585: To mock the behavior from the actual function https://pkg.go.dev/github.com/coreos/go-iptables/iptables#IPTables.DeleteChain

pkg/networkutils/network_test.go
jchen6585: yes

pkg/iptableswrapper/mocks/iptables_maps.go
jchen6585: Update: resolved this

pkg/iptableswrapper/mocks/iptables_maps.go
```diff
@@ -44,7 +44,13 @@ func (ipt *MockIptables) Insert(table, chain string, pos int, rulespec ...string
 	if ipt.DataplaneState[table] == nil {
 		ipt.DataplaneState[table] = map[string][][]string{}
 	}
-	ipt.DataplaneState[table][chain] = append(ipt.DataplaneState[table][chain], rulespec)
+	fmt.Println(len(ipt.DataplaneState[table][chain]))
```
jdn5126, line 47: nit: we'll want to remove this, else it will show up in every unit test output

pkg/networkutils/network_test.go
jdn5126: It seems a bit weird to pass the chain name twice, as it is hard to tell that the name as the first entry maps to a `-N` entry. I think we should either pass the string `-N`, or infer from the chain

pkg/iptableswrapper/mocks/iptables_maps.go
```diff
@@ -91,6 +97,11 @@ func (ipt *MockIptables) List(table, chain string) ([]string, error) {
 	var chains []string
 	chainContents := ipt.DataplaneState[table][chain]
 	for _, ruleSpec := range chainContents {
+		// If a rule spec is just the chain name, it is the chain creation rule
+		if len(ruleSpec) == 1 {
+			chains = append(chains, fmt.Sprintf("-N %s", chain))
```
jdn5126, line 102: The `List` API should not be modifying the chain. Shouldn't we be adding the `-N` entry when the chain is created? This is the part that is confusing me

jdn5126 (changes requested):

pkg/iptableswrapper/mocks/iptables_maps.go
```diff
@@ -101,10 +111,14 @@ func (ipt *MockIptables) List(table, chain string) ([]string, error) {
...
 	return chains, nil
-
 }
 
 func (ipt *MockIptables) NewChain(table, chain string) error {
+	exists, _ := ipt.ChainExists(table, chain)
+	if exists {
+		return errors.New("Chain already exists")
+	}
+	ipt.Append(table, chain, "-N", chain)
```
jdn5126, line 121: nit: let's add a comment explaining why we are doing this

pkg/networkutils/network_test.go
```diff
@@ -466,18 +470,20 @@ func TestSetupHostNetworkWithExcludeSNATCIDRs(t *testing.T) {
 		map[string]map[string][][]string{
 			"nat": {
 				"AWS-SNAT-CHAIN-0": [][]string{
-					{"-d", "10.10.0.0/16", "-m", "comment", "--comment", "AWS SNAT CHAIN", "-j", "RETURN"},
-					{"-d", "10.11.0.0/16", "-m", "comment", "--comment", "AWS SNAT CHAIN", "-j", "RETURN"},
-					{"-d", "10.12.0.0/16", "-m", "comment", "--comment", "AWS SNAT CHAIN EXCLUSION", "-j", "RETURN"},
+					{"-N", "AWS-SNAT-CHAIN-0"},
```
jdn5126, line 473: Why are there two entries here? Isn't the `-N` enough? Since we have the chain name already

jchen6585: Just for more clarity since the rule itself is `-N chainName`, but yeah i can modify it to just be -N

jdn5126: I see, it's up to you, I don't have that strong of a preference

jdn5126 (approved):

PR #2729, opened 2023-12-27
Enable running tests with --endpoint flag.

orsenthil:

When we run this flag against beta clusters, we need to send the `--endpoint $ENDPOINT` so that describe cluster operation will be successful.

- [x] Tested against Beta
- [ ] Tested against Prod


jdn5126 (dismissed):

scripts/run-static-canary.sh
```diff
@@ -10,6 +10,13 @@ GINKGO_TEST_BUILD="$SCRIPT_DIR/../test/build"
 # TEST_IMAGE_REGISTRY is the registry in test-infra-* accounts where e2e test images are stored
 TEST_IMAGE_REGISTRY=${TEST_IMAGE_REGISTRY:-"617930562442.dkr.ecr.us-west-2.amazonaws.com"}
 
+# If $ENDPOINT is set, as in it is for beta clusters, then add the --endpoint flag to the ginkgo test command
+ENDPOINT_FLAG=""
+if [ -n "$ENDPOINT" ]; then
+  ENDPOINT_FLAG="--endpoint=$ENDPOINT"
```
jaydeokar, line 16: I think we need `eks-endpoint` instead of `--endpoint`

https://github.com/aws/amazon-vpc-cni-k8s/blob/a42364a2e48ae7564d1ea8ad395144908717be06/test/framework/options.go#L60

orsenthil: @jaydeokar , yes. It should be `eks-endpoint` . Thanks for catching.
I was testing it correctly - https://github.com/orsenthil/amazon-vpc-cni-k8s/commit/e65646bdbc4b00b85ee2b6d17113c69c8785a9ae

Missed this in the PR.

jaydeokar (approved):
lgtm

PR #2731, opened 2023-12-29
Fix endpoint flag, use $ENDPOINT_OPTION already defined in lib/canary.sh

orsenthil:

lib/canary.sh was already defining two variables "ENDPOINT_FLAG" and "ENDPOINT_OPTION" 
and it was confusing another introduction of this flag in this script.

We have to use what is already defined here

https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/lib/canary.sh#L13

Integration test suite uses ENDPOINT_OPTION like this
https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/run-ginkgo-integration-suite.sh#L46

- [] Tested in Beta


jaydeokar (approved):
lgtm

PR #2744, opened 2024-01-05
Faster eni scaleup

jchen6585:
**What type of PR is this?**

enhancement

N/A

We would previously allocate 1 ENI max per loop: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/ipamd.go#L621-L628

This PR modifies the `updateIpPoolIfRequired` function to have faster ENI scale up to avoid the 5 second sleep.

CNI suite: passing
IPAMD suite: passing

N/A

No, Yes

No

No

```release-note

```

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -652,7 +652,11 @@ func (c *IPAMContext) updateIPPoolIfRequired(ctx context.Context) {
 	log.Debugf("IP stats - total IPs: %d, assigned IPs: %d, cooldown IPs: %d", stats.TotalIPs, stats.AssignedIPs, stats.CooldownIPs)
 
 	if datastorePoolTooLow {
-		c.increaseDatastorePool(ctx)
+		// Allow for rapid scale up to decrease time it takes for pod to retrieve an ip
+		// but conservative scale down to account for pod churn
+		for datastorePoolStillTooLow := datastorePoolTooLow; datastorePoolStillTooLow; datastorePoolStillTooLow, _ = c.isDatastorePoolTooLow() {
+			c.increaseDatastorePool(ctx)
+		}
 	} else if c.isDatastorePoolTooHigh(stats) {
 		c.decreaseDatastorePool(decreaseIPPoolInterval)
 	}


pkg/ipamd/ipamd.go
```diff
@@ -529,7 +529,11 @@ func (c *IPAMContext) nodeInit() error {
 
 	// On node init, check if datastore pool needs to be increased. If so, attach CIDRs from existing ENIs and attach new ENIs.
 	if !c.disableENIProvisioning && c.isDatastorePoolTooLow() {
-		if err := c.increaseDatastorePool(ctx); err != nil {
```
jdn5126, line 532: nit: I am realizing now that `increaseDatastorePool` is an expensive function, so the better check is to loop on `c.isDatastorePoolTooLow` returning true. As long as we are supposed to increase the datastore pool, we should do it.

Now, the question to think about is whether we would do this only during the main IP pool loop, or during initialization as well

jchen6585: ack

pkg/ipamd/ipamd.go
```diff
@@ -636,6 +636,10 @@ func (c *IPAMContext) updateIPPoolIfRequired(ctx context.Context) {
 
 	if c.isDatastorePoolTooLow() {
 		c.increaseDatastorePool(ctx)
+		// Speeds up ENI scale up
+		for c.isDatastorePoolTooLow() {
```
jdn5126, line 640: Let's wait for https://github.com/aws/amazon-vpc-cni-k8s/pull/2745 to merge, as we should be able to make this a bit cleaner, i.e. something like:
```
poolTooLow, stats := c.isDatastorePoolTooLow()
if poolTooLow {
    for stillTooLow := poolTooLow; stillTooLow; stillTooLow, _ = c.isDatastorePoolTooLow(nil) {
        c.increaseDatastorePool(ctx)
    }
} else if c.isDatastorePoolTooHigh(stats) {
    ...
}
```

pkg/ipamd/ipamd.go
```diff
@@ -652,7 +652,9 @@ func (c *IPAMContext) updateIPPoolIfRequired(ctx context.Context) {
 	log.Debugf("IP stats - total IPs: %d, assigned IPs: %d, cooldown IPs: %d", stats.TotalIPs, stats.AssignedIPs, stats.CooldownIPs)
 
 	if datastorePoolTooLow {
-		c.increaseDatastorePool(ctx)
+		for datastorePoolStillTooLow := datastorePoolTooLow; datastorePoolStillTooLow; datastorePoolStillTooLow, _ = c.isDatastorePoolTooLow() {
```
jdn5126, line 655: nit: let's add a comment here explaining why we want to rapidly scale up, but we do not want to rapidly scale down

jdn5126 (changes requested):

jdn5126 (approved):

orsenthil:
This change as the cause of performance issue mentioned in https://github.com/aws/amazon-vpc-cni-k8s/issues/2807 and has been reverted in https://github.com/aws/amazon-vpc-cni-k8s/pull/2810

PR #2745, opened 2024-01-05
Track max pods, simplify warm IP pool management

jdn5126:
**What type of PR is this?**
enhancement

N/A

I set out trying to do two things in this PR:
1. Handle the low-hanging fruit of not allocating more IPs to the node than `max-pods`, so as not to waste VPC IPs.
2. Cleanup and simplify the warm IP pool management logic

On point 1, I partially succeeded, as we now do not increase the datastore pool when more than `max-pods` IPs are allocated to the node. This check cannot be exact since IPs can be allocated in chunks (prefix mode or "full-ENI" mode). Note that this check only needs to be enforced on the allocation side, not the deallocation side.

On point 2, I had very limited success, as the warm IP pool management logic is a behemoth. I made some simplifications, primarily around limiting how many times we calculate `DataStoreStats`. Calls to `GetIPStats` acquire a datastore lock that can slow down pod IP acquisition. The conclusion I came to is that the warm pool logic has to be significantly refactored and broken up.

N/A

Manually verified that all integration tests pass after this change. Also verified that when max pods is set to a low value, excess IPs are not allocated.

No

No, Yes

No

Yes

```release-note
Track max pods, simplify warm IP pool management
```

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -154,10 +154,10 @@ const (
 	ipV4AddrFamily = "4"
 	ipV6AddrFamily = "6"
 
-	//insufficientCidrErrorCooldown is the amount of time reconciler will wait before trying to fetch
-	//more IPs/prefixes for an ENI. With InsufficientCidr we know the subnet doesn't have enough IPs so
-	//instead of retrying every 5s which would lead to increase in EC2 AllocIPAddress calls, we wait for
-	//120 seconds for a retry.
+	// insufficientCidrErrorCooldown is the amount of time reconciler will wait before trying to fetch
+	// more IPs/prefixes for an ENI. With InsufficientCidr we know the subnet doesn't have enough IPs so
+	// instead of retrying every 5s which would lead to increase in EC2 AllocIPAddress calls, we wait for
+	// 120 seconds for a retry.
 	insufficientCidrErrorCooldown = 120 * time.Second
 
 	// envManageUntaggedENI is used to determine if untagged ENIs should be managed or unmanaged
@@ -216,6 +216,7 @@ type IPAMContext struct {
 	lastInsufficientCidrError time.Time
 	enableManageUntaggedMode  bool
 	enablePodIPAnnotation     bool
+	maxPods                   int // maximum number of pods that can be scheduled on the node
 }
 
 // setUnmanagedENIs will rebuild the set of ENI IDs for ENIs tagged as "no_manage"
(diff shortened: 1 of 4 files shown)


pkg/ipamd/ipamd.go
```diff
@@ -2122,44 +2106,48 @@ func (c *IPAMContext) hasRoomForEni() bool {
 	return c.dataStore.GetENIs() < (c.maxENI - c.unmanagedENI - trunkEni)
 }
 
-func (c *IPAMContext) isDatastorePoolTooLow() bool {
-	short, _, warmTargetDefined := c.datastoreTargetState()
-	if warmTargetDefined {
-		return short > 0
+func (c *IPAMContext) isDatastorePoolTooLow() (bool, *datastore.DataStoreStats) {
+	stats := c.dataStore.GetIPStats(ipV4AddrFamily)
+	// If max pods has been reached, pool is not too low
```
orsenthil, line 2111: > If max pods has been reached, pool is not too low

I think, this comment is bit confusing, as max pods has been mentioned without the context. I got confused while reading this.

If we write,  if `c.maxPods <= stats.TotalIPs` then it seems alright.

Here we are saying, the nodes maxPods is less than or equal to the total ip available in the datastore, and so the isDatastorePoolTooLow too low is false.

jdn5126: Yeah, what this function checks is a bit confusing. `isDatastorePoolTooLow` looks at the number of currently available IPs and compares that number to the warm/minimum targets. If there are not enough available IPs, we return false.

The `maxPods` check here is short-circuiting that logic by saying, "forgot about warm targets, do not allocate any more IPs if we have already allocated more than maxPods worth"

orsenthil: I understood it better, when thinking like this 
> c.maxPods <= stats.TotalIPs

Because we are sending stats as the argument, I think, you went with the validation as  `stats.TotalIPs >= c.maxPods`; explaining the we have more than maxPods worth. 

Sounds good to me.

orsenthil (approved):
LGTM.

pkg/ipamd/ipamd.go
```diff
@@ -499,17 +500,26 @@ func (c *IPAMContext) nodeInit() error {
 		}, 30*time.Second)
 	}
 
+	// Make a k8s client request for the current node so that max pods can be derived
+	node, err := k8sapi.GetNode(ctx, c.k8sClient)
```
jayanthvn, line 504: We should ideally remove the need of k8sClient usage unless any feature like custom networking would need it. We can enhance CNINode for this information. /cc @haouc

jdn5126: Yeah, I thought about that, but we already have a non-optional k8s client call for the node when we attach an ENI - to check if the node is currently marked as schedulable. It would be nice to have this node state available in `CNINode`, but that would still require a k8s client call to retrieve.

jayanthvn: sure..with secondary IPs we would never go over max pods since max pods take secondary IP into account. This would be beneficial for prefix delegation but it is still tricky since allocation happens in chunks of 16..

jdn5126: Right, that's we don't enforce a hard check on increase (or decrease). We simply do not increase, even when the warm targets tell us to, when max pods has been reached

achevuru: Also, in PD mode - Let's say we have WARM_PREFIX set to `1` and the node is already @ Max Pods. Now, with this change, we will not be honoring the WARM PREFIX in this state as the node is already at capacity w.r.t maxPods. If the user now churns the pods on this node (Let's say rollout restart of all deployments) - IPs assigned to recently deleted pods can't be reused for next 30s and so the new pods will not have any IPs and we would need to pull in a new prefix anyways? Current workflow, would've assigned an additional prefix and the new pods would've consumed IPs form the additional prefix and we would've released the old prefix after the cool down period. Did I understand the new flow right?

jdn5126: Your understanding is correct. Over-provisioning could help in a case where the user restarts all pods at once. I do not think that is a case that we should design for, though. The downsides of over-provisioning (cost and resource utilization) seem much higher than slower recovery after restarting all pods, which should be an extremely rare event.

PR #2748, opened 2024-01-10
cni-metrics-helper add podAnnotation value

prysmakou:
**What type of PR is this?**

feature
cni-metrics-helper add deployment.podAnnotation value

N/A

Allows to add arbitrary annotations to pods (eg DataDog requires specific annotation to scrap Prometheus metrics)

manual deploy

No

No

It changes the deployment/daemonset

It introduces a new parameter in cni-metrics-helper helm chart (with a default value {} )

```release-note

```

### DIFF
--- a/charts/cni-metrics-helper/values.yaml
+++ b/charts/cni-metrics-helper/values.yaml
@@ -34,3 +34,5 @@ revisionHistoryLimit: 10
 podSecurityContext: {}
 
 containerSecurityContext: {}
+
+podAnnotations: {}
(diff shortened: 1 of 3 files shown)


jdn5126:
Thanks for the contribution @prysmakou! Let's follow the example set by the `aws-vpc-cni` chart: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/charts/aws-vpc-cni/values.yaml#L117 and https://github.com/aws/amazon-vpc-cni-k8s/blob/master/charts/aws-vpc-cni/templates/daemonset.yaml#L24

Also, can you update the title and description to be **cni** instead of **csi**?

charts/cni-metrics-helper/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: cni-metrics-helper
-version: 1.16.0
+version: 1.16.1
```
jdn5126, line 3: nit: you can remove this change, as this gets updated automatically when a new release is cut

charts/cni-metrics-helper/values.yaml
```diff
@@ -34,3 +34,6 @@ revisionHistoryLimit: 10
 podSecurityContext: {}
 
 containerSecurityContext: {}
+
+deployment:
```
jdn5126, line 38: nit: let's remove the `deployment` parent and just use `podAnnotations`, same as VPC CNI

jdn5126 (approved):

PR #2751, opened 2024-01-16
Update Helm adoption script

JoeNorth:
Updates the Helm adoption Bash script to include the `amazon-vpc-cni` configmap introduced in #2517

**What type of PR is this?**

N/A

Customers using this script to adopt the VPC CNI resources to be managed by Helm run into an error like the following:

```
Error: INSTALLATION FAILED: rendered manifests contain a resource that already exists. Unable to continue with install: ConfigMap "amazon-vpc-cni" in namespace "kube-system" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "aws-vpc-cni"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "kube-system"
```

This ConfigMap was added in v1.14 but the script was not updated to reflect this change thus leading to Helm install failures.

No

No

No

No


jdn5126 (approved):
Change overall LGTM, thanks for adding @JoeNorth!

PR #2782, opened 2024-02-03
Fix max eni limit for multicard

jchen6585:
**What type of PR is this?**

bug

N/A

When using an instance type with multi-card, CNI will be unable to create new ENIs since our `GetENILimit` function currently only checks the max ENIs for a singular network card.

Only ran units tests

N/A

No, Yes

No

No

```release-note

```

### DIFF
--- a/pkg/vpc/vpc.go
+++ b/pkg/vpc/vpc.go
@@ -59,6 +59,13 @@ func GetENILimit(instanceType string) (int, error) {
 		log.Errorf("%s: %s", instanceType, ErrInstanceTypeNotExist)
 		return -1, ErrInstanceTypeNotExist
 	}
+	if len(instance.NetworkCards) > 1 {
+		eniLimit := 0
+		for _, networkCards := range instance.NetworkCards {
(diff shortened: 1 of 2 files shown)


pkg/vpc/vpc.go
```diff
@@ -59,6 +59,13 @@ func GetENILimit(instanceType string) (int, error) {
 		log.Errorf("%s: %s", instanceType, ErrInstanceTypeNotExist)
 		return -1, ErrInstanceTypeNotExist
 	}
+	if len(instance.NetworkCards) > 1 {
+		eniLimit := 0
+		for _, networkCards := range instance.NetworkCards {
+			eniLimit = eniLimit + int(networkCards.MaximumNetworkInterfaces)
+		}
+		return eniLimit, nil
```
jayanthvn, line 67: We don't support multi cards yet..

jayanthvn (changes requested):

jchen6585:
Closed as we do not support multi-card yet so this change is unnecessary.

PR #2784, opened 2024-02-05
Ignore non-zero cards

jchen6585:
**What type of PR is this?**

bug

N/A

The previous logic `c.dataStore.GetENIs() < (c.maxENI - c.unmanagedENI - trunkEni)` would return negative for multicard instance types that had EFA devices installed on non-zero cards. This adds another value to the inequality to negate ENIs on non-zero cards as CNI does not support multi-card yet.

CNI and IPAMD suites look good

No

No, Yes

No

No

```release-note

```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -141,6 +141,9 @@ type APIs interface {
 	// GetENILimit returns the number of ENIs that can be attached to an instance
 	GetENILimit() int
 
+	// GetNetworkCards returns the network cards the instance has
+	GetNetworkCards() []vpc.NetworkCard
+
 	// GetPrimaryENImac returns the mac address of the primary ENI
 	GetPrimaryENImac() string
 
@@ -153,11 +156,11 @@ type APIs interface {
 	// WaitForENIAndIPsAttached waits until the ENI has been attached and the secondary IPs have been added
 	WaitForENIAndIPsAttached(eni string, wantedSecondaryIPs int) (ENIMetadata, error)
 
-	//SetCNIunmanaged ENI
-	SetCNIUnmanagedENIs(eniID []string) error
+	//SetMultiCardENIs ENI
+	SetMultiCardENIs(eniID []string) error
 
-	//IsCNIUnmanagedENI
-	IsCNIUnmanagedENI(eniID string) bool
+	//IsMultiCardENI
+	IsMultiCardENI(eniID string) bool
 
 	//IsPrimaryENI
 	IsPrimaryENI(eniID string) bool
@@ -200,7 +203,7 @@ type EC2InstanceMetadataCache struct {
 
 	unmanagedENIs          StringSet
 	useCustomNetworking    bool
-	cniunmanagedENIs       StringSet
+	multiCardENIs          StringSet
 	enablePrefixDelegation bool
 
 	clusterName       string
@@ -497,7 +500,7 @@ func (cache *EC2InstanceMetadataCache) RefreshSGIDs(mac string) error {
 		newENIs := StringSet{}
 		newENIs.Set(eniIDs)
 
-		tempfilteredENIs := newENIs.Difference(&cache.cniunmanagedENIs)
+		tempfilteredENIs := newENIs.Difference(&cache.multiCardENIs)
 		filteredENIs := tempfilteredENIs.Difference(&cache.unmanagedENIs)
 
 		sgIDsPtrs := aws.StringSlice(sgIDs)
@@ -703,13 +706,16 @@ func (cache *EC2InstanceMetadataCache) awsGetFreeDeviceNumber() (int, error) {
 	inst := result.Reservations[0].Instances[0]
 	var device [maxENIs]bool
 	for _, eni := range inst.NetworkInterfaces {
-		if aws.Int64Value(eni.Attachment.DeviceIndex) > maxENIs {
-			log.Warnf("The Device Index %d of the attached ENI %s > instance max slot %d",
-				aws.Int64Value(eni.Attachment.DeviceIndex), aws.StringValue(eni.NetworkInterfaceId),
-				maxENIs)
-		} else {
-			log.Debugf("Discovered device number is used: %d", aws.Int64Value(eni.Attachment.DeviceIndex))
-			device[aws.Int64Value(eni.Attachment.DeviceIndex)] = true
+		// We don't support multi-card yet, so only account for network card zero
+		if aws.Int64Value(eni.Attachment.NetworkCardIndex) == 0 {
+			if aws.Int64Value(eni.Attachment.DeviceIndex) > maxENIs {
+				log.Warnf("The Device Index %d of the attached ENI %s > instance max slot %d",
+					aws.Int64Value(eni.Attachment.DeviceIndex), aws.StringValue(eni.NetworkInterfaceId),
+					maxENIs)
+			} else {
+				log.Debugf("Discovered device number is used: %d", aws.Int64Value(eni.Attachment.DeviceIndex))
+				device[aws.Int64Value(eni.Attachment.DeviceIndex)] = true
+			}
 		}
 	}
 
@@ -780,6 +786,7 @@ func (cache *EC2InstanceMetadataCache) attachENI(eniID string) (string, error) {
 		DeviceIndex:        aws.Int64(int64(freeDevice)),
 		InstanceId:         aws.String(cache.instanceID),
 		NetworkInterfaceId: aws.String(eniID),
+		NetworkCardIndex:   aws.Int64(0),
 	}
 	start := time.Now()
 	attachOutput, err := cache.ec2SVC.AttachNetworkInterfaceWithContext(context.Background(), attachInput)
@@ -1447,6 +1454,15 @@ func (cache *EC2InstanceMetadataCache) GetENILimit() int {
 	return eniLimit
 }
 
+// GetNetworkCards returns the network cards the instance has
+func (cache *EC2InstanceMetadataCache) GetNetworkCards() []vpc.NetworkCard {
+	networkCards, err := vpc.GetNetworkCards(cache.instanceType)
+	if err != nil {
+		return nil
+	}
+	return networkCards
+}
+
 // GetInstanceHypervisorFamily returns hypervisor of EC2 instance type
 func (cache *EC2InstanceMetadataCache) GetInstanceHypervisorFamily() string {
 	hypervisor, err := vpc.GetHypervisorType(cache.instanceType)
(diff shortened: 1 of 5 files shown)


pkg/ipamd/ipamd.go
```diff
@@ -198,13 +198,15 @@ type IPAMContext struct {
...
 	maxPrefixesPerENI         int
 	unmanagedENI              int
-	warmENITarget             int
-	warmIPTarget              int
-	minimumIPTarget           int
-	warmPrefixTarget          int
-	primaryIP                 map[string]string // primaryIP is a map from ENI ID to primary IP of that ENI
-	lastNodeIPPoolAction      time.Time
-	lastDecreaseIPPool        time.Time
+	nonZeroCardENI            int
```
jdn5126, line 201: nit: for this to be extensible in the future, lets consider a field tracking the number of network cards on the instance

pkg/ipamd/ipamd.go
```diff
@@ -2107,7 +2112,7 @@ func (c *IPAMContext) hasRoomForEni() bool {
 	if c.enablePodENI && c.dataStore.GetTrunkENI() == "" {
 		trunkEni = 1
 	}
-	return c.dataStore.GetENIs() < (c.maxENI - c.unmanagedENI - trunkEni)
+	return c.dataStore.GetENIs() < (c.maxENI - c.unmanagedENI - trunkEni + c.nonZeroCardENI)
```
jdn5126, line 2115: nit: I would rather see an `if c.numNetworkCards > 1` condition here, as then we can add a comment explaining how we do the calculation for instance with multiple network cards

pkg/ipamd/ipamd.go
```diff
@@ -2107,6 +2114,10 @@ func (c *IPAMContext) hasRoomForEni() bool {
 	if c.enablePodENI && c.dataStore.GetTrunkENI() == "" {
 		trunkEni = 1
 	}
+	if c.numNetworkCards > 1 {
+		// ENIs on non-zero network cards do not count
+		return c.dataStore.GetENIs() < (c.maxENI - c.unmanagedENI - trunkEni + c.nonZeroCardENI)
```
jdn5126, line 2119: I am a bit confused by this logic. Since we have more than 1 network card, I can see why we want to use `c.numNetworkCards` in this equation.

So to check if there is room, we start by getting how many ENIs are attached. Then we check if that is less than the max number of ENIs the instance can support, subtracting the number of ENIs that are unmanaged by IPAM (including trunk ENI). How does adding in the number of network cards help in this comparison? Is this so that we can attach an ENI to an additional network card in the future?

pkg/ipamd/ipamd.go
```diff
@@ -1771,10 +1776,12 @@ func (c *IPAMContext) filterUnmanagedENIs(enis []awsutils.ENIMetadata) []awsutil
 		} else if c.awsClient.IsCNIUnmanagedENI(eni.ENIID) {
 			log.Debugf("Skipping ENI %s: since on non-zero network card", eni.ENIID)
 			numFiltered++
+			nonZeroCard++
```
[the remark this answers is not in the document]
jdn5126: Spoke offline and this is required, but we need to comment that this counts ENIs not on network card index 0.

pkg/awsutils/awsutils.go
```diff
@@ -703,13 +706,16 @@ func (cache *EC2InstanceMetadataCache) awsGetFreeDeviceNumber() (int, error) {
...
 	var device [maxENIs]bool
 	for _, eni := range inst.NetworkInterfaces {
-		if aws.Int64Value(eni.Attachment.DeviceIndex) > maxENIs {
-			log.Warnf("The Device Index %d of the attached ENI %s > instance max slot %d",
-				aws.Int64Value(eni.Attachment.DeviceIndex), aws.StringValue(eni.NetworkInterfaceId),
-				maxENIs)
-		} else {
-			log.Debugf("Discovered device number is used: %d", aws.Int64Value(eni.Attachment.DeviceIndex))
-			device[aws.Int64Value(eni.Attachment.DeviceIndex)] = true
+		// We don't support multi-card yet, so only account for network card zero
```
jdn5126, line 709: nit: "IPAMD does not support multi-card instances, so skip network cards other than the zero-indexed card"

pkg/ipamd/ipamd.go
```diff
@@ -198,13 +198,16 @@ type IPAMContext struct {
...
 	unmanagedENI              int
-	warmENITarget             int
-	warmIPTarget              int
-	minimumIPTarget           int
-	warmPrefixTarget          int
-	primaryIP                 map[string]string // primaryIP is a map from ENI ID to primary IP of that ENI
-	lastNodeIPPoolAction      time.Time
-	lastDecreaseIPPool        time.Time
+	numNetworkCards           int
+	nonZeroCardENI            int
```
jdn5126, line 202: Discussed offline, and we still want to cache the number of ENIs that are behind the non-zero network card. We just need a better variable name and a comment explaining this. Something like `multiCardENIs` and `ENIs attached to non-zero network cards`

pkg/ipamd/ipamd.go
jdn5126, line 1779: Discussed offline, and while we still need to filter for ENIs behind the non-zero network card, there is no need to count the number here since we already have that number in the result of the describe call, i.e. `metadataResult.MultiCardENIIDs`

jdn5126: We also want to rename `IsCNIUnmanagedENI` to something that clearly indicates that we are checking if this is an ENI that is not behind Network Card 0, i.e. a multi-card ENI

pkg/ipamd/ipamd.go
jdn5126: Discussed offline, and we can change this equation to:
```
// For multi-card instances, c.unmanagedENI includes all ENIs attached to non-zero network cards. We must exclude these ENIs from the calculation as they cannot be used.
return c.dataStore.GetENIs() < (c.maxENI - (c.unmanagedENI - c.multiCardENIs) - trunkEni)
```
for more readability

pkg/awsutils/awsutils.go
```diff
@@ -1922,16 +1938,16 @@ func (cache *EC2InstanceMetadataCache) getENIsFromPaginatedDescribeNetworkInterf
 	return innerErr
 }
 
-// SetCNIUnmanagedENIs Set unmanaged ENI set
-func (cache *EC2InstanceMetadataCache) SetCNIUnmanagedENIs(eniID []string) error {
+// SetMultiCardENIs Set unmanaged ENI set
+func (cache *EC2InstanceMetadataCache) SetMultiCardENIs(eniID []string) error {
 	if len(eniID) != 0 {
 		cache.cniunmanagedENIs.Set(eniID)
```
jdn5126, line 1944: nit: while we are here, let's update this `cniunmanagedENIs` variable name to be `multiCardENIs`

pkg/awsutils/awsutils.go
```diff
@@ -1922,18 +1938,18 @@ func (cache *EC2InstanceMetadataCache) getENIsFromPaginatedDescribeNetworkInterf
 	return innerErr
 }
 
-// SetCNIUnmanagedENIs Set unmanaged ENI set
-func (cache *EC2InstanceMetadataCache) SetCNIUnmanagedENIs(eniID []string) error {
+// SetMultiCardENIs Set unmanaged ENI set
```
jdn5126, line 1941: nit: we should fix this comment to be something like: `SetMultiCardENIs creates a StringSet tracking ENIs not behind the default network card index`

pkg/awsutils/awsutils.go
```diff
@@ -1922,18 +1938,18 @@ func (cache *EC2InstanceMetadataCache) getENIsFromPaginatedDescribeNetworkInterf
...
 	if len(eniID) != 0 {
-		cache.cniunmanagedENIs.Set(eniID)
+		cache.multiCardENIs.Set(eniID)
 	}
 	return nil
 }
 
-// IsCNIUnmanagedENI returns if the eni is unmanaged
-func (cache *EC2InstanceMetadataCache) IsCNIUnmanagedENI(eniID string) bool {
+// IsMultiCardENI returns if the eni is unmanaged
```
jdn5126, line 1949: nit: `IsMultiCardENI returns if the ENI is not behind the default network card index (multi-card ENI)`

pkg/ipamd/ipamd.go
```diff
@@ -198,13 +198,16 @@ type IPAMContext struct {
...
 	unmanagedENI              int
-	warmENITarget             int
-	warmIPTarget              int
-	minimumIPTarget           int
-	warmPrefixTarget          int
-	primaryIP                 map[string]string // primaryIP is a map from ENI ID to primary IP of that ENI
-	lastNodeIPPoolAction      time.Time
-	lastDecreaseIPPool        time.Time
+	numNetworkCards           int
+	multiCardENIs             int //ENIs attached to non-zero cards
```
jdn5126, line 202: big nit: space after comment and "non-zero network card"

jdn5126 (dismissed):

pkg/ipamd/ipamd.go
```diff
@@ -2107,6 +2113,10 @@ func (c *IPAMContext) hasRoomForEni() bool {
 	if c.enablePodENI && c.dataStore.GetTrunkENI() == "" {
 		trunkEni = 1
 	}
+	if c.numNetworkCards > 1 {
+		// For multi-card instances, c.unmanagedENI includes all ENIs attached to non-zero network cards. We must exclude these ENIs from the calculation as they cannot be used.
+		return c.dataStore.GetENIs() < (c.maxENI - (c.unmanagedENI - c.multiCardENIs) - trunkEni)
```
achevuru, line 2118: Instead of modifying this, why don't we skip counting multi card ENIs towards `unmanaged ENIs` (i.e.,) we will not increment `numFiltered` here - https://github.com/aws/amazon-vpc-cni-k8s/blob/502605c3ce75c355ed50878ccfb1c9d1f0bc46a1/pkg/ipamd/ipamd.go#L1773 . Are we missing out on anything else by doing so?

pkg/awsutils/awsutils.go
```diff
@@ -703,13 +706,16 @@ func (cache *EC2InstanceMetadataCache) awsGetFreeDeviceNumber() (int, error) {
...
 	for _, eni := range inst.NetworkInterfaces {
-		if aws.Int64Value(eni.Attachment.DeviceIndex) > maxENIs {
-			log.Warnf("The Device Index %d of the attached ENI %s > instance max slot %d",
-				aws.Int64Value(eni.Attachment.DeviceIndex), aws.StringValue(eni.NetworkInterfaceId),
-				maxENIs)
-		} else {
-			log.Debugf("Discovered device number is used: %d", aws.Int64Value(eni.Attachment.DeviceIndex))
-			device[aws.Int64Value(eni.Attachment.DeviceIndex)] = true
+		// We don't support multi-card yet, so only account for network card zero
+		if aws.Int64Value(eni.Attachment.NetworkCardIndex) == 0 {
```
jayanthvn, line 710: Do we still need this change?

jchen6585: Yep so we don't mark a device index on card 0 as used when it is really not.

jayanthvn: Ahh right since we do a describe call..later on we can actually enhance it..should be fine for now..

jayanthvn (approved):
lgtm

PR #2785, opened 2024-02-06
gRPC call for networkpolicy agent [only for main branch]

jayanthvn:
**What type of PR is this?**

feature

**Which issue does this PR fix**: N/A

We are looking into certain scenarios where a gRPC call has to be made to network policy agent. This will be the RPC auto generated code. Do not merge to release branch yet.

N/A

Yes, these are auto generated.

No

No

No

No

```release-note

```


pkg/rpcwrapper/client.go
```diff
@@ -19,9 +19,10 @@ import (
 	grpc "google.golang.org/grpc"
 )
 
-// RPC is the wrapper interface for the CNI gRPC backend client
+// RPC is the wrapper interface for the CNI and network policy agen gRPC backend client
```
achevuru, line 22: typo: `agent`

rpc/rpc.proto
```diff
@@ -38,7 +38,8 @@ message AddNetworkReply {
   int32 ParentIfIndex = 10;
   // end of pod-eni parameters
 
-  // next field: 13
+  bool EnableNetworkPolicy = 13;
```
jdn5126, line 41: Is the motivation for this field so that the CNI knows whether to make the NP RPC call?

jayanthvn: yes similar to SGPP, external SNAT..

achevuru (approved):

Issue #2606, opened 2023-10-06
Ability to set pod MTU separate from ENI MTU (or eth0)


archoversight:
**What would you like to be added**:

I'd like to have the ability to set the MTU for my pod's virtual interfaces to a lower MTU than the MTU for my eth0. I am running an IPv6 only EKS cluster, and attempting to deploy Cillium on top with Wireguard encryption so that pod to pod traffic is transparently encrypted.

Unfortunately the overhead from the Wireguard tunnel and lack of path MTU means that currently traffic is silently dropped if it tries to send packets of MTU 9001 when the Wireguard tunnel is set to 8921 MTU.

I tried setting `AWS_VPC_ENI_MTU` to `8000` as an example, but it seems to also change the MTU when starting for `eth0` which is not what I want.

**Why is this needed**:

When chaining CNI's that add transparent encryption or encapsulation on IPv6 hosts where path MTU does not function, it would be nice to have an escape hatch.

**References**:

https://github.com/cilium/cilium/issues/28413#issuecomment-1749803719
https://github.com/cilium/cilium/issues/28387
https://aws.amazon.com/blogs/containers/transparent-encryption-of-node-to-node-traffic-on-amazon-eks-using-wireguard-and-cilium/

jdn5126:
@archoversight setting `AWS_VPC_ENI_MTU` should set the MTU for the pod's virtual interfaces. `eth0` is the name of the pod veth endpoint in the pod networking namespace. Are you not seeing the MTU get set on that interface?

archoversight:
> @archoversight setting `AWS_VPC_ENI_MTU` should set the MTU for the pod's virtual interfaces. `eth0` is the name of the pod veth endpoint in the pod networking namespace. Are you not seeing the MTU get set on that interface?

I am using EKS in IPv6 mode, eth0 has a prefix delegated to it. I am looking at the host (pod launched with `kubectl debug`) and it is showing that eth0's MTU is changing, not just the veths that are created.

I am seeing the MTU set on the veth interfaces correctly, but I also see the MTU change on eth0.

mmerickel:
@jdn5126 we're referring to the host eth0, not in a pod. It's true that in the pods the veth interface shows up as eth0.

jdn5126:
The host `eth0` is the primary ENI, which the VPC CNI does not manage. If you want to change the MTU on the primary ENI, you would need to do so in the AMI, or in the node group template

[...]

PR #2791, opened 2024-02-08
Pod MTU

jchen6585:
**What type of PR is this?**

feature

#2606 

Implements a new env var `POD_MTU` that sets the pod MTU. This will default to `AWS_VPC_ENI_MTU` value if not set.

Ran CNI, IPAMD, and IPv6 test suites.

No

No, Yes

No

Yes

```release-note
By setting `POD_MTU`, you can have different MTU values for the pod and ENI.

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,6 +267,15 @@ Default: empty
 Specify a comma-separated list of IPv4 CIDRs to exclude from SNAT. For every item in the list an `iptables` rule and off\-VPC
 IP rule will be applied. If an item is not a valid ipv4 range it will be skipped. This should be used when `AWS_VPC_K8S_CNI_EXTERNALSNAT=false`.
 
+#### `POD_MTU` (v1.x.x+)
+
+Type: Integer as a String
+
+*Note*: The default value is set to AWS_VPC_ENI_MTU, which defaults to 9001 if unset.
+Default: 9001
+
+Used to configure the MTU size for pod virtual interfaces. The valid range is from `576` to `9001`.
+
 #### `WARM_ENI_TARGET`
 
 Type: Integer as a String
--- a/cmd/aws-vpc-cni/main.go
+++ b/cmd/aws-vpc-cni/main.go
@@ -88,6 +88,7 @@ const (
 	envHostCniConfDirPath    = "HOST_CNI_CONFDIR_PATH"
 	envVethPrefix            = "AWS_VPC_K8S_CNI_VETHPREFIX"
 	envEniMTU                = "AWS_VPC_ENI_MTU"
+	envPodMTU                = "POD_MTU"
 	envEnablePodEni          = "ENABLE_POD_ENI"
 	envPodSGEnforcingMode    = "POD_SECURITY_GROUP_ENFORCING_MODE"
 	envPluginLogFile         = "AWS_VPC_K8S_PLUGIN_LOG_FILE"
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -267,6 +267,14 @@ Default: empty
 Specify a comma-separated list of IPv4 CIDRs to exclude from SNAT. For every item in the list an `iptables` rule and off\-VPC
 IP rule will be applied. If an item is not a valid ipv4 range it will be skipped. This should be used when `AWS_VPC_K8S_CNI_EXTERNALSNAT=false`.
 
+#### `POD_MTU` (v1.x.x+)
+
+Type: Integer as a String
+
+Default: 9001
```
jdn5126, line 274: nit: let's add a comment that the default value is whatever `AWS_VPC_ENI_MTU` is set to

cmd/aws-vpc-cni/main.go
```diff
@@ -279,6 +280,10 @@ func generateJSON(jsonFile string, outFile string, getPrimaryIP func(ipv4 bool)
 	}
 	vethPrefix := utils.GetEnv(envVethPrefix, defaultVethPrefix)
 	mtu := utils.GetEnv(envEniMTU, defaultMTU)
+	_, found := os.LookupEnv("POD_MTU")
+	if found {
+		mtu = utils.GetEnv(envPodMTU, defaultMTU)
+	}
```
jdn5126, line 286: nit: let's rename this to `podMTU`, and let's do:
```
// Derive pod MTU from ENI MTU by default
eniMTU := utils.GetEnv(envEniMTU, defaultMTU)
// If pod MTU environment variable is set, overwrite ENI MTU.
podMTU := utils.GetEnv(envPodMTU, eniMTU)
```

test/integration/cni/host_networking_test.go
```diff
@@ -150,6 +154,54 @@ var _ = Describe("test host networking", func() {
 			By("validating host networking is teared down correctly")
 			common.ValidateHostNetworking(common.NetworkingTearDownSucceeds, input, primaryNode.Name, f)
 		})
+
+		It("Validate Host Networking setup after changing Pod MTU and Veth Prefix", func() {
```
jdn5126, line 158: nit: since this is a copy of the test case above, I am wondering if there is a way that we could avoid copying code by using a new `Context` and `JustBeforeEach` to set the environment variables

test/integration/ipv6/ipv6_host_networking_test.go
```diff
@@ -144,6 +150,54 @@ var _ = Describe("[CANARY] test ipv6 host netns setup", func() {
 			By("validating host networking is teared down correctly")
 			ValidateHostNetworking(NetworkingTearDownSucceeds, input)
 		})
+
+		It("Validate host netns setup after changing Pod MTU and Veth Prefix", func() {
```
jdn5126, line 154: nit: same here, it would be good to avoid copying

jdn5126 (approved):

README.md
```diff
@@ -267,6 +267,15 @@ Default: empty
...
 IP rule will be applied. If an item is not a valid ipv4 range it will be skipped. This should be used when `AWS_VPC_K8S_CNI_EXTERNALSNAT=false`.
 
+#### `POD_MTU` (v1.x.x+)
+
+Type: Integer as a String
+
+*Note*: The default value is set to AWS_VPC_ENI_MTU, which defaults to 9001 if unset.
+Default: 9001
+
+Used to configure the MTU size for pod virtual interfaces. The valid range is from `576` to `9001`.
```
archoversight, line 277: `576` is only valid for IPv6, otherwise the minimum MTU has to be `1280`

jdn5126: Good call @archoversight. @jchen6585 we should update the README for `AWS_VPC_ENI_MTU` and `POD_MTU` at the very least, and we can consider adding runtime logging to error if IPv6 is configured and the MTU is < 1280

Issue #2792, opened 2024-02-13
aws eniconfig not being honored and pods using trunk instead of ENI


sstarcher:
**What happened**:
Upgraded from aws-vpc-cni v1.12.6 to v1.16.0.  Pods sometimes get assigned to the trunk interface instead of to the ENI.  This causes them to not get the correct security groups from the ENIConfig.  A small sample size this seems to be pods that got assigned to the node just as it is coming up.

**Attach logs**

snippet of logs remainder sent 

aws-node-f5x7z aws-node {"level":"info","ts":"2024-02-13T13:19:12.381Z","caller":"ipamd/ipamd.go:822","msg":"Found ENI Config Name: eni-config-ds-subnet-0318d75ae06a34052"}
aws-node-f5x7z aws-node {"level":"info","ts":"2024-02-13T13:19:12.483Z","caller":"ipamd/ipamd.go:793","msg":"ipamd: using custom network config: [sg-066233d33bbd94a21 sg-03d0fde3a6a691a6d], subnet-0318d75ae06a34052"}
aws-node-f5x7z aws-node {"level":"info","ts":"2024-02-13T13:19:12.483Z","caller":"awsutils/awsutils.go:728","msg":"Trying to allocate 10 IP addresses on new ENI"}
aws-node-f5x7z aws-node {"level":"info","ts":"2024-02-13T13:19:12.483Z","caller":"awsutils/awsutils.go:728","msg":"Using a custom network config for the new ENI"}
aws-node-f5x7z aws-node {"level":"info","ts":"2024-02-13T13:19:12.483Z","caller":"awsutils/awsutils.go:728","msg":"Creating ENI with security groups: [sg-066233d33bbd94a21 sg-03d0fde3a6a691a6d] in subnet: subnet-0318d75ae06a34052"}
aws-node-f5x7z aws-node {"level":"info","ts":"2024-02-13T13:19:12.910Z","caller":"awsutils/awsutils.go:728","msg":"Created a new ENI: eni-0a9379b23fe4ae3e1"}
aws-node-f5x7z aws-node {"level":"info","ts":"2024-02-13T13:19:14.216Z","caller":"ipamd/ipamd.go:838","msg":"Successfully created and attached a new ENI eni-0a9379b23fe4ae3e1 to instance"}

[...]

PR #2801, opened 2024-02-20
Do not allocate IPs or prefixes to trunk ENIs or EFA ENIs

jdn5126:
**What type of PR is this?**
bug

https://github.com/aws/amazon-vpc-cni-k8s/issues/2792

This PR updates the `GetENINeedsIP` function, which tries to find an ENI in the datastore that new IPs/prefixes can be assigned to. This function needs to skip trunk ENIs, as they are managed by the VPC Resource Controller. It also skips EFA ENIs, as those are unmanaged by the VPC CNI.

N/A

All existing unit tests and integration tests pass. I spun up a 1.28 cluster with Custom Networking and Security Groups for Pods configured, and I verified the security groups assigned to trunk ENIs, the placement of new pods, etc. From the issue description in https://github.com/aws/amazon-vpc-cni-k8s/issues/2792, everything seems to be resolved. 

No

No, Yes

No

Yes

```release-note
The VPC CNI should not assign IPs to trunk ENIs or EFA ENIs.
```


jchen6585 (dismissed):
LGTM

orsenthil:
Updating the README with this detail (keeping that comprehensive might help), and aws docs separately with this detail might help too.

orsenthil (dismissed):
Please see if you would like to update the README in appropriate section.

pkg/ipamd/datastore/data_store.go
```diff
@@ -978,8 +978,8 @@ func (ds *DataStore) GetENINeedsIP(maxIPperENI int, skipPrimary bool) *ENI {
 	ds.lock.Lock()
 	defer ds.lock.Unlock()
 	for _, eni := range ds.eniPool {
-		if skipPrimary && eni.IsPrimary {
-			ds.log.Debugf("Skip the primary ENI for need IP check")
+		if (skipPrimary && eni.IsPrimary) || eni.IsTrunk || eni.IsEFA {
```
jayanthvn, line 981: You need to remove `IsEFA` check... primary ENI can be EFA enabled and still have secondary Ips/prefixes for pods.

jayanthvn (changes requested):

pkg/ipamd/datastore/data_store.go
haouc: Why did we check `EFA` in the first place? Removing the check doesn't provide the check on non primary ENIs anymore. Is that what we want?

haouc (approved):
lgtm

pkg/ipamd/datastore/data_store.go
jdn5126: The AWS VPC CNI will not allocate EFA ENIs, but it can assign IP addresses to them after they are attached. Typically, customers attaching EFA ENIs already have all IP addresses assigned, but we still cannot exclude them in this check.

Shadowssong:
Hey @jdn5126 , has this been officially released yet?  I was trying to find it in the changelog or in the github diffs but I can't seem to find it anywhere.  We ran into this when upgrading our CNI from 1.15 -> 1.16 -> 1.17.  Thanks!

jdn5126:
@Shadowssong this PR was released with https://github.com/aws/amazon-vpc-cni-k8s/releases/tag/v1.16.4

Shadowssong:
@jdn5126 thanks!  We were on 1.17 and still experiencing the exact same issue described in the associated issue.  We reverted to 1.15 and the issue resolved.  I will open a support ticket to further debug this issue.

jayanthvn:
When you upgraded from 1.15 to 1.16, can you share which patch version of 1.15 and 1.16?

Shadowssong:
@jayanthvn We specified `v1.16.4-eksbuild.2` for 1.16 and then for 1.17 we did `v1.17.1-eksbuild.1`.  We were on 1.17 when we observed the behavior was occurring and reverting the EKS addon back to 1.15 resolved the issue once the daemonset restarted.

PR #2809, opened 2024-02-23
Fix VPC CNI weekly tests for Bottlerocket and kops

jdn5126:
**What type of PR is this?**
test fixes

N/A

This PR resolves failures in the weekly Bottlerocket and kops tests.
- Bottlerocket - tests were failing on cluster deletion, as `eksctl` was trying to drain nodes before deleting them. Nodes could not successfully drain due to the Coredns Pod Disruption Budget. The fix is to pass `--disable-nodegroup-eviction` on cluster deletion.
- kops - integration tests were failing. The combination of updating the kops version and removing bad/unnecessary CLI flags resolved these failures.  

Manually ran tests and validated that they passed.

No

No, Yes

No

No

```release-note

```


jaydeokar (approved):
lgtm

Issue #2807, opened 2024-02-22
v1.16.3 High CPU Usage on Medium Instance Size


aballman:
**What happened**:

I updated to v1.16.3. Specifically on `medium` instance size CPU usage by `aws-node` pod is maxed out

On a `t3.medium` (I have also seen this happen on a `c6g.medium`)
![Screenshot 2024-02-22 at 1 41 19 PM](https://github.com/aws/amazon-vpc-cni-k8s/assets/35912177/06fcf1ba-ae93-4ba7-90c6-a033f6aac327)

On an `m7i.2xlarge`
![Screenshot 2024-02-22 at 1 42 14 PM](https://github.com/aws/amazon-vpc-cni-k8s/assets/35912177/5db020b5-c7eb-42f4-bc33-26b5aa28ca34)

**Attach logs**

This log is spewing constantly in `/var/log/aws-routed-eni/ipamd.log` on the `medium` other instance sizes are typical volume of logs
```
{"level":"debug","ts":"2024-02-22T21:34:38.560Z","caller":"ipamd/ipamd.go:779","msg":"Node found \"ip-10-146-63-142.ec2.internal\" - no of taints - 0"}
{"level":"debug","ts":"2024-02-22T21:34:38.560Z","caller":"ipamd/ipamd.go:658","msg":"Skipping ENI allocation as the max ENI limit is already reached"}
{"level":"debug","ts":"2024-02-22T21:34:38.560Z","caller":"ipamd/ipamd.go:658","msg":"Starting to increase pool size"}
{"level":"debug","ts":"2024-02-22T21:34:38.560Z","caller":"ipamd/ipamd.go:779","msg":"Node found \"ip-10-146-63-142.ec2.internal\" - no of taints - 0"}
{"level":"debug","ts":"2024-02-22T21:34:38.560Z","caller":"ipamd/ipamd.go:658","msg":"Skipping ENI allocation as the max ENI limit is already reached"}
{"level":"debug","ts":"2024-02-22T21:34:38.560Z","caller":"ipamd/ipamd.go:658","msg":"Starting to increase pool size"}
{"level":"debug","ts":"2024-02-22T21:34:38.560Z","caller":"ipamd/ipamd.go:779","msg":"Node found \"ip-10-146-63-142.ec2.internal\" - no of taints - 0"}
[...]

PR #2810, opened 2024-02-23
Revert #2744 to prevent livelock when attempting to increase datastore pool

jdn5126:
**What type of PR is this?**
bug

https://github.com/aws/amazon-vpc-cni-k8s/issues/2807

This PR fixes a livelock condition that was reported in https://github.com/aws/amazon-vpc-cni-k8s/issues/2807. When the datastore pool is too low, we try to aggressively scale up to the desired state, hence the for loop. Under certain conditions, the datastore pool could return "too low" for a prolonged period of time (or indefinitely).

The change here is to revert https://github.com/aws/amazon-vpc-cni-k8s/pull/2744 so that only one scale up attempt happens during each iteration of the activity loop.

N/A

Ran all integration tests to ensure that they passed.

No

No, Yes

No

Yes

```release-note
Fix potential livelock while trying to increase the size of the datastore pool
```

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -660,11 +660,7 @@ func (c *IPAMContext) updateIPPoolIfRequired(ctx context.Context) {
 	log.Debugf("IP stats - total IPs: %d, assigned IPs: %d, cooldown IPs: %d", stats.TotalIPs, stats.AssignedIPs, stats.CooldownIPs)
 
 	if datastorePoolTooLow {
-		// Allow for rapid scale up to decrease time it takes for pod to retrieve an ip
-		// but conservative scale down to account for pod churn
-		for datastorePoolStillTooLow := datastorePoolTooLow; datastorePoolStillTooLow; datastorePoolStillTooLow, _ = c.isDatastorePoolTooLow() {
-			c.increaseDatastorePool(ctx)
-		}
+		c.increaseDatastorePool(ctx)
 	} else if c.isDatastorePoolTooHigh(stats) {
 		c.decreaseDatastorePool(decreaseIPPoolInterval)
 	}


pkg/ipamd/ipamd.go
```diff
@@ -654,11 +654,13 @@ func (c *IPAMContext) updateIPPoolIfRequired(ctx context.Context) {
...
 	log.Debugf("IP stats - total IPs: %d, assigned IPs: %d, cooldown IPs: %d", stats.TotalIPs, stats.AssignedIPs, stats.CooldownIPs)
 
+	// Scale up rapidly to decrease the time it takes for new pods to get an IP assigned. Scale down conservatively to account for pod churn.
 	if datastorePoolTooLow {
-		// Allow for rapid scale up to decrease time it takes for pod to retrieve an ip
-		// but conservative scale down to account for pod churn
-		for datastorePoolStillTooLow := datastorePoolTooLow; datastorePoolStillTooLow; datastorePoolStillTooLow, _ = c.isDatastorePoolTooLow() {
+		// Prevent livelock by limiting the number of attempts at scaling up.
+		count := 0
+		for datastorePoolStillTooLow := datastorePoolTooLow; datastorePoolStillTooLow && count < 3; datastorePoolStillTooLow, _ = c.isDatastorePoolTooLow() {
```
jayanthvn, line 661: Why don't we prevent this loop when MaxENIs are reached?

jdn5126: 3 was chosen to be more aggressive than our previous implementation, which used 1, while still being less than infinity. There is no other real reason or metrics behind it. We want to support rapid scale up, but there's no number that is ideal for all cases, so 3 seemed a safe choice.

pkg/ipamd/ipamd.go
```diff
@@ -659,11 +659,13 @@ func (c *IPAMContext) updateIPPoolIfRequired(ctx context.Context) {
...
```
orsenthil, line 666: Could you move number 3 to a constant with a proper name and comment, that could be adjusted later or removed if we desire? Keeping the number 3 here in the loop can confuse the future reader.

I am fine with the current logic.

orsenthil: A module level constant is fine too.

pkg/ipamd/ipamd.go
jayanthvn: I mean we have maxENI, maxIPsPerENI, maxPrefixesPerENI and TotalIPs(Assigned+cooldownIPs)...based on this why won't we decide if this optimization is needed or can be skipped? Just need to make sure we handle prefix delegation and Ipv6 scenarios...

pkg/ipamd/ipamd.go
jdn5126: Good call, I added a constant with a comment that hopefully explains its use well enough

pkg/ipamd/ipamd.go
jdn5126: It is theoretically possible to get `isDatastorePoolTooLow` to return false in all cases where we do not want to attach any more IPs, but that would add a lot more complexity to the function. What that doesn't cover, though, is real errors, like EC2 API errors that could happen for a short amount of time, which would similarly lead to high CPU consumption. I think we want to be defensive here due to the possibility of errors.

As for conditionally adding this optimization, that is also an option, but I again think that increases the complexity for little payoff. The goal is to do as much work as possible in one iteration of the activity loop without creating a livelock scenario. If you think it is better to remove this entirely and restore the old behavior, I think that is fine.

jayanthvn: Yeah I feel let's revert it and take it up offline.

jdn5126: The latest change reverts https://github.com/aws/amazon-vpc-cni-k8s/pull/2744

jayanthvn (approved):
lgtm..

orsenthil:
LGTM too.

PR #2841, opened 2024-03-12
Mount /run/xtables.lock as FileOrCreate in Helm chart

### LINKED ISSUE #2840: /run/xtables.lock created as directory when installed with Helm
**What happened**

The file `/run/xtables.lock` is created as a directory on the host machine. This breaks things that expect it to be a file, including `kube-proxy`. `kube-proxy` being unavailable causes the CNI to be unable to reach the API-server, and the node remains stuck in `NotReady` state.

**Attach logs**

Events from `kube-proxy pod`:

```
Events:
  Type     Reason       Age                   From     Message
  ----     ------       ----                  ----     -------
  Warning  FailedMount  4m2s (x60 over 109m)  kubelet  MountVolume.SetUp failed for volume "xtables-lock" : hostPath type check failed: /run/xtables.lock is not a file
```

I expect `/run/xtables.lock` to be created as a file.

This is a race condition that occurs fairly rarely, it depends on whether `kube-proxy` or the AWS CNI daemonset is started first. But it could probably be reproduced by:

1. Installing the AWS CNI from the Helm chart
2. Deleting `/run/xtables.lock` on the host
3. Restart the AWS CNI DaemonSet pod on the host

I'll create a PR with a fix shortly.

**Environment**:
- Kubernetes version (use `kubectl version`): `1.28.7`
- CNI Version: Helm Chart `1.13.0`
- OS (e.g: `cat /etc/os-release`): `ubuntu 22.04`
- Kernel (e.g. `uname -a`): `5.15.0-1055-aws`

kwohlfahrt:
**What type of PR is this?**
bug

fixes #2840 

This PR updates the Helm chart, to specify that `/run/xtables.lock` must be mounted as a file. See the issue for why this is a problem.

Forked the chart, and applied the change. After this, no nodes got stuck in `NotReady` state on startup.

No

Possibly - if `/run/xtables.lock` has already been created as a directory, the CNI pod will fail to start. However, if this occurs, `kube-proxy` will also fail to start, so the node will already not be functional.

I'm not sure there is any way to avoid this potential breakage though, but am open to suggestions!

No

No

```release-note
Always mount `/run/xtables.lock` as a file, even if it does not already exist
```


charts/aws-vpc-cni/templates/daemonset.yaml
```diff
@@ -183,6 +183,7 @@ spec:
       - name: xtables-lock
         hostPath:
           path: /run/xtables.lock
+          type: FileOrCreate
```
orsenthil, line 186: Should we have reasons to believe that `/run/xtables.lock` *file* should be have been created by the iptables package in the operating system?

There is note of caution in kubelet volume docs - https://kubernetes.io/docs/concepts/storage/volumes/

> Caution: The FileOrCreate mode does not create the parent directory of the file. If the parent directory of the mounted file does not exist, the pod fails to start. To ensure that this mode works, you can try to mount directories and files separately, as shown in the [FileOrCreate example](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath-fileorcreate-example) for hostPath.

kwohlfahrt: Just tested this on a fresh Ubuntu 22.04 VM. This file is not created until the first run of the `iptables-legacy` command.

The default `iptables` is a compatibility wrapper around nftables, which never creates this file at all, it is only created when interacting with actual iptables with `iptables-legacy`. So if nothing on startup calls the legacy iptables interface, the file will not exist.

kwohlfahrt: > If the parent directory of the mounted file does not exist, the pod fails to start. To ensure that this mode works, you can try to mount directories and files separately, as shown in the [FileOrCreate example](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath-fileorcreate-example) for hostPath.

I would expect `/run` to exist on all Linux hosts, as it is part of [FHS](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s15.html), so I don't think this is a risk? `kube-proxy` (as set up by `kubeadm`) already uses `FileOrCreate` for this path, so if `/run` doesn't exist, `kube-proxy` won't start anyway, and the node won't function properly.

orsenthil:
Verified the fix. LGTM.

orsenthil (approved):
LGTM

Kyslik:
Anyone using the AWS addon(s); you will need to wait until https://github.com/aws/eks-charts/pull/1101 is merged as that's the chart being used.

Issue #2887, opened 2024-04-18
Improve VPC CNI memory by reducing number of things it is caching 


GnatorX:
**What would you like to be added**:
Narrow down what VPC CNI is caching to reduce memory utilization in large clusters. Currently we see pretty high memory utilization and seems to scale with nodes. 

**Why is this needed**:
The current behavior is pretty problematic when cluster size gets large (5000+ nodes) causing us to increase memory request for the CNI even though it isn't necessary for the CNI to use all that memory.

I believe simply adding a new ByObject Filter on Node object is enough to reduce cache utilization. Since https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/k8sapi/k8sutils.go#L183 only gets the node object the CNI is running on.

orsenthil:
Thanks for the report and the Pull Request. Have you done any measurements with and without this change? Could you share the differences?

GnatorX:
Not yet. Will update once I have tested this

GnatorX:
I took a pprof of the issue. 
![Screenshot 2024-04-29 at 10 30 58 AM](https://github.com/aws/amazon-vpc-cni-k8s/assets/6425977/d5d4a776-825b-4508-a308-79446c928f4b)

It seems like the issue is with the stream watcher is consuming memory during cluster size increase. It seems to require quite a bit of memory in order to process all nodes and store it in the memory. Even though the memory consumption isn't very high, its still unnecessary to store all node information in cache. 

I need to re-test this with my change however I do believe the real solution is to avoid performing list watch against all nodes and only watch for node events specific to the CNI.

orsenthil:
>  K8s caches which usesList + watches on startup are extremely expensive calls

>  Even though the memory consumption isn't very high, its still unnecessary to store all node information in cache.

> I do believe the real solution is to avoid performing list watch against all nodes and only watch for node events specific to the CNI.

It is pretty standard for k8s client calls to use the cached client. It will be good to measure difference in the memory usage *and the performance* of the various operations in the large clusters before we decide to not use the cache.

With your changes, if you see any different in both memory and performance, please share an update here.

GnatorX:
> It is pretty standard for k8s client calls to use the cached client. It will be good to measure difference in the memory usage and the performance of the various operations in the large clusters before we decide to not use the cache.

Agreed. 

[...]

PR #2888, opened 2024-04-18
Add byobject filter on nodes

GnatorX:
**What type of PR is this?**

improvement

https://github.com/aws/amazon-vpc-cni-k8s/issues/2887

This PR adds a filter to reduce the number of node object VPC CNI watches for to just the node it is managing.

Tested on our cluster of 3-4k node by @dl-stripe 

No
No.

No

No


orsenthil:
Closing per the discussion here - https://github.com/aws/amazon-vpc-cni-k8s/issues/2887

dl-stripe:
A couple more datapoints in favor of this change. We have a few large node count clusters with a higher rate of churn for these nodes. The number of Node events pushed to the daemonset was causing excesive bandwidth usage. Adding the filter drastically reduced the number of events distributed to the system. Similarly the number of bytes pushed fell drastically (sharing a screenshot without absolute numbers, happy to share in private, but you can see the relative drop in bytes processed).

```promql
sum by (cluster, kind) (rate(apiserver_watch_events_total{kind="Node"}[5m]))
```
image
image

jayanthvn:
@GnatorX - The PR is still in draft mode. Please feel free to move to review. Also can you run `make format` ?

GnatorX:
ya i can do that

GnatorX:
ran `make format`

jayanthvn:
Updated the branch and started the test workflow..

GnatorX:
Thanks

pkg/k8sapi/k8sutils.go
```diff
@@ -37,7 +37,11 @@ func getIPAMDCacheFilters() map[client.Object]cache.ByObject {
 		return map[client.Object]cache.ByObject{
 			&corev1.Pod{}: {
 				Field: fields.Set{"spec.nodeName": nodeName}.AsSelector(),
-			}}
+			},
```
orsenthil, line 40: The previous two related to changes to the cache improvements were these.

- https://github.com/aws/amazon-vpc-cni-k8s/pull/1419/files#diff-f48443f100e2f9072cbb1924ed0a9090831ade5986d10535641ad5c0a63f51b4
- https://github.com/aws/amazon-vpc-cni-k8s/commit/ea10123e8105df97d2b2fd9916fc7d255458075c#diff-f48443f100e2f9072cbb1924ed0a9090831ade5986d10535641ad5c0a63f51b4

I am trying to understand why the only `Pod{}` to the filter here in the first place.

orsenthil:
@dl-stripe - What is the LHS value in the first graph here - https://github.com/aws/amazon-vpc-cni-k8s/pull/2888#issuecomment-2406032407

dl-stripe:
> @dl-stripe - What is the LHS value in the first graph here - [#2888 (comment)](https://github.com/aws/amazon-vpc-cni-k8s/pull/2888#issuecomment-2406032407)

It's the number of Node events pushed through informers (based off `apiserver_watch_events_total`) over a 5 minute period for 3 sample clusters we run the AWS VPC CNI in. They're all large clusters but have slightly different node counts.

orsenthil:
When we were using node labels to signal security group for pods feature - https://docs.aws.amazon.com/eks/latest/userguide/security-groups-pods-deployment.html

this change for letting kubernetes client cache the node calls, could have had a impact, and dependent on the cache invalidation by the client when the labels change. Right now since, we don't depend on the CNI node labels, this change looks good to me.

@dl-stripe  / @GnatorX  - Do you have Security Groups for Pods in your cluster under test and did you see any impact this change?

GnatorX:
> this change for letting kubernetes client cache the node calls, could have had a impact, and dependent on the cache invalidation by the client when the labels change. Right now since, we don't depend on the CNI node labels, this change looks good to me.

This is filter down the call for nodes when informer performs the  list + watch. It shouldn't affect any label changes. It should narrow the informer cache to just the node the VPC CNI is running on and ignore all other node's updates and events

orsenthil:
> This is filter down the call for nodes when informer performs the list + watch. It shouldn't affect any label changes. It should narrow the informer cache to just watch for the node the VPC CNI is running on and ignore all other node's updates and events

Got it. This makes sense looks correct and helpful too.

orsenthil (approved):
This change LGTM. 

The filtermap used here - https://github.com/aws/amazon-vpc-cni-k8s/blob/d49c8a3f48b70f408713c2a5caced96c4b426be6/pkg/k8sapi/k8sutils.go#L68 and this will add a filter to list/watch events for node operations, restricting it to node that CNI runs on. 

This correct. 

I understand adding an additional cache field can increase the memory usage by VPC CNI, but it did not per here - https://github.com/aws/amazon-vpc-cni-k8s/issues/2887

We will verify memory usage VPC CNI after this change before the release of this change.

GnatorX:
The reason initially I wasn't able to get the memory improvement from this change is because our internal setup vendored the code, so when I made the change it didn't actually take effect. 
At a high level what is happening here is we are reducing the amount of data VPC CNI's informer would process and save.
https://github.com/kubernetes/sample-controller/raw/master/docs/images/client-go-controller-interaction.jpeg
The field selector changes the reflector's list & watch to just the node it is running on. This changes 2 things.

1) We no longer get watch events from other nodes this reduces network calls and CPU processing
2) We no longer cache all node information in the cluster within the store. This reduce memory consumption

GnatorX:
https://github.com/kubernetes-sigs/controller-runtime/blob/main/designs/use-selectors-at-cache.md

pkg/k8sapi/k8sutils.go
GnatorX: I suspect it wasn't as obvious because this requires large clusters to show the increase in memory and networking bandwidth. We only saw this because our cluster is 4k node count

orsenthil:
Thank you for the explanation and the links to behavior. This change looks good and helpful.

PR #2895, opened 2024-04-25
Filter Managed ENI.

orsenthil:
**What type of PR is this?**

_Feature_

**Which issue does this PR fix?**:

https://github.com/aws/amazon-vpc-cni-k8s/issues/2875

**What does this PR do / Why do we need it?**:

If the SG reconciler runs before the ENI/IP reconciler it will modify the security groups as the ENI/IP reconciler hasn't had a chance to check the tags on the ENI yet.

Without relying on cache, when the SG reconciler is run, it will not update the ENI with the node.k8s.amazonaws.com/no_manage: true tag

**Testing done on this change**:

- [x] Unit Tests are successful.
- [x] Integration Tests need to be verified. 

Current Mocks do not verify this.

### Testing Done.

* Mock / Unit Tests are successful.
* Canary Tests are OK.
image

* Unmanaged ENI's don't have a coveraged - Reopened this - https://github.com/aws/amazon-vpc-cni-k8s/issues/2474

**Manual Testing*

- Attached a new ENI to an existing instance with `no_manage: true`
- Since the new ENIs are gotten from the datastore, noticed that RefreshSG followed the `no_manage`: tag, and skipped changing any Security Groups for newly added ENI with `no_manage` tag.

Verified with a debugging output against a custom CNI build

1. `eni-025db9eced8c995b0` - ENI with no_manage tag.

2. IPAM Marking it as unmanaged.

```
{"level":"debug","ts":"2024-05-07T01:25:19.021Z","caller":"ipamd/ipamd.go:430","msg":"Marking ENI eni-025db9eced8c995b0 as being unmanaged"}
{"level":"debug","ts":"2024-05-07T01:25:19.021Z","caller":"ipamd/ipamd.go:431","msg":"Skipping ENI eni-025db9eced8c995b0: since it is unmanaged"}
```

3. Refresh SG Skipping (Temporarily added a debug statement).

```
[root@ip-192-168-51-243 aws-routed-eni]# tail -f ipamd.log |grep DEBUG
{"level":"debug","ts":"2024-05-07T01:41:22.262Z","caller":"ipamd/ipamd.go:510","msg":"DEBUG: ADDING ENI FROM DATASTORE eni-09c8c893dd7b75844"}
{"level":"debug","ts":"2024-05-07T01:41:22.262Z","caller":"ipamd/ipamd.go:510","msg":"DEBUG: ADDING ENI FROM DATASTORE eni-039b6089c57710b3e"}
[...]

PR #2905, opened 2024-05-03
Fix go dependency with go-cose library.

orsenthil:
**What type of PR is this?**

This indirect dependency was causing reference failures in IntelliJ.

go list -modfile=go.mod -m -json -mod=go.mod all #gosetup

fails with

> go: github.com/veraison/go-cose@v1.2.0: invalid version: unknown revision v1.2.0

As older modules are retracted

https://pkg.go.dev/github.com/veraison/go-cose?tab=versions

And this needs to be pinned to v1.1.0

So that go list is successful.

```
{
	"Path": "github.com/veraison/go-cose",
	"Version": "v1.2.0",
	"Replace": {
		"Path": "github.com/veraison/go-cose",
		"Version": "v1.1.0",
		"Time": "2023-05-18T12:32:47Z",
		"GoMod": "/Users/senthilx/go/pkg/mod/cache/download/github.com/veraison/go-cose/@v/v1.1.0.mod",
		"GoVersion": "1.18"
	},
	"Indirect": true,
	"GoMod": "/Users/senthilx/go/pkg/mod/cache/download/github.com/veraison/go-cose/@v/v1.1.0.mod",
	"GoVersion": "1.18"
}
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -180,3 +180,6 @@ replace golang.org/x/crypto => golang.org/x/crypto v0.17.0
 
 // Cannot be removed until all dependencies use net library v0.23.0 or higher
 replace golang.org/x/net => golang.org/x/net v0.23.0
+
+// Version of go-cose v1.2.0 and v1.2.1 have been deprecated in favor v1.1.0
+replace github.com/veraison/go-cose => github.com/veraison/go-cose v1.1.0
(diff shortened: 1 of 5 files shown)


orsenthil:
No need. As the change was introduced in https://github.com/aws/amazon-vpc-cni-k8s/pull/2895

PR #2919, opened 2024-05-20
Switch to counter type metic for awscni_no_available_ip_addresses

liptanbiswas:
**What type of PR is this?**

bug

The `NoAvailableIPAddrs` metic is being used couple times [here](https://github.com/aws/amazon-vpc-cni-k8s/blob/3daa258714406f9d321d9c1c379bb533268d6615/pkg/ipamd/datastore/data_store.go#L663) and [here](https://github.com/aws/amazon-vpc-cni-k8s/blob/3daa258714406f9d321d9c1c379bb533268d6615/pkg/ipamd/datastore/data_store.go#L733),  both to increment the gauge. The metric is advertised as gauge but there's no calls for decrementing the gauge and its behaving like a counter. Since its behaving like a counter, let make it counter instead.  

We need the PR so we correctly identify the metric as counter.

```release-note
Fixed metric type for `NoAvailableIPAddrs` to accurately reflect its usage as a counter rather than a gauge.
```


orsenthil (approved):
LGTM.

PR #2922, opened 2024-05-21
Skip Soak Test while running other tests.

orsenthil:
### Skip Soak Test while running other tests.

SOAK_TEST is run separately and on it's own. Running it with other tests will consume time and will cause timeout for other tests.


jchen6585 (approved):
lgtm

PR #2925, opened 2024-05-24
Expose network policy log file location to be configured using helm

orsenthil:
Expose network policy log file location to be configured using helm chart values.

### DIFF
--- a/charts/aws-vpc-cni/values.yaml
+++ b/charts/aws-vpc-cni/values.yaml
@@ -43,7 +43,7 @@ nodeAgent:
     privileged: true
   enableCloudWatchLogs: "false"
   enablePolicyEventLogs: "false"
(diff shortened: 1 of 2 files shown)


charts/aws-vpc-cni/values.yaml
```diff
@@ -43,6 +43,7 @@ nodeAgent:
     privileged: true
   enableCloudWatchLogs: "false"
   enablePolicyEventLogs: "false"
+  policyEventLogFileLocation: "/var/log/aws-routed-eni/network-policy-agent.log"
```
achevuru, line 46: Log file will have both NP agent logs & policy event logs and so I think it is better we use a generic name such as `networkPolicyAgentLogFileLocation`

achevuru (approved):

PR #2934, opened 2024-05-29
Helpful Make target to login to public ECR.

orsenthil:
We often need to login to ECR to do the build. This make target can used to login to public ecr

```
make ecr-public-login
```

Not adding any dependency on purpose as it goal is to use this manually.


jaydeokar (approved):
lgtm

PR #2946, opened 2024-06-05
Add extraEnv for add additional env from configmap or secrets to daem…

gawsoftpl:
**What type of PR is this?**
improvement

Add feature to inject env from configmap or secrets

Add feature to inject env from configmap or secrets

no

no

no

```release-note

```


orsenthil:
Why do we need this extraEnv vars? CNI plugin isn't customizable by customers or application other than what is exposed by the plugin.

charts/aws-vpc-cni/values.yaml
```diff
@@ -88,6 +88,25 @@ env:
...
+#   valueFrom:
+#     configMapKeyRef:
+#       name: example-config
+#       key: ENV_VAR2
+# - name: SECRET_VAR1
+#   valueFrom:
+#     secretKeyRef:
+#       name: example-secret
+#       key: SECRET_VAR1
+extraEnv: []
```
orsenthil, line 107: Please move the above example comment below line 107.
You could keep it as single example instead 3, and comment it as
> Optional: Additional environment variables for helm chart.

orsenthil: Also, as I understand, this is helm chart only feature. This wont be supported by self managed addon (applying yaml directly, in which case, they will have to edit the environment variables directly and with MAO.

The cost of maintaining this difference needs to be considered.


If it is strictly not required, I would consider dropping this, if any other workaround is available with helm chart.

orsenthil (approved):
It seems alright to bring the change here.

LGTM.

PR #2951, opened 2024-06-11
use __MTU__ variable for IPv4 egress-cni too

### LINKED ISSUE #2949: Setting MTU on ipv4 egress on a ipv6 only cluster
**What happened**:

We are trying to change the MTU values on all interfaces on our ipv6 only cluster with ipv4 egress. When we set the AWS_VPC_ENI_MTU value, this only affects the eth0 interface but not the v4if0 interface where ipv4 egress is supported. Which values do we need to set so the MTU values are persisted on both interfaces?

**Environment**:
- Kubernetes version (use `kubectl version`): v1.30.0-eks-036c24b
- CNI Version: v1.18.1-eksbuild.3
- Kernel (e.g. `uname -a`): 5.10.213-201.855.amzn2.x86_64

benben:
**What type of PR is this?**
bug

fixes #2949 

When setting `AWS_VPC_ENI_MTU` and/or `POD_MTU` to `"1500"` before, the network interfaces in a pod look like this:
```
$ ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
    inet6 ::1/128 scope host
3: eth0@if17: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default
    link/ether ae:97:27:f0:0e:2a brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet6 2600:<snip>::1/128 scope global
    inet6 fe80::ac97:27ff:fef0:e2a/64 scope link
4: v4if0@if18: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9001 qdisc noqueue state UP group default
    link/ether b2:7c:f4:d2:81:7c brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 169.254.172.9/22 brd 169.254.175.255 scope global v4if0
    inet6 fe80::b07c:f4ff:fed2:817c/64 scope link
```
after this change, the interfaces correctly show 1500 for both:
```
$ ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
    inet6 ::1/128 scope host
3: eth0@if17: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default
    link/ether ae:97:27:f0:0e:2a brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet6 2600:<snip>::1/128 scope global
    inet6 fe80::ac97:27ff:fef0:e2a/64 scope link
4: v4if0@if18: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default
    link/ether b2:7c:f4:d2:81:7c brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 169.254.172.9/22 brd 169.254.175.255 scope global v4if0
    inet6 fe80::b07c:f4ff:fed2:817c/64 scope link
```

It runs in our test environments and a single production region for ~8h. This is roughly 100 nodes.

no

People having set something else than MTU 9001 will also get the new MTU for the IPv4 interface.

Not sure, we rebuild the container and init container and deployed them at once.

People having set something else than MTU 9001 will also get the new MTU for the IPv4 interface.

```release-note
- fix MTU setting for the egress-cni: egress-cni is now also picking up on a custom `AWS_VPC_ENI_MTU` setting
```

### DIFF
--- a/misc/10-aws.conflist
+++ b/misc/10-aws.conflist
@@ -15,7 +15,7 @@
     {
       "name": "egress-cni",
       "type": "egress-cni",
-      "mtu": "9001",
+      "mtu": "__MTU__",
       "enabled": "__EGRESSPLUGINENABLED__",
       "randomizeSNAT": "__RANDOMIZESNAT__",
       "nodeIP": "__NODEIP__",


misc/10-aws.conflist
```diff
@@ -15,7 +15,7 @@
     {
       "name": "egress-cni",
       "type": "egress-cni",
-      "mtu": "9001",
```
orsenthil, line 18: Thanks for the contribution, @benben .

This change can covered by a test case addition to 
https://github.com/aws/amazon-vpc-cni-k8s/blob/master/cmd/aws-vpc-cni/main_test.go#L51

looks like we didn't cover this initially when we had the opportunity to do so. Please see if you can add a test case for this in the above linked file.

orsenthil (approved):
LGTM

misc/10-aws.conflist
orsenthil: We will bring this test separately.

PR #2952, opened 2024-06-12
Upgrade to latest versions of GitHub actions

electrocucaracha:
**What type of PR is this?**
dependency update

NA

It seems like `dependabot` is not detecting the latest versions released by the GitHub actions used by this project. This change upgrades to the latest stable versions, resulting in fixing issues and vulneratibilties on the CI jobs.

This testing has to be performed on the CI, which limits the possibility to run local tests.

No

No

No

No

```release-note

```

### DIFF
--- a/.github/workflows/issue-stale-pr.yaml
+++ b/.github/workflows/issue-stale-pr.yaml
@@ -12,7 +12,7 @@ jobs:
   stale:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/stale@main
+      - uses: actions/stale@28ca1036281a5e5922ead5184a1bbf96e5fc984e # refs/tags/v9.0.0
         id: stale
         with:
           ascending: true
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -1,0 +1,32 @@
+---
+# SPDX-license-identifier: Apache-2.0
+##############################################################################
+# Copyright (c) 2024
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Apache License, Version 2.0
+# which accompanies this distribution, and is available at
+# http://www.apache.org/licenses/LICENSE-2.0
+##############################################################################
+name: Scheduled Update Versions
+# yamllint disable-line rule:truthy
+on:
+  schedule:
+    - cron: '0 0 * * 5'
+  workflow_dispatch:
+jobs:
+  check-versions:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # refs/tags/v4.1.7
+      - uses: technote-space/create-pr-action@91114507cf92349bec0a9a501c2edf1635427bc5 # refs/tags/v2.1.4
+        with:
+          EXECUTE_COMMANDS: |
+            gh_actions=$(grep -r "uses: [a-z\-]*/[\_a-z\-]*@" .github/workflows/ | sed 's/@.*//' | awk -F ': ' '{ print $3 }' | sort | uniq)
+            for action in $gh_actions; do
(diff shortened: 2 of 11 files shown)


orsenthil:
@electrocucaracha - Was this auto generated or manually updated? How can I verify these are latest? Also for ubuntu why do you suggest using the a version than main?

.github/workflows/issue-stale-pr.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
   stale:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/stale@main
+      - uses: actions/stale@v9
```
orsenthil, line 15: Why this is a versioned tag instead of main?

electrocucaracha:
> @electrocucaracha - Was this auto generated or manually updated? How can I verify these are latest? Also for ubuntu why do you suggest using the a version than main? 

This was manually generated. Usually, these changes are detected by `dependabot`, but it seems not working as expected.

.github/workflows/issue-stale-pr.yaml
electrocucaracha: It's a good practice to not use the main branch, so I suggest using the latest stable

orsenthil:
@electrocucaracha - would you be interested in pinning the commit hash instead of the tags of these packages; there is a reference to use commit hash here.

electrocucaracha:
> @electrocucaracha - would you be interested in pinning the commit hash instead of the tags of these packages; there is a reference to use commit hash here.

Ok, I can give it a try. Maybe this can be achieved configuring the `dependabot` [configuration](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/.github/dependabot.yml), properly, for now it requires adding this section:

```
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "monthly"
```

orsenthil:
Yes, we can add that section in the configuration too.

electrocucaracha:
@orsenthil Apparently `dependabot` doesn't support commit hashes feature, and some [third party tools](https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash) offer that lack of functionality. But this can be easily created with the following bash script.

```bash
gh_actions=$(grep -r "uses: [a-z\-]*/[a-z\-]*@" .github/workflows/ | sed 's/@.*//' | awk -F ': ' '{ print $3 }' | sort | uniq)

for action in $gh_actions; do
    commit_hash=$(git ls-remote --tags "https://github.com/$action" | awk '{ print $NF,$0 }' | sort -k1,1 -n | cut -f2- -d' ' | grep -oh '.*refs/tags/v[0-9]*$' | tail -1 | awk '{ printf "%s # %s\n",$1,$2 }')
    egrep -lRZ "uses: $action@" .github/workflows/ | xargs -0 -l sed -i -e "s|uses: $action@.*|uses: $action@$commit_hash|g"
done
```

Maybe, those bash scripts instructions can be included in a GitHub action.

orsenthil:
@electrocucaracha If the bash scripts help update to the latest action commit sha, we can include them as a makefile target as a first step. We want to control on when to update

electrocucaracha:
> @electrocucaracha If the bash scripts help update to the latest action commit sha, we can include them as a makefile target as a first step. We want to control on when to update

Well, I had in mind to use it in a GitHub action that uses this https://github.com/marketplace/actions/create-pr-action

orsenthil:
>  I had in mind to use it in a GitHub action that uses this https://github.com/marketplace/actions/create-pr-action

Sounds good to me.

orsenthil (approved):

.github/workflows/update.yml
```diff
@@ -0,0 +1,35 @@
...
+name: Scheduled Update Versions
+# yamllint disable-line rule:truthy
+on:
+  schedule:
+    - cron: '0 0 * * 5'
+  workflow_dispatch:
+jobs:
+  check-versions:
+    permissions:
+      contents: write  # for technote-space/create-pr-action to push code
```
orsenthil, line 19: I am wondering if write action is required here?

orsenthil (changes requested):
Is write to repo authorization required for the GH action?

.github/workflows/update.yml
electrocucaracha: I think that was required for [protected tags](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-tag-protection-rules), I'll remove it to confirm that.

orsenthil (approved):
LGTM.

Issue #2908, opened 2024-05-06
Create secondary ENI when previous ENI isn't full due to lack of IPs in subnet - enhanced subnet discovery


jchen6585:
**What would you like to be added**:
When enhanced subnet discovery is enabled, we should create a secondary ENI using a new subnet (if there is one available) when the previous ENI is failing to add more IPs due to lack of IPs in subnet.

**Why is this needed**:
This will prevent IPAMD from being stuck trying to allocate IPs to the previous ENI even if the subnet is out of IPs.

yieon-lyon:
If you provide this method, I think there will be any problems with IP loss for existing instances.
But before understanding this, we must first carefully manage subnet configuration. with cidr ...

orsenthil:
This is addressed in https://github.com/aws/amazon-vpc-cni-k8s

dougbyrne:
I think the previous link was meant to be https://github.com/aws/amazon-vpc-cni-k8s/pull/2954

PR #2954, opened 2024-06-13
Subnet Discovery - Unfilled ENI fix

jchen6585:
**What type of PR is this?**

#2908 

If subnet discovery is on, we will proceed to create a new ENI if existing ENI is unfilled but the subnet is out of IPs.

Manual Testing: Will be pasted offline
Integration: CNI, IPAMD, Custom Networking

- Screenshot 2024-06-19 at 2 42 46 PM
- Screenshot 2024-06-19 at 3 15 30 PM
- Screenshot 2024-06-19 at 4 32 15 PM

No

No, Yes

No

No

```release-note

```

### DIFF
--- a/pkg/ipamd/datastore/data_store.go
+++ b/pkg/ipamd/datastore/data_store.go
@@ -973,8 +973,9 @@ func (e *ENI) hasPods() bool {
 	return e.AssignedIPv4Addresses() != 0
 }
 
-// GetENINeedsIP finds an ENI in the datastore that needs more IP addresses allocated
-func (ds *DataStore) GetENINeedsIP(maxIPperENI int, skipPrimary bool) *ENI {
+// GetAllocatableENIs finds ENIs in the datastore that needs more IP addresses allocated
+func (ds *DataStore) GetAllocatableENIs(maxIPperENI int, skipPrimary bool) []*ENI {
+	var enis []*ENI
 	ds.lock.Lock()
 	defer ds.lock.Unlock()
 	for _, eni := range ds.eniPool {
@@ -985,10 +986,10 @@ func (ds *DataStore) GetENINeedsIP(maxIPperENI int, skipPrimary bool) *ENI {
 		if len(eni.AvailableIPv4Cidrs) < maxIPperENI {
 			ds.log.Debugf("Found ENI %s that has less than the maximum number of IP/Prefixes addresses allocated: cur=%d, max=%d",
 				eni.ID, len(eni.AvailableIPv4Cidrs), maxIPperENI)
-			return eni
+			enis = append(enis, eni)
 		}
 	}
-	return nil
+	return enis
 }
 
 // RemoveUnusedENIFromStore removes a deletable ENI from the data store.
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -915,6 +915,7 @@ func (c *IPAMContext) tryAssignCidrs() (increasedPool bool, err error) {
 // For an ENI, try to fill in missing IPs on an existing ENI.
 // PRECONDITION: isDatastorePoolTooLow returned true
 func (c *IPAMContext) tryAssignIPs() (increasedPool bool, err error) {
+
 	// If WARM_IP_TARGET is set, only proceed if we are short of target
 	short, _, warmIPTargetsDefined := c.datastoreTargetState(nil)
 	if warmIPTargetsDefined && short == 0 {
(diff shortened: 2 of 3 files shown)


pkg/ipamd/datastore/data_store.go
```diff
@@ -974,21 +974,28 @@ func (e *ENI) hasPods() bool {
...
-			return eni
+			if useSubnetDiscovery {
+				return eni, len(ds.eniPool)
+			}
+			return eni, 1
 		}
 	}
-	return nil
+	// Returning 1 since we want one iteration of IP allcocation if subnet discovery is off
+	return nil, 1
```
orsenthil, line 998: Can this scenario be covered using unit-test (return 1 iteration of ip allocation). Same for line 994.

pkg/ipamd/ipamd.go
```diff
@@ -915,6 +915,8 @@ func (c *IPAMContext) tryAssignCidrs() (increasedPool bool, err error) {
 // For an ENI, try to fill in missing IPs on an existing ENI.
 // PRECONDITION: isDatastorePoolTooLow returned true
 func (c *IPAMContext) tryAssignIPs() (increasedPool bool, err error) {
+	insufficientENI := make(map[*datastore.ENI]bool)
```
jayanthvn, line 918: Each time we call `tryAssignIps` or `tryAssignPrefixes` we create a new map.. `c.dataStore.GetENINeedsIP(c.maxIPsPerENI, c.useCustomNetworking, c.useSubnetDiscovery, insufficientENI)` so here, we are always passing new map isn't it or am I missing something?

pkg/ipamd/ipamd.go
```diff
@@ -928,45 +930,52 @@ func (c *IPAMContext) tryAssignIPs() (increasedPool bool, err error) {
...
-	if eni != nil && len(eni.AvailableIPv4Cidrs) < c.maxIPsPerENI {
-		currentNumberOfAllocatedIPs := len(eni.AvailableIPv4Cidrs)
-		// Try to allocate all available IPs for this ENI
-		resourcesToAllocate := min((c.maxIPsPerENI - currentNumberOfAllocatedIPs), toAllocate)
-		output, err := c.awsClient.AllocIPAddresses(eni.ID, resourcesToAllocate)
-		if err != nil && !containsPrivateIPAddressLimitExceededError(err) {
-			log.Warnf("failed to allocate all available IP addresses on ENI %s, err: %v", eni.ID, err)
-			// Try to just get one more IP
-			output, err = c.awsClient.AllocIPAddresses(eni.ID, 1)
+	eni, numOfENIs := c.dataStore.GetENINeedsIP(c.maxIPsPerENI, c.useCustomNetworking, c.useSubnetDiscovery, insufficientENI)
```
jayanthvn, line 933: Why don't we return just an array of eni or an empty array?

jayanthvn: Nit: GetAllocatableENIs or something like that? since it is returning more than one ENI

jchen6585: Ack

pkg/ipamd/ipamd.go
```diff
@@ -928,45 +929,50 @@ func (c *IPAMContext) tryAssignIPs() (increasedPool bool, err error) {
...
+			output, err := c.awsClient.AllocIPAddresses(eni.ID, resourcesToAllocate)
 			if err != nil && !containsPrivateIPAddressLimitExceededError(err) {
-				ipamdErrInc("increaseIPPoolAllocIPAddressesFailed")
-				return false, errors.Wrap(err, fmt.Sprintf("failed to allocate one IP addresses on ENI %s, err ", eni.ID))
+				log.Warnf("failed to allocate all available IP addresses on ENI %s, err: %v", eni.ID, err)
+				// Try to just get one more IP
+				output, err = c.awsClient.AllocIPAddresses(eni.ID, 1)
+				if err != nil && !containsPrivateIPAddressLimitExceededError(err) {
+					ipamdErrInc("increaseIPPoolAllocIPAddressesFailed")
+					if c.useSubnetDiscovery && containsInsufficientCIDRsOrSubnetIPs(err) {
```
orsenthil, line 945: > containsInsufficientCIDRsOrSubnetIPs(err)

In the testing, if this condition is covered for all scenarios (including prefix-delegation), then we are good.

You could paste the output of the testing in the PR description too.

orsenthil (approved):
LGTM

pkg/ipamd/datastore/data_store.go
```diff
@@ -985,10 +986,14 @@ func (ds *DataStore) GetENINeedsIP(maxIPperENI int, skipPrimary bool) *ENI {
 		if len(eni.AvailableIPv4Cidrs) < maxIPperENI {
 			ds.log.Debugf("Found ENI %s that has less than the maximum number of IP/Prefixes addresses allocated: cur=%d, max=%d",
 				eni.ID, len(eni.AvailableIPv4Cidrs), maxIPperENI)
-			return eni
+			enis = append(enis, eni)
+			if useSubnetDiscovery {
```
jayanthvn, line 990: Wondering should we even do this? we could always get all ENIs and for non-subnet-Discovery scenario we just pick eni[0] in the caller.. also running the loop should be fine since the number of ENis is always capped by the max enis supported per instance..

orsenthil: Either seems fine with me. The `useSubnetDiscovery` seems to scope operation of getting all subnets to the specific feature, which in other words, closer to previous behavior.

jchen6585: Now that I think about it we don't need this conditional.. Non-subnet discovery would only have one unfilled ENI at a time since we only have one subnet. We can just return all unfilled ENIs

orsenthil: Either seems fine. 

> running the loop should be fine since the number of ENis is always capped by the max enis supported per instance

And we aren't doing any API call here. So, seems fine.

jchen6585: In the case of custom networking removing this conditional would work as well (e.g. primary ENI is unfilled and subnet is exhausted, secondary ENI has space for more IPs and is allocated on subnet from ENIConfig). The caller would attempt to allocate IPs on primary ENI --> fail --> attempt on secondary ENI (ENIConfig Subnet)

pkg/ipamd/datastore/data_store.go
```diff
@@ -973,8 +973,9 @@ func (e *ENI) hasPods() bool {
 	return e.AssignedIPv4Addresses() != 0
 }
 
-// GetENINeedsIP finds an ENI in the datastore that needs more IP addresses allocated
-func (ds *DataStore) GetENINeedsIP(maxIPperENI int, skipPrimary bool) *ENI {
+// GetAllocatableENIs finds ENIs in the datastore that needs more IP addresses allocated
+func (ds *DataStore) GetAllocatableENIs(maxIPperENI int, skipPrimary bool, useSubnetDiscovery bool) []*ENI {
```
orsenthil, line 977: The flag `useSubnetDiscovery` is not used anymore.

orsenthil (approved):
LGTM

PR #2955, opened 2024-06-14
Update test options default value and help.

orsenthil:
We use `&options.NgNameLabelKey` to identify *nodes* and not node group. See:  https://github.com/search?q=repo%3Aaws%2Famazon-vpc-cni-k8s%20f.Options.NgNameLabelKey&type=code 

This has been confusing to me, and changing the default value and help message. Not changing option name itself (for now).


test/framework/options.go
```diff
@@ -56,8 +56,8 @@ func (options *Options) BindFlags() {
 	flag.StringVar(&options.ClusterName, "cluster-name", "", `Kubernetes cluster name (required)`)
 	flag.StringVar(&options.AWSRegion, "aws-region", "", `AWS Region for the kubernetes cluster`)
 	flag.StringVar(&options.AWSVPCID, "aws-vpc-id", "", `AWS VPC ID for the kubernetes cluster`)
-	flag.StringVar(&options.NgNameLabelKey, "ng-name-label-key", "eks.amazonaws.com/nodegroup", "label key used to identify nodegroup name")
-	flag.StringVar(&options.NgNameLabelVal, "ng-name-label-val", "", "label value with the nodegroup name")
+	flag.StringVar(&options.NgNameLabelKey, "ng-name-label-key", "kubernetes.io/os", "label key used to identify node")
+	flag.StringVar(&options.NgNameLabelVal, "ng-name-label-val", "linux", "label value with the node")
```
achevuru, line 60: So, we currently ignore the value field for `eks.amazonaws.com/nodegroup`? and solely rely on the key string?

orsenthil: Yes, we ignore the default value here and set these options (both key and value) in our test scripts like these 
- https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/run-canary-test.sh#L22
- https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/run-cni-release-tests.sh#L25

The important thing to note is these are not *node group* identifier key and values, they are *node* identifier key value pairs. Ideally, we have to change the option name from `options.NgNameLabelKey` to `options.NodeNameLabelKey` and similar for value too.

orsenthil: I was initially thinking this was designed for nodegroup selector which can be used for nodes, but looking at all references of the code, this is designed for node selector, hence this change seems helpful.

achevuru (approved):

PR #2956, opened 2024-06-15
Use ECR Mirror for Curl Test Image.

orsenthil:
**What type of PR is this?**
Testing

I realized for all test images, with curl left out, we were using Internal ECR registry.
curlimages/curl was being downloaded from dockerhub, and it was facing issues with tests running in the china region.

This change ensures that the curl images used in the tests are available from the test-registry similar to other images images.

1. curl images are published to internal registries (including ECR registry for china).

2. This PR will be merged only after internal registry sync is complete, and test job can be launched using this image; and also tested against beta.

**Do not merge before the last two steps as it can result in test failures**


M00nF1sh (dismissed):

jaydeokar (dismissed):
lgtm

jaydeokar (approved):
lgtm

Issue #2963, opened 2024-06-19
valid openAPIV3 Schema


cyberox:
**What would you like to be added**:
Update the ENIConfig CRD to include a valid openAPIV3 Schema

**Why is this needed**:
When creating ENIConfig resources using Terraform manifest resources, I receive a warning:
```
│ Warning: This custom resource does not have an associated OpenAPI schema.
│
│ We could not find an OpenAPI schema for this custom resource. Updates to this resource will cause a forced replacement.
```

Terraform issue: https://github.com/hashicorp/terraform-provider-kubernetes/issues/1382

orsenthil:
aws-vpc-cni will need to add complete schema definition here - https://github.com/aws/amazon-vpc-cni-k8s/blob/e65d256c6dfa3e2dc318c6319b3c1a70addfa764/config/master/aws-k8s-cni.yaml#L33 

similar to AWS LB controller does here - https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/install/v2_2_4_full.yaml#L37

orsenthil:
@cyberox  - would you like to use these changes in your terraform config https://github.com/aws/amazon-vpc-cni-k8s/pull/2969 and see if the terraform behavior is fixed?

cyberox:
@orsenthil - yes, those changes fix the issue.

orsenthil:
This is addressed in https://github.com/aws/amazon-vpc-cni-k8s/releases/tag/v1.18.3

PR #2969, opened 2024-06-26
Update the APISpec Schema definition for ENIConfig.

orsenthil:
**What type of PR is this?**

Addresses the issue raised in https://github.com/aws/amazon-vpc-cni-k8s/issues/2963
Installer / Config Manager Terraform wasn't gracefully creating ENIConfig without this Schema Spec.

https://github.com/aws/amazon-vpc-cni-k8s/issues/2963

**This was auto generated**

```
kubebuilder create api --group webapp --version v1 --kind ENIConfig
```

Copied the existing type from https://github.com/aws/amazon-vpc-cni-k8s/blob/189f00f64f3e0b0bd0f3b8eff070aed3f0ba32b6/pkg/apis/crd/v1alpha1/eniconfig_types.go

```
make manifests
```

No

No

No

SchemaSpec API docs will have information.

```release-note

```

### DIFF
--- a/charts/aws-vpc-cni/crds/customresourcedefinition.yaml
+++ b/charts/aws-vpc-cni/crds/customresourcedefinition.yaml
@@ -43,7 +43,6 @@ spec:
                 subnet:
                   type: string
               required:
-              - securityGroups
               - subnet
               type: object
             status:
--- a/charts/aws-vpc-cni/crds/customresourcedefinition.yaml
+++ b/charts/aws-vpc-cni/crds/customresourcedefinition.yaml
@@ -14,6 +14,41 @@ spec:
         openAPIV3Schema:
           type: object
           x-kubernetes-preserve-unknown-fields: true
+          description: ENIConfig is the Schema for the eniconfigs API
+          properties:
+            apiVersion:
+              description: |-
+                APIVersion defines the versioned schema of this representation of an object.
+                Servers should convert recognized schemas to the latest internal value, and
+                may reject unrecognized values.
+                More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
+              type: string
+            kind:
+              description: |-
+                Kind is a string value representing the REST resource this object represents.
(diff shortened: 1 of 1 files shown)


cyberox:
@orsenthil - the new lines should indent two more spaces if I'm correct.

orsenthil:
@cyberox- Updated. Yes, they were wrong in the first version.

charts/aws-vpc-cni/crds/customresourcedefinition.yaml
```diff
@@ -14,6 +14,41 @@ spec:
...
+              description: ENIConfigSpec defines the desired state of ENIConfig
+              properties:
+                securityGroups:
+                  items:
+                    type: string
+                  type: array
+                subnet:
+                  type: string
+              required:
+              - securityGroups
```
jchen6585, line 46: Security groups aren't required

orsenthil: This was auto generated from this - https://github.com/aws/amazon-vpc-cni-k8s/blob/189f00f64f3e0b0bd0f3b8eff070aed3f0ba32b6/pkg/apis/crd/v1alpha1/eniconfig_types.go#L28

Since both securityGroup and subnets are listed in spec, it is taking them as required.

jchen6585: Once this is enforced ENIConfigs can't be applied unless securityGroups are specified

jchen6585 (approved):
lgtm

charts/aws-vpc-cni/crds/customresourcedefinition.yaml
jayanthvn: Yeah, we discussed internally about this since it would be needed for upgrades and downgrades..

jayanthvn (approved):
lgtm, during release we need to test upgrades and downgrades with both self managed and eks addons..

PR #2979, opened 2024-07-03
Add unit test and readme update for AWS_VPC_ENI_MTU for Egress plugin behavior.

orsenthil:
**What type of PR is this?**

* Testing

Adds Test code for this change - 
https://github.com/aws/amazon-vpc-cni-k8s/commit/fe4c7a1d8b8dd6c73b84338825a762c739df472e


cmd/aws-vpc-cni/main_test.go
```diff
@@ -48,6 +49,33 @@ func TestGenerateJSONPlusBandwidthAndTuning(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+// Validate setting environment AWS_VPC_ENI_MTU, takes effect for egress-cni plugin
+func TestEgressCNIPlugin(t *testing.T) {
```
achevuru: Test function appears to be validating the MTU value passed to the egress CNI plugin and not the plugin functionality itself, so I would suggest renaming the function to reflect that

jchen6585 (commented):
Could we add an integration test case? Besides that LGTM

orsenthil:
> Could we add an integration test case? Besides that LGTM

This was covering a recently introduced change - https://github.com/aws/amazon-vpc-cni-k8s/pull/2951 I see the egress-cni plugin can see additional integration tests, including MTU default and change. We will bring that separately.

cmd/aws-vpc-cni/main_test.go
orsenthil: Yeah, it was covering this change only 
https://github.com/aws/amazon-vpc-cni-k8s/commit/fe4c7a1d8b8dd6c73b84338825a762c739df472e

achevuru (approved):

PR #2998, opened 2024-07-31
WIP: use protobuf content type instead of json

bhavi-koduru:
In favor of elevating single eks cluster performance, migrating the EKS components to use protobuf instead of json for performance gain.

**What type of PR is this?**

```release-note

```

### DIFF
--- a/pkg/k8sapi/k8sutils.go
+++ b/pkg/k8sapi/k8sutils.go
@@ -92,6 +92,9 @@ func CreateKubeClient(appName string) (client.Client, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


orsenthil:
@bhavi-koduru - could you update the description with motivation behind this change, and why this is required?

hakuna-matatah:
Motivation is to use **protobuf** over **json** to help gain performance and reduce footprint on server side i.e controlplane/apiserver.

bhavi-koduru:
vpc cni uses controller runtime which inturn uses protobuf by default. Hence this change is not necessary.
https://github.com/aws/amazon-vpc-cni-k8s/blob/27ce1362636567592f006b987f3820c6b0fef55e/pkg/k8sapi/k8sutils.go#L172

PR #3015, opened 2024-08-27
Move KUBE_CONFIG_PATH variable to KUBECONFIG variable

852hamza:
What type of PR is this?

Type: Bug
Which issue does this PR fix?

This PR addresses an issue where the unconventional KUBE_CONFIG_PATH environment variable was being used across multiple scripts and configurations, causing confusion. The conventional KUBECONFIG variable is now used instead to align with standard practices.
What does this PR do / Why do we need it?

The main objective of this PR is to replace all instances of the KUBE_CONFIG_PATH variable with the conventional KUBECONFIG variable. This change reduces potential confusion and aligns the project with standard Kubernetes practices.
Key Changes:

    Replaced all occurrences of KUBE_CONFIG_PATH with KUBECONFIG in the following scripts:
        scripts/lib/canary.sh
        scripts/run-static-canary.sh
        scripts/run-ginkgo-integration-suite.sh
        scripts/lib/cluster.sh
        scripts/run-canary-test.sh
        scripts/run-multus-tests.sh
        test/integration/README.md
        scripts/run-ipv6-canary-test.sh
        scripts/run-cni-release-tests.sh
        scripts/run-soak-test.sh
        scripts/run-ipv6-integration-tests.sh

Testing done on this change:

Manual and integration tests were run to ensure that the changes did not introduce any issues and that all scripts and configurations functioned as expected with the KUBECONFIG variable. Logs were reviewed to confirm successful execution.
Test Results:

    All tests passed without any errors.
    Verified that the scripts correctly reference the KUBECONFIG variable.

Will this PR introduce any new dependencies?

No, this PR does not introduce any new dependencies.
Will this break upgrades or downgrades? Has updating a running cluster been tested?

No, this change will not break upgrades or downgrades. Updating a running cluster was tested and confirmed to be unaffected by this change.
Does this change require updates to the CNI daemonset config files to work?

No, this change does not require updates to the CNI daemonset config files.
Does this PR introduce any user-facing change?

No user-facing changes are introduced. The changes are internal to the scripts and configurations.


orsenthil:
@852hamza - We need this is backwards compatible manner, can you see if there is any reference that can be take `KUBE_CONFIG_PATH` and assign it to `KUBECONFIG` ?

852hamza:
@orsenthil I've added a fallback mechanism in the script to ensure compatibility with previous versions that still use KUBE_CONFIG_PATH. The logic checks if KUBE_CONFIG_PATH is set, then assigns KUBECONFIG to the value of KUBE_CONFIG_PATH. This ensures that the script works seamlessly with both old and new configurations without requiring any changes to existing setups. Let me know if you need any further adjustments!

852hamza:
@orsenthil any update plz

scripts/lib/canary.sh
```diff
@@ -4,8 +4,15 @@
 
 SECONDS=0
 
+# Fallback to KUBE_CONFIG_PATH if KUBECONFIG is not set
+
+if [ -n "$KUBE_CONFIG_PATH" ]; then
```
orsenthil, line 9: This looks good. let's slight modify it like this.

```
# Fallback to KUBE_CONFIG_PATH,if set, and and KUBECONFIG is not defined.
if [ -n "$KUBE_CONFIG_PATH" ] && [ -z "$KUBECONFIG" ]; then
  export KUBECONFIG=$KUBE_CONFIG_PATH
fi
```

* This way, we can preserve the backwards compatible if some script is using older variable.
* And safely remove the dependency on `KUBE_CONFIG_PATH` by using `KUBECONFIG`

orsenthil: Please replace this in other places of this usage too, and this looks good to me.

852hamza:
@orsenthil 
- Created a new script, set_kubeconfig.sh, to centralize the Kubernetes configuration setting. This script checks and sets the KUBECONFIG environment variable based on KUBE_CONFIG_PATH if KUBECONFIG is not already set.

- Updated several scripts to source this new script for setting KUBECONFIG. This change ensures all scripts consistently use the correct configuration without duplicating code across multiple files. This centralization also simplifies maintenance and potential future changes to the environment configuration logic.

Scripts updated to source set_kubeconfig.sh:
- scripts/run-canary-test.sh
- scripts/run-ginkgo-integration-suite.sh
- scripts/run-static-canary.sh
- etc.

This change aids in maintaining a cleaner and more manageable codebase by reducing redundancy and centralizing configuration logic.

scripts/lib/canary.sh
852hamza: Kindly check my latest commit

scripts/lib/canary.sh
```diff
@@ -4,8 +4,13 @@
 
 SECONDS=0
 
+# Fallback to KUBE_CONFIG_PATH if KUBECONFIG is not set
+
+source ./set_kubeconfig.sh
```
orsenthil, line 9: Using `./set_kubeconfig.sh`make dependent on the path.
The previous version better with the change suggested; so that we could remove that whole condition later.

If you still prefer this version, please change it to use `SCRIPT_DIR` reference (like SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )") or you use the previous rev with just the change suggested too.

Sorry for this back and forth, my idea was to temporarily remove the dependency on KUBE_CONFIG_PATH path, while not breaking any runners and then getting rid of it. Adding another script doesn't help much.

852hamza: @orsenthil can you check updated commit

852hamza:
@orsenthil can you check the latest commit code

orsenthil (approved):
LGTM. 

We can remove `$SCRIPT_DIR/set_kubeconfig.sh` when no other helper scripts are using the old `KUBE_CONFIG_PATH`.

PR #3035, opened 2024-09-23
Fix fetching enimetadata

Pavani-Panakanti:
**What type of PR is this?**
improvement

Tested cni and ipamd is working as expected by scaling pods and creating new ENIs

No

No

No

No

```release-note

```


pkg/awsutils/awsutils.go
```diff
@@ -602,90 +608,96 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
...
-			awsAPIErrInc("GetSubnetIPv6CIDRBlocks", err)
-		} else {
-			subnetV6Cidr = v6cidr.String()
-		}
+	var ec2ipv4Prefixes []*ec2.Ipv4PrefixSpecification
+	var ec2ipv6Prefixes []*ec2.Ipv6PrefixSpecification
 
-		imdsIPv6s, err := cache.imds.GetIPv6s(ctx, eniMAC)
+	if networkCard == 0 {
+		// Get IPv4 and IPv6 addresses assigned to interface
```
jayanthvn, line 621: Need to remove IPv6 from the comment

pkg/awsutils/awsutils.go
```diff
@@ -1356,7 +1368,7 @@ func (cache *EC2InstanceMetadataCache) DescribeAllENIs() (DescribeAllENIsResult,
 		if interfaceType == "trunk" {
 			trunkENI = eniID
 		}
-		if interfaceType == "efa" {
+		if interfaceType == "efa" || interfaceType == "efa-only" {
```
jayanthvn, line 1371: eventually we should move the interfaceType to some enum for trunk, efa, efa-only..

jayanthvn (approved):
lgtm

pkg/awsutils/awsutils.go
```diff
@@ -602,90 +608,96 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
...
-		if err != nil {
-			awsAPIErrInc("GetSubnetIPv6CIDRBlocks", err)
-		} else {
-			subnetV6Cidr = v6cidr.String()
-		}
+	var ec2ipv4Prefixes []*ec2.Ipv4PrefixSpecification
+	var ec2ipv6Prefixes []*ec2.Ipv6PrefixSpecification
 
-		imdsIPv6s, err := cache.imds.GetIPv6s(ctx, eniMAC)
+	if networkCard == 0 {
```
orsenthil, line 620: What are we verifying with a `if networkCard == 0` here? 

1. Can this condition ever be false? 
2. Should we add a comment that explains why we are checking this here?

Also, we need to test this change in some variety of instances - Multi Card Instance, regular instance, and ensuring this is not leading to any unexpected behavior.

orsenthil:
Code change looks good to me. I think, we need to explain that `if networkCard == 0` is the safe operation to do in the comments. Assuming it is.

pkg/awsutils/awsutils.go
Pavani-Panakanti: @orsenthil On a multicard instance this condition would be false for non-0 cards. For non-0 cards we are just getting basic eni info and not fetching any ip related info as some of the efa interfaces might not have ip  info

Here https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/awsutils/awsutils.go#L1344 we already assume CNI always manages only network card 0 and rest are just added as unmanaged ENI's so it is safe to make the assumption that we need complete ENIMetadata only for network card 0  

I have tested this on multiple instances including multi-card, single card, different efa interfaces and verified the functionality.

I will add comments in code explaining why we are checking for network card 0

Pavani-Panakanti:
> Code change looks good to me. I think, we need to explain that `if networkCard == 0` is the safe operation to do in the comments. Assuming it is.

pkg/awsutils/awsutils.go
Pavani-Panakanti: Makes sense @jayanthvn. I will add a backlog task for this

pkg/awsutils/awsutils.go
Pavani-Panakanti: @jayanthvn we are fetching both IPv4 and IPv6 addresses for ENI's on network card 0. Comment is same as earlier to this change. Github side by side comparison is a bit messed for this PR

orsenthil (approved):
LGTM

jayanthvn (approved):

PR #3041, opened 2024-09-26
Update enimetadata for interfaces with no ip info

Pavani-Panakanti:
**What type of PR is this?**
bug

Network card field is not being present in imds in all instances. We can rely on it to filter out ENI's on non-0 network cards

Check if ipv4 and ipv6 fields are present in imds. Fetch these data only if these fields are present

Verified aws-node is coming up as expected and pods are being scheduled for both multi card and single card instances. Verified with interface with no ips

No

No

No

No

```release-note

```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -585,12 +585,6 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
 		return ENIMetadata{}, err
 	}
 
-	networkCard, err := cache.imds.GetNetworkCard(ctx, eniMAC)
-	if err != nil {
-		awsAPIErrInc("GetNetworkCard", err)
-		return ENIMetadata{}, err
-	}
-
 	deviceNum, err = cache.imds.GetDeviceNumber(ctx, eniMAC)
 	if err != nil {
 		awsAPIErrInc("GetDeviceNumber", err)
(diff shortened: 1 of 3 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -608,7 +602,37 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
...
+	// Efa-only interfaces do not have any ipv4s or ipv6s associated with it
+	for _, field := range macImdsFields {
+		if field == "local-ipv4s" {
+			imdsIPv4s, err := cache.imds.GetLocalIPv4s(ctx, eniMAC)
+			if err != nil {
+				awsAPIErrInc("GetLocalIPv4s", err)
+				return ENIMetadata{}, err
+			}
+			if len(imdsIPv4s) > 0 {
+				isEFAOnlyInterface = false
```
jayanthvn, line 622: My concern is if the ENI is truly missing `local-ipv4`

pkg/awsutils/imds.go
```diff
@@ -136,6 +136,24 @@ func (imds TypedIMDS) GetMACs(ctx context.Context) ([]string, error) {
 	return list, err
 }
 
+// GetMACImdsFields returns the imds fields present for a MAC
```
orsenthil, line 139: Couldn't we use `GetMACs` function and check for it. 

- Introducing yet another `GetMACImdsFields` - only to check for the `fields` might be introduced a too specialized a function for our purpose here.

pkg/awsutils/awsutils.go
orsenthil: > field == "local-ipv4s"

I agree. If these kind of 'structural' definitions is supported by any documentation or spec, that will be best to rely upon.

orsenthil:
The integration test is successful on this change - https://github.com/aws/amazon-vpc-cni-k8s/actions/runs/11053939700

But we have to make sure that we are doing this accurately - with the support of the spec which defines the multi-card instance characteristics.

pkg/awsutils/imds.go
orsenthil: It will be preferable to use `GetMACs` to continue with coverage here.

Pavani-Panakanti: @orsenthil Can you add more details on using GetMACs.. So efa-only will not have any IP fields associated with them. So checking local-ipv4s field. Want to understand how we can use GetMACs here

orsenthil: @Pavani-Panakanti  - you are right. I thought, we were using the `network/interfaces/macs/` and some property of that output. This is different. Sorry for the distraction.

Pavani-Panakanti:
Following with this change in new PR https://github.com/aws/amazon-vpc-cni-k8s/pull/3047

PR #3047, opened 2024-09-26
Filter for interfaces with no ip info

Pavani-Panakanti:
**What type of PR is this?**
feature support
-->

Do not fetch ip info from imds for interfaces that will not have ip fields

Tested with multi card and single card instance by deploying pods

normal interface imds fields 
```
sh-4.2$ curl http://169.254.169.254/latest/meta-data/network/interfaces/macs/0e:bf:84:ab:27:65/
device-number
interface-id
local-hostname
local-ipv4s
mac
network-card
owner-id
security-group-ids
security-groups
subnet-id
subnet-ipv4-cidr-block
vpc-id
vpc-ipv4-cidr-block
vpc-ipv4-cidr-blocks
```

imds fields for interfaces with no ip info 
```
sh-4.2$ curl http://169.254.169.254/latest/meta-data/network/interfaces/macs/0e:46:1b:72:19:89
device-number
interface-id
local-hostname
mac
network-card
owner-id
security-group-ids
security-groups
subnet-id
subnet-ipv4-cidr-block
vpc-id
vpc-ipv4-cidr-block
vpc-ipv4-cidr-blocks
```
No

No

No

No

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -603,14 +603,15 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
 	}
 
 	log.Debugf("Found ENI: %s, MAC %s, device %d", eniID, eniMAC, deviceNum)
+
 	// Get IMDS fields for the interface
 	macImdsFields, err := cache.imds.GetMACImdsFields(ctx, eniMAC)
 	if err != nil {
 		awsAPIErrInc("GetMACImdsFields", err)
 		return ENIMetadata{}, err
 	}
-	isEFAOnlyInterface := true
-	// Efa-only interfaces do not have any ipv4s or ipv6s associated with it. If we don't find any local-ipv4 or ipv6 info in imds we assume it to be efa-only interface and validate this later via ec2 call n
+	ipInfoAvailable := false
+	// Efa-only interfaces do not have any ipv4s or ipv6s associated with it. If we don't find any local-ipv4 or ipv6 info in imds we assume it to be efa-only interface and validate this later via ec2 call
 	for _, field := range macImdsFields {
 		if field == "local-ipv4s" {
 			imdsIPv4s, err := cache.imds.GetLocalIPv4s(ctx, eniMAC)
@@ -619,7 +620,7 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
 				return ENIMetadata{}, err
 			}
 			if len(imdsIPv4s) > 0 {
-				isEFAOnlyInterface = false
+				ipInfoAvailable = true
 				log.Debugf("Found IPv4 addresses associated with interface. This is not efa-only interface")
 				break
 			}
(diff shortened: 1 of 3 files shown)


Pavani-Panakanti:
Few comments on earlier PR https://github.com/aws/amazon-vpc-cni-k8s/pull/3041

Pavani-Panakanti:
Integration tests successful https://github.com/aws/amazon-vpc-cni-k8s/actions/runs/11061575263/job/30734454177

pkg/awsutils/awsutils.go
```diff
@@ -603,89 +603,130 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
...
+		awsAPIErrInc("GetMACImdsFields", err)
 		return ENIMetadata{}, err
 	}
-
-	ec2ip4s := make([]*ec2.NetworkInterfacePrivateIpAddress, len(imdsIPv4s))
-	for i, ip4 := range imdsIPv4s {
-		ec2ip4s[i] = &ec2.NetworkInterfacePrivateIpAddress{
-			Primary:          aws.Bool(i == 0),
-			PrivateIpAddress: aws.String(ip4.String()),
+	isEFAOnlyInterface := true
```
orsenthil, line 612: Why don't we flip the logic here?
```
	isEFAOnlyInterface := false
```
And if it true, then immediately return with appropriate empty fields

```
return ENIMetadata{
		ENIID:          eniID,
		MAC:            eniMAC,
		DeviceNumber:   deviceNum,
		SubnetIPv4CIDR: [],
		IPv4Addresses:  [],
		IPv4Prefixes:   [],
		SubnetIPv6CIDR: [],
```


Rest look good to me.

orsenthil:
In the description, please paste the output of the IMDS call made against an EFA instance that shows the `"local-ipv4s"` field vs a normal instance not having. For testing verification.

pkg/awsutils/awsutils.go
Pavani-Panakanti: Yeah we can do that

Pavani-Panakanti:
> In the description, please paste the output of the IMDS call made against an EFA instance that shows the `"local-ipv4s"` field vs a normal instance not having. For testing verification.

Added

pkg/awsutils/awsutils.go
Pavani-Panakanti: Updated

pkg/awsutils/awsutils.go
```diff
@@ -604,6 +604,53 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
...
+			}
+			if len(imdsIPv4s) > 0 {
+				ipInfoAvailable = true
+				log.Debugf("Found IPv4 addresses associated with interface. This is not efa-only interface")
+				break
+			}
+		}
+		if field == "ipv6s" {
+			imdsIPv6s, err := cache.imds.GetIPv6s(ctx, eniMAC)
+			if err != nil {
```
M00nF1sh, line 630: why we treat IPv4 and IPv6 differently.
If GetLocalIPv4s failed, we return error directly. If GetIPv6s failed, we didn't return error.

If there is no particular reason, we should refactor the logic & code to make it have similar structure/logic for ipv4&ipv6. (otherwise such difference will became a maintaince burden and could easily cause bugs when changes are made here)

Pavani-Panakanti: @M00nF1sh  We are not erroring out at missing ipv6 info in other parts of the code https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/awsutils/awsutils.go#L631. Did the same here to maintain consistency to same as earlier irrespective of this change.

Might be we should add some context and comments around why it is different for ipv6

pkg/awsutils/awsutils.go
```diff
@@ -1356,9 +1403,16 @@ func (cache *EC2InstanceMetadataCache) DescribeAllENIs() (DescribeAllENIsResult,
 		if interfaceType == "trunk" {
 			trunkENI = eniID
 		}
-		if interfaceType == "efa" {
+		if interfaceType == "efa" || interfaceType == "efa-only" {
 			efaENIs[eniID] = true
 		}
+		if interfaceType != "efa-only" {
+			if len(eniMetadata.IPv4Addresses) == 0 && len(eniMetadata.IPv6Addresses) == 0 {
```
orsenthil, line 1410: Should this condition be a `&&` or an `||` ? 

This is the only new Error we are introducing here.  Do we strictly require this ? cc: @jayanthvn (this was introduced from the review comment).


Rest of the change looks good to me with capturing EFA only instance.

orsenthil (approved):
Changes LGTM.

pkg/awsutils/awsutils.go
jayanthvn: There are 2 cases, `local-ipv4` field is missing or `local-ipv4` value is empty

`local-ipv4` value is empty -> we return error in 618

`local-ipv4` field is missing (Eg: IMDS out of sync on new node or aws-node restart) -> previously we used to error out - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/awsutils/imds.go#L293-L299 but now we return the ENI (line 640) and call describe ENI. So to keep the behavior consistent this additional check i.e, if not `efa-only` and missing primary IP we error out... But I think we don't need to check `IPv6Addresses` since we have been ignoring ipv6 missing data (need to check why that was introduced though)

Pavani-Panakanti: Updated. We will not be returning error if we miss ipv6addresses to maintain the behavior

orsenthil: Okay, this reasoning sounds good to me.

orsenthil (approved):
LGTM.

PR #3122, opened 2024-11-22
Fix issues with handling unmanaged ENIs with IPv6 only

gavinbunney:
**What type of PR is this?**
bug

We have unmanaged trunk ENIs attached to our worker nodes which are designed for ipv6 only networks. These ENIs are tagged with `node.k8s.amazonaws.com/no_manage`, however the listing of attached ENIs happens before the IPAMD process filters those ENIs. As such, the metadata retrieval process fails when looking up the ipv4 address details for these ENIs, and causes the `aws-k8s-agent` process to exit with an initialization failure.

In this log, `eni-084b51xxxxxx` / `0e:7c:f9:xx:xx:xx` is the aws-vpc-cni managed ENI, and `eni-0b2cf8b6xxxxxx` / `0e:f3:73:xx:xx:xx` is our managed eni:

```
{"level":"debug","ts":"2024-11-22T16:49:26.314Z","caller":"awsutils/awsutils.go:1317","msg":"Total number of interfaces found: 2 "}
{"level":"debug","ts":"2024-11-22T16:49:26.314Z","caller":"awsutils/awsutils.go:567","msg":"Found ENI MAC address: 0e:7c:f9:xx:xx:xx"}
{"level":"debug","ts":"2024-11-22T16:49:26.316Z","caller":"awsutils/awsutils.go:567","msg":"Found ENI: eni-084b51xxxxxx, MAC 0e:7c:f9:xx:xx:xx, device 0"}
{"level":"debug","ts":"2024-11-22T16:49:26.318Z","caller":"awsutils/awsutils.go:567","msg":"Found IPv6 addresses associated with interface. This is not efa-only interface"}
{"level":"debug","ts":"2024-11-22T16:49:26.322Z","caller":"awsutils/awsutils.go:567","msg":"Found ENI MAC address: 0e:f3:73:xx:xx:xx"}
{"level":"debug","ts":"2024-11-22T16:49:26.324Z","caller":"awsutils/awsutils.go:567","msg":"Found ENI: eni-0b2cf8b6xxxxxx, MAC 0e:f3:73:xx:xx:xx, device 1"}
{"level":"debug","ts":"2024-11-22T16:49:26.326Z","caller":"awsutils/awsutils.go:567","msg":"Found IPv6 addresses associated with interface. This is not efa-only interface"}
{"level":"warn","ts":"2024-11-22T16:49:26.327Z","caller":"awsutils/imds.go:376","msg":"failed to retrieve network/interfaces/macs/0e:f3:73:xx:xx:xx/subnet-ipv4-cidr-block from instance metadata EC2MetadataError: failed to make EC2Metadata request\n<?xml version=\"1.0\" encoding=\"iso-8859-1\"?>\n<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Transitional//EN\"\n\t\t \"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd\">\n<html xmlns=\"http://www.w3.org/1999/xhtml\" xml:lang=\"en\" lang=\"en\">\n <head>\n  <title>404 - Not Found</title>\n </head>\n <body>\n  <h1>404 - Not Found</h1>\n </body>\n</html>\n\n\tstatus code: 404, request id: "}
{"level":"error","ts":"2024-11-22T16:49:26.327Z","caller":"aws-k8s-agent/main.go:42","msg":"Initialization failure: ipamd init: failed to retrieve attached ENIs info: DescribeAllENIs: failed to get local ENI metadata: get attached ENIs: failed to retrieve ENI metadata for ENI: 0e:f3:73:xx:xx:xx: EC2MetadataError: failed to make EC2Metadata request\n<?xml version=\"1.0\" encoding=\"iso-8859-1\"?>\n<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Transitional//EN\"\n\t\t \"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd\">\n<html xmlns=\"http://www.w3.org/1999/xhtml\" xml:lang=\"en\" lang=\"en\">\n <head>\n  <title>404 - Not Found</title>\n </head>\n <body>\n  <h1>404 - Not Found</h1>\n </body>\n</html>\n\n\tstatus code: 404, request id: "}
```

This PR better handles missing IPv4 information for ENIs which are IPv6 only (using same pattern that the IPv6 IPs lookup uses)

Added unit tests to cover the new paths. Running in our EKS cluster the ENIs are now retrieved successfully (In this log, `eni-084b51xxxxxx` / `0e:7c:f9:xx:xx:xx` is the aws-vpc-cni managed ENI, and `eni-0b2cf8b6xxxxxx` / `0e:f3:73:xx:xx:xx` is our managed eni):

```
{"level":"debug","ts":"2024-11-22T17:45:13.815Z","caller":"awsutils/awsutils.go:1325","msg":"Total number of interfaces found: 2 "}
{"level":"debug","ts":"2024-11-22T17:45:13.815Z","caller":"awsutils/awsutils.go:567","msg":"Found ENI MAC address: 0e:7c:f9:8b:06:fd"}
{"level":"debug","ts":"2024-11-22T17:45:13.817Z","caller":"awsutils/awsutils.go:567","msg":"Found ENI: eni-084b51xxxxxx, MAC 0e:7c:f9:xx:xx:xx, device 0"}
{"level":"debug","ts":"2024-11-22T17:45:13.818Z","caller":"awsutils/awsutils.go:567","msg":"Found IPv6 addresses associated with interface. This is not efa-only interface"}
{"level":"debug","ts":"2024-11-22T17:45:13.819Z","caller":"awsutils/awsutils.go:567","msg":"Found ENI MAC address: 0e:f3:73:xx:xx:xx"}
{"level":"debug","ts":"2024-11-22T17:45:13.821Z","caller":"awsutils/awsutils.go:567","msg":"Found ENI: eni-0b2cf8b6xxxxxx, MAC 0e:f3:73:xx:xx:xx, device 1"}
{"level":"debug","ts":"2024-11-22T17:45:13.822Z","caller":"awsutils/awsutils.go:567","msg":"Found IPv6 addresses associated with interface. This is not efa-only interface"}
{"level":"info","ts":"2024-11-22T17:45:14.003Z","caller":"ipamd/ipamd.go:424","msg":"Got network card index 0 for ENI eni-084b51xxxxxx"}
{"level":"info","ts":"2024-11-22T17:45:14.004Z","caller":"ipamd/ipamd.go:424","msg":"eni-084b51xxxxxx is of type: interface"}
{"level":"info","ts":"2024-11-22T17:45:14.004Z","caller":"ipamd/ipamd.go:424","msg":"Got network card index 0 for ENI eni-0b2cf8b6xxxxxx"}
{"level":"info","ts":"2024-11-22T17:45:14.004Z","caller":"ipamd/ipamd.go:424","msg":"eni-0b2cf8b6xxxxxx is of type: trunk"}
{"level":"debug","ts":"2024-11-22T17:45:14.004Z","caller":"ipamd/ipamd.go:385","msg":"DescribeAllENIs success: ENIs: 2, tagged: 2"}
```

n/a

Tested with upgrading inplace without issues

n/a

n/a

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -610,7 +610,9 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
 		awsAPIErrInc("GetMACImdsFields", err)
 		return ENIMetadata{}, err
 	}
-	ipInfoAvailable := false
+
+	ipv4Available := false
+	ipv6Available := false
 	// Efa-only interfaces do not have any ipv4s or ipv6s associated with it. If we don't find any local-ipv4 or ipv6 info in imds we assume it to be efa-only interface and validate this later via ec2 call
 	for _, field := range macImdsFields {
 		if field == "local-ipv4s" {
@@ -620,7 +622,7 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
 				return ENIMetadata{}, err
 			}
 			if len(imdsIPv4s) > 0 {
-				ipInfoAvailable = true
+				ipv4Available = true
 				log.Debugf("Found IPv4 addresses associated with interface. This is not efa-only interface")
 				break
 			}
@@ -630,14 +632,14 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
 			if err != nil {
 				awsAPIErrInc("GetIPv6s", err)
 			} else if len(imdsIPv6s) > 0 {
-				ipInfoAvailable = true
+				ipv6Available = true
 				log.Debugf("Found IPv6 addresses associated with interface. This is not efa-only interface")
 				break
 			}
 		}
 	}
 
-	if !ipInfoAvailable {
+	if !ipv4Available && !ipv6Available {
 		return ENIMetadata{
 			ENIID:          eniID,
 			MAC:            eniMAC,
@@ -652,23 +654,29 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
 	}
 
 	// Get IPv4 and IPv6 addresses assigned to interface
-	cidr, err := cache.imds.GetSubnetIPv4CIDRBlock(ctx, eniMAC)
-	if err != nil {
-		awsAPIErrInc("GetSubnetIPv4CIDRBlock", err)
-		return ENIMetadata{}, err
-	}
+	var ec2ip4s []*ec2.NetworkInterfacePrivateIpAddress
+	var subnetV4Cidr string
+	if ipv4Available {
+		cidr, err := cache.imds.GetSubnetIPv4CIDRBlock(ctx, eniMAC)
+		if err != nil {
+			awsAPIErrInc("GetSubnetIPv4CIDRBlock", err)
+			return ENIMetadata{}, err
+		}
 
-	imdsIPv4s, err := cache.imds.GetLocalIPv4s(ctx, eniMAC)
-	if err != nil {
-		awsAPIErrInc("GetLocalIPv4s", err)
-		return ENIMetadata{}, err
-	}
+		subnetV4Cidr = cidr.String()
 
-	ec2ip4s := make([]*ec2.NetworkInterfacePrivateIpAddress, len(imdsIPv4s))
-	for i, ip4 := range imdsIPv4s {
-		ec2ip4s[i] = &ec2.NetworkInterfacePrivateIpAddress{
-			Primary:          aws.Bool(i == 0),
-			PrivateIpAddress: aws.String(ip4.String()),
+		imdsIPv4s, err := cache.imds.GetLocalIPv4s(ctx, eniMAC)
+		if err != nil {
+			awsAPIErrInc("GetLocalIPv4s", err)
+			return ENIMetadata{}, err
+		}
+
+		ec2ip4s = make([]*ec2.NetworkInterfacePrivateIpAddress, len(imdsIPv4s))
+		for i, ip4 := range imdsIPv4s {
+			ec2ip4s[i] = &ec2.NetworkInterfacePrivateIpAddress{
+				Primary:          aws.Bool(i == 0),
+				PrivateIpAddress: aws.String(ip4.String()),
+			}
 		}
 	}
 
@@ -732,7 +740,7 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
 		ENIID:          eniID,
 		MAC:            eniMAC,
 		DeviceNumber:   deviceNum,
-		SubnetIPv4CIDR: cidr.String(),
+		SubnetIPv4CIDR: subnetV4Cidr,
 		IPv4Addresses:  ec2ip4s,
 		IPv4Prefixes:   ec2ipv4Prefixes,
 		SubnetIPv6CIDR: subnetV6Cidr,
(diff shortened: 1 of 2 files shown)


gavinbunney:
@orsenthil @jaydeokar Would you be able to take a look? We are attempting to work around issues with our primarily-ipv6 network so would be good to get this merged in

orsenthil:
@gavinbunney - I will review this shortly. It is my radar.

pkg/awsutils/awsutils.go
```diff
@@ -1399,22 +1407,25 @@ func (cache *EC2InstanceMetadataCache) DescribeAllENIs() (DescribeAllENIsResult,
 		interfaceType := aws.StringValue(ec2res.InterfaceType)
 		log.Infof("%s is of type: %s", eniID, interfaceType)
 
-		// This assumes we only have one trunk attached to the node..
-		if interfaceType == "trunk" {
+		// The primary trunk eni requires an IPv4 address
+		if interfaceType == "trunk" && len(eniMetadata.IPv4Addresses) > 0 {
```
orsenthil, line 1411: This is the only change and assumption (even the previous that assumed only one trunk and now we are making assertion that one trunk with ipv4 address) that is causing a bit of concern to me.

What if you don't have this `&& len(eniMetadata.IPv4Addresses) > 0` assertion. Wouldn't the rest of the changes be sufficient? I see that `getENIMetadata` is now protected and you wont run into issue with ipamd.

gavinbunney: The issue is in the caller from ipamd when it's checking for a trunk eni, it uses the passed back `metadataResult.TrunkENI` to search for it; in our case, it would pickup our managed ENI and not the aws-vpc-cni one - ref: https://github.com/aws/amazon-vpc-cni-k8s/blob/64748b4594ddefd938f3a127a30ea3112a576f3e/pkg/ipamd/ipamd.go#L1317-L1329

An alternative might be to pull up the filtering of ENIs into the `awsutils.go` file itself so it's filtered right when fetching from the AWS APIs, so they are dropped beforehand, so the aws-vpc-cni wouldn't need to worry about unique setup (like the other fix in this PR where there are no IPv4 addresses).

orsenthil: The function `func (c *IPAMContext) checkForTrunkENI() bool` is only called from 

https://github.com/aws/amazon-vpc-cni-k8s/blob/5c471fe47af7df8a34520d82692f1e6e40920589/pkg/ipamd/ipamd.go#L644

and is gated through `c.enablePodENI` . 

I assume you had `ENABLE_POD_ENI` and `IPV6`, and instead of AWS created trunk, this call returned your trunk and that caused a problem for you. A problem that is not shown in the traceback above. Is that correct?

--- 

In, 
https://github.com/aws/amazon-vpc-resource-controller-k8s/blob/b5a054e051e0a223001c3008a6f1245fbbbcf176/pkg/aws/ec2/api/helper.go#L125C1-L130C3

when we create a trunk interface, we do not specify ` EnablePrimaryIpv6 ` which will guarantee that there will always be IPv4 Address.  So, checking for IPV4 address won't break the behavior.


`if interfaceType == "trunk" && len(eniMetadata.IPv4Addresses) > 0 ` is, but this introducing some special information due a to unique cluster configuration.

How about using `tagMap[eniMetadata.ENIID] = convertSDKTagsToTags(ec2res.TagSet)` early and verifying that the trunk as `no_manage` tag? Would that be more explicit ?

Other approaches like filtering early, or even documenting a bit further for clarity is fine with me.

gavinbunney: Ah got it thanks. Let me have another look at filtering it out. I'll pull that change out of this PR in the meantime, so can get the IPv4 fixes merged in

orsenthil (approved):
LGTM

pkg/awsutils/awsutils.go
```diff
@@ -652,23 +654,29 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
...
 
 	// Get IPv4 and IPv6 addresses assigned to interface
-	cidr, err := cache.imds.GetSubnetIPv4CIDRBlock(ctx, eniMAC)
-	if err != nil {
-		awsAPIErrInc("GetSubnetIPv4CIDRBlock", err)
-		return ENIMetadata{}, err
-	}
+	var ec2ip4s []*ec2.NetworkInterfacePrivateIpAddress
+	var subnetV4Cidr string
+	if ipv4Available {
```
orsenthil, line 659: One problem I realized much later is, the way in which we check for `ipv4Available` and `ipv6Available` - seems to depend on the order in which imdsFields are returned.
```
	for _, field := range macImdsFields {
		if field == "local-ipv4s" {
			imdsIPv4s, err := cache.imds.GetLocalIPv4s(ctx, eniMAC)
			if err != nil {
				awsAPIErrInc("GetLocalIPv4s", err)
				return ENIMetadata{}, err
			}
			if len(imdsIPv4s) > 0 {
				ipv4Available = true
				log.Debugf("Found IPv4 addresses associated with interface. This is not efa-only interface")
				break
			}
		}
		if field == "ipv6s" {
			imdsIPv6s, err := cache.imds.GetIPv6s(ctx, eniMAC)
			if err != nil {
				awsAPIErrInc("GetIPv6s", err)
			} else if len(imdsIPv6s) > 0 {
				ipv6Available = true
				log.Debugf("Found IPv6 addresses associated with interface. This is not efa-only interface")
				break
			}
		}
	}
```

Previously we assumed the ENI (or primary) always will have an IPV4 and went with that approach. Now, since we are checking `ipv4Available` explicitly, if for some reason order of checking of imds returned IPV6 and not ipv4 (due to break condition in the above code. then this introduces bug. 

We will need to remove the break condition in the above loop to remain compatible with the previous behavior, that is, to always get the IPV4 address for the primary ENI.

pkg/awsutils/awsutils.go
orsenthil: Fixing this in master here - https://github.com/aws/amazon-vpc-cni-k8s/pull/3156

PR #3141, opened 2024-12-08
run make generate-limits to update the max pods file

tzneal:
**What type of PR is this?** dependency update

I ran `make generate-limits` to get the new instance types

```release-note

```


misc/eni-max-pods.txt
```diff
@@ -592,6 +606,7 @@ p4d.24xlarge 737
 p4de.24xlarge 737
 p5.48xlarge 100
 p5e.48xlarge 100
+p5en.48xlarge 198
```
jayanthvn, line 609: @Pavani-Panakanti can you double check p5, p5e and p5en instances here? 

Ideally we should have just 50 Ips per ENI and script does skip calculation - 

https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/gen_vpc_ip_limits.go#L155

Screenshot 2024-12-08 at 10 30 48 AM

misc/eni-max-pods.txt
```diff
@@ -828,6 +843,7 @@ t4g.xlarge 58
 trn1.2xlarge 58
 trn1.32xlarge 247
 trn1n.32xlarge 247
+trn2.48xlarge 100
```
jayanthvn, line 846: Screenshot 2024-12-08 at 10 33 10 AM

@Pavani-Panakanti  - even this..

misc/eni-max-pods.txt
jayanthvn: Ignore my comments, I missed the number of ENIs..

PR #3143, opened 2024-12-10
Handle EKS Service for the Beta Endpoint.

orsenthil:
The DescribeCluster against the Beta endpoint cluster was failing with `operation error STS: AssumeRoleWithWebIdentity` 

For the tests we exercise against a custom endpoint,  for EKS we need to send the PartitionID, and SigningRegion, so that STS can use those to sign the request for the EKS cluster with the custom endpoint.

## Testing Done

* Locally was able to describe cluster against a custom endpoint.
* Tested in internal CI.

### DIFF
--- a/test/framework/resources/aws/services/eks.go
+++ b/test/framework/resources/aws/services/eks.go
@@ -47,12 +47,20 @@ type AddonInput struct {
 
 func NewEKS(cfg aws.Config, endpoint string) (EKS, error) {
 	var err error
-
+	var customResolver aws.EndpointResolverWithOptions
 	if endpoint != "" {
-		customResolver := aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
-			return aws.Endpoint{
-				URL: endpoint,
-			}, nil
+		// EKS Custom endpoint resolver needs PartitionID, SingingRegion and URL for handling STS requests.
+		// TODO: default to "aws" partition for now as it handled only tests. Provide option to pass partitionID.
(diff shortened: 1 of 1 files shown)


test/framework/resources/aws/services/eks.go
```diff
@@ -47,12 +47,20 @@ type AddonInput struct {
...
-		customResolver := aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
-			return aws.Endpoint{
-				URL: endpoint,
-			}, nil
+		// EKS Custom endpoint resolver needs PartitionID, SingingRegion and URL for handling STS requests.
+		// TODO: default to "aws" partition for now as it handled only tests. Provide option to pass partitionID.
+		customResolver = aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
+			if service == eks.ServiceID {
+				return aws.Endpoint{
+					PartitionID:   "aws",
```
sushrk, line 57: Will this work for regions in `aws-cn/aws-gov` partition?

orsenthil: Good point. I will have to test. I will see how we are sending the endpoint in those partitions.

orsenthil: If we are not sending a custom `--eks-endpoint` (that is a custom endpoint cluster in the particular partion, we should be good), otherwise partition id needs to sent separately.

sushrk: Chatted offline; endpoint is only set in beta tests so this is good to go.

sushrk (approved):
LGTM, thanks!

Issue #2797, opened 2024-02-18
pods with attached security groups cannot reach Pod Identity Agent link local address


tmehlinger:
When using security groups for pods, pods with security groups attached cannot reach the Pod Identity agent on 169.254.170.23. Any pods _without_ security groups can reach the agent without issue. The agent pods have no security groups associated and I've ensured that the security groups on failing pods permit egress traffic on TCP port 80, and my node security group permits ingress traffic on port 80 from cluster subnets. I've tried various combinations of egress/ingress from pod/node security groups, and even a blanket policy that permits traffic to/from `0/0` with no success.

I'm using the EKS Addon with the following configuration:
```
            {
                "enableNetworkPolicy": "true",
                "env": {
                    "ENABLE_POD_ENI": "true",
                },
                "init": {
                    "env": {
                        "DISABLE_TCP_EARLY_DEMUX": "true"
                    }
                }
            }
```

I've tried running the CNI with `POD_SECURITY_GROUP_ENFORCING_MODE` set to `standard` but this causes traffic to a peered VPCs to be denied in addition to pod identity traffic being dropped (and I want strict enforcement, regardless).

Reading the documentation for standard mode behavior:

> inbound/outbound traffic from another pod on the same host or another service on the same host(such as kubelet/nodeLocalDNS) won't be enforced by security group rules.

My totally wild guess about what's happening is strict mode requires enforcement of security group rules and the node security group is dropping traffic destined for a link local address as invalid.

Could someone point me the right direction? Thanks!

**Environment**:
Kubernetes version (use `kubectl version`):
`Server Version: version.Info{Major:"1", Minor:"28+", GitVersion:"v1.28.5-eks-5e0fdde", GitCommit:"e78a4be9da4c375a87a109e0f4a5f4a8d2bc17c0", GitTreeState:"clean", BuildDate:"2024-01-02T20:34:46Z", GoVersion:"go1.20.12", Compiler:"gc", Platform:"linux/amd64"}`

CNI Version:
`v1.16.2-eksbuild.1`

OS (e.g: `cat /etc/os-release`):
AWS EKS 1.28.5 AMI.
```
AME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
[...]

PR #3148, opened 2024-12-11
Add a high priority rule to handle link-Local traffic via local route table in SGPP Mode for both IPV4 and IPV6.

orsenthil:
**What type of PR is this?**
bug-fix

https://github.com/aws/amazon-vpc-cni-k8s/issues/2797

When POD_SECURITY_GROUP_ENFORCING_MODE is set to strict, pods with SGPP configured have their traffic routed over a branch ENI and thus entirely bypass the primary interface on the node. 
Thus pods with SGPP strict mode, will fail to reach link-local addresses. This change ensures that when SGPP is configured in the strict mode, the link-local traffic goes through local route table.  This is similar to how ICMPv6 packets from the gateway is handled currently for strict mode.

### DIFF
--- a/pkg/networkutils/network.go
+++ b/pkg/networkutils/network.go
@@ -341,6 +341,8 @@ func (n *linuxNetwork) SetupHostNetwork(vpcv4CIDRs []string, primaryMAC string, 
 
 	// In strict mode, packets egressing pod veth interfaces must route via the trunk ENI in order for security group
 	// rules to be applied. Therefore, the rule to lookup the local routing table is moved to a lower priority than VLAN rules.
+	// A high priority rule is added to handle to/from link-local traffic by the local routing table
+	// A high priority rule is added for ICMPv6 packets from the gateway to lookup in the local routing table
 	if enablePodENI && n.podSGEnforcingMode == sgpp.EnforcingModeStrict {
 		localRule := n.netLink.NewRule()
 		localRule.Table = localRouteTable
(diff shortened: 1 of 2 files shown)


orsenthil:
One comment on adding high priority rule is, this changes changes the definition strict mode, and makes communication possible which wasn't previously, by definition.


We introduced `POD_SECURITY_GROUP_ENFORCING_MODE` knob for these specific user cases as communication of pods on the same host / another service on the same host. The outbound traffic for pods using SGPP when using the standard mode will have the Security Group enforced by branch ENI - https://github.com/aws/amazon-vpc-cni-k8s/issues/2797

We need to evaluate this if this change is required at all.

orsenthil:
Allowing link-local traffic will require more changes than this PR. That brought in the changes that is almost close the introduction of standard mode that was introduced - https://github.com/aws/amazon-vpc-cni-k8s/pull/1907  - So, the `standard` mode seems to be _only_ way right now for the link-local traffic.

bleggett:
> Allowing link-local traffic will require more changes than this PR. That brought in the changes that is almost close the introduction of standard mode that was introduced - #1907 - So, the `standard` mode seems to be _only_ way right now for the link-local traffic.

Yes, using `standard` is the only way at this time. I think that must be fixed in some manner tho. It's not really acceptable for a CNI to mishandle an entire IETF defined CIDR range by default, that at least should be feature-flaggable behavior.

It's not a security concern either - the fundamental problem is that link-local traffic _cannot_ make it far enough for SGPP enforcement to apply to it in the first place, so there is nothing to assert policy against.

cb-troydai:
Today, under the strict mode, we can apply security group rules to allow host process access to pod health check pod. Drop link local prevent us from doing that.

pkg/networkutils/network.go
```diff
@@ -358,8 +360,18 @@ func (n *linuxNetwork) SetupHostNetwork(vpcv4CIDRs []string, primaryMAC string,
 			return errors.Wrap(err, "ChangeLocalRulePriority: failed to delete priority 0 local rule")
 		}
 
-		// In IPv6 strict mode, ICMPv6 packets from the gateway must lookup in the local routing table so that branch interfaces can resolve their gateway.
+		// Add a high priority rule, to handle all link-local traffic by local table.
```
orsenthil, line 363: Self Review: This is not the correct way to handle the link-local traffic. 
After adding this I realized that link-local traffic will not go through this route table. 

We need to use SGP "standard mode" for link local traffic.

orsenthil:
Closing this, a SGP "standard mode" is the recommendation to handle Link-Local traffic.

PR #3156, opened 2024-12-18
Fix issue with primary ENI ip lookup when an ENI has both IPv4 and IPv6 Address Assigned.

orsenthil:
**What type of PR is this?**

Fixes a regression introduced in this https://github.com/aws/amazon-vpc-cni-k8s/pull/3122

Previously we assumed the ENI (primary ENI) always will have an IPV4 and went with getting the IPv4 address.

Primary ENI will always have an IPV4 address.

However, when https://github.com/aws/amazon-vpc-cni-k8s/pull/3122 was introduced, we are checking ipv4Available explicitly. The way in which the ipv4 or ipv6 is checked in based on the value of IMDS metadata on MAC. It dependent on the order in which IMDS returned the value - *due* to using a break condition.

https://github.com/aws/amazon-vpc-cni-k8s/pull/3122/files#r1889412276

*The break condition is removed and this is fixed.**

Added unit-test to catch this issue. Without the change in `awsutils.go`the unit test will fail.


M00nF1sh (approved):
/lgtm
/approve

PR #3160, opened 2024-12-23
fix(test): add volume mount for docker-func-test target

omerap12:
**What type of PR is this?**
/kind bug

fixes: #3076

The docker-func-test Make target fails because it cannot access /host/opt/cni/bin
directory which is expected to be mounted from the host system.

NONE

NONE

NONE

```release-note
NONE
```


orsenthil (approved):
LGTM.

But looks like this test will need to removed as it was probably written when we had full build base image. Right now, it will exercise if container can be run, and since the entrypoint is a daemon, it wont exit.

Issue #3017, opened 2024-08-28
CRD policyendpoints does not pass kubeconform schema validation


phihos:
**What happened**:

After rendering resources via aws-vpc-cni helm chart [kubeconform](https://github.com/yannh/kubeconform) checks it against  the `CustomResourceDefinition` schema and fails with

```
policyendpoints.networking.k8s.aws is invalid: For field metadata.creationTimestamp: Invalid type. Expected: string, given: null
```

It refers to [this line](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/charts/aws-vpc-cni/crds/customresourcedefinition.yaml#L61):

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  annotations:
    controller-gen.kubebuilder.io/version: v0.11.3
  creationTimestamp: null  <--- this one
  labels:
    app.kubernetes.io/name: amazon-network-policy-controller-k8s
  name: policyendpoints.networking.k8s.aws
spec:
```

Can this attribute be safely deleted? If not, what is the purpose of setting `creationTimestamp: null `?

**What you expected to happen**:

The CRD should pass kubeconform.

**How to reproduce it (as minimally and precisely as possible)**:

```
git clone https://github.com/aws/amazon-vpc-cni-k8s.git
cd amazon-vpc-cni-k8s/charts/aws-vpc-cni
cat crds/customresourcedefinition.yaml| kubeconform -schema-location default -schema-location 'https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master/{{.NormalizedKubernetesVersion}}/{{.ResourceKind}}.json'
```

Output: 

```
stdin - CustomResourceDefinition policyendpoints.networking.k8s.aws is invalid: problem validating schema. Check JSON formatting: jsonschema: '/metadata/creationTimestamp' does not validate with https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master/master/customresourcedefinition.json#/properties/metadata/$ref/properties/creationTimestamp/$ref/type: expected string, but got null
```

orsenthil:
I think, we just have to remove this line

https://github.com/aws/amazon-vpc-cni-k8s/blob/master/charts/aws-vpc-cni/crds/customresourcedefinition.yaml#L61

And it should be set automatically by Kubernetes when the resource is created.

hpatel292-seneca:
Hi @orsenthil, can I work on this?

[...]

PR #3163, opened 2024-12-31
fix: remove null creationTimestamp from CRD metadata

omerap12:
**What type of PR is this?**
bug

issue #3017

Removes metadata.creationTimestamp: null field from CustomResourceDefinition to resolve kubeconform validation errors. This field is automatically managed by Kubernetes and should not be explicitly set to null in the CRD definition.

```release-note
NONE
```


orsenthil:
Looks like `createTimestamp` was added automatically when using `controller-gen.kubebuilder.io/version: v0.11.3` 
- This is tracked as an upstream kubernetes issue - https://github.com/kubernetes/kubernetes/issues/109427 - giving the details, and possible fix here https://github.com/kubernetes/kubernetes/issues/67610 that can satisfy kubeconform tools.

Since creationTimeStamp is populated by system https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L179-L188 it is okay to remove it from CRD file.


```
	// CreationTimestamp is a timestamp representing the server time when this object was
	// created. It is not guaranteed to be set in happens-before order across separate operations.
	// Clients may not set this value. It is represented in RFC3339 form and is in UTC.
	//
	// Populated by the system.
	// Read-only.
	// Null for lists.
	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#metadata
	// +optional
	CreationTimestamp Time `json:"creationTimestamp,omitempty" protobuf:"bytes,8,opt,name=creationTimestamp"`

```

orsenthil (approved):
LGTM

Issue #3158, opened 2024-12-19
VpcCNI is making list calls on all nodes & CNINode in cluster and causing scalability issues on large clusters. 


hakuna-matatah:
**What would you like to be added**:

Based on the code it seems  like vpc-cni is using [informer for k8s client](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/k8sapi/k8sutils.go#L119) and warms up the cache on startup iiuc for certain [resources like nodes, cninodes etc](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/k8sapi/k8sutils.go#L100-L104) 
I think ask is to  change this to use a direct k8s client or a informer cache on specific object name/namespace if possible  as vpc-cni only need single Node object from my understanding given its a deamonset.

**Why is this needed**:

LIST calls are expensive for  `nodes` and `cninodes` on large scale clusters especially when warming up client side cache as it causes lot of data to be exchanged especially when cni-node crash loops on very large cluster and overload APIServer.

hakuna-matatah:
It appears [node filters]( https://github.com/aws/amazon-vpc-cni-k8s/commit/0703d03dec8afb8f83a7ff0c9d5eb5cc3363026e ) are added recently. We should be doing the same for cninodes.

PR #3164, opened 2024-12-31
Add CNINode to cache filter

dims:
We should reduce the number of CNINode object VPC CNI watches for to just the node it is managing as well.

**What type of PR is this?**
improvement

Please see https://github.com/aws/amazon-vpc-cni-k8s/issues/3158

Very similar to https://github.com/aws/amazon-vpc-cni-k8s/issues/2887 and https://github.com/aws/amazon-vpc-cni-k8s/pull/2888 but this one is to limit `CNINode`'s

This PR adds a filter to reduce the number of CNINode object VPC CNI watches for to just the node it is managing.

EKS scalability clusters by @hakuna-matatah 

No.

No.

No.

No.

No.

```release-note
NONE
```


jayanthvn (approved):
Thanks @dims

PR #3202, opened 2025-02-11
Add grpc call to fetch networkpolicymode from NP

Pavani-Panakanti:
**What type of PR is this?**
improvement

This will help np agent fetch networkpolicymode which is env_var available in ipamd

This adds a configserverbackend which listens on GetNetworkPolicyConfig request and responses with networkPolicyMode value. This will only be called when there is ebpf probe binary change in NP agent which is not very frequent. It is possible only during NP agent image upgrades and very rare

Made calls from np agent to ipamd multiple times to make sure it responds as expected

```
{"level":"info","ts":"2025-02-11T02:34:02.580Z","caller":"rpc/rpc.pb.go:1296","msg":"Received request for Network Policy Agent configs"}
{"level":"info","ts":"2025-02-11T02:34:02.580Z","caller":"rpc/rpc.pb.go:1296","msg":"Send NetworkPolicyAgentConfigReply: NetworkPolicyMode: %!t(string=strict)"}
```

No

It should not break any upgrades/downgrades. Tested

No

No

```release-note

```


pkg/ipamd/rpc_handler.go
```diff
@@ -320,6 +321,18 @@ func (s *server) DelNetwork(ctx context.Context, in *rpc.DelNetworkRequest) (*rp
...
 
+func (s *server) GetNetworkPolicyConfigs(ctx context.Context, e *emptypb.Empty) (*rpc.NetworkPolicyAgentConfigReply, error) {
+
+	log.Infof("Received request for Network Policy Agent configs")
+
+	resp := &rpc.NetworkPolicyAgentConfigReply{
+		NetworkPolicyMode: s.ipamContext.networkPolicyMode,
+	}
+
+	log.Infof("Send NetworkPolicyAgentConfigReply: NetworkPolicyMode: %t", resp.NetworkPolicyMode)
```
jayanthvn, line 332: nit -> maybe '%v' ?

jayanthvn (approved):

pkg/ipamd/rpc_handler.go
```diff
@@ -320,6 +321,18 @@ func (s *server) DelNetwork(ctx context.Context, in *rpc.DelNetworkRequest) (*rp
 	return &rpc.DelNetworkReply{Success: err == nil, IPv4Addr: ipv4Addr, IPv6Addr: ipv6Addr, DeviceNumber: int32(deviceNumber)}, err
 }
 
+func (s *server) GetNetworkPolicyConfigs(ctx context.Context, e *emptypb.Empty) (*rpc.NetworkPolicyAgentConfigReply, error) {
```
orsenthil, line 324: We should add a comment here that this `GetNetworkPolicyConfigs` is invoked from the Network Policy Agent.

orsenthil (approved):
LGTM.

Issue #3203, opened 2025-02-13
Connection Issues with VPC CNI Network Policy Enforcement


janavenkat:
**What happened**:

We recently enabled network policy in AWS VPC CNI. After enabling it, we started experiencing intermittent connectivity issues when EKS pods attempt to connect to RabbitMQ hosted behind VPC peering. The issue occurs randomly, even though we do not have any deny network policies—only `allow_all ingress` and `allow_all egress` policies are in place.

Disabling network policy enforcement resolves the issue, while re-enabling it causes the problem to return. The network-policy-agent logs indicate a possible reversal in connection direction, suggesting a potential issue with conntrack.

We managed to get the `DENY` logs by enabling `enablePolicyEventLogs` 

Also we're using Otterize to automate our network polices. 

**Attach logs**
```  
 {
       "@timestamp": "2025-02-04 10:36:49.034",
       "@message": "Node: REDACTED.compute.internal; SIP: <INTERNAL_IP>; SPORT: 53912; DIP: <EXTERNAL_IP>; DPORT: 5671; PROTOCOL: TCP; PolicyVerdict: ACCEPT"
     },
     {
       "@timestamp": "2025-02-04 10:36:48.986",
       "@message": "Node: REDACTED.compute.internal; SIP: <EXTERNAL_IP>; SPORT: 5671; DIP: <INTERNAL_IP>; DPORT: 53912; PROTOCOL: TCP; PolicyVerdict: DENY"
     }
}
```
**What you expected to happen**:

We shouldn't expect any denied connections, as we're allowing all ingress and egress rules

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): 1.29
- CNI Version: 1.19.0
- OS (e.g: `cat /etc/os-release`): Amazon linux 2
- Kernel (e.g. `uname -a`): 5.10.233-233.887.amzn2.x86_64

orsenthil:
@janavenkat - this might be related to https://github.com/aws/aws-network-policy-agent/issues/345 please follow that issue.

janavenkat:
@orsenthil Thank you and the fix seems merged. When we can expect the release in EKS Addons? but for now, I can only see version 1.19.2 available.

orsenthil:
EKS Addons release is being made available, AWS docs will be updated. once the addons is available in all regions.

Pavani-Panakanti:
[...]

PR #3206, opened 2025-02-18
Changes to attach probes at pod start

haouc:
What type of PR is this?
bug

Which issue does this PR fix?:
https://github.com/aws/aws-network-policy-agent/issues/345

This PR is updated from https://github.com/aws/amazon-vpc-cni-k8s/pull/3188

What does this PR do / Why do we need it?:

    On every pod add, make a call to NP agent to configure ebpf probes. If NP is enabled, np agent will configure probes accordingly for the pod. If NP is disabled, np agent returns success. This is the same behavior for strict mode earlier, we are making same changes for standard mode too to fix the above github issue
    On every pod del, make a call to NP agent to clean up ebpf probes. If NP is enabled, np agent will clean up caches and probes associated with this pod. If not it will return success. Pod del will not throw any errors from NP.
    NP agent PR to handle these calls - 

    https://github.com/aws/aws-network-policy-agent/pull/361

Testing done on this change:

    Added unit tests
    Manual tests of pod creation and deletion with NP disabled and NP enabled.

Setup a generic ipv4 cluster with 3 nodes

EnforceNpToPod Testing - NP enabled
Created a pod and verified the EnforceNptoPod calls and probes are being attached to pods as expected in default allow. Verified pod connectivity in above case
Created multiple pods (10) at once and verified locking is happening in NP as expected and probes are being attached correctly
Did a large scale up and verified calls are being made as expected

EnforceNpToPod Testing - NP disabled
Created pods and verified that NP is returning success immediately

DeletePodNp - NP enabled
Deleted a pod and verified caches are being cleaned up and bpf programs are being deleted as expected
Deleted multiple pods at once and verified cleanup of caches and programs is working as expected
Verified locking on NP side when multiple calls from cni to NP happen at same time
Did a huge scale down and verified the functionality

DeletePodNp Testing - NP disabled
Deleted pods and verified that NP is returning success immediately

Tested below upgrade/downgrade functionality
If a pod is created before upgrade and deleted after upgrade - we will call DeletePodNp. If pod has policies applied, there will be probes attached and we will clean the caches and programs. If pod does not have any policies applied, we will not find any entries in the internal caches in NP related to this pod and we will just return success
If pod is created after upgrade and then deleted after downgrade - No call will be made from cni to np after downgrade. If pod has policies applied, probes will be cleaned up as part of reconcile code flow (This is how it works before these changes). If no policies are applied, there will be pod info left in internal caches and bpf program will not be deleted. Probes will be anyway deleted as part of pod delete. This should not be a concern. Internal caches will be reset on np agent restart

Tested NP enable -> disable and disable -> enable functionality as well
Tested all of the above with ipv6 cluster

    Integration tests successful - https://github.com/aws/amazon-vpc-cni-k8s/actions/runs/13122307097

Will this PR introduce any new dependencies?:
No

Will this break upgrades or downgrades? Has updating a running cluster been tested?:
Tested upgrades and downgrades. It will not break any updates or downgrades. CNI and NP have to be compatible after this change

No

No

```release-note

```


jayanthvn (dismissed):
lgtm, this is the same PR as - https://github.com/aws/amazon-vpc-cni-k8s/pull/3188

jayanthvn (approved):

janavenkat:
@haouc Still I have the issue even updating to version `1.19.3` and my k8s version `1.31` and https://github.com/aws/amazon-vpc-cni-k8s/issues/3203 explained already here.

I can reproduce:
1. A script that connects to RabbitMQ and other sites.
2. I've created a cron job that creates multiple pods (40), which also creates new nodes, running every 5 minutes.
3. However, I'm experiencing connection errors as described in the issue."

PR #3218, opened 2025-02-28
Init container changes for getting all the primary attached interfaces

jaydeokar:
**What type of PR is this?**
feature 

N/A

This adds changes to Init container for CNI. When Multiple NICs are initialized, the init container will identify these interfaces to set kernel parameters. 
We skip efa-only interfaces since these cannot be used for IP traffic

Tested on single card instance and multi-card instance 

```
time="2025-02-28T07:26:45Z" level=info msg="Copying CNI plugin binaries ..."
time="2025-02-28T07:26:45Z" level=info msg="Copied all CNI plugin binaries to /host/opt/cni/bin"
time="2025-02-28T07:26:45Z" level=info msg="found primary interfaces for the host [eth0]"
time="2025-02-28T07:26:45Z" level=info msg="Found primaryIF [eth0]"
time="2025-02-28T07:26:45Z" level=info msg="Updated net/ipv4/conf/eth0/rp_filter to 2\n"
time="2025-02-28T07:26:45Z" level=info msg="Updated net/ipv4/tcp_early_demux to 1\n"
time="2025-02-28T07:26:45Z" level=info msg="CNI init container done"
```

No new dependencies.. We already made IMDS calls. But we do have a bit more IMDS calls than previous

No, should not break upgrade or downgrade.. 

No

Not with this change

### DIFF
--- a/cmd/aws-vpc-cni-init/main.go
+++ b/cmd/aws-vpc-cni-init/main.go
@@ -15,12 +15,15 @@
 package main
 
 import (
+	"context"
 	"os"
 
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/awsutils"
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/awsutils/awssession"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/procsyswrapper"
 	"github.com/aws/amazon-vpc-cni-k8s/utils"
 	"github.com/aws/amazon-vpc-cni-k8s/utils/cp"
-	"github.com/aws/amazon-vpc-cni-k8s/utils/imds"
+	ec2metadata "github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
 
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


cmd/aws-vpc-cni-init/main.go
```diff
@@ -41,45 +44,89 @@ const (
...
+	ctx := context.Background()
+
+	awsconfig, err := awssession.New(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create aws session")
+	}
+	ec2Metadata := ec2metadata.NewFromConfig(awsconfig)
+	imds := awsutils.TypedIMDS{ec2Metadata}
+
+	macAddresses, err := imds.GetMACs(ctx)
```
jayanthvn, line 60: Won't this return all ENIs - i.e, even secondary ENIs - https://github.com/aws/amazon-vpc-cni-k8s/blob/0dccb2296cc25828b2772daae0f8297185b04686/pkg/awsutils/imds.go#L210

jaydeokar: Yes.. But we are filtering the ENIs which are at device-number 0..

jaydeokar:
Closing this PR as we don't need these changes

PR #3227, opened 2025-03-03
Update helm chart to ensure that created eniconfig name is always a string

cheeseandcereal:
**What type of PR is this?**
bug

no issue # created for this

There's a bug that's possible with the current helm chart, where if `.Values.eniConfig.subnets` is specified, and the keys for this `subnets` objects contains any strings that are numbers, or other strings like "true", "false", "null", etc, then when the helm chart tries to apply these ENIConfigs, the `metadata.name` field is not properly quoted, and the name value ends up being the wrong data type.

This PR updates to simply wrap this key in quotes, so even if these sorts of keys are used, they are properly treated as strings.

tried templating/installing the helm chart and confirmed the output. For example:

In values, set `eniconfig` as:

```yaml
eniConfig:
  # Specifies whether ENIConfigs should be created
  create: true
  region: us-west-2
  subnets:
    2:
      id: subnet-123
      securityGroups:
      - sg-123
```

Output from `helm template charts/aws-vpc-cni`

```yaml
...
---
# Source: aws-vpc-cni/templates/eniconfig.yaml
kind: ENIConfig
metadata:
  name: "2"
spec:
  securityGroups:
    - sg-123
  subnet: subnet-123
```

no

no

no

no

```release-note

```

### DIFF
--- a/charts/aws-vpc-cni/templates/eniconfig.yaml
+++ b/charts/aws-vpc-cni/templates/eniconfig.yaml
@@ -3,7 +3,7 @@
 apiVersion: crd.k8s.amazonaws.com/v1alpha1
 kind: ENIConfig
 metadata:
-  name: {{ $key }}
+  name: "{{ $key }}"
 spec:
   {{- if $value.securityGroups }}
   securityGroups:


orsenthil:
> There's a bug that's possible with the current helm chart, where if .Values.eniConfig.subnets is specified, and the keys for this subnets objects contains any strings that are numbers, or other strings like "true", "false", "null", etc, then when the helm chart tries to apply these ENIConfigs, the metadata.name field is not properly quoted, and the name value ends up being the wrong data type.

This looks like more a test or negative test to me than an actual use case. 
Subnets caannot be things like `2` true or false. Perhaps the customer will have to fix their input.

cheeseandcereal:
> Subnets caannot be things like 2 true or false. Perhaps the customer will have to fix their input.

This isn't for the actual id of the subnet, this is for the key of the object, which is used as the eniconfig name. It can be any arbitrary string as far as I understand. It's a user-supplied arbitrary string value that doesn't have to map to an actual subnet identifier of any sort.

orsenthil:
> It can be any arbitrary string as far as I understand. It's a user-supplied arbitrary string value that doesn't have to map to an actual subnet identifier of any sort.

Technically, yes. Personally, I am bit ambivalent on this change.
We haven't quoted anything like this for invalid user input here and this will stand out as an exception.

Perhaps getting another opinion from @M00nF1sh  or @jayanthvn might help.

cheeseandcereal:
Wrapping these templated values in quotes is fairly standard practice for helm charts because of this exact mis-typing reason. Will defer to other reviewers, but yeah. Providing a valid string key for the object name can result in the helm chart failing to install which definitely doesn't seem right to me.

M00nF1sh (approved):
Even though I doubt there are real use cases to use a number or (true/false) as eniConfig's name, i think quote them is no harm anyway since it's not breaking existing valid use case.

If in the future we want pose some restrictions on the eniConfig's name, we should check explicitly instead of rely on this YAML rendering issue

Issue #3196, opened 2025-02-10
Update netlink to v1.3.1


orsenthil:
We have noticed that when CNI attempts to list the IPv6 addresses using the netlink library, it can return an  EINTR (interrupted system call) error if the kernel's data structures change during the address listing process.

 The current version of the netlink library used by the VPC CNI (v1.3.0) does not handle this error properly, leading to the observed error message.

This is fixed by an enhancement to Netlink Library in this change - https://github.com/vishvananda/netlink/commit/084abd93d350e97ee5410b5b6311bcc211f7ea05 and the netlink PR - https://github.com/vishvananda/netlink/pull/1018 

and is made available only since v1.3.1

**What would you like to be added**:

We are tracking as PR merged into master/v1.3.1

https://github.com/vishvananda/netlink/pull/1018

vadasambar:
waiting for review: https://github.com/aws/amazon-vpc-cni-k8s/pull/3237

yash97:
I think we should try to follow approach taken here https://github.com/containernetworking/plugins/pull/1154. To add retry wrapper if interrupt error is returned.

jaydeokar:
Changes are merged. This will be released with the next version

PR #3237, opened 2025-03-18
fix: update netlink dep to latest commit

vadasambar:
- this fixes interrupted sys call errors in the plugin

**What type of PR is this?**
bug

#3196 

This PR fixes the frequent interrupted sys call errors like these:
```
Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox \"xxxxxxxx\": plugin type=\"aws-cni\" name=\"aws-cni\" failed (add): add command: failed to setup network: SetupPodNetwork: failed to setup veth pair: failed to setup veth network: setup NS network: failed while waiting for v6 addresses to be stable: could not list addresses: interrupted system call
```

No

No

No

```release-note
fix: interrupted system call error
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,7 +32,7 @@ require (
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.10.0
-	github.com/vishvananda/netlink v1.3.0
+	github.com/vishvananda/netlink v1.3.1-0.20250303224720-0e7078ed04c8
 	go.uber.org/zap v1.27.0
 	golang.org/x/net v0.33.0
 	golang.org/x/sys v0.28.0
--- a/go.sum
+++ b/go.sum
@@ -428,8 +428,8 @@ github.com/stretchr/testify v1.8.0/go.mod h1:yNjHg4UonilssWZ8iaSj1OCr/vHnekPRkoO
 github.com/stretchr/testify v1.8.1/go.mod h1:w2LPCIKwWwSfY2zedu0+kehJoqGctiVI29o6fzry7u4=
 github.com/stretchr/testify v1.10.0 h1:Xv5erBjTwe/5IxqUQTdXv5kgmIvbHo3QQyRwhJsOfJA=
 github.com/stretchr/testify v1.10.0/go.mod h1:r2ic/lqez/lEtzL7wO/rwa5dbSLXVDPFyf8C91i36aY=
-github.com/vishvananda/netlink v1.3.0 h1:X7l42GfcV4S6E4vHTsw48qbrV+9PVojNfIhZcwQdrZk=
-github.com/vishvananda/netlink v1.3.0/go.mod h1:i6NetklAujEcC6fK0JPjT8qSwWyO0HLn4UKG+hGqeJs=
+github.com/vishvananda/netlink v1.3.1-0.20250303224720-0e7078ed04c8 h1:Y4egeTrP7sccowz2GWTJVtHlwkZippgBTpUmMteFUWQ=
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -32,7 +32,7 @@ require (
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.10.0
-	github.com/vishvananda/netlink v1.3.0
+	github.com/vishvananda/netlink v1.3.1-0.20250303224720-0e7078ed04c8
```
vadasambar, line 35: v1.3.1 hasn't been released yet. This uses the latest commit at the time of creating this PR: https://github.com/vishvananda/netlink/commit/0e7078ed04c84cea47daea45be061544e565ec49

vadasambar: We could handle the error in the code here: https://github.com/vishvananda/netlink/commit/0e7078ed04c84cea47daea45be061544e565ec49

with something like:
```go
		if errors.Is(err, unix.EINTR) || err != nil {
			return fmt.Errorf("could not list addresses: %v", err)
		}
```
but we might end up ignoring unix.EINTR's which really matter.

haouc:
@vadasambar thanks for updating this. Did you get chance to test? If you did, can you comment with the reprod you had in the overview?

vadasambar:
> @vadasambar thanks for updating this. Did you get chance to test? If you did, can you comment with the reprod you had in the overview?

This PR is only updating a dependency. If the tests are passing (for this repo), even if the issue is not fixed (I think the issue will be fixed), we would be in a better state. 

The only problem I see is using a master/trunk version of a dependency. Provided upstream dependency is doing its due diligence, we should be fine? Let me know what you think.

vadasambar:
@haouc ^

yash97:
I don't see the urgency to use package commit while upgrading dependency. We can update package when it is officially released. Even with this update we only get benefit of error categorization. But we either we retry silently  using this new update which is not officially released yet or we let kubelet retry on failure. It doesn't seem worth the risk of importing package with a tag with no official release as of now.

vadasambar:
> I don't see the urgency to use package commit while upgrading dependency. We can update package when it is officially released. Even with this update we only get benefit of error categorization. 

We could wait for the upstream to update the dependency but we don't know when that would happen. The fix was merged in Sep 2024. 

> Even with this update we only get benefit of error categorization. But we either we retry silently using this new update which is not officially released yet or we let kubelet retry on failure. It doesn't seem worth the risk of importing package with a tag with no official release as of now.

We are throwing an error in the CNI (now) and letting kubelet retry vs using the result from call (before) even if the system call was interrupted (and then keep on retrying until the result contains the the value we were looking for). If we let kubelet retry, it's going to add to the pod startup time. 

I guess we have 3 options now:
1. Wait for upstream to release the fix (no timeline on this; last releases: 2024, 2022, 2020)
2. Use the upstream's trunk version (comes with risk of using an unreleased version of a dependency)
3. Handle the error in CNI (need to be careful with dropping `sys call interrupted` errors; might need to replicate upstream's fix until there's a new release, something like [this](https://github.com/vishvananda/netlink/pull/1018/files#diff-fce4aa7fac10c3abf3797195b2d1b10a59caece6fe79d30ed9c7fed56088eeeaR192) and [this](https://github.com/vishvananda/netlink/pull/1018/files#diff-26442c63ef579f249eb12f2fd0bd406a1adc54f5d774c44d3a9c8dfed7aae380R51))

yash97:
regarding this PR I think we should try to follow approach taken here https://github.com/containernetworking/plugins/pull/1154. To add retry wrapper if interrupt error is returned.

jaydeokar:
Closing in favor of this PR https://github.com/aws/amazon-vpc-cni-k8s/pull/3286

PR #3253, opened 2025-04-10
Add env for node agent enabled

Pavani-Panakanti:
**What type of PR is this?**
bug

We should fail on pod creation if we can't connect to np agent and fail to attach probes at pod start. 

Some of the cx might have deleted NP container if they are not using network policies, so adding env_var NODE_AGENT_ENABLED. If cx is deleting np container they have to set this to false. When this is set to false we don't try to connect to np container at all

nodeAgentEnabled: true, removed np container and verified that pod creation is failing 
nodeAgentEnabled: true, np agent running - verified that pod creation is successful 
nodeAgentEnabled: false, np agent running - no calls being made to np agent 
nodeAgentEnabled: false, np agent not running - no calls being made to np agent

No

No

Adding a new env_var NODE_AGENT_ENABLED

Yes. We are adding a new env_var. We need to add release notes and also update docs

```release-note

```

### DIFF
--- a/charts/aws-vpc-cni/templates/daemonset.yaml
+++ b/charts/aws-vpc-cni/templates/daemonset.yaml
@@ -84,6 +84,8 @@ spec:
             - name: {{ $key }}
               value: {{ $value | quote }}
 {{- end }}
+            - name: NODE_AGENT_ENABLED
+              value: {{ .Values.nodeAgent.enabled | quote }}
 {{- with .Values.extraEnv }}
       {{- toYaml .| nindent 12 }}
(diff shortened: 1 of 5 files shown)


oliviassss:
I may not have a full context here for existing issues, but it looks unnecessary and and a bit manual for UX to introduce a new ENV var. the vpc cni should be able to self-recover if the NPA container exists or not.

Pavani-Panakanti:
Going with a different approach and using existing variable here https://github.com/aws/amazon-vpc-cni-k8s/pull/3254

PR #3254, opened 2025-04-11
Skip configuring NP related if network_policy_enforcing_mode is not set

Pavani-Panakanti:
**What type of PR is this?**
Enhancement

We should fail on pod creation if we can't connect to np agent and fail to attach probes at pod start.

Some of the cx might have deleted NP container if they are not using network policies, so we are using env_var NETWORK_POLICY_ENFORCING_MODE to check if cx wants network policy configuration or not. If cx is deleting np container they have to not set this env_var. When this is not set we don't try to connect to np container at all

standard mode, np enabled
strict mode, np enabled
"" mode, np enabled --> invalid mode fails
not set, np container present --> we skip reaching out to np container
not set, np container not present --> we skip call to np container 
set to some value, np container not present --> pod creates fail

No

No

No

Cx who have removed NP container should not set NETWORK_POLICY_ENFORCING_MODE

```release-note
Cx who have removed NP container should not set NETWORK_POLICY_ENFORCING_MODE
```


charts/aws-vpc-cni/templates/daemonset.yaml
```diff
@@ -81,9 +81,13 @@ spec:
             timeoutSeconds: {{ .Values.readinessProbeTimeoutSeconds }}
           env:
 {{- range $key, $value := .Values.env }}
+{{- if and (eq $key "NETWORK_POLICY_ENFORCING_MODE") (not .Values.nodeAgent.enabled) }}
```
jayanthvn, line 84: I'm fine with this but a minor nit, we could make it readable wdyt? something like this -

```
{{- $skipKey := and (eq $key "NETWORK_POLICY_ENFORCING_MODE") (not .Values.nodeAgent.enabled) }}
{{- if not $skipKey }}
  - name: {{ $key }}
    value: {{ $value | quote }}
{{- else }}
  # Skipping NETWORK_POLICY_ENFORCING_MODE because nodeAgent is disabled
{{- end }}

```

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -451,6 +454,10 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 		log.Warnf("Container %s did not have a valid IP %s", args.ContainerID, r.IPv4Addr)
 	}
 
+	if r.NetworkPolicyMode == "" {
+		log.Infof("NETWORK_POLICY_ENFORCING_MODE is not set")
+		return nil
+	}
```
jayanthvn, line 460: In line 466, we need the same fix as add for logs right?

```
		log.Errorf("Failed to connect to network policy agent: %v", err)
		return errors.New("add cmd: failed to setup network policy")
```

jayanthvn:
Can you check metrics helper compilation is failing?

Also can you add a note here - https://github.com/aws/amazon-vpc-cni-k8s/tree/master?tab=readme-ov-file#network_policy_enforcing_mode-v1171 for use case of someone removing the NP container?

charts/aws-vpc-cni/templates/daemonset.yaml
Pavani-Panakanti: yeah this looks more readable. Updated

cmd/routed-eni-cni-plugin/cni.go
Pavani-Panakanti: We were never failing in NP when there are any errors in delete path so followed the same here but it might be better to make that more strict to avoid leaving any maps behind. Updated

Pavani-Panakanti:
Fixed metrics.go and also updated README. We still have one vulncheck failing for updating goversion to 1.24.2 but we don't have the base image for 1.24.2 yet

pkg/ipamd/ipamd.go
```diff
@@ -1846,13 +1845,14 @@ func EnablePodENI() bool {
...
-		return "", errors.New("invalid Network policy mode, supported modes: none, strict, standard")
+	value, exists := os.LookupEnv(envNetworkPolicyMode)
+	if !exists {
+		return "", nil
+	}
+	if utils.IsValidNetworkPolicyEnforcingMode(value) {
+		return value, nil
 	}
-	return defaultNetworkPolicyMode, nil
+	return "", errors.New("invalid Network policy mode, supported modes: none, strict, standard")
```
oliviassss, line 1855: just to confirm backward compatibility, but default the helm chart has the `NETWORK_POLICY_ENFORCING_MODE=strict` right? So the change should not break any existing cx behavior

Pavani-Panakanti: By default network_policy_enforcing_mode is set to standard https://github.com/aws/amazon-vpc-cni-k8s/blob/master/charts/aws-vpc-cni/values.yaml#L90. It should not break any existing behavior. Cx who explicitly remove the env_var entirely will be impacted and we upgraded doc to state that

jaydeokar (approved):
lgtm

ghost:
FYI, no issue, but this was a breaking change for us and wasn't obvious in the changelog or in the diff between previous and new manifests. Note that we are not using Helm for deployment.

Prior to this change (running v1.19.3), we had the value `none` for the env. var. It wasn't clear what the expected behaviour was when no policy agent was running (we are using Calico). When upgrading to 1.20.4, this became a breaking change. It would have been nice if the CNI pod failed to start due to the invalid value of the environment variable, rather than silently assuming the policy agent was desired.

(I've just seen that it was called out in the v1.19.4 release notes).

PR #3256, opened 2025-04-14
Merged master into release-1.19 with master versions for all conflicts

oliviassss:
**What type of PR is this?**

```release-note

```


charts/aws-vpc-cni/values.yaml
```diff
@@ -46,6 +46,7 @@ nodeAgent:
   networkPolicyAgentLogFileLocation: "/var/log/aws-routed-eni/network-policy-agent.log"
   enableIpv6: "false"
   metricsBindAddr: "8162"
+  metricsBindPort: "8162"
```
jaydeokar, line 49: This should not have made it to the original PR. I am not sure why this was allowed.

pkg/ipamd/rpc_handler.go
```diff
@@ -316,9 +317,21 @@ func (s *server) DelNetwork(ctx context.Context, in *rpc.DelNetworkRequest) (*rp
...
 	}
 
-	log.Infof("Send DelNetworkReply: IPv4Addr: %s, IPv6Addr: %s, DeviceNumber: %d, err: %v", ipv4Addr, ipv6Addr, deviceNumber, err)
+	log.Infof("Send DelNetworkReply: IPv4Addr: %s, IPv6Addr: %s, DeviceNumber: %d, NetworkPolicyMode: %s, err: %v", ipv4Addr, ipv6Addr, deviceNumber, s.ipamContext.networkPolicyMode, err)
 
-	return &rpc.DelNetworkReply{Success: err == nil, IPv4Addr: ipv4Addr, IPv6Addr: ipv6Addr, DeviceNumber: int32(deviceNumber)}, err
+	return &rpc.DelNetworkReply{Success: err == nil, IPv4Addr: ipv4Addr, IPv6Addr: ipv6Addr, DeviceNumber: int32(deviceNumber), NetworkPolicyMode: s.ipamContext.networkPolicyMode}, err
+}
+
+func (s *server) GetNetworkPolicyConfigs(ctx context.Context, e *emptypb.Empty) (*rpc.NetworkPolicyAgentConfigReply, error) {
```
jaydeokar, line 325: This seems duplicated

oliviassss: just remove this

charts/aws-vpc-cni/values.yaml
oliviassss: yeah this looks weird...
it's added by this commit: https://github.com/aws/amazon-vpc-cni-k8s/commit/28c99c9443d2c3a3408cf9d1d67f7e395dd92ee8

oliviassss: picked up the fix in master

oliviassss:
the image `public.ecr.aws/eks-distro-build-tooling/golang:1.24.2-gcc-al2` was released yesterday, will upgrade it in a separate PR.

pkg/ipamd/ipamd.go
```diff
@@ -523,21 +535,32 @@ func (c *IPAMContext) nodeInit() error {
...
+			maxPods, isInt64 := node.Status.Capacity.Pods().AsInt64()
+			if !isInt64 {
+				log.Errorf("Failed to parse max pods: %s", node.Status.Capacity.Pods().String)
+				podENIErrInc("nodeInit")
+				return errors.New("error while trying to determine max pods")
+			}
+			c.maxPods = int(maxPods)
+		}
+	} else {
+		maxPods, err := c.getMaxPodsFromFile()
```
jaydeokar, line 556: How will this work if we use PD ? 
Won't it mean we can schedule more Pods on the node ?

oliviassss: Synced offline. During init we wait 30s for apiserver connectivity, during which most of the cases should be connected, so expect minimal impact on existing cx.
Without API server, IPAMD will run in a limited mode, meaning no k8s cache, and reading maxPods from static file. Another option is to copy kubelet config to the container and read the maxPods from there, but that may not be accurate either, and may require security review.
Once the apiserver connected in the background, we will re-read maxPods from node status, and re-create k8s cache for accuracy.

jaydeokar (dismissed):
lgtm

jaydeokar (approved):

PR #3277, opened 2025-04-30
fix node init failure, add sgpp test in script

oliviassss:
**What type of PR is this?**

Bug

1. The node object got redeclared and local to the `c.withApiServer{}` block, which lead to IPAMD failed to discover ENIConfig of the node because the node obj passed to the `GetNodeSpecificENIConfigName()` [here](https://github.com/aws/amazon-vpc-cni-k8s/blob/release-1.19/pkg/ipamd/ipamd.go#L569) is a nil.
2. Add sgpp+customnetworking test suites in integration test.

Ran the custom-networking-sgpp integration test manually and it passed
```

```

```release-note

```


jaydeokar (dismissed):

haouc (approved):
lgtm

PR #3287, opened 2025-05-21
[Draft]feat: adding DISABLE_ENI_METADATA_RECONCILE configuration so ipamd will not interact with EC2 for node init and reconcile

bhaoz:
(feat: adding DISABLE_ENI_METADATA_RECONCILE configuration so ipamd will not interact with EC2 for node init and reconcile)

**What type of PR is this?**

feature

```release-note

```


yash97:
Please provide description, reason, use case for this feature.

pkg/awsutils/awsutils.go
```diff
@@ -396,6 +397,7 @@ func New(useSubnetDiscovery, useCustomNetworking, disableLeakedENICleanup, v4Ena
 	log.Infof("Subnet discovery enabled %v", cache.useSubnetDiscovery)
 	cache.v4Enabled = v4Enabled
 	cache.v6Enabled = v6Enabled
+	cache.disableENIMetadataReconcile = disableENIMetadataReconcile
```
bluekvirus, line 400: I guess same question here, should we avoid adding to 'cache' obj. Looks like we just need to read the input param and skip the section in DescribeAllENIs()

pkg/ipamd/ipamd.go
```diff
@@ -134,6 +134,10 @@ const (
 	// disableLeakedENICleanup is used to specify that the task checking and cleaning up leaked ENIs should not be run.
 	envDisableLeakedENICleanup = "DISABLE_LEAKED_ENI_CLEANUP"
 
+	// disableENIMetadataReconcile is used to specify that EC2 API calls should be skipped and only IMDS metadata should be
+	// used to get ENI information.
+	envDisableENIMetadataReconcile = "DISABLE_ENI_METADATA_RECONCILE"
```
jaydeokar, line 139: Lets name it as `DISABLE_EC2_API_CALLS` instead..

bluekvirus (changes requested):
I'm thinking if possible, lets minimize the change for now. Add reading of ENV and honor this through the new input param of util function, without changing any objects. Just pure routine skip by single ENV read at the time of calling the util function?

pkg/ipamd/ipamd.go
bhaoz: Renamed to ```ENABLE_IMDS_ONLY_MODE``` but let me know which name you feel better

pkg/awsutils/awsutils.go
```diff
@@ -1353,6 +1355,25 @@ func (cache *EC2InstanceMetadataCache) DescribeAllENIs() (DescribeAllENIsResult,
 		eniMap[eni.ENIID] = eni
 	}
 
+	// If DISABLE_ENI_METADATA_RECONCILE is enabled, skip EC2 API call and return IMDS data only
+	if cache.disableENIMetadataReconcile {
+		log.Debug("DISABLE_ENI_METADATA_RECONCILE is enabled, skipping EC2 API call and using IMDS metadata only")
+		// Collect the verified ENIs
+		var verifiedENIs []ENIMetadata
+		for _, eniMetadata := range eniMap {
+			verifiedENIs = append(verifiedENIs, eniMetadata)
```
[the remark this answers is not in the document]
bhaoz: The thing is, we have a lot of logic checking EC2 response in the following code. I'm changing it to include multi-cards logic in the next revision

pkg/awsutils/awsutils.go
```diff
@@ -633,6 +642,7 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
 	}
 	ipv4Available := false
 	ipv6Available := false
+	networkCard := 0
```
bluekvirus, line 645: Is there a way to avoid tackling the network card info? I'm thinking we want to skip the whole network cards / device > 1 thingy. Just focus on disabling api calls.

pkg/awsutils/awsutils.go
```diff
@@ -655,6 +665,14 @@ func (cache *EC2InstanceMetadataCache) getENIMetadata(eniMAC string) (ENIMetadat
 				log.Debugf("Found IPv6 addresses associated with interface. This is not efa-only interface")
 			}
 		}
+		if field == "network-card" {
+			networkCard, err = cache.imds.GetNetworkCard(ctx, eniMAC)
+			if err != nil {
+				awsAPIErrInc("GetNetworkCard", err)
+				log.Errorf("Network Card data not found from", networkCard)
```
jaydeokar, line 672: nit - format specifier missing. And we should log error, not `networkCard`

pkg/ipamd/ipamd.go
```diff
@@ -1829,15 +1833,19 @@ func getMinimumIPTarget() int {
...
 func disableENIProvisioning() bool {
-	return utils.GetBoolAsStringEnvVar(envDisableENIProvisioning, false)
+	return enableImdsOnlyMode() || utils.GetBoolAsStringEnvVar(envDisableENIProvisioning, false)
 }
 
 func disableLeakedENICleanup() bool {
 	// Cases where leaked ENI cleanup is disabled:
 	// 1. IPv6 is enabled, so no ENIs are attached
 	// 2. ENI provisioning is disabled, so ENIs are not managed by IPAMD
 	// 3. Environment var explicitly disabling task is set
```
jaydeokar, line 1843:
```suggestion
	// 3. CNI is operating in IMDS only mode
	// 4. Environment var explicitly disabling task is set
```

pkg/awsutils/awsutils.go
```diff
@@ -78,6 +80,10 @@ const (
 
 	// the default page size when paginating the DescribeNetworkInterfaces call
 	describeENIPageSize = 1000
+
+	// enableImdsOnlyMode is used to specify that EC2 API calls should be skipped and only IMDS metadata should be
+	// used to get ENI information.
```
jaydeokar, line 85: You don't need to define the variable here again

pkg/awsutils/awsutils.go
bhaoz: What format specifier missing you're referring to? From existing code, it seems we always return error to upstream and handled by GetAttachedENIs() method. I think it's ok to follow existing pattern?

pkg/awsutils/awsutils.go
bhaoz: but we cannot import from ```ipamd``` as that will introduce circular imports. Didn't find a file for putting constant in current repo. I can have a new file for shared constant

pkg/awsutils/awsutils.go
jaydeokar: log.Errorf("Network Card data not found from %v",  err)

jaydeokar: Something like this^

bhaoz: The thing is, error is passed above to [here](https://github.com/aws/amazon-vpc-cni-k8s/blob/4cf248870605ac8eec5030f6bfd7927b773cdf05/pkg/awsutils/awsutils.go#L590) and then handled [here](
https://github.com/aws/amazon-vpc-cni-k8s/blob/4cf248870605ac8eec5030f6bfd7927b773cdf05/pkg/ipamd/ipamd.go#L1375) with logging of the error messages. We're not lostign track of the error. Is duplicating a error logging helps here? Also the above error handling is following this similar pattern.

jaydeokar: The only thing missing is format specifier "%v" is what I was calling out before.

bhaoz: Updated in latest revision

jaydeokar (approved):
lgtm

PR #3303, opened 2025-06-03
Allow configuration of the logLevel for network-policy-agent

nclaeys:
The agent logs by default on info, it would be good to allow users to overwrite this using through the helm chart.

**What type of PR is this?**
improvement of chart

The node-policy agent is verbose and the helm chart does not allow overwriting this.

Give users the option to tweak the configuration of the network policy agent.

Generating the helm chart and applying

/

No, kept the default as it exists now

/

/


viveksb007 (approved):
lgtm

jaydeokar:
This was already merged. Closing

PR #3305, opened 2025-06-05
feat: expose max allocatable IPs thru grpc server

mselim00:
**What type of PR is this?**
feature

N/A

Exposes the maximum number of IP addresses that IPAMD believes can be allocated in its current environment, given the applied VPC CNI configurations.

Deployed this to my own cluster and wrote a simple application to make the GetMaxAllocatableIPs request to the GRPC server.

No

No

No

```release-note
NONE
```


go.mod
```diff
@@ -177,8 +177,3 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.6.0 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
-
-replace gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776 => gopkg.in/yaml.v3 v3.0.1
-
-// Version of go-cose v1.2.0 and v1.2.1 have been deprecated in favor v1.1.0
-replace github.com/veraison/go-cose => github.com/veraison/go-cose v1.1.0
```
mselim00: go-cose isn't used AFAICT, and I'm not yet clear on why the replace was added for yaml. It was merged in an integration tests fix (ref: https://github.com/aws/amazon-vpc-cni-k8s/commit/56d95cd3120ba2e5117393af6a310de111561159)

adding replace directives blocks other projects from adding dependencies on this one (which blocks re-use of the grpc client constructs, or others), so it'd be best to avoid them if possible. still trying to figure out the yaml one though

mselim00: looks like the yaml replace was actually added in https://github.com/aws/amazon-vpc-cni-k8s/commit/ecfe8fa3a2fe1a7201d493ff46ebb14a23934aca but still no details on why. there's no change to the go.sum though, looks like the specified version isn't being used anyway

mselim00: I'll leave this separate, it's not immediately blocking to use the new method in the grpc server that's exposed here

rpc/mocks/rpc_mocks.go
```diff
@@ -153,46 +173,3 @@ func (mr *MockNPBackendClientMockRecorder) EnforceNpToPod(arg0, arg1 interface{}
 	varargs := append([]interface{}{arg0, arg1}, arg2...)
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "EnforceNpToPod", reflect.TypeOf((*MockNPBackendClient)(nil).EnforceNpToPod), varargs...)
 }
-
```
jaydeokar: Can you rebase ? These shouldn't be deleted

mselim00: good catch, rebased and re-generated to fix that

jaydeokar:
@mselim00 can you rebase the changes and fix the conflicts ?

mselim00:
We went a different direction with this one, so this is no longer needed. See https://github.com/awslabs/amazon-eks-ami/pull/2446 for reference

Issue #3311, opened 2025-06-18
Version v1.19.4+ gets TLS timeouts on start when using AWS Firewall due to X25519MLKEM768


brinnjoyce:
**What happened**:
When using an AWS firewall with TLS SNI rules to allow traffic versions v1.19.4 and v1.19.5 cannot start as it gets blocked by the AWS firewall.

**Attach logs**
Relevant section of ipamd.log:
```
{"level":"error","ts":"2025-06-18T05:43:01.767Z","caller":"ipamd/ipamd.go:440","msg":"Failed to call ec2:DescribeNetworkInterfaces for [eni-097b09b811ad42866 eni-02dfb197b8da23113]: operation error EC2: DescribeNetworkInterfaces, exceeded maximum number of attempts, 3, https response error StatusCode: 0, RequestID: , request send failed, Post \"https://ec2.ap-southeast-2.amazonaws.com/\": net/http: TLS handshake timeout"}
```

**What you expected to happen**:
It should not get blocked by the firewall.

**Anything else we need to know?**:

I have confirmed that this is the issue by setting the env var `GODEBUG=tlsmlkem=0` like so:
```
kubectl -n kube-system set env daemonset aws-node --containers=aws-node GODEBUG=tlsmlkem=0
```

This should be set in the go.mod as other projects have done to avoid this problem

Related to https://github.com/golang/go/issues/67061 https://github.com/hashicorp/terraform-provider-aws/pull/41740

brinnjoyce:
We have fixed this on the Network Firewall side by adding the `ssl_state:client_hello` keyword to our tls drop rule to ensure it waits for the Client Hello to complete before it drops the traffic.

PR #3312, opened 2025-06-18
Set godebug tlsmlkem=0 to avoid AWS Firewall issues

brinnjoyce:
**What type of PR is this?**
bug

[3311](https://github.com/aws/amazon-vpc-cni-k8s/issues/3311)

Allows the vpc-cni addon to work in cases where AWS Firewall is allowing traffic based on TLS SNI rules

N/A - tested via setting this as an environment variable to confirm the fix

No

No

No

No

```release-note
Set godebug tlsmlkem=0 to avoid AWS Firewall issues
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,10 @@ module github.com/aws/amazon-vpc-cni-k8s
 
 go 1.24.4
 
+// At the time of adding this configuration, the new Go feature introduced here https://github.com/golang/go/issues/67061,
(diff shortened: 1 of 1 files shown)


brinnjoyce:
We have fixed this on the Network Firewall side by adding the `ssl_state:client_hello` keyword to our tls drop rule to ensure it waits for the Client Hello to complete before it drops the traffic.

PR #3319, opened 2025-06-20
Enhanced Subnet Discovery

dshehbaj:
add testing doc here

add integration test screenshot here

add edge cases here

**What type of PR is this?**

```release-note

```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -61,8 +61,10 @@ const (
 
 	// AllocENI need to choose a first free device number between 0 and maxENI
 	// 100 is a hard limit because we use vlanID + 100 for pod networking table names
-	maxENIs                 = 100
-	clusterNameEnvVar       = "CLUSTER_NAME"
+	maxENIs           = 100
+	clusterNameEnvVar = "CLUSTER_NAME"
+	// clusterTagKeyPrefix is the prefix for the cluster-specific subnet tags
+	clusterTagKeyPrefix     = "kubernetes.io/cluster/"
 	eniNodeTagKey           = "node.k8s.amazonaws.com/instance_id"
 	eniCreatedAtTagKey      = "node.k8s.amazonaws.com/createdAt"
 	eniClusterTagKey        = "cluster.k8s.amazonaws.com/name"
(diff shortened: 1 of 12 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -966,28 +969,65 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
+			if checkErr != nil {
+				// If we can't determine exclusion status, log warning and proceed
+				log.Warnf("Failed to check if primary subnet is excluded: %v. Proceeding with ENI creation attempt.", checkErr)
+			} else if excluded {
+				// Primary subnet is explicitly excluded
+				err = errors.New("primary subnet is tagged with kubernetes.io/role/cni=0 and subnet discovery is disabled - no valid subnets available for ENI creation")
+				log.Error(err.Error())
+				return "", err
+			}
+
```
dshehbaj, line 1179: Deals with misconfiguration scenario where customer tags subnet for exclusion but does not have subnet discovery enabled.

pkg/awsutils/awsutils.go
```diff
@@ -966,28 +969,65 @@ func (cache *EC2InstanceMetadataCache) createENI(useCustomCfg bool, sg []*string
...
+				// Even in fallback, check if primary subnet is excluded
+				excluded, checkErr := cache.isPrimarySubnetExcluded()
+				if checkErr != nil {
+					log.Warnf("Failed to check if primary subnet is excluded: %v. Proceeding with ENI creation attempt.", checkErr)
+				} else if excluded {
+					// Primary subnet is explicitly excluded
+					err = errors.New("primary subnet is tagged with kubernetes.io/role/cni=0 - no valid subnets available for ENI creation")
+					log.Error(err.Error())
+					return "", err
+				}
```
dshehbaj, line 1122: Dealing with the case where describe subnet call fails, but customer has also opted to exclude the primary subnet.

pkg/ipamd/datastore/data_store.go
```diff
@@ -797,6 +843,11 @@ func (ds *DataStore) GetIPStats(addressFamily string) *DataStoreStats {
 		TotalPrefixes: ds.allocatedPrefix,
 	}
 	for _, eni := range ds.eniPool {
+		// Skip excluded ENIs when calculating available IPs for pod allocation
+		if eni.IsExcludedForPodIPs {
+			continue
+		}
+
```
dshehbaj, line 850: We skip the excluded ENI (which would the Primary ENI in this case practically speaking) from IP stat calculation. 

Doing it here would automatically take care of warm target calculation in other places such as `isDatastorePoolTooLow`, `shouldRemoveExtraENIs`.

pkg/awsutils/awsutils.go
```diff
@@ -506,6 +510,205 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+		awsAPIErrInc("DiscoverCustomSecurityGroups", err)
+		log.Warnf("Failed to discover custom security groups: %v", err)
+		log.Info("Falling back to using primary security groups for ENIs in secondary subnets")
+
+		// Clear custom security groups cache to trigger fallback behavior
+		cache.customSecurityGroups.Set([]string{})
+
+		// Apply primary security groups to ENIs in secondary subnets as fallback
+		return cache.applyFallbackSecurityGroups(store)
+	}
```
dshehbaj: Fallback to primary security group and show logs to customer if we fail to describe any secondary security groups.

We do the same fallback for primary subnets.

PR #3320, opened 2025-06-20
fix npe when npr grpc call failed

M00nF1sh:
**What type of PR is this?**
bug

N/A

bug fix to avoid segment fault when GPRC call failed(npr will be nil in that case)

```release-note

```


jaydeokar (approved):
lgtm

mengqiy (approved):
LGTM

PR #3323, opened 2025-06-25
Update multus-daemonset-thick.yml

odural:
There is a race condition when the Multus daemonset comes up after a node reboot, which makes the initContainer hang up because of line number 215. It was supposed to be “cp -f,” but the original upstream used “cp” only, which makes it hang if the shim directory already exists. In customer sites, we made a workaround fix to update this using kubectl patch, but it would be better to fix it in the manifest file (since this is not touching the upstream code itself). The Multus daemonset gets killed by OOM occasionally. We think it would be better to increase the memory limits to 200Mi from the current 50Mi (line 171).

**What type of PR is this?**

```release-note

```


jaydeokar (approved):
lgtm

PR #3328, opened 2025-06-30
update the VPC limits files

tzneal:
**What type of PR is this?**
improvement

Updates the limits files for the new c8gn. It also removed a few U* types, but per https://docs.aws.amazon.com/ec2/latest/instancetypes/mo.html

> The U-9tb1, U-12tb1, U-18tb1, and U-24tb1 instance types are no longer available for new instance launches. If your workload requires a high-memory instance, we recommend that you use a U7i instance type instead.

```release-note

```


jaydeokar (approved):
lgtm

jayanthvn (approved):

PR #3339, opened 2025-07-02
Add retry on for netlink when it receives a ErrDumpInterrupted

dcoppa:
**What type of PR is this?**

improvement

Errors like:

"Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox \"b413c572e7bac7a27767123be3da3e031a3ac5b3695efc995c15057996b2c930\": plugin type=\"aws-cni\" name=\"aws-cni\" failed (add): add command: failed to setup network: SetupPodNetwork: failed to setup veth pair: failed to setup veth network: setup NS network: failed while waiting for v6 addresses to be stable: could not list addresses: results may be incomplete or inconsistent"

This PR adds a retry mechanism when netlink fails with error ErrDumpInterrupted.
Adapted from https://github.com/containernetworking/plugins/blob/main/pkg/netlinksafe/netlink.go

Unit tests passed:

```code
PASS
ok  	github.com/aws/amazon-vpc-cni-k8s/pkg/vpc	0.003s	coverage: 100.0% of statements
```

Also tested on my sandbox EKS cluster.

No

No

No

```release-note
Implement a retry mechanism for netlink that, on an IPv6 EKS cluster, should fix (or mitigate) the following error:

Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "e868238d122df610228688f51fee6a187544a45ce43a9d613307b6b08246e65a": plugin type="aws-cni" name="aws-cni" failed (add): add command: failed to setup network: SetupPodNetwork: failed to setup veth pair: failed to setup veth network: setup NS network: failed while waiting for v6 addresses to be stable: could not list addresses: results may be incomplete or inconsistent

No additional actions from users required.
```

### DIFF
--- a/misc/eni-max-pods.txt
+++ b/misc/eni-max-pods.txt
@@ -682,6 +682,8 @@ m8i.large 59
 m8i.metal-48xl 1514
 m8i.metal-96xl 1514
 m8i.xlarge 118
+mac-m4.metal 234
+mac-m4pro.metal 234
 mac1.metal 234
 mac2-m1ultra.metal 234
 mac2-m2.metal 234
--- a/pkg/vpc/vpc_ip_resource_limit.go
+++ b/pkg/vpc/vpc_ip_resource_limit.go
@@ -8482,6 +8482,32 @@ var instanceNetworkingLimits = map[string]InstanceTypeLimits{
 		HypervisorType: "nitro",
 		IsBareMetal:    false,
 	},
+	"mac-m4.metal": {
+		ENILimit:                8,
+		IPv4Limit:               30,
+		DefaultNetworkCardIndex: 0,
+		NetworkCards: []NetworkCard{
+			{
+				MaximumNetworkInterfaces: 8,
+				NetworkCardIndex:         0,
+			},
+		},
+		HypervisorType: "unknown",
+		IsBareMetal:    true,
+	},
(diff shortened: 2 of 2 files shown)


yash97:
Thanks for contributing this change. Can you make sure that netlink wrapper is always used in code base. For this we need to make changes where vishvananda/netlink is directly used in code to use this wrapper. Also have linter rule to avoid any one in future to use that library directly.

pkg/netlinkwrapper/netlink.go
```diff
@@ -103,15 +126,27 @@ func (*netLink) LinkSetUp(link netlink.Link) error {
...
 
 func (*netLink) LinkList() ([]netlink.Link, error) {
-	return netlink.LinkList()
+	var links []netlink.Link
+	var err error
+	err = retryOnErrDumpInterrupted(func() error {
+		links, err = netlink.LinkList()
+		return err
+	})
+	return links, discardErrDumpInterrupted(err)
```
yash97: What is the rationale behind discarding this error type?
If we already retried and the error still occurred, shouldn't the user be informed about it? I understand that we’re logging the error, but the user won’t see that. Also, when this error occurs, are the links stale or simply empty?

PR #3347, opened 2025-07-09
Adding support in CNI for managing multiple network interface card on the instance

jaydeokar:
**What type of PR is this?** 
feature

N/A

Amazon VPC CNI only manages Network Card 0 on all instances (including multicard supported instances). This restricts the bandwidth usage to only network card 0 as pod interfaces are only connected to NIC 0. With this change, CNI now starts to manage all the available network cards. 
A pod can now request access to these network cards via an annotation to make use of all the available bandwidth. A pod which requires this support can do so via annotation. This will create interfaces in the pod namespace equal to the number of network cards available for use on the instance. The pods can then use these interfaces for their egress traffic which has certain BW requirements.

Describing major changes as the change log is significant 

## CNI

Add flow
- CNI looks for annotations on the pod using the capabilities option. If the pod contains annotation `k8s.amazonaws.com/nicConfig: multi-nic-attachment`, CNI will ask for multiple IPs from network cards available on the instance.
- IPAMD now returns a list of IPs (IPAddress field in GRPC response). The number of IPs returned equals the number of interfaces created by CNI inside the Pod.
- Host veth name for multi-nic interfaces is generated by `<pod-namespace>.<pod-name>.<index>`. However we still retain the original naming convention for the first interface of the pod `<pod-namespace>.<pod-name>`
- Container veth name for multi-nic interfaces has a prefix `mNicIf` eg `mNicf1`, `mNicIf2` ...
- IPAMD now returns route table number instead of just the device number. The Route table number is calculated as `(network card index * max ENIs per NIC) + device-number + 1`. Note Route table number 1 (device 0, network card 0) is the main route table for CNI
- CNI now stores the Route table number of the interface in the container interface struct `MAC`, which is used to cleanup pod networking when IPAMD is down (del with prev result)
- CNI runs a loop over the IP address and attaches to the Pod network.

Delete flow
- CNI relies on IPAMD to return the number of IPs and interface to delete
- If IPAMD is down, CNI uses the prev result to delete the Pod and Host Network

## IPAMD

Node Init Flow
1. IPAMD now initializes all the network card on the instance and creates datastore for network cards which it needs to manage
2. CNI by default assigns 1 WARM IP and 1 MIN IP for secondary ENIs on Network card > 0
3. CNI manages all network cards by default except the following cases 
    a. Network card just has an `efa-only` ENI attached with no additional ENIs
4. CNI manages the network cards when 
    a. It is empty i.e no ENIs are attached to it
    b. It has an EFA ENI attached
    c. It has an EFA-only ENI attached to an device number but also has an ENA device attached on the network card
3. For IPv4 and IPv6, it initializes the host network for the primary interface
  a. For IPv6, all traffic will egress out of the ENIs the traffic comes to i.e we don't add any SNAT or Connmark rules for IPv6
3. IPAMD now includes checks for UpdatingCIDRRules when the VPC CIDR or the Security Group associated with the ENIs changes for IPv6 (which already existed for IPv4)
4. On node init, IPAMD initializes all the available datastores and if the datastore is low, it creates interfaces for each network card. It has a new function for IPv6 ENIs where we create 1 ENI for each network card with IPv6 address and then assign an Prefix of size /80. If the creation of ENI fails (due to missing permission), we don't fail but you can see the permisisons error in ipamd logs

Yes, ran all the test suites on a single card instance and ran manual tests on a multicard- instance

No

Upgrades should be fine. Downgrade requires to delete the pods using multi-nic annotation and then downgrade otherwise the Pod IPs/Host Networking setup can leak

No

Yes, customers will now see interfaces attached to NIC > 0 on supported instances

```release-note

```


cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -194,47 +218,69 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
...
+	var vethMetadata []driver.VirtualInterfaceMetadata
+	var podIPs []*current.IPConfig
+	var podInterfaces []*current.Interface
 	var dummyInterface *current.Interface
 
+	for index, ip := range r.IPAddress {
+
+		var hostVethName string
+
+		addr := parseIPAddress(ip)
```
M00nF1sh: nit: this ip structure seems more like an "ipAllocation" instead of "ipAddress". i'd prefer to rename it.
also, rename this function to be something like getIPAddressFromIPAllocation to be specific

pkg/networkutils/network.go
```diff
@@ -1015,21 +1020,27 @@ func GetIPv4Gateway(eniSubnetCIDR *net.IPNet) net.IP {
...
 		return errors.New("setupENINetwork should never be called on the primary ENI")
 	}
-	tableNumber := deviceNumber + 1
-	log.Infof("Setting up network for an ENI with IP address %s, MAC address %s, CIDR %s and route table %d",
-		eniIP, eniMAC, eniSubnetCIDR, tableNumber)
-	link, err := linkByMac(eniMAC, netLink, retryLinkByMacInterval)
+
+	// CNI will never have route table 1 as we don't setup device number 0 on network card 0
+	// So, table number 1 for CNI is same as main table
+	tableNumber := (networkCard * maxENIPerNIC) + deviceNumber + 1
```
M00nF1sh: let's make a dedicated function for this to make this explicit.
calculateCNIRouteTableID(networkCardIndex, deviceNumber)

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -575,13 +617,40 @@ func teardownPodNetworkWithPrevResult(driverClient driver.NetworkAPIs, conf *Net
...
 	}
-	containerIP, err := getContainerIP(prevResult, contVethName)
-	if err != nil {
-		log.Errorf("Failed to get container IP: %v", err)
-		return false
+
+	// Since we are always using dummy interface to store the device number of network card 0 IP
+	// RT ID for NC-0 is also stored in the container interface entry. So we have a path for migration where
+	// getting the device number from dummy interface can be deprecated entirely. This is currently done to keep it backwards compatible
+	routeTableId := deviceNumber + 1
```
M00nF1sh: this shall be `calculateCNIRouteTableID(0, deviceNumber)`, so the logic of this calculation is centralized

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -575,13 +617,40 @@ func teardownPodNetworkWithPrevResult(driverClient driver.NetworkAPIs, conf *Net
...
 		return false
 	}
-	containerIP, err := getContainerIP(prevResult, contVethName)
-	if err != nil {
-		log.Errorf("Failed to get container IP: %v", err)
-		return false
+
+	// Since we are always using dummy interface to store the device number of network card 0 IP
+	// RT ID for NC-0 is also stored in the container interface entry. So we have a path for migration where
+	// getting the device number from dummy interface can be deprecated entirely. This is currently done to keep it backwards compatible
```
M00nF1sh: seems we are still creating "dummy interface" for both old & new(multi-nic) path, however, for multi-nic, you already encoded the deviceNumber in container mac.

However on deletes path, your new logic are still depends on "dummy interface" first(which will exist since you created it) 

i think you should structure the code such that "new logic" no longer depends on dummy interface.

we need a clear migration here

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -575,13 +617,40 @@ func teardownPodNetworkWithPrevResult(driverClient driver.NetworkAPIs, conf *Net
...
+		if err != nil {
+			log.Errorf("container interface name %s does not exist %v", containerInterfaceName, err)
+			continue
+		}
+
+		// If this property is set, that means the container metadata has the route table ID which we can use
+		// If this is not set, it is a pod launched before this change was introduced.
+		// So it is only managing network card 0 at that time and device number + 1 is the route table ID which we calculate from device number
+		if containerInterface.Mac != "" {
+			routeTableId, err = strconv.Atoi(containerInterface.Mac)
```
M00nF1sh, line 645: nit: could it be something like below to make this logic explicit
```
encodeRouteTableIDIntoMac
decodeRouteTableIDFromMac
```

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -45,18 +45,25 @@ const (
 	v6DADTimeout = 10 * time.Second
 )
 
+type VirtualInterfaceMetadata struct {
+	IPAddress         *net.IPNet
+	DeviceNumber      int
```
M00nF1sh, line 50: why we only have DeviceNumber here but not networkCardNumber

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -162,35 +170,58 @@ func (createVethContext *createVethPairContext) run(hostNS ns.NetNS) error {
 	var maskLen int
 	var addr *netlink.Addr
 	var defNet *net.IPNet
+	var family int
 
-	if createVethContext.v4Addr != nil {
-		gw = net.IPv4(169, 254, 1, 1)
+	if createVethContext.ipAddr.IP.To4() != nil {
+		gw = net.IPv4(169, 254, 1, byte(createVethContext.index)+1)
```
M00nF1sh: let's have a helper function for this as well.

bonus point, group those helpful function that define our network setup for given card/index in a central place.

pkg/awsutils/awsutils.go
```diff
@@ -529,19 +541,19 @@ func (cache *EC2InstanceMetadataCache) RefreshSGIDs(mac string, store *datastore
...
+				eniIDs = append(eniIDs, eniID)
+			}
 		}
 
 		newENIs := StringSet{}
 		newENIs.Set(eniIDs)
-
-		tempfilteredENIs := newENIs.Difference(&cache.multiCardENIs)
-		filteredENIs := tempfilteredENIs.Difference(&cache.unmanagedENIs)
+		// JD: REVISIT HERE
```
yash97: lets not leave this comment here.

cmd/routed-eni-cni-plugin/cni.go
jaydeokar: Just realized, we will have to keep the dummy interface logic for downgrade scenarios. I'll add the path to migration

pkg/awsutils/imds.go
```diff
@@ -463,6 +463,30 @@ func (typedimds TypedIMDS) GetLocalIPv4s(ctx context.Context, mac string) ([]net
 		imdsErr := new(imdsRequestError)
 		oe := new(smithy.OperationError)
 		if errors.As(err, &imdsErr) || errors.As(err, &oe) {
+			if IsNotFound(err) {
+				// No IPv4 address on the interface, not an error
+				return nil, nil
+			}
+			log.Warnf("%v", err)
```
M00nF1sh: needs clear log message or remove it

pkg/awsutils/imds.go
```diff
@@ -510,6 +534,25 @@ func (typedimds TypedIMDS) GetIPv6Prefixes(ctx context.Context, mac string) ([]n
...
+	ips, err := typedimds.getIP(ctx, "ipv6")
+	if err != nil {
+		imdsErr := new(imdsRequestError)
+		oe := new(smithy.OperationError)
+		if errors.As(err, &imdsErr) || errors.As(err, &oe) {
+			if IsNotFound(err) {
+				// No IPv6.  Not an error, just a disappointment :(
+				return nil, nil
+			}
+			log.Warnf("%v", err)
```
M00nF1sh: needs clear log message or remove it

pkg/awsutils/imds.go
```diff
@@ -510,6 +534,25 @@ func (typedimds TypedIMDS) GetIPv6Prefixes(ctx context.Context, mac string) ([]n
 	return prefixes, err
 }
 
+// GetLocalIPv6 returns the IPv6 addresses associated with the primary interface.
+func (typedimds TypedIMDS) GetLocalIPv6(ctx context.Context) (net.IP, error) {
```
M00nF1sh, line 538: where is this function called?

pkg/ipamd/datastore/data_store.go
```diff
@@ -1534,3 +1551,49 @@ func (ds *DataStore) DeleteFromContainerRule(entry *CheckpointEntry) {
...
+	}
+
+	return &DataStoreAccess{
+		DataStores: dsList,
+	}
+}
+
+func (ds *DataStoreAccess) GetDataStore(networkCard int) *DataStore {
+
+	for index, datastore := range ds.DataStores {
```
M00nF1sh, line 1583: why not store a mapping of index -> ds?

cmd/routed-eni-cni-plugin/driver/driver.go
jaydeokar: We will eventually remove this field once we migrate to v1.20+ 
This is currently added to support downgrade scenarios

pkg/awsutils/imds.go
jaydeokar: This is called in the init flow to get the IPv6 primary address. We were previously getting ipv4 address when using ipv6 which is not really needed

cmd/routed-eni-cni-plugin/driver/driver.go
jaydeokar: Updated this

cmd/routed-eni-cni-plugin/cni.go
jaydeokar: Added the details about how we can remove this dummy interface. We have to keep it for now for downgrade path

cmd/routed-eni-cni-plugin/cni.go
jaydeokar: done

cmd/routed-eni-cni-plugin/cni.go
jaydeokar: Updated to IPAllocationMetadata.

cmd/routed-eni-cni-plugin/cni.go
jaydeokar: Will take these as follow up

M00nF1sh:
/lgtm
/approve
as discussed offline, this code is currently in prod usage by Cx, we decided to fast-forward with this change and take improvements as follow up

M00nF1sh (approved):

pkg/networkutils/network.go
```diff
@@ -429,19 +435,16 @@ func (n *linuxNetwork) updateHostIptablesRules(vpcCIDRs []string, primaryMAC str
 
 	ipProtocol := iptables.ProtocolIPv4
 	if v6Enabled {
-		// Essentially a stub function for now in V6 mode. We will need it when we support v6 in secondary IP and
-		// custom networking modes. We don't need to install any SNAT rules in v6 mode and currently there is no need
-		// to mark packets entering via Primary ENI as all the pods in v6 mode will be behind primary ENI. Will have to
-		// start doing that once we start supporting custom networking mode in v6.
-		ipProtocol = iptables.ProtocolIPv6
+		// For v6, we don't add any SNAT or Connmark rules as traffic enters and exits from the ENI it came from
+		return nil
```
dshehbaj, line 439: Did we decide not have the external VPC traffic from pods behind secondary ENI go out via primary ENI's IP in IPv6 mode?

jaydeokar: Yes, that was removed

PR #3348, opened 2025-07-11
assigning Mac addr to veth pair to bypass udev interference

yash97:
**What type of PR is this?**

bug

This will help resolve dependency on Mac address policy dependency for cni. It will not break pod networking. Ref code in udev where it will not assign Mac if user has provided mac address. https://github.com/systemd/systemd/blob/v257.7/src/udev/net/link-config.c#L609. 

```release-note

```

### DIFF
--- a/cmd/routed-eni-cni-plugin/driver/driver.go
+++ b/cmd/routed-eni-cni-plugin/driver/driver.go
@@ -15,6 +15,7 @@
 package driver
 
 import (
+	"crypto/rand"
 	"fmt"
 	"net"
 	"os"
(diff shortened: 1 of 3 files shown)


cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -646,3 +659,44 @@ func buildVlanLink(vlanName string, vlanID int, parentIfIndex int, eniMAC string
...
+func NewMACGenerator() MACGenerator {
+	return MACGenerator{netlink: netlinkwrapper.NewNetLink(), randMACfn: generateRandomMAC}
+}
+
+func generateRandomMAC() string {
+	mac := make([]byte, 6)
+	rand.Read(mac)
+	// Set the local bit and unset the multicast bit
+	mac[0] = (mac[0] | 2) & 0xfe
+
```
yash97, line 677: for this ref https://datatracker.ietf.org/doc/html/rfc7042#section-2.1

PR #3354, opened 2025-07-16
bypassing udev interference by assigning mac address to host side veth

### LINKED ISSUE #2103: Pods cannot talk to cluster IPs on Ubuntu 2204

yash97:
**What type of PR is this?**

bug

Fixes #2103 

This PR assigns mac address to host side veth when creating veth pair for container. 
This will help resolve dependency on Mac address policy dependency for cni. It will not break pod networking. Ref code in udev where it will not assign Mac if user has provided mac address. https://github.com/systemd/systemd/blob/v257.7/src/udev/net/link-config.c#L609.

- Ran Integration test and unit test.
- Tested pod egress connectivity in standard, strict, and vanilla mode when Mac address policy is persistent.

Added New integration test  which checks the egress connectivity for pod with Mac address policy to `None` and to `Persistent` have added as `skip` block. Will remove it after release. 

```release-note

```


cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -677,3 +690,44 @@ func buildVlanLink(vlanName string, vlanID int, parentIfIndex int, eniMAC string
...
+		return "", err
+	}
+	macMap := make(map[string]struct{})
+	for _, link := range ll {
+		if link.Attrs() != nil {
+			macMap[link.Attrs().HardwareAddr.String()] = struct{}{}
+		}
+	}
+
+	for i := 0; i < 10; i++ {
```
haouc: How large is the possibility that we need try > 10 to get a not-existing mac? My guess is very close to 0 but want to confirm.

yash97: there are total 48 bits in mac address. We keep local and unicast bit same for all. so 46 bits available for randomness. so there 2<sup>46</sup> possible mac address. If there are 10 links present on node probability should be 10 out of 2<sup>46</sup> possibility. We try to 10 times if it is still same we fail and retry. We can go into math of probability of collision after 10 tries, too but chance of collision will be close to 0. also we just retry. Our scope of uniqueness is limited to host level.

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -677,3 +690,44 @@ func buildVlanLink(vlanName string, vlanID int, parentIfIndex int, eniMAC string
...
+}
+
+func generateRandomMAC() string {
+	mac := make([]byte, 6)
+	rand.Read(mac)
+	// Set the local bit and unset the multicast bit
+	mac[0] = (mac[0] | 2) & 0xfe
+
+	return fmt.Sprintf("%02x:%02x:%02x:%02x:%02x:%02x",
+		mac[0], mac[1], mac[2], mac[3], mac[4], mac[5])
```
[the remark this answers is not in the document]
yash97: Rand.Read always return nil error.

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -387,8 +394,15 @@ func (n *linuxNetwork) setupVeth(hostVethName string, contVethName string, netns
 		}
 		log.Debugf("Successfully deleted old hostVeth %s", hostVethName)
 	}
-
-	createVethContext := newCreateVethPairContext(contVethName, hostVethName, ipAddr, mtu, index)
+	macAddrStr, err := NewMACGenerator().generateUniqueRandomMAC()
```
M00nF1sh, line 397: nit: I'd like to let generateUniqueRandomMAC return a net.HardwareAddress directly, so it frees the consumer code to parse and handle error.

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -677,3 +691,44 @@ func buildVlanLink(vlanName string, vlanID int, parentIfIndex int, eniMAC string
...
+	// Set the local bit and unset the multicast bit
+	mac[0] = (mac[0] | 2) & 0xfe
+
+	return fmt.Sprintf("%02x:%02x:%02x:%02x:%02x:%02x",
+		mac[0], mac[1], mac[2], mac[3], mac[4], mac[5])
+}
+
+// generateUniqueRandomMAC will compare randomly generated Mac to mac addresses of veth already present in host.
+func (m MACGenerator) generateUniqueRandomMAC() (string, error) {
+	ll, err := m.netlink.LinkList()
```
jaydeokar, line 716: We can't cache this ? I'm thinking about latency since we will be making a list call with every pod creation.

test/integration/cni/pod_traffic_test.go
```diff
@@ -265,6 +274,134 @@ var _ = Describe("test pod networking", func() {
...
+		Context("When MAC address Policy is None", func() {
+			// check if current policy is none, if not make it none
+			It("can ping to 8.8.8.8", func() {
+				Skip("Skipping pod egress traffic until addon release")
+				Expect(setMACAddressPolicy(primaryNode.Name, None)).Error().ShouldNot(HaveOccurred())
+				// deploy pod on this node.
+				pod = manifest.NewDefaultPodBuilder().
+					NodeName(primaryNode.Name).
+					Container(manifest.NewBusyBoxContainerBuilder(f.Options.TestImageRegistry).Build()).
+					Namespace("default").
```
jaydeokar: nit - You should change this to run in the test namespace which we create

jaydeokar (approved):

test/integration/cni/pod_traffic_test.go
```diff
@@ -265,6 +274,132 @@ var _ = Describe("test pod networking", func() {
...
+					NodeName(primaryNode.Name).
+					Container(manifest.NewBusyBoxContainerBuilder(f.Options.TestImageRegistry).Build()).
+					Build()
+				pod, err = f.K8sResourceManagers.PodManager().CreateAndWaitTillRunning(pod)
+				Expect(err).ToNot(HaveOccurred())
+
+				command := []string{"ping", "-c", "1", "8.8.8.8"}
+				stdout, stderr, err := f.K8sResourceManagers.PodManager().PodExec("default", pod.Name, command)
+				Expect(err).ToNot(HaveOccurred(), stderr, stdout)
+				Expect(stderr).To(BeEmpty(), stdout)
```
jaydeokar, line 330: nit - Change this to string matching instead of checking empty and not empty as a followup

jaydeokar (approved):

cmd/routed-eni-cni-plugin/driver/driver.go
yash97: had discussion internally. Can be improved if there are issues. Limitation with caching is we cannot track all link device on node.

PR #3357, opened 2025-07-17
bumping kops tag to v1.33.0-beta.1

yash97:
**What type of PR is this?**

dependency

Issue https://github.com/kubernetes/kops/issues/17267. is fixed in  v1.33.0-beta.1

```release-note

```


viveksb007 (dismissed):

.github/workflows/kops-test.yaml
```diff
@@ -43,8 +43,8 @@ jobs:
           ROLE_ARN: ${{ secrets.EKS_CLUSTER_ROLE_ARN }}
           RUN_CNI_INTEGRATION_TESTS: false
           RUN_KOPS_TEST: true
-          K8S_VERSION: v1.33.0-beta.0
-          KOPS_VERSION: v1.32.1
+          K8S_VERSION: 1.33.0-beta.0
```
viveksb007, line 46: is k8s version change intended?

yash97: yes. here in integration script we download by adding `v` if we do not remove it, test fails. https://github.com/aws/amazon-vpc-cni-k8s/blob/master/scripts/run-integration-tests.sh#L263

PR #3378, opened 2025-07-30
Updating static config for instances and better error handling

jaydeokar:
**What type of PR is this?**
fix 

Missing static config for instances which are no longer publicly available

We have a fallback to describe the instance in case we don't have the data, but that fails here

N/A

**Does this PR introduce any user-facing change?**: N/A

```release-note

```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -1723,7 +1723,9 @@ func (cache *EC2InstanceMetadataCache) GetENILimit() int {
 func (cache *EC2InstanceMetadataCache) GetNetworkCards() []vpc.NetworkCard {
 	networkCards, err := vpc.GetNetworkCards(cache.instanceType)
 	if err != nil {
-		return nil
+		// fallback to default for network card index 0 as all instances have at least one network card
+		// this needs be changed when an instance can have multiple network cards each with different maxENI limits
+		return []vpc.NetworkCard{{NetworkCardIndex: 0}}
 	}
 	return networkCards
 }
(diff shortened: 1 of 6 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -1723,7 +1723,9 @@ func (cache *EC2InstanceMetadataCache) GetENILimit() int {
 func (cache *EC2InstanceMetadataCache) GetNetworkCards() []vpc.NetworkCard {
 	networkCards, err := vpc.GetNetworkCards(cache.instanceType)
 	if err != nil {
-		return nil
+		// fallback to default for network card index 0 as all instances have at least one network card
+		// this needs be changed when an instance can have multiple network cards each with different maxENI limits
+		return []vpc.NetworkCard{{NetworkCardIndex: 0}}
```
haouc, line 1728: How do we handle this default case when it happens? This looks like a dummy card to me. Are we falling back to somewhere to pull cards' metadata, or fail?

jaydeokar: This is only so that we don't run into issues where Network Card information is missing and to not run into a panic
Most nodes will be using a single network card by default

haouc: Does this mean we could still face an inconsistency for some types when we miss data and default to 1 card?

haouc: Synced with Jay offline and looks good to me to merge this change and following up with other improvements.

haouc (approved):

haouc (approved):
lgtm

pkg/vpc/vpc_test.go
```diff
@@ -114,17 +114,13 @@ func TestGetNetworkCards(t *testing.T) {
 		},
 	})
 	assert.NoError(t, err)
-
-	networkCards, err = GetNetworkCards("c5a.metal")
```
jaydeokar, line 118: This case is no longer valid

Issue #3067, opened 2024-10-11
Subnet discovery enhancements to skip node primary ENI and support secondary ENI security groups


mikestef9:
**What would you like to be added**:

Enhanced subnet discovery (launch [blog](https://aws.amazon.com/blogs/containers/amazon-vpc-cni-introduces-enhanced-subnet-discovery/)) provides an improved UX compared to "custom networking", but doesn't yet support all of the capabilities of custom networking, notably, ability to run pods in separate subnets from nodes, and attach separate security group(s) to the pods running on secondary ENIs in alternate subnets. 

*To run pods only in secondary subnets and not primary node subnet:*
Support tag key/value of `kubernetes.io/role/cni=0` which would instruct VPC CNI to skip using that subnet for pods. Users could tag node subnets with this key/value pair. In this case, the value of the tag key (1 vs 0) now matters with enhanced subnet discovery. 

*To apply alternate security groups to pods running in secondary subnets:*
Support tag key/value of `kubernetes.io/role/cni=1` on security groups in the VPC. The VPC CNI would discover these security groups at startup, and these would only be applied to ENIs launched in secondary subnets discovered using the subnet discovery feature.

**Why is this needed**:

To support the use case of pods and nodes running in separate subnets that is possible with custom networking, but with the improved UX of subnet discovery.

orsenthil:
We can add this in v1.18.8 release.

paramanandd:
Its really important to have this frature.

- Currently we have limited IPS i vpc around 1500-2000 and this isnt a good way to add new CIDR's everytime we start reaching exhaustion of IP's, and it also need a lot of process and approval to include new CIDR, because that include lots of network changes.
- The Custom Networking approach if implemented as stated above will be better option to proceed further. 
- kubernetes.io/role/cni=0 tag on subnets for routable and nodes in a cluster
- kubernetes.io/role/cni=1 tag on subnets for the pods in a cluster and this subnets can be non-routable and present withing every vpc.
- kubernetes.io/role/cni=1 & additional tag to determine which cluster pods or kubernetes.io/role/cni={cluster-name}  tag on  security groups to be assigned to pods based from specific clusters
-  how will be the scenario when custom networking is implemented and we introduced new security group, will that update the sg for pod inplace or instance refresh will be done by aws karpenter for such changes.

mikestef9:
Thanks for the feedback

[...]

Issue #2904, opened 2024-05-02
Enhanced subnet discovery should use configurable tags


dougbyrne:
**What would you like to be added**: The enhanced subnet discovery should use configurable tags. Currently the CNI will use any subnet in a matching VPC and AZ with the tag `kubernetes.io/role/cni: 1`. The CNI should have additional configurable tag filters.

**Why is this needed**: If multiple EKS clusters share a VPC, they will all share subnets. It is desirable to specify additional tags unique to each cluster to limit the discovery scope. An example filter might be `kubernetes.io/cluster/my-example-cluster: shared`.

| Subnet ID  | tag `kubernetes.io/role/cni` | tag `kubernetes.io/cluster/my-example-cluster` | tag `kubernetes.io/cluster/some-other-cluster` | Selected by `my example cluster` | Selected by `some other cluster` |
|------------|------------------------------|------------------------------------------------|------------------------------------------------|----------------------------------|----------------------------------|
| subnet-123 | 1                            | shared                                         | _untagged_                                     | :white_check_mark: Yes           | :x: No                           |
| subnet-456 | 1                            | _untagged_                                     | shared                                         | :x: No                           | :white_check_mark: Yes           |
| subnet-789 | 1                            | shared                                         | shared                                         | :white_check_mark: Yes           | :white_check_mark: Yes           |
| subnet-abc | _untagged_                   | shared                                         | _untagged_                                     | :x: No                           | :x: No                           |
| subnet-def | _untagged_                   | _untagged_                                     | _untagged_                                     | :x: No                           | :x: No                           |

dougbyrne:
Unstale

dougbyrne:
Unstale

lorengordon:
Unstale

dougbyrne:
Unstale

santinsinga:
It is becoming more common to share VPC among clusters and we cannot use this feature.

SandeepCitigori:
Enhanced subnet discovery is a good feature, but we could not use it due to multiple clusters in the same VPC.

santinsinga:
unstale

santinsinga:
Use case: having different environments (test, dev, pre) in the same non production account

lorengordon:
Unstale

dougbyrne:
Unstale

dougbyrne:
Still valid

haouc:
[...]

PR #3380, opened 2025-07-30
Enhanced Subnet Discovery

dshehbaj:
**What type of PR is this?**
feature

https://github.com/aws/amazon-vpc-cni-k8s/issues/3067
https://github.com/aws/amazon-vpc-cni-k8s/issues/2904

This PR adds a requested customer feature to allow customers to 
1. Include or exclude subnets/security groups by tagging them `kubernetes.io/role/cni=1` or `kubernetes.io/role/cni=0` respectively.
2. Apply custom security groups for secondary ENIs created using `kubernetes.io/role/cni=1` tagged subnet if the alternate security group is also tagged `kubernetes.io/role/cni=1`.
3. Reserve subnets for specific cluster based on the `kubernetes.io/cluster/<my-example-cluster>` tag. See https://github.com/aws/amazon-vpc-cni-k8s/issues/2904 more.

1. Added unit tests and integration tests.
2. Manual testing pending.

- Manual testing pending.

- Requires `ENABLE_SUBNET_DISCOVERY` enabled which is already done by default.

- Users were already tagging their subnets as the part of the original subnet discovery feature. This enhanced feature will also allow users to tag alternate security groups.

- Make sure that IPv6 documentation and managed IPv4 policy are updated.

```release-note

```

Integration tests passing in IPv4 cluster.

image

 Currently adapting our test framework to work with IPv6 cluster. Will paste results here.


pkg/ipamd/ipamd.go
```diff
@@ -1252,7 +1335,16 @@ func (c *IPAMContext) setupENI(eni string, eniMetadata awsutils.ENIMetadata, isT
 		return errors.Wrapf(err, "failed to add ENI %s to data store", eni)
 	}
 
-	// Store the addressable IP for the ENI.
+	// Mark primary ENI as excluded if primary subnet is excluded
+	if eni == primaryENI && c.isPrimarySubnetExcluded {
+		log.Infof("Marking primary ENI %s as excluded from pod IP allocation", eni)
+		err := c.dataStoreAccess.GetDataStore(eniMetadata.NetworkCard).SetENIExcludedForPodIPs(eni, true)
```
dshehbaj: Check if `eniMetadata.NetworkCard` is the same network card as 0.

pkg/awsutils/awsutils.go
```diff
@@ -1061,12 +1271,75 @@ func (cache *EC2InstanceMetadataCache) getVpcSubnets() ([]ec2types.Subnet, error
 	return subnetResult.Subnets, nil
 }
 
-func validTag(subnet ec2types.Subnet) bool {
+// validTag checks if subnet should be used for ENI/IP allocation
+// For primary subnet: include by default (no tag), exclude only if tag value is "0"
+// For secondary subnets: exclude by default (no tag), include only if tag exists with non-"0" value
+// If the subnet has cluster-specific tags, it will only be used by the matching cluster
+func validTag(subnet ec2types.Subnet, isPrimarySubnet bool) bool {
```
dshehbaj, line 1278: Refactor this function and make it simpler

pkg/ipamd/ipamd.go
```diff
@@ -571,10 +642,22 @@ func (c *IPAMContext) nodeInit() error {
...
+		}
+
 		// Refresh security groups and VPC CIDR blocks in the background
 		// Ignoring errors since we will retry in 30s
 		go wait.Forever(func() {
 			c.awsClient.RefreshSGIDs(primaryENIMac, c.dataStoreAccess)
+			// Also refresh custom security groups for secondary subnets
+			if c.useSubnetDiscovery {
+				c.awsClient.RefreshCustomSGIDs(c.dataStoreAccess.GetDataStore(DefaultNetworkCardIndex))
+			}
```
dshehbaj, line 660: Update Refresh Custom SGIDs to handle all network cards

pkg/ipamd/ipamd.go
```diff
@@ -454,12 +452,60 @@ func New(k8sClient client.Client, withApiServer bool) (*IPAMContext, error) {
...
 
+// primaryENIWasPreviouslyUsed checks if the primary ENI has any assigned pod IPs
+// This is used to determine if we should re-include an excluded primary ENI for backward compatibility
+func (c *IPAMContext) primaryENIWasPreviouslyUsed() bool {
+	primaryENI := c.awsClient.GetPrimaryENI()
+	eniInfos := c.dataStoreAccess.GetDataStore(DefaultNetworkCardIndex).GetENIInfos()
+
+	if primaryENIInfo, exists := eniInfos.ENIs[primaryENI]; exists {
+		// Check if CNI has any IPs currently assigned to pods
+		assignedIPs := primaryENIInfo.AssignedIPv4Addresses()
```
dshehbaj, line 463: This is only checking IPv4, make sure it also checks IPv6

pkg/ipamd/ipamd.go
```diff
@@ -454,12 +452,60 @@ func New(k8sClient client.Client, withApiServer bool) (*IPAMContext, error) {
...
+
+		if assignedIPs > 0 {
+			log.Infof("Primary ENI %s has %d assigned IPv4 addresses, re-including for backward compatibility despite subnet exclusion",
+				primaryENI, assignedIPs)
+			return true
+		} else {
+			log.Infof("Primary ENI %s has no assigned IPv4 addresses, respecting primary subnet exclusion tag", primaryENI)
+
+			// Check if there are unassigned secondary IPs and log for customer awareness
+			totalSecondaryIPs := len(primaryENIInfo.AvailableIPv4Cidrs)
```
dshehbaj, line 473: Also do IPv4

pkg/awsutils/awsutils.go
```diff
@@ -966,7 +1149,7 @@ func (cache *EC2InstanceMetadataCache) createENI(sg []*string, eniCfgSubnet stri
 	eniDescription := eniDescriptionPrefix + cache.instanceID
 	tags := cache.createENITags()
 
-	var needIPs = numIPs
+	needIPs := numIPs
 
 	if !cache.v6Enabled {
```
dshehbaj, line 1154: Enable it for ipv6

pkg/awsutils/awsutils_test.go
```diff
@@ -2726,6 +2726,199 @@ func TestAllocENIWithSubnetExclusion(t *testing.T) {
...
+					},
+				},
+			},
+			useSubnetDiscovery: true,
+			v6Enabled:          true,
+			expectError:        false,
+		},
+		{
+			name: "IPv6 enabled with subnet discovery and mixed IPv4/IPv6 subnets",
+			subnets: []ec2types.Subnet{
```
dshehbaj, line 2809: Need to check edge case in code if customer is on IPv4 cluster and the code picks up an IPv6 subnet because that's all they had tagged in the console and vice versa.

pkg/awsutils/awsutils.go
```diff
@@ -234,6 +245,7 @@ type EC2InstanceMetadataCache struct {
 	useCustomNetworking      bool
 	unmanagedNICs            []bool
 	efaOnlyENIsByNetworkCard []string
+	multiCardENIs            StringSet
```
dshehbaj: left over from rebase, remove this

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +525,248 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+			}
+
+			eniIDs = append(eniIDs, eniID)
+		}
+	}
+
+	// Apply standard filters (unmanaged and multi-card ENIs)
+	newENIs := StringSet{}
+	newENIs.Set(eniIDs)
+	tempfilteredENIs := newENIs.Difference(&cache.multiCardENIs)
```
dshehbaj: left over from rebase, remove

pkg/awsutils/awsutils.go
```diff
@@ -992,33 +1175,79 @@ func (cache *EC2InstanceMetadataCache) createENI(sg []*string, eniCfgSubnet stri
...
+			excluded, checkErr := cache.isPrimarySubnetExcluded()
+			if checkErr != nil {
+				// If we can't determine exclusion status, log warning and proceed
+				log.Warnf("Failed to check if primary subnet is excluded: %v. Proceeding with ENI creation attempt.", checkErr)
+			} else if excluded {
+				// Primary subnet is explicitly excluded
+				err = errors.New("primary subnet is tagged with kubernetes.io/role/cni=0 and subnet discovery is disabled - no valid subnets available for ENI creation")
+				log.Error(err.Error())
+				return "", err
+			}
```
dshehbaj: This checks if Cx has a tagged subnet, but have not enable subnet discovery. Do we want to do this check?

pkg/ipamd/ipamd.go
```diff
@@ -454,12 +452,73 @@ func New(k8sClient client.Client, withApiServer bool) (*IPAMContext, error) {
...
+		// Check if CNI has any IPs currently assigned to pods
+		var assignedIPs int
+		var ipType string
+		if c.enableIPv6 {
+			assignedIPs = primaryENIInfo.AssignedIPv6Addresses()
+			ipType = "IPv6"
+		} else {
+			assignedIPs = primaryENIInfo.AssignedIPv4Addresses()
+			ipType = "IPv4"
+		}
```
dshehbaj: See if we can use `hasPods()` function.

Currently `hasPods()` only checks for IPv4 assigned addresses, will need to update that func to support IPv6.

pkg/ipamd/datastore/data_store.go
```diff
@@ -991,6 +1063,11 @@ func (ds *DataStore) GetAllocatableENIs(maxIPperENI int, skipPrimary bool) []*EN
 			ds.log.Debugf("Skip needs IP check for trunk ENI of primary ENI when Custom Networking is enabled")
 			continue
 		}
+		// Skip ENIs that are excluded from pod IP allocation
+		if eni.IsExcludedForPodIPs {
+			ds.log.Debugf("Skip needs IP check for ENI %s as it is excluded from pod IP allocation", eni.ID)
+			continue
+		}
```
dshehbaj, line 1119: This will skip the primary ENI in `GetAllocatableENIs` function and prevent it from getting IPs assigned/allocated

pkg/ipamd/ipamd_test.go
```diff
@@ -3129,3 +3143,87 @@ func TestPrimaryENIWasPreviouslyUsedIPv6(t *testing.T) {
...
+	// Expect both IP and prefix cleanup
+	m.awsutils.EXPECT().DeallocIPAddresses(primaryENIid, gomock.Any()).Return(nil).AnyTimes()
+	m.awsutils.EXPECT().DeallocPrefixAddresses(primaryENIid, []string{"10.0.0.0/28"}).Return(nil)
+	m.awsutils.EXPECT().GetPrimaryENI().Return(primaryENIid)
+
+	wasUsed := mockContext.primaryENIWasPreviouslyUsed()
+	assert.False(t, wasUsed, "Primary ENI should be excluded and auto-cleanup prefixes")
+}
+
+func TestPrimaryENIAutoCleanupFailureHandling(t *testing.T) {
```
dshehbaj: Add case for IPv6 mode too

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
 
-// RefreshSGIDs retrieves security groups
-func (cache *EC2InstanceMetadataCache) RefreshSGIDs(mac string, dsAccess *datastore.DataStoreAccess) error {
-	ctx := context.TODO()
+// discoverCustomSecurityGroups discovers security groups with the cni-role tag
+func (cache *EC2InstanceMetadataCache) discoverCustomSecurityGroups() ([]string, error) {
+	describeSGInput := &ec2.DescribeSecurityGroupsInput{
+		Filters: []ec2types.Filter{
+			{
+				Name:   aws.String("vpc-id"),
```
haouc, line 540: We need support global SGs which could be in another VPCs.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+				Values: []string{cache.vpcID},
+			},
+			{
+				Name:   aws.String("tag:" + subnetDiscoveryTagKey),
+				Values: []string{"1"},
```
haouc: nit: let's make them to const, such as selected = 1.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+			},
+		},
+	}
 
-	sgIDs, err := cache.imds.GetSecurityGroupIDs(ctx, mac)
+	result, err := cache.ec2SVC.DescribeSecurityGroups(context.Background(), describeSGInput)
```
haouc: Why don't use `ctx` here? Also we should consider using paginated API.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
 	if err != nil {
-		awsAPIErrInc("GetSecurityGroupIDs", err)
-		return err
+		return nil, errors.Wrap(err, "discoverCustomSecurityGroups: unable to describe security groups")
```
haouc: Let's use fmt.Errorf("unable to describe security groups: %w", err)

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+	}
+
+	var sgIDs []string
+	for _, sg := range result.SecurityGroups {
```
haouc, line 556: You can use `lo.FromSlicePtr(result.SecurityGroups)`

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+	return sgIDs, nil
+}
+
+// getENISubnetID gets the subnet ID for an ENI from AWS
+func (cache *EC2InstanceMetadataCache) getENISubnetID(eniID string) (string, error) {
+	describeInput := &ec2.DescribeNetworkInterfacesInput{
+		NetworkInterfaceIds: []string{eniID},
+	}
+
+	result, err := cache.ec2SVC.DescribeNetworkInterfaces(context.Background(), describeInput)
```
haouc: I think a shared context object should be created for those calls which will be easier for tracking and managing.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+	if err != nil {
+		return "", errors.Wrap(err, "getENISubnetID: unable to describe network interface")
```
haouc: Same here, let's using `fmt.Errorf()`

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+	}
+
+	if len(result.NetworkInterfaces) == 0 {
+		return "", errors.New("getENISubnetID: no interfaces found")
```
haouc: Same here, use fmt.Errorf()

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+	}
+
+	return *result.NetworkInterfaces[0].SubnetId, nil
```
haouc, line 578: You can use `lo.FromPtr()` which handles nil point issue and avoid panic.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+}
+
+// Helper function to check if an ENI is in a secondary subnet
+func (cache *EC2InstanceMetadataCache) isENIInSecondarySubnet(eniID string) bool {
+	eniDetails, err := cache.getENISubnetID(eniID)
```
haouc: I think `eniDetails` is a string of subnet ID? The name is misleading.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+	return err == nil && eniDetails != cache.subnetID
```
haouc: I assume `eniDetails` is guaranteed not being empty?

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+			eniIDs = append(eniIDs, eniID)
+		}
+	}
+
+	// Apply standard filters (unmanaged and multi-card ENIs)
+	newENIs := StringSet{}
+	newENIs.Set(eniIDs)
+	filteredENIs := newENIs.Difference(&cache.unmanagedENIs)
+
+	return filteredENIs.SortedList()
```
haouc, line 618: I am not sure why ENIs need to be sorted and based on what sorting key?

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+}
+
+// Helper function to apply security groups to a list of ENIs
+func (cache *EC2InstanceMetadataCache) applySecurityGroupsToENIs(eniIDs []string, sgIDs []string, logPrefix string) {
+	for _, eniID := range eniIDs {
+		log.Debugf("%s ENI %s with security groups %v", logPrefix, eniID, sgIDs)
+
+		attributeInput := &ec2.ModifyNetworkInterfaceAttributeInput{
```
haouc, line 626: Probably referred from somewhere else, but I don't know why ENIs need to be modified with other SGs? I am not aware that the feature allows SGs update after ENIs are created.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+			NetworkInterfaceId: aws.String(eniID),
+		}
+		start := time.Now()
+		_, err := cache.ec2SVC.ModifyNetworkInterfaceAttribute(context.Background(), attributeInput)
+		prometheusmetrics.Ec2ApiReq.WithLabelValues("ModifyNetworkInterfaceAttribute").Inc()
+		prometheusmetrics.AwsAPILatency.WithLabelValues("ModifyNetworkInterfaceAttribute", fmt.Sprint(err != nil), awsReqStatus(err)).Observe(msSince(start))
+
+		if err != nil {
+			if errors.As(err, &awsAPIError) {
+				if awsAPIError.ErrorCode() == "InvalidNetworkInterfaceID.NotFound" {
```
haouc, line 637: Can you check if errors type check have been handled in an existing file?

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
 	}
-	cache.securityGroups.Set(sgIDs)
 
-	if !cache.useCustomNetworking && (addedSGsCount != 0 || deletedSGsCount != 0) {
-		var eniIDs []string
+	return addedCount, deletedCount
+}
+
+// RefreshCustomSGIDs discovers and refreshes security groups tagged for use with the CNI
+func (cache *EC2InstanceMetadataCache) RefreshCustomSGIDs(dsAccess *datastore.DataStoreAccess) error {
```
haouc: Is this func called somewhere? I don't think I saw it was used.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
-		newENIs.Set(eniIDs)
-		filteredENIs := newENIs.Difference(&cache.unmanagedENIs)
+// applyFallbackSecurityGroupsForAllDatastores applies primary security groups to ENIs in secondary subnets across all datastores
+func (cache *EC2InstanceMetadataCache) applyFallbackSecurityGroupsForAllDatastores(dsAccess *datastore.DataStoreAccess) error {
+	log.Info("Applying primary security groups as fallback for ENIs in secondary subnets across all datastores")
 
-		// This will update SG for managed ENIs created by EKS.
-		for _, eniID := range filteredENIs.SortedList() {
-			log.Debugf("Update ENI %s", eniID)
+	primarySGs := cache.securityGroups.SortedList()
```
haouc, line 719: Why do we need to sort them?

pkg/awsutils/awsutils.go
```diff
@@ -1061,15 +1288,79 @@ func (cache *EC2InstanceMetadataCache) getVpcSubnets() ([]ec2types.Subnet, error
...
-	for _, tag := range subnet.Tags {
-		if *tag.Key == subnetDiscoveryTagKey {
-			return true
+// validTag checks if subnet should be used for ENI/IP allocation
+// For primary subnet: include by default (no tag), exclude only if tag value is "0"
+// For secondary subnets: exclude by default (no tag), include only if tag exists with non-"0" value
+// If the subnet has cluster-specific tags, it will only be used by the matching cluster
+func validTag(subnet ec2types.Subnet, isPrimarySubnet bool) bool {
+	// Get cluster name for cluster-specific tag checks
+	localClusterName := os.Getenv(clusterNameEnvVar)
```
haouc, line 1298: I don't think this is guaranteed. Will we have a prerequisite with this feature support?

pkg/awsutils/awsutils.go
```diff
@@ -1061,15 +1288,79 @@ func (cache *EC2InstanceMetadataCache) getVpcSubnets() ([]ec2types.Subnet, error
...
+	if !hasClusterTags {
+		// No cluster tags means subnet is available to all clusters
+		log.Debugf("Subnet %s has no cluster-specific tags, making it available for all clusters", *subnet.SubnetId)
+		return true
+	}
+
+	if belongsToThisCluster {
+		// Subnet is tagged for this specific cluster
+		log.Debugf("Subnet %s has matching cluster tag %s, including it for ENI creation", *subnet.SubnetId, localClusterTagKey)
+		return true
```
haouc, line 1326: Looks like this func just need
```
if !hasClusterTags || belingsToThisCluster {
return true
}
return false
```

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
-					if awsAPIError.ErrorCode() == "InvalidNetworkInterfaceID.NotFound" {
-						awsAPIErrInc("IMDSMetaDataOutOfSync", err)
+	cache.applySecurityGroupsToENIs(eniIDs, primarySGs, "Applying primary security groups to")
+
+	return nil
+}
+
+// RefreshSGIDs retrieves security groups
+func (cache *EC2InstanceMetadataCache) RefreshSGIDs(mac string, dsAccess *datastore.DataStoreAccess) error {
+	ctx := context.TODO()
```
haouc: We should generalize the context is used across API calls.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
-				if errors.As(err, &awsAPIError) {
-					if awsAPIError.ErrorCode() == "InvalidNetworkInterfaceID.NotFound" {
-						awsAPIErrInc("IMDSMetaDataOutOfSync", err)
+	cache.applySecurityGroupsToENIs(eniIDs, primarySGs, "Applying primary security groups to")
+
+	return nil
+}
+
+// RefreshSGIDs retrieves security groups
+func (cache *EC2InstanceMetadataCache) RefreshSGIDs(mac string, dsAccess *datastore.DataStoreAccess) error {
```
haouc: What event triggers this func?

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
-				// If update failed due to stale ENI then returning error will prevent updating SG
-				// for following ENIs since the list is sorted
-				log.Debugf("refreshSGIDs: unable to update the ENI %s SG - %v", eniID, err)
 			}
+		} else {
+			// Original behavior: apply to all managed ENIs when subnet discovery is disabled
+			for _, ds := range dsAccess.DataStores {
+				eniInfos := ds.GetENIInfos()
+				for eniID := range eniInfos.ENIs {
+					eniIDs = append(eniIDs, eniID)
```
haouc, line 763: Would `eniIDs := append(eniIDs, ds.GetENIInfos()...)` just work?

pkg/awsutils/awsutils.go
```diff
@@ -992,33 +1173,79 @@ func (cache *EC2InstanceMetadataCache) createENI(sg []*string, eniCfgSubnet stri
...
+					// We already determined isPrimarySubnet above, just reuse the variable
+					if !isPrimarySubnet && len(cache.customSecurityGroups.SortedList()) > 0 {
+						log.Infof("Using custom security groups for ENI in secondary subnet %s", *subnet.SubnetId)
+						input.Groups = cache.customSecurityGroups.SortedList()
+					} else if !isPrimarySubnet {
+						// Secondary subnet but no custom security groups available - use primary SGs as fallback
+						log.Infof("No custom security groups available, using primary security groups for ENI in secondary subnet %s", *subnet.SubnetId)
 					}
-					log.Infof("Creating ENI with security groups: %v in subnet: %s", input.Groups, aws.ToString(input.SubnetId))
+					log.Infof("Creating ENI with security groups: %v in subnet: %s", input.Groups, aws.ToString(subnet.SubnetId))
```
haouc, line 1224: Looks we added many more log lines, can you recheck them to maintain them to minimal level.

pkg/awsutils/awsutils.go
```diff
@@ -992,33 +1173,79 @@ func (cache *EC2InstanceMetadataCache) createENI(sg []*string, eniCfgSubnet stri
...
 					input.SubnetId = subnet.SubnetId
 					networkInterfaceID, err = cache.tryCreateNetworkInterface(input)
 					if err == nil {
 						return networkInterfaceID, nil
 					}
 				}
+
+				// If no valid subnets found, return appropriate error
+				if !validSubnetsFound {
+					err = errors.New("no valid subnets available for ENI creation - all subnets are either not tagged or tagged with kubernetes.io/role/cni=0")
```
haouc: Can you check the similar pattern to use one line return instead of three lines.

pkg/ipamd/datastore/data_store.go
```diff
@@ -477,6 +488,38 @@ func (ds *DataStore) AddENI(eniID string, deviceNumber int, isPrimary, isTrunk,
...
+	defer ds.lock.Unlock()
+
+	eni, ok := ds.eniPool[eniID]
+	if !ok {
+		return errors.New(UnknownENIError)
+	}
+
+	eni.IsExcludedForPodIPs = excluded
+	if excluded {
+		ds.log.Infof("ENI %s marked as excluded from pod IP allocation", eniID)
```
haouc, line 503: Let's avoid logging if possible especially when the code is under lock mechanism.

pkg/ipamd/datastore/data_store.go
```diff
@@ -636,6 +679,12 @@ func (ds *DataStore) AssignPodIPv6Address(ipamKey IPAMKey, ipamMetadata IPAMMeta
 
 	// In IPv6 Prefix Delegation mode, eniPool will only have Primary ENI.
 	for _, eni := range ds.eniPool {
+		// Skip ENIs that are excluded from pod IP allocation
+		if eni.IsExcludedForPodIPs {
+			ds.log.Debugf("Skipping ENI %s as it is excluded from pod IP allocation", eni.ID)
```
haouc, line 733: Unnecessary logging

pkg/awsutils/awsutils.go
dshehbaj: We call this here in IPAMD: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/ipamd.go#L565-L579

SInce I also made a `RefreshCustomSGID`, I refactored this function to share the logic between the two.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+		eniIDs = append(eniIDs, cache.getFilteredENIs(ds, true)...) // only secondary subnet ENIs
+	}
 
-			if err != nil {
-				if errors.As(err, &awsAPIError) {
-					if awsAPIError.ErrorCode() == "InvalidNetworkInterfaceID.NotFound" {
-						awsAPIErrInc("IMDSMetaDataOutOfSync", err)
+	cache.applySecurityGroupsToENIs(eniIDs, primarySGs, "Applying primary security groups to")
+
+	return nil
```
haouc: If not returning errors, the func doesn't need error in return type.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +524,247 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+				eniInfos := ds.GetENIInfos()
+				for eniID := range eniInfos.ENIs {
+					eniIDs = append(eniIDs, eniID)
+				}
+			}
+
+			newENIs := StringSet{}
+			newENIs.Set(eniIDs)
+			filteredENIs := newENIs.Difference(&cache.unmanagedENIs)
+			eniIDs = filteredENIs.SortedList()
```
haouc, line 770: I saw ENI IDs and SGs being sorted everywhere and not sure why they need to be sorted.

pkg/awsutils/awsutils.go
```diff
@@ -992,33 +1173,79 @@ func (cache *EC2InstanceMetadataCache) createENI(sg []*string, eniCfgSubnet stri
...
 				log.Warnf("Failed to call ec2:DescribeSubnets: %v", vpcErr)
 				log.Info("Defaulting to same subnet as the primary interface for the new ENI")
+
+				// Even in fallback, check if primary subnet is excluded
+				excluded, checkErr := cache.isPrimarySubnetExcluded()
+				if checkErr != nil {
+					log.Warnf("Failed to check if primary subnet is excluded: %v. Proceeding with ENI creation attempt.", checkErr)
+				} else if excluded {
+					// Primary subnet is explicitly excluded
+					err = errors.New("primary subnet is tagged with kubernetes.io/role/cni=0 - no valid subnets available for ENI creation")
```
haouc: `return fmt.Errorf("")` does the same with one line.

pkg/awsutils/awsutils.go
```diff
@@ -992,33 +1173,79 @@ func (cache *EC2InstanceMetadataCache) createENI(sg []*string, eniCfgSubnet stri
 			return networkInterfaceID, nil
 		}
 	} else {
-		if cache.useSubnetDiscovery && !cache.v6Enabled {
```
haouc, line 995: Should v6 check be removed at here? If new logic need check v6, should v6 check be maintained somewhere to keep existing flow?

pkg/awsutils/awsutils.go
```diff
@@ -2253,3 +2541,38 @@ func checkAPIErrorAndBroadcastEvent(err error, api string) {
...
+
+	subnet := subnetResult.Subnets[0]
+	// Check if the subnet has the exclusion tag
+	return !validTag(subnet, true), nil
+}
+
+// IsPrimarySubnetExcluded implements the APIs interface to check if primary subnet is excluded
+func (cache *EC2InstanceMetadataCache) IsPrimarySubnetExcluded() (bool, error) {
+	// Always check tags - explicit user intent should be respected regardless of subnet discovery setting
+	return cache.isPrimarySubnetExcluded()
```
haouc: Not sure why we need two func which are doing same thing. If no extra logic, can't just use `IsPrimarySubnetExcluded` with `isPrimarySubnetExcluded`'s code?

pkg/awsutils/awsutils.go
dshehbaj: Fixed.

Had it separated before so it's easier to understand the relationship between `hasClusterTags` and `belongsToThisCluster` since we need to exclude the subnet if it has the cluster tag of cluster B, but not on the current cluster A.

pkg/awsutils/awsutils.go
dshehbaj: Added shared context for all functions across `awsutils` and `ipamd`.

pkg/awsutils/awsutils.go
dshehbaj: Added shared context for all functions across `awsutils` and `ipamd`.

pkg/awsutils/awsutils.go
dshehbaj: We use this function alongside `RefreshSGIDs` to refresh the custom security groups in `ipamd`.

pkg/awsutils/awsutils.go
dshehbaj: Added shared context for all functions across `awsutils` and `ipamd`.

pkg/awsutils/awsutils.go
dshehbaj: Combined these two functions.

Was trying to encapsulate the logic and have one function called internal only, and the other one externally, but since there is no difference in logic makes sense to combine them to keep things simple.

pkg/awsutils/awsutils.go
dshehbaj: Yes, we would need to remove the v6 check here to make it work with secondary ENIs/subnets. 

Right now it would just create the secondary ENI using the same settings as the primary ENI: https://github.com/aws/amazon-vpc-cni-k8s/blob/312b1d2c907c8109c5fe72cccae7755df64f6b67/pkg/awsutils/awsutils.go#L1020-L1026

pkg/awsutils/awsutils.go
dshehbaj: Fixed

pkg/awsutils/awsutils.go
dshehbaj: Fixed

pkg/awsutils/awsutils.go
dshehbaj: Fixed

pkg/awsutils/awsutils.go
dshehbaj: Fixed

pkg/awsutils/awsutils.go
dshehbaj: Fixed

pkg/awsutils/awsutils.go
dshehbaj: Fixed

pkg/awsutils/awsutils.go
dshehbaj: `SubnetId` is a string, and we are already checking for `	if len(result.NetworkInterfaces) == 0 {
` in the line above.

pkg/awsutils/awsutils.go
dshehbaj: `result.SecurityGroups` is a slice of values, and not pointers. So this will not work.

pkg/awsutils/awsutils.go
dshehbaj: switched to `eniSubnetID`

pkg/awsutils/awsutils.go
dshehbaj: It is a string, switched the variable name to `eniSubnetID`

pkg/awsutils/awsutils.go
dshehbaj: Currently the only two functions to convert a `StringSet` to a `[]string` is either calling `StringSet.Set()` or `StringSet.SortedList()`.

We have used `SortedList()` everywhere we are using `StringSet` even where sorting is not required so I just decided to stick with that convention.

pkg/awsutils/awsutils.go
dshehbaj: Currently the only two functions to convert a `StringSet` to a `[]string` is either calling `StringSet.Set()` or `StringSet.SortedList()`.

We have used `SortedList()` everywhere we are using `StringSet` even where sorting is not required so I just decided to stick with that convention.

pkg/awsutils/awsutils.go
dshehbaj: Currently the only two functions to convert a `StringSet` to a `[]string` is either calling `StringSet.Set()` or `StringSet.SortedList()`.

We have used `SortedList()` everywhere we are using `StringSet` even where sorting is not required so I just decided to stick with that convention.

pkg/awsutils/awsutils.go
dshehbaj: `ds.GetENIInfos()` returns `*datastore.ENIInfos`, and then you iterate through each ENI by going over the map 

`*datastore.ENIInfos.ENIs`.

So not straightforward to do it in one line unless we want to do something like this.

```
eniIDs := slices.Collect(maps.Keys(ds.GetENIInfos().ENIs))
```

pkg/awsutils/awsutils.go
dshehbaj: Fixed

pkg/awsutils/awsutils.go
dshehbaj: Are you talking about the case where customers do not pass in the cluster name?

If they do not tag a subnet with 

`clusterTagKeyPrefix = "kubernetes.io/cluster/"`

then it would not matter if they pass in the cluster name or not as the function will see that there are no cluster specific tags so it would not do that check.


Now if the customer does tag `kubernetes.io/cluster/<cluster-name>=shared` and then forgets to pass in the cluster name as the env var I would assume that would be a customer misconfiguration, right? 

The best we can do here is just log it out, which we are already doing.

pkg/awsutils/awsutils.go
dshehbaj: We already did this in the `RefreshSGIDs` function here: https://github.com/aws/amazon-vpc-cni-k8s/blob/312b1d2c907c8109c5fe72cccae7755df64f6b67/pkg/awsutils/awsutils.go#L561-L564

Since I created `RefreshCustomSGIDs` function. I refactored some of the logic into a shared function.

pkg/awsutils/awsutils.go
dshehbaj: We used the same string in existing logic

https://github.com/aws/amazon-vpc-cni-k8s/blob/312b1d2c907c8109c5fe72cccae7755df64f6b67/pkg/awsutils/awsutils.go#L571C5-L575C6

Let me add this to a file/find better way to handle this.

pkg/awsutils/awsutils.go
dshehbaj: I don't believe we support VPC peering in CNI. I could not find any existing logic that I could leverage. 

I think keeping it local VPC only might keep things simpler, and we could do VPC peering as a follow up.

What do you think?

pkg/awsutils/awsutils.go
haouc: Some wrapper methods returning `bool` should be sufficient. We can have all errros handled in same pattern and same package though.

cmd/aws-k8s-agent/main.go
```diff
@@ -159,7 +159,7 @@ func _main() int {
 	}
 
 	// Pool manager
-	go ipamContext.StartNodeIPPoolManager()
+	go ipamContext.StartNodeIPPoolManager(context.Background())
```
haouc, line 162: If want to use background(), I don't think you need to pass it to the func. It is global context.

pkg/awsutils/awsutils.go
haouc: global SGs is not VPC peering. EC2 support using SG from another VPC.

pkg/awsutils/awsutils.go
```diff
@@ -427,7 +440,7 @@ func New(useSubnetDiscovery, useCustomNetworking, disableLeakedENICleanup, v4Ena
 
 	// Clean up leaked ENIs in the background
 	if !disableLeakedENICleanup {
-		go wait.Forever(cache.cleanUpLeakedENIs, time.Hour)
+		go wait.Forever(func() { cache.cleanUpLeakedENIs(context.Background()) }, time.Hour)
```
haouc, line 449: I think I may have confused you. When I asked if you can pass context around, I was trying to say if the context is managed and should be passed through the tree for better visibility. If they are all the same context like background(), I think it is ok to to get them when need them. This change looks like unnecessarily adding complexity.

pkg/awsutils/awsutils.go
haouc: `SubnetId` is a pointer (https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#NetworkInterface)

pkg/awsutils/awsutils.go
haouc: SecurityGroups is slice of pointers (https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#DescribeSecurityGroupsOutput)

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +526,246 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+			},
+			{
+				Name:   aws.String("tag:" + subnetDiscoveryTagKey),
+				Values: []string{subnetDiscoveryTagValueIncluded},
+			},
+		},
+	}
 
-	sgIDs, err := cache.imds.GetSecurityGroupIDs(ctx, mac)
+	result, err := cache.ec2SVC.DescribeSecurityGroups(ctx, describeSGInput)
```
haouc, line 550: Since you are not using ID, I would suggest to use pagination.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +526,246 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+		return "", fmt.Errorf("getENISubnetID: no interfaces found")
+	}
+
+	return *result.NetworkInterfaces[0].SubnetId, nil
+}
+
+// Helper function to check if an ENI is in a secondary subnet
+func (cache *EC2InstanceMetadataCache) isENIInSecondarySubnet(ctx context.Context, eniID string) bool {
+	eniSubnetID, err := cache.getENISubnetID(ctx, eniID)
+	return err == nil && eniSubnetID != cache.subnetID
```
haouc, line 584: nit: considering change cache.subnetID to cache.primarySubnetID or cache.nodeSubnetID.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +526,246 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+
+			isSecondarySubnet := cache.isENIInSecondarySubnet(ctx, eniID)
+
+			// Filter based on subnet type
+			if onlySecondarySubnets && !isSecondarySubnet {
+				continue
+			} else if !onlySecondarySubnets && isSecondarySubnet {
+				// When we want primary subnet ENIs, skip secondary subnet ENIs
+				continue
+			}
```
haouc, line 607: Probably ?
```
if onlySecondarySubnets != isSecondarySubnet {
    continue
}
```

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +526,246 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+func (cache *EC2InstanceMetadataCache) applySecurityGroupsToENIs(ctx context.Context, eniIDs []string, sgIDs []string, logPrefix string) {
+	for _, eniID := range eniIDs {
+		log.Debugf("%s ENI %s with security groups %v", logPrefix, eniID, sgIDs)
+
+		attributeInput := &ec2.ModifyNetworkInterfaceAttributeInput{
+			Groups:             sgIDs,
+			NetworkInterfaceId: aws.String(eniID),
+		}
+		start := time.Now()
+		_, err := cache.ec2SVC.ModifyNetworkInterfaceAttribute(ctx, attributeInput)
```
haouc, line 631: What are you going to do with the error? An error means updating SGs failed. Should some action be taken?

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +526,246 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
-		addedSGsCount++
+		log.Infof("Found %s SG %s, added to ipamd cache", sgType, sg)
+		addedCount++
 	}
+
+	deletedCount := 0
 	for _, sg := range deletedSGs.SortedList() {
-		log.Infof("Removed %s from ipamd cache", sg)
-		deletedSGsCount++
+		log.Infof("Removed %s SG %s from ipamd cache", sgType, sg)
```
haouc, line 664: nit: looks like not deleting them yet?

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +526,246 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+	return addedCount, deletedCount
+}
+
+// RefreshCustomSGIDs discovers and refreshes security groups tagged for use with the CNI
+func (cache *EC2InstanceMetadataCache) RefreshCustomSGIDs(ctx context.Context, dsAccess *datastore.DataStoreAccess) error {
+	sgIDs, err := cache.discoverCustomSecurityGroups(ctx)
+	if err != nil {
+		awsAPIErrInc("DiscoverCustomSecurityGroups", err)
+		log.Warnf("Failed to discover custom security groups: %v", err)
+		log.Info("Falling back to using primary security groups for ENIs in secondary subnets")
```
haouc, line 677: Can you combine two logs?

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +526,246 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+
+		// Clear custom security groups cache to trigger fallback behavior
+		cache.customSecurityGroups.Set([]string{})
```
haouc, line 680: Not following why need to flush the cache here instead of handling them in fallback call below?

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +526,246 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+
+	// Check if no custom security groups were found (empty list)
+	if len(sgIDs) == 0 {
+		log.Info("No custom security groups found, using primary security groups for ENIs in secondary subnets")
+
+		// Clear custom security groups cache
+		cache.customSecurityGroups.Set([]string{})
+
+		// Apply primary security groups to ENIs in secondary subnets as fallback
+		cache.applyFallbackSecurityGroupsForAllDatastores(ctx, dsAccess)
```
haouc, line 695: Is this falling back to node SGs?

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +526,246 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+
+		return nil
+	}
+
+	addedCount, deletedCount := cache.detectSecurityGroupChanges(sgIDs, &cache.customSecurityGroups, "custom")
```
haouc, line 700: Let's try not to use arbitary string to pass conditions.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +526,246 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+	if addedCount != 0 || deletedCount != 0 {
+		var eniIDs []string
 		for _, ds := range dsAccess.DataStores {
-			eniInfos := ds.GetENIInfos()
-			for eniID := range eniInfos.ENIs {
-				eniIDs = append(eniIDs, eniID)
-			}
+			eniIDs = append(eniIDs, cache.getFilteredENIs(ctx, ds, true)...) // only secondary subnet ENIs
 		}
+		cache.applySecurityGroupsToENIs(ctx, eniIDs, sgIDs, "Update")
```
haouc, line 709: Same suggestion here as well.

haouc: Also I expect some error handling here since it basically calls upsteam API.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +526,246 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
-			if err != nil {
-				if errors.As(err, &awsAPIError) {
-					if awsAPIError.ErrorCode() == "InvalidNetworkInterfaceID.NotFound" {
-						awsAPIErrInc("IMDSMetaDataOutOfSync", err)
+	var eniIDs []string
+	for _, ds := range dsAccess.DataStores {
+		eniIDs = append(eniIDs, cache.getFilteredENIs(ctx, ds, true)...) // only secondary subnet ENIs
+	}
+
+	cache.applySecurityGroupsToENIs(ctx, eniIDs, primarySGs, "Applying primary security groups to")
```
haouc, line 729: We should have error to handle.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +526,246 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+
+// RefreshSGIDs retrieves security groups
+func (cache *EC2InstanceMetadataCache) RefreshSGIDs(ctx context.Context, mac string, dsAccess *datastore.DataStoreAccess) error {
+	sgIDs, err := cache.imds.GetSecurityGroupIDs(ctx, mac)
+	if err != nil {
+		awsAPIErrInc("GetSecurityGroupIDs", err)
+		return err
+	}
+
+	addedCount, deletedCount := cache.detectSecurityGroupChanges(sgIDs, &cache.securityGroups, "primary")
```
haouc, line 740: Same arbitary argument being passed through.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +526,246 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+		var eniIDs []string
+
+		// When subnet discovery is enabled, only apply primary SGs to primary subnet ENIs
+		if cache.useSubnetDiscovery {
+			for _, ds := range dsAccess.DataStores {
+				// Get only primary subnet ENIs (onlySecondarySubnets=false)
+				primarySubnetENIs := cache.getFilteredENIs(ctx, ds, false)
+				for _, eniID := range primarySubnetENIs {
+					// Filter out unmanaged ENIs
+					if !cache.unmanagedENIs.Has(eniID) {
```
haouc, line 753: nit: I feel unmanaged ENIs should be handled in `cache.getFilteredENIs()`.

pkg/awsutils/awsutils.go
haouc: You can fo it with lo package. But it is ok to use common for loop since this is just simplifying syntax.

pkg/awsutils/awsutils.go
```diff
@@ -513,75 +526,246 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+			}
+
+			newENIs := StringSet{}
+			newENIs.Set(eniIDs)
+			filteredENIs := newENIs.Difference(&cache.unmanagedENIs)
+			eniIDs = filteredENIs.SortedList()
 		}
+
+		// Apply security groups to the filtered ENIs
+		cache.applySecurityGroupsToENIs(ctx, eniIDs, sgIDs, "Update")
```
haouc, line 774: Same here.

pkg/awsutils/awsutils.go
```diff
@@ -987,39 +1169,79 @@ func (cache *EC2InstanceMetadataCache) createENI(sg []*string, eniCfgSubnet stri
...
 				log.Info("Defaulting to same subnet as the primary interface for the new ENI")
-				networkInterfaceID, err = cache.tryCreateNetworkInterface(input)
+
+				// Even in fallback, check if primary subnet is excluded
+				excluded, checkErr := cache.IsPrimarySubnetExcluded(ctx)
+				if checkErr != nil {
+					log.Warnf("Failed to check if primary subnet is excluded: %v. Proceeding with ENI creation attempt.", checkErr)
+				} else if excluded {
+					// Primary subnet is explicitly excluded
+					return "", fmt.Errorf("primary subnet is tagged with kubernetes.io/role/cni=0 - no valid subnets available for ENI creation")
```
haouc, line 1195: I suggest surface this error to customers.

pkg/awsutils/awsutils.go
haouc: Are you saying if no clustername ENV set, the subnet will be shared between clusters?

pkg/awsutils/awsutils.go
```diff
@@ -1175,13 +1184,13 @@ func (cache *EC2InstanceMetadataCache) createENI(ctx context.Context, sg []*stri
...
 		if cache.useSubnetDiscovery {
-			subnetResult, vpcErr := cache.getVpcSubnets(ctx)
+			subnetResult, vpcErr := cache.GetVpcSubnets(ctx)
 			if vpcErr != nil {
 				log.Warnf("Failed to call ec2:DescribeSubnets: %v", vpcErr)
 				log.Info("Defaulting to same subnet as the primary interface for the new ENI")
 
 				// Even in fallback, check if primary subnet is excluded
-				excluded, checkErr := cache.IsPrimarySubnetExcluded(ctx)
+				excluded, checkErr := cache.IsSubnetExcluded(ctx, cache.subnetID)
```
dshehbaj, line 1190: use primarySubnetID = cache.subnetID

PR #3404, opened 2025-08-14
Add and configure soak tests workflow

Supreeth095:
**What type of PR is this?**

```release-note

```

### DIFF
--- a/.github/workflows/soak-tests.yml
+++ b/.github/workflows/soak-tests.yml
@@ -15,10 +15,6 @@ jobs:
     steps:
       - name: Checkout latest commit in the PR
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # refs/tags/v4.1.7
-      - name: Set up Docker QEMU
-        uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 # refs/tags/v3.6.0
-      - name: Set up Docker Buildx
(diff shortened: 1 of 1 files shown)


.github/workflows/soak-tests.yml
```diff
@@ -0,0 +1,43 @@
+name: Soak Tests
+on:
+  push:
+    branches:
+      - "master"
+      - "release*"
```
yash97: why do we need to run this on all push? should it be nightly?

yash97: also add this so we can manualy run it too https://github.com/aws/amazon-vpc-cni-k8s/blob/master/.github/workflows/nightly-cron-tests.yaml#L4

Supreeth095: Yup, Nightly should be good enough. Replaced all push with nightly. Added option to run manually

.github/workflows/soak-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+  id-token: write
+  contents: read
+
+jobs:
+  integration-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout latest commit in the PR
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # refs/tags/v4.1.7
+      - name: Set up Docker QEMU
```
yash97: do we build image here too, if not then we should remove it.

Supreeth095: Got rid of Build. I think we don't need it. Let me commit and check.

yash97 (approved):

PR #3411, opened 2025-08-20
Removing pinning of golang version

jaydeokar:
**What type of PR is this?**
dependency update and adds a automation to auto-generate vpc-limits.go file for new instances

**Which issue does this PR fix?**: golang CVE 

**What does this PR do / Why do we need it?**:  It uses the latest golang patch version instead of hard coding the patch release version 

**Will this PR introduce any new dependencies?**:  N/A

**Will this break upgrades or downgrades? Has updating a running cluster been tested?**:  N/A

**Does this change require updates to the CNI daemonset config files to work?**:  N/A

**Does this PR introduce any user-facing change?**:  N/A

```release-note

```


jaydeokar:
eg run - https://github.com/jaydeokar/amazon-vpc-cni-k8s/pull/14

.github/workflows/generate-limits.yaml
```diff
@@ -0,0 +1,67 @@
+name: Generate Limits
+
+on:
+  workflow_dispatch: {}
+  schedule:
+    - cron: "0 10 * * 1" # Every Monday at 10:00 UTC
```
viveksb007: why every monday and not daily? is there some pre-determined schedule from aws side?

viveksb007 (approved):

.github/workflows/generate-limits.yaml
jaydeokar: No reason, we don't release CNI that frequently so started with weekly. I can make it more frequent

viveksb007: i see, makes sense. somehow i thought these change will immediately be available to use

haouc (approved):
lgtm

PR #3428, opened 2025-09-02
Fix malformed YAML in aws-k8s-cni-cn.yaml

### LINKED ISSUE #3239: Malformed yaml included in config yaml
**What happened**:

I was attempting to use a yaml file contained in the config directory to install the aws vpc cni. This line: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/config/master/aws-k8s-cni-cn.yaml#L208 should be joined with the preceding line.  

Errors like this could be prevented in the future by including a step in CI to have the files processed a tools like yq.

----- 

As a note, I am aware of the helm chart that is available. I prefer to use files like those included in the projects config directory and make the require modifications through kustomize.

This malformed line has been around for a few releases without anyone complaining so I would bet that most people are using the provided helm chart.

For any yaml contained in the project to be well formed. Specifically, I would expect that config files included in the config directory are suitable to use as a base for installing.
```
cat aws-k8s-cni-cn.yaml | yq
```
Observe the outputted error when trying to process the malformed  yaml.
Not relevant

musharafmaqbool:
Fixes #3239 - Join securityContext line with empty object

**What type of PR is this?**

```release-note

```

### DIFF
--- a/config/master/aws-k8s-cni-cn.yaml
+++ b/config/master/aws-k8s-cni-cn.yaml
@@ -436,8 +436,7 @@ spec:
       terminationGracePeriodSeconds: 10
       tolerations:
         - operator: Exists
-      securityContext:
-        {}
+      securityContext: {}
       containers:
         - name: aws-node
           image: 961992271922.dkr.ecr.cn-northwest-1.amazonaws.com.cn/amazon-k8s-cni:v1.20.1


musharafmaqbool:
hey kindly review my pr request

jaydeokar:
Can you add details on how you are fixing these ?
 
This needs to be fixed in the helm chart instead of the yaml files. Also it will be helpful if it can be a part of github action instead of doing this manually. Refer to this as an example. 

https://github.com/aws/amazon-vpc-cni-k8s/blob/master/.github/workflows/generate-limits.yaml

musharafmaqbool:
Hi @jaydeokar, thanks for the review!

**How I'm fixing this:**
- Found the malformed YAML where `securityContext:` and `{}` were split across two lines
- Joined them into a single line: `securityContext: {}` 
- This resolves the YAML parsing error mentioned in issue #3239

**Regarding your suggestions:**
1. **Helm chart fix**: I can look into updating the source Helm templates as well
2. **GitHub Action**: The workflow example you shared for automated YAML validation looks useful

Should I:
1. Update this PR to also fix the Helm chart source?
2. Create a separate issue/PR for adding the GitHub Action workflow?

Let me know your preference for how to proceed!

musharafmaqbool:
@jaydeokar hey can you please review it

PR #3429, opened 2025-09-03
Enhance IP allocation error diagnostics with detailed ENI and fragmentation info

musharafmaqbool:
Follow-up to enhance IP allocation error diagnostics with detailed ENI and fragmentation information.

**What type of PR is this?**
improvement

**Which issue does this PR fix?**
Addresses feedback on improving IP allocation error messages to provide better diagnostic information for troubleshooting.

**What does this PR do / Why do we need it?**
This PR enhances IP allocation failure error messages in ipamd.go to include:
- ENI count and limit information 
- IPv4/IPv6 prefix allocation status with current usage statistics
- Better context for prefix delegation scenarios
- Enhanced diagnostic information to help users quickly identify root causes of IP assignment failures

**Testing done on this change:**
Enhanced error messages have been tested to ensure they provide actionable diagnostic information without breaking existing functionality.

**Will this PR introduce any new dependencies?**
No

**Will this break upgrades or downgrades? Has updating a running cluster been tested?**
No, this only enhances error message content and does not affect core functionality.

**Does this change require updates to the CNI daemonset config files to work?**
No

**Does this PR introduce any user-facing changes?**
Yes, users will see more detailed and helpful error messages when IP allocation fails, making troubleshooting easier.

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -13,6 +13,7 @@
 
 package ipamd
 
+
 import (
 	"context"
 	"fmt"
@@ -23,13 +24,9 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
-
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/k8sapi"
-
 	"github.com/aws/smithy-go"
-
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
 	"github.com/aws/aws-sdk-go-v2/aws"
 	ec2types "github.com/aws/aws-sdk-go-v2/service/ec2/types"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


erezzarum:
I'm not sure it gives the details i described in the issue.
Have you run tests when subnet is fragmented? how did you run it?

pkg/ipamd/ipamd.go
jaydeokar, line 1: This will just add more verbose errors on ipamd log which was not the intention to begin with. 

What we need is message/event on pod events or some Custom Resource which gives the reason why IP allocation to the pod failed. 
Right now it's a generic message of "failing to get an IP address". There are two async processes. One which scales up the IPs/ENIs and an GRPC handler which just picks an IP from the datastore and assigns it to the pod. The log enhancement is for the latter part when pod fails to get an IP address

PR #3430, opened 2025-09-03
Add VPC sweeper for cleaning up orphaned test resources

Supreeth095:
- Add vpc-sweeper.sh script to clean up test VPCs older than 7 days
- Add GitHub workflow to run sweeper weekly on Sunday 9am PST
- Add Makefile targets for vpc-sweeper and vpc-sweeper-force
- Includes dry-run mode by default for safety

**What type of PR is this?**
improvement

Cron Jobs never run because VPC count is always exceeding quota. We need a sweeper that cleans up Stale VPCs and corresponding resources.

Cron Jobs never run because VPC count is always exceeding quota. We need a sweeper that cleans up Stale VPCs and corresponding resources.

No

No

No

```release-note

```

### DIFF
--- a/.github/workflows/vpc-sweeper.yaml
+++ b/.github/workflows/vpc-sweeper.yaml
@@ -1,0 +1,48 @@
+name: VPC Sweeper
+
+on:
+  schedule:
+    - cron: "0 17 * * 0" # Weekly on Sunday at 9 AM PST (17:00 UTC)
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: 'Run in dry-run mode'
+        required: false
+        default: 'false'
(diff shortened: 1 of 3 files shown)


yash97:
will this script delete eks cluster created resources only or will also cleanup stale resources from cluster created through kops

Supreeth095:
> will this script delete eks cluster created resources only or will also cleanup stale resources from cluster created through kops
This deletes only resources created with test cluster tags.

jaydeokar:
Can you please add testing details here ?  Maybe run it against the test account to check if everything is getting cleaned up and add it here. Also this script does not cover kops clusters which gets leaked

PR #3437, opened 2025-09-10
fetch cni plugins from s3 bucket for internal builds

jupdec:
**What type of PR is this?**

```release-note

```

### DIFF
--- a/.github/workflows/kops-test.yaml
+++ b/.github/workflows/kops-test.yaml
@@ -43,8 +43,8 @@ jobs:
           ROLE_ARN: ${{ secrets.EKS_CLUSTER_ROLE_ARN }}
           RUN_CNI_INTEGRATION_TESTS: false
           RUN_KOPS_TEST: true
-          K8S_VERSION: 1.33.0-beta.0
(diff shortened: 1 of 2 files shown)


yash97:
can you please add testing done to validate this change in description.

Makefile
```diff
@@ -293,7 +293,15 @@ plugins: FETCH_VERSION=1.7.1
 plugins: FETCH_URL=https://github.com/containernetworking/plugins/archive/refs/tags/v$(FETCH_VERSION).tar.gz
 plugins: VISIT_URL=https://github.com/containernetworking/plugins/tree/v$(FETCH_VERSION)/plugins/
 plugins: CORE_PLUGINS=bandwidth,host-local,loopback,portmap,sbr
+plugins: S3_BUCKET=cni-plugins-dev-artifacts-us-west-2
```
viveksb007: nit - if this is specific to dev, then maybe inline in line 300 instead of global declaration of S3_BUCKET var?

viveksb007 (commented):
please add details on why is this change needed for internal build? is there any issue in regular source of CNI plugins?

PR #3440, opened 2025-09-16
feat: add retry logic for netlink dump operations

dcoppa:
- Wrap netlink.ErrDumpInterrupted errors with retry logic in netlinkwrapper
- Implement retry mechanism for LinkList, RouteList, AddrList, and RuleList operations
- Update files to use netlinkwrapper instead of direct netlink calls

**What type of PR is this?**

improvement

Errors like:

`Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "b413c572e7bac7a27767123be3da3e031a3ac5b3695efc995c15057996b2c930": plugin type="aws-cni" name="aws-cni" failed (add): add command: failed to setup network: SetupPodNetwork: failed to setup veth pair: failed to setup veth network: setup NS network: failed while waiting for v6 addresses to be stable: could not list addresses: results may be incomplete or inconsistent`

This PR adds a retry mechanism when netlink fails with error ErrDumpInterrupted.

Adapted from https://github.com/containernetworking/plugins/blob/main/pkg/netlinksafe/netlink.go

This new PR should address all the remarks made on my previous submission.
Sorry for the delay, life got in the way.

Unit tests passed.
Also tested on my sandbox EKS cluster.

No

No / Yes

No

Implement a retry mechanism for netlink that, on an IPv6 EKS cluster, should fix (or mitigate) the following error:

`Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "e868238d122df610228688f51fee6a187544a45ce43a9d613307b6b08246e65a": plugin type="aws-cni" name="aws-cni" failed (add): add command: failed to setup network: SetupPodNetwork: failed to setup veth pair: failed to setup veth network: setup NS network: failed while waiting for v6 addresses to be stable: could not list addresses: results may be incomplete or inconsistent`

No additional actions from users required.

### DIFF
--- a/pkg/netlinkwrapper/netlink_test.go
+++ b/pkg/netlinkwrapper/netlink_test.go
@@ -105,9 +105,9 @@ func TestRetryOnErrDumpInterrupted_Delay(t *testing.T) {
 	firstToSecond := timestamps[1].Sub(timestamps[0])
 	assert.Less(t, firstToSecond, 50*time.Millisecond, "First retry should happen immediately without delay")
 
-	for i := 2; i < len(timestamps); i++ {
-		delay := timestamps[i].Sub(timestamps[i-1])
-		assert.GreaterOrEqual(t, delay, 100*time.Millisecond, "Delay before attempt %d should be at least 100ms", i+1)
+	for attempt := 2; attempt < len(timestamps); attempt++ {
+		delay := timestamps[attempt].Sub(timestamps[attempt-1])
+		assert.GreaterOrEqual(t, delay, 100*time.Millisecond, "Delay before attempt %d should be at least 100ms", attempt+1)
 	}
 }
 


viveksb007:
Thanks @dcoppa for the PR.

can you also open a corresponding issue for this PR? (under what conditions in IPv6 cluster do you see this issue? if we can re-produce this issue on our side, that will be great)

dcoppa:
> Thanks @dcoppa for the PR.
> 
> can you also open a corresponding issue for this PR? (under what conditions in IPv6 cluster do you see this issue? if we can re-produce this issue on our side, that will be great)

Hi!

Here's the issue: https://github.com/aws/amazon-vpc-cni-k8s/issues/3461

The problem has also already been discussed in issue number https://github.com/aws/amazon-vpc-cni-k8s/issues/3196.
Take a look in particular at @yash97 ’s comments.

jaydeokar:
Thanks for raising this PR. We will take a look

pkg/netlinkwrapper/netlink.go
```diff
@@ -69,6 +75,42 @@ type NetLink interface {
...
 
+const (
+	// Environment variable to configure netlink max retry attempts
+	envNetlinkMaxRetries = "AWS_VPC_K8S_CNI_NETLINK_MAX_RETRIES"
+	defaultMaxAttempts   = 5
+)
+
+// getMaxAttempts returns the configured maximum retry attempts for netlink operations
+func getMaxAttempts() int {
+	maxAttempts, _, _ := utils.GetIntFromStringEnvVar(envNetlinkMaxRetries, defaultMaxAttempts)
```
jaydeokar: I believe CNI won't be able to access this env variable since it's running as a binary on the host. Might need to pass this via conf file or the GRPC call

yash97: We had discussion offline, We think we should follow upstream. Keep it simple and stupid, make it constant 5 times.

dcoppa: > We had discussion offline, We think we should follow upstream. Keep it simple and stupid, make it constant 5 times.

Done.

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -564,7 +564,7 @@ func (n *linuxNetwork) teardownIPBasedContainerRouteRules(containerAddr *net.IPN
 	log.Debugf("Successfully deleted toContainer rule, containerAddr=%s, rtTable=%v", containerAddr.String(), "main")
 
 	if rtTable != unix.RT_TABLE_MAIN {
-		fromContainerRule := netlink.NewRule()
+		fromContainerRule := n.netLink.NewRule()
```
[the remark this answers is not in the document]
yash97: ~~@dcoppa  need to look into this change.~~  netlink wrapper interface does have this method https://github.com/aws/amazon-vpc-cni-k8s/blob/a49107121e4c0337372dbe18d65f15d761973ebf/pkg/netlinkwrapper/netlink.go#L58-L59. Bad copilot.

pkg/netlinkwrapper/netlink.go
```diff
@@ -69,6 +74,26 @@ type NetLink interface {
 type netLink struct {
 }
 
+func retryOnErrDumpInterrupted(f func() error) error {
+	var lastErr error
+	const maxAttempts = 5
+	for attempt := 0; attempt < maxAttempts; attempt++ {
+		err := f()
```
yash97, line 82: can we add really small delay after 1 st attempt, so kernel is not overwhelmed.

dcoppa: > can we add really small delay after 1 st attempt, so kernel is not overwhelmed.

Is an exponential backoff better, or only after the first attempt?
How small the delay?

yash97: I think this is critical part, do not want to add exponential here. I think 100 millisecond should be good. Also there should be no delay in first attempt. I think it is implied but only sleep if error is because of interrupt.

dcoppa: Does it look good?

yash97: looking good. One more thing lol, I just noticed there are no unit test for this function retryOnErrDumpInterrupted. Let's please add that. This function is getting larger and critical for netlink operations.. Pr is shaping up nicely. Thank you for being so responsive.

dcoppa: Added netlink_test.go

PR #3452, opened 2025-09-21
Optionally setting up Route table for non zero Network Cards if AMI is already configuring at bootup

jaydeokar:
**What type of PR is this?** 
improvement

CNI SetupENINetwork changes to let ec2-net-utils handle ENI setup on the host where multiple cards are in use

This allows ec2-net-utils to setup all the ENIs on NC > 0 instead of depending on enabling multi NIC feature of CNI.
CNI will verify if the ENI is already setup on the host and reuse the route tables configured by EC2 

**Testing done on this change**: Yes, tested with upgrade scenarios

This introduces an dependency for downgrades where nodes need to be refreshed if you are trying to downgrade the version and are already using the new AMI which does the ENI setup. This is only needed if you are using `ENABLE_MULTI_NIC` feature of CNI. 


pkg/ipamd/ipamd.go
```diff
@@ -1275,7 +1275,7 @@ func (c *IPAMContext) setupENI(eni string, eniMetadata awsutils.ENIMetadata, isT
 		if c.enableIPv6 {
 			subnetCidr = eniMetadata.SubnetIPv6CIDR
 		}
-		err = c.networkClient.SetupENINetwork(c.primaryIP[eni], eniMetadata.MAC, eniMetadata.DeviceNumber, eniMetadata.NetworkCard, subnetCidr, c.maxENI, isTrunkENI)
+		routeTableID, err := c.networkClient.SetupENINetwork(c.primaryIP[eni], eniMetadata.MAC, eniMetadata.DeviceNumber, eniMetadata.NetworkCard, subnetCidr, c.maxENI, isTrunkENI)
```
M00nF1sh: can we move this to earlier, thus we can call `Datastore.AddENI` and pass in the routeTableID directly.

This is better than separately call AddRouteTableIDForENI, which ensures ENIs in datastore have valid route table ID.

pkg/ipamd/rpc_handler_test.go
```diff
@@ -194,8 +204,8 @@ func TestServer_AddNetwork(t *testing.T) {
 					{
 						IPv4Addr:     "192.168.16.100",
 						IPv6Addr:     "",
-						DeviceNumber: int32(0),
-						RouteTableId: int32(1*0 + maxENIsPerCard + 1), // Network Card * DeviceNumber + Max ENIs per card + 1
+						DeviceNumber: int32(1),
+						RouteTableId: int32(10101), // Network Card * DeviceNumber + Max ENIs per card + 1
```
haouc, line 208: Do you intentionally change to a number from variables?

pkg/networkutils/helper.go
```diff
@@ -2,11 +2,21 @@ package networkutils
...
+func CalculateOldRouteTableId(deviceNumber int, networkCardIndex int, maxENIsPerNetworkCard int) int {
+	return deviceNumber + 1 + (networkCardIndex * maxENIsPerNetworkCard)
+}
+
+func CalculateRouteTableId(deviceNumber int, networkCardIndex int) int {
 	if networkCardIndex == 0 {
 		return deviceNumber + 1
 	} else {
-		return deviceNumber + 1 + (networkCardIndex * maxENIsPerNetworkCard)
+		// https://github.com/amazonlinux/amazon-ec2-net-utils/blob/main/lib/lib.sh#L285
```
haouc: I think the link doesn't work, at least not for main.

pkg/networkutils/network.go
```diff
@@ -1020,93 +1020,87 @@ func GetIPv4Gateway(eniSubnetCIDR *net.IPNet) net.IP {
 }
 
 // SetupENINetwork adds default route to route table (eni-<eni_table>), so it does not need to be called on the primary ENI
-func (n *linuxNetwork) SetupENINetwork(eniIP string, eniMAC string, deviceNumber int, networkCard int, eniSubnetCIDR string, maxENIPerNIC int, isTrunkENI bool) error {
+func (n *linuxNetwork) SetupENINetwork(eniIP string, eniMAC string, deviceNumber int, networkCard int, eniSubnetCIDR string, maxENIPerNIC int, isTrunkENI bool) (int, error) {
 	return setupENINetwork(eniIP, eniMAC, deviceNumber, networkCard, eniSubnetCIDR, n.netLink, retryLinkByMacInterval, retryRouteAddInterval, n.mtu, maxENIPerNIC, isTrunkENI)
 }
 
-// Even if this is called for primary IF on multicard instance
+func getNetworkConfig(isV6 bool, eniSubnetIPNet *net.IPNet) (int, int, net.IP, net.IP) {
```
haouc: Looks like this is no need to be an individual function? Would an inline condition statement work at LN1050?

pkg/networkutils/network.go
```diff
@@ -1020,93 +1020,87 @@ func GetIPv4Gateway(eniSubnetCIDR *net.IPNet) net.IP {
...
+	_, eniSubnetIPNet, err := net.ParseCIDR(eniSubnetCIDR)
+	if err != nil {
+		return 0, errors.Wrapf(err, "setupENINetwork: invalid IP CIDR block %s", eniSubnetCIDR)
+	}
+
+	family, mask, zeroAddr, gw = getNetworkConfig(isV6, eniSubnetIPNet)
+
+	tableNumber, ruleExists, err := getRouteTableNumberForENI(networkCard, eniIP, mask, deviceNumber, maxENIPerNIC, isV6, netLink)
+	if err != nil {
+		return 0, err
```
haouc: Do you want to wrap this error as well? Like LN 1062/1066?

pkg/ipamd/ipamd.go
jaydeokar: Moved this to before AddENI

pkg/ipamd/rpc_handler_test.go
jaydeokar: Yes, as per the new calculation that AMI uses

pkg/networkutils/network.go
jaydeokar: Updated to an inline function

pkg/networkutils/network.go
jaydeokar: Changed this to return error similar to others

pkg/networkutils/helper.go
jaydeokar: Updated this

pkg/networkutils/network.go
```diff
@@ -1019,24 +1021,39 @@ func GetIPv4Gateway(eniSubnetCIDR *net.IPNet) net.IP {
 	return gw
 }
 
+func (n *linuxNetwork) GetRouteTableNumberForENI(networkCard int, eniIP string, deviceNumber int, maxENIPerNIC int, isV6 bool) (int, bool, error) {
+	return getRouteTableNumberForENI(networkCard, eniIP, lo.Ternary(isV6, 128, 32), deviceNumber, maxENIPerNIC, isV6, n.netLink)
+}
```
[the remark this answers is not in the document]
jaydeokar: It's an inline helper function, we aren't passing twice

PR #3463, opened 2025-09-29
feat: enhanced subnet discovery rebase from #3380

haouc:
**What type of PR is this?**

The is a feature PR and rebased from #3380 

```release-note

```


pkg/awsutils/awsutils.go
```diff
@@ -516,75 +532,242 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
-		var eniIDs []string
+	return addedCount, deletedCount
+}
+
+// RefreshCustomSGIDs discovers and refreshes security groups tagged for use with the CNI
+func (cache *EC2InstanceMetadataCache) RefreshCustomSGIDs(ctx context.Context, dsAccess *datastore.DataStoreAccess) error {
+	sgIDs, err := cache.discoverCustomSecurityGroups(ctx)
+	if err != nil {
+		awsAPIErrInc("DiscoverCustomSecurityGroups", err)
+		log.Warnf("Failed to discover custom security groups: %v. Falling back to using primary security groups for ENIs in secondary subnets", err)
```
Pavani-Panakanti, line 685: Here user might have applied some custom security groups but EC2 call might have failed to fetch this data. In such cases wouldn't falling back to primary SG be security concern ? 
err != nil -> can we throw err or retry in this case

pkg/awsutils/awsutils.go
```diff
@@ -516,75 +532,242 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+	return err == nil && eniSubnetID != cache.subnetID
+}
+
+// Helper function to get ENIs that match specific criteria
+func (cache *EC2InstanceMetadataCache) getFilteredENIs(ctx context.Context, store *datastore.DataStore, onlySecondarySubnets bool) []string {
+	eniInfos := store.GetENIInfos()
+	var eniIDs []string
+
+	for eniID := range eniInfos.ENIs {
+		if eniInfo, ok := eniInfos.ENIs[eniID]; ok {
```
Pavani-Panakanti, line 605: This seems a bit confusing 
```
for eniID, eniInfo := range eniInfos.ENIs
```

pkg/awsutils/awsutils.go
```diff
@@ -516,75 +532,242 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
+			// Skip primary ENI for secondary subnet operations
+			if onlySecondarySubnets && eniInfo.IsPrimary {
+				continue
+			}
+
+			isSecondarySubnet := cache.isENIInSecondarySubnet(ctx, eniID)
```
Pavani-Panakanti: Looks like we will be making EC2 call here for every reconcile loop. As subnet id wont change for an ENI should we cache the subnet ID in ENIMetadata to avoid EC2 calls in every loop which will be expensive

pkg/awsutils/awsutils.go
```diff
@@ -516,75 +532,242 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
...
-		newENIs.Set(eniIDs)
-		filteredENIs := newENIs.Difference(&cache.unmanagedENIs)
+	return nil
+}
 
-		// This will update SG for managed ENIs created by EKS.
-		for _, eniID := range filteredENIs.SortedList() {
-			log.Debugf("Update ENI %s", eniID)
+// applyFallbackSecurityGroupsForAllDatastores applies primary security groups to ENIs in secondary subnets across all datastores
+func (cache *EC2InstanceMetadataCache) applyFallbackSecurityGroupsForAllDatastores(ctx context.Context, dsAccess *datastore.DataStoreAccess) {
```
Pavani-Panakanti: can we rename this to "applyPrimarySGsToSecondarySubnetENIs" ?

pkg/awsutils/awsutils.go
```diff
@@ -995,39 +1176,79 @@ func (cache *EC2InstanceMetadataCache) createENI(sg []*string, eniCfgSubnet stri
...
+						continue
 					}
-					log.Infof("Creating ENI with security groups: %v in subnet: %s", input.Groups, aws.ToString(input.SubnetId))
+					validSubnetsFound = true
+					// If this is a secondary subnet and we have custom security groups, use those instead
+					// We already determined isPrimarySubnet above, just reuse the variable
+					if !isPrimarySubnet && len(cache.customSecurityGroups.SortedList()) > 0 {
+						log.Infof("Using custom security groups for ENI in secondary subnet %s", *subnet.SubnetId)
+						input.Groups = cache.customSecurityGroups.SortedList()
+					} else if !isPrimarySubnet {
```
Pavani-Panakanti, line 1242: we are not overriding input.Groups here. If earlier subnet goes through first "if" case then input will be carried from there to next call which might go through "else" case

pkg/awsutils/awsutils.go
```diff
@@ -995,39 +1176,79 @@ func (cache *EC2InstanceMetadataCache) createENI(sg []*string, eniCfgSubnet stri
...
+		if cache.useSubnetDiscovery {
+			subnetResult, vpcErr := cache.GetVpcSubnets(ctx)
 			if vpcErr != nil {
 				log.Warnf("Failed to call ec2:DescribeSubnets: %v", vpcErr)
 				log.Info("Defaulting to same subnet as the primary interface for the new ENI")
-				networkInterfaceID, err = cache.tryCreateNetworkInterface(input)
+
+				// Even in fallback, check if primary subnet is excluded
+				excluded, checkErr := cache.IsSubnetExcluded(ctx, cache.subnetID)
+				if checkErr != nil {
```
Pavani-Panakanti, line 1210: In this case if primary subnet is excluded and we go ahead with ENI creation it would be a security concern right ?

pkg/awsutils/awsutils.go
```diff
@@ -1069,15 +1290,71 @@ func (cache *EC2InstanceMetadataCache) getVpcSubnets() ([]ec2types.Subnet, error
...
 
-func validTag(subnet ec2types.Subnet) bool {
-	for _, tag := range subnet.Tags {
-		if *tag.Key == subnetDiscoveryTagKey {
-			return true
+// validTag checks if subnet should be used for ENI/IP allocation
+// For primary subnet: include by default (no tag), exclude only if tag value is "0"
+// For secondary subnets: exclude by default (no tag), include only if tag exists with non-"0" value
+// If the subnet has cluster-specific tags, it will only be used by the matching cluster
+func validTag(subnet ec2types.Subnet, isPrimarySubnet bool) bool {
```
Pavani-Panakanti: How about renaming to "isSubnetValidForENICreation" ?

pkg/awsutils/awsutils.go
```diff
@@ -1069,15 +1290,71 @@ func (cache *EC2InstanceMetadataCache) getVpcSubnets() ([]ec2types.Subnet, error
...
+			log.Debugf("Primary subnet %s has no %s tag, including it for ENI creation (backwards compatibility)", *subnet.SubnetId, subnetDiscoveryTagKey)
+		} else {
+			// Secondary subnets require explicit opt-in via CNI tag
+			log.Debugf("Subnet %s has no %s tag, excluding it from ENI creation", *subnet.SubnetId, subnetDiscoveryTagKey)
+			return false
 		}
 	}
+
+	// Rule 3: Check cluster-specific tags
+	if !hasClusterTags || belongsToThisCluster {
```
Pavani-Panakanti: curious if someone can add multiple cluster tags to same subnet ?

pkg/awsutils/awsutils.go
```diff
@@ -2270,3 +2591,35 @@ func checkAPIErrorAndBroadcastEvent(err error, api string) {
 		}
 	}
 }
+
+// IsSubnetExcluded checks if a subnet is excluded by examining its kubernetes.io/role/cni tag
+func (cache *EC2InstanceMetadataCache) IsSubnetExcluded(ctx context.Context, subnetID string) (bool, error) {
```
Pavani-Panakanti, line 2625: should IsSubnetExcluded() check for cluster ownership as well similar to validTag()?

pkg/ipamd/ipamd.go
```diff
@@ -1262,14 +1269,43 @@ func (c *IPAMContext) setupENI(eni string, eniMetadata awsutils.ENIMetadata, isT
 	if err != nil && err.Error() != datastore.DuplicatedENIError {
 		return errors.Wrapf(err, "failed to add ENI %s to data store", eni)
 	}
-	// Keep track of the primary IP for this ENI
-	c.primaryIP[eni] = primaryIP
+
+	// Check if this ENI (primary or secondary) is in an excluded subnet and mark it for exclusion
+	if c.useSubnetDiscovery {
+		subnetID, err := c.awsClient.GetENISubnetID(ctx, eni)
+		if err != nil {
```
Pavani-Panakanti: In case of err we will miss the subnet exclusion if specified by user ? should we throw error ?

pkg/ipamd/ipamd.go
```diff
@@ -1262,14 +1269,43 @@ func (c *IPAMContext) setupENI(eni string, eniMetadata awsutils.ENIMetadata, isT
...
+			log.Warnf("setupENI: failed to get subnet ID for ENI %s: %v", eni, err)
+		} else {
+			excluded, err := c.awsClient.IsSubnetExcluded(ctx, subnetID)
+			if err != nil {
```
Pavani-Panakanti: same here

pkg/ipamd/ipamd.go
```diff
@@ -2440,12 +2480,67 @@ func (c *IPAMContext) tryUnassignPrefixFromENI(eniID string, networkCard int) {
...
+// cleanupExcludedENI cleans up unassigned secondary IPs and prefixes from excluded ENI
+func (c *IPAMContext) cleanupExcludedENI(ctx context.Context, eniID string) {
+	log.Debugf("cleanupExcludedENI: cleaning up resources from excluded ENI %s", eniID)
+
+	// Clean up based on the current IP mode
+	if c.enableIPv4 {
+		if c.enablePrefixDelegation {
+			// In prefix delegation mode, cleanup unassigned IPv4 prefixes
+			log.Debugf("Cleaning up unassigned IPv4 prefixes from excluded ENI %s", eniID)
+			c.tryUnassignPrefixFromENI(ctx, eniID, DefaultNetworkCardIndex)
```
Pavani-Panakanti, line 2507: We are iterating over all datastores and network cards in "checkAndHandleAllENIExclusion", then why are we defaulting to DefaultNetworkCardIndex here ? That should be passed from datastore to cleanupExcludeENI ?
Is this feature supported with multi NIC and was it tested ?

pkg/ipamd/ipamd.go
```diff
@@ -2440,12 +2480,67 @@ func (c *IPAMContext) tryUnassignPrefixFromENI(eniID string, networkCard int) {
...
+// checkAndHandleAllENIExclusion checks all existing ENIs (primary and secondary) across all network cards for subnet exclusion
+func (c *IPAMContext) checkAndHandleAllENIExclusion(ctx context.Context) {
+	// Iterate over all datastores (one per network card)
+	for _, ds := range c.dataStoreAccess.DataStores {
+		// Get ENI information using the public method
+		eniInfos := ds.GetENIInfos()
+		for eniID, eni := range eniInfos.ENIs {
+
+			// Only process ENIs that are already marked as excluded
+			if !eni.IsExcludedForPodIPs {
```
Pavani-Panakanti, line 2533: if someone adds exclusion tag to subnet after eni is created and CNI is up and running..we will not be handling it ? I see this is "sExcludedForPodIPs" setup only once at startup. So even if marked for exclusion later we will continue to use for pod ips till aws-node is restarted. Is this understanding right and expected behavior ?

pkg/ipamd/datastore/data_store.go
```diff
@@ -480,6 +491,38 @@ func (ds *DataStore) AddENI(eniID string, deviceNumber int, isPrimary, isTrunk,
...
+	return nil
+}
+
+// IsENIExcludedForPodIPs returns whether an ENI is excluded from pod IP allocation
+func (ds *DataStore) IsENIExcludedForPodIPs(eniID string) bool {
+	ds.lock.Lock()
+	defer ds.lock.Unlock()
+
+	eni, ok := ds.eniPool[eniID]
+	if !ok {
```
Pavani-Panakanti, line 522: should we return err here instead ?

haouc:
> What is the rollback behavior to previous versions which won't have this feature ?
> 
>     * Will there be any leaked ENI's that were created in secondary subnets as older version of CNI will not be aware of these ?
> 
>     * Will there be any leaked pod ips which were assigned from secondary subnet ENI's ?

New ENIs have the same tags as today's ENIs and will be cleaned up by sweepers in case of leaking. Rollback will need downgrading the version and recycle nodes.

pkg/awsutils/awsutils.go
haouc: I think this is following the same pattern calling the GetENIInfos() and iterating through keys.

pkg/awsutils/awsutils.go
haouc: I think this makes sense. Let me update the structs and support subnetID field.

pkg/awsutils/awsutils.go
haouc: Yes, makes sense.

pkg/awsutils/awsutils.go
haouc: yeah, I agree. We should fail the ENI creation and retry with reconciliation.

pkg/awsutils/awsutils.go
haouc: Makes sense. The SGs should be preset and overridden if needed.

pkg/awsutils/awsutils.go
haouc: yeah, it should be supported. We could in future to support values. But as other practices, we only look up keys.

pkg/awsutils/awsutils.go
haouc: Yes, refactored the logic.

pkg/awsutils/awsutils.go
haouc: Updated the PR to address the issue.

pkg/ipamd/ipamd.go
haouc: After the refactor, the API call is no longer needed.

pkg/ipamd/ipamd.go
haouc: Makes sense, updated them.

pkg/ipamd/ipamd.go
haouc: I don't think we want to support multi-cards for now. Let me update and check with folks offline.

pkg/ipamd/ipamd.go
haouc: Yes. We don't support "backwards" to update existing ENIs if this is your question. Only new ENIs should be applied with updated subnets.

haouc: Where a ENI can be created need to be decided at the moment that the ENI is created.

pkg/ipamd/datastore/data_store.go
haouc: This is looking up a key/value from a map.

pkg/awsutils/awsutils.go
haouc: Make sense. We shouldn't update to primary on error. Updated the PR for change.

pkg/ipamd/ipamd.go
haouc: After rethinking this, I think we have to support multi-card to avoid different behavior across cards.

pkg/awsutils/awsutils.go
```diff
@@ -2270,3 +2612,36 @@ func checkAPIErrorAndBroadcastEvent(err error, api string) {
...
+	for _, subnet := range subnets {
+		if *subnet.SubnetId == subnetID {
+			validForCluster := ValidSubnetForCluster(subnet, isPrimarySubnet)
+			// Check if the subnet has the exclusion tag kubernetes.io/role/cni=0
+			for _, tag := range subnet.Tags {
+				if *tag.Key == "kubernetes.io/role/cni" {
+					tagValue := *tag.Value
+					tagExcluded := tagValue == "0"
+					log.Debugf("IsSubnetExcluded: subnet %s has tag kubernetes.io/role/cni=%s, excluded=%t, clusterValid=%t", subnetID, tagValue, tagExcluded, validForCluster)
+					return tagExcluded && validForCluster, nil
```
Pavani-Panakanti: if subnet is not validForCluster we should return true and exclude it irrespective of whatever tag is present right ? Logic seems incorrect. Can you check ?

Pavani-Panakanti:
Added a comment, rest of the changes looks good to me

yash97:
can you also refer to this code coverage report https://app.codecov.io/github/aws/amazon-vpc-cni-k8s/pull/3463. I think in awsutils.go there many missed lines. If you have added them do consider adding unit test to them. Lets make sure we atleast cover critical or tricky parts.

pkg/awsutils/awsutils.go
haouc: makes sense. I have updated the code. Thanks

Pavani-Panakanti:
Thanks Hao for the changes !!

Issue #3351, opened 2025-07-15
Failed to assign an IP address to container


maxsxu:
**What happened**:

Pod event
```
"aws-cni" failed (add): add cmd: failed to assign an IP address to container
```

**Attach logs**

```
{"level":"debug","ts":"2025-07-15T09:32:42.904Z","caller":"rpc/rpc.pb.go:1105","msg":"AddNetworkRequest: K8S_POD_NAME:\"myapp-7dbfd86df4-gsnsn\" K8S_POD_NAMESPACE:\"streamnative\" K8S_POD_INFRA_CONTAINER_ID:\"23296d3b8d9673814fd2039c316826aa531272bccb6d04a4ef6d910e20640b23\" ContainerID:\"23296d3b8d9673814fd2039c316826aa531272bccb6d04a4ef6d910e20640b23\" IfName:\"eth0\" NetworkName:\"aws-cni\" Netns:\"/var/run/netns/cni-b7d93dd1-57f9-6c58-4f56-f70ae8b09678\""}
{"level":"debug","ts":"2025-07-15T09:32:42.904Z","caller":"datastore/data_store.go:607","msg":"AssignPodIPv4Address: IP address pool stats: total 64, assigned 64"}
{"level":"debug","ts":"2025-07-15T09:32:42.904Z","caller":"datastore/data_store.go:687","msg":"Get free IP from prefix failed no free IP available in the prefix - 10.50.62.224/fffffff0"}
{"level":"debug","ts":"2025-07-15T09:32:42.904Z","caller":"datastore/data_store.go:607","msg":"Unable to get IP address from CIDR: no free IP available in the prefix - 10.50.62.224/fffffff0"}
{"level":"debug","ts":"2025-07-15T09:32:42.904Z","caller":"datastore/data_store.go:687","msg":"Get free IP from prefix failed no free IP available in the prefix - 10.50.59.144/fffffff0"}
{"level":"debug","ts":"2025-07-15T09:32:42.904Z","caller":"datastore/data_store.go:607","msg":"Unable to get IP address from CIDR: no free IP available in the prefix - 10.50.59.144/fffffff0"}
{"level":"debug","ts":"2025-07-15T09:32:42.904Z","caller":"datastore/data_store.go:687","msg":"Get free IP from prefix failed no free IP available in the prefix - 10.50.60.96/fffffff0"}
[...]

PR #3523, opened 2025-11-25
Adding function/logging for subnet CIDR fragmentation visibility

cdirubbio:
**What type of PR is this?** improvement

**Which issue does this PR fix?**: Providing visibility into Subnet CIDR Fragmentation https://github.com/aws/amazon-vpc-cni-k8s/issues/3351

**What does this PR do / Why do we need it?**: Creates functions to:
1. Retrive subnet by ENI ID 
2. Use that subnet to gather the already assigned prefixes and IPs from ENIs in subnet
3. Calculate free prefixes based on already assigned IPs
4. Log number of available prefixes in subnet

**Testing done on this change**: I've tested this in my cluster and the additional logs looks as follows:
```
{"level":"info","ts":"2025-11-24T23:27:05.902Z","caller":"ipamd/rpc_handler.go:232","msg":"Subnet prefix availability: 2 free /28 prefixes in subnet subnet-abcdefgxyz (ENI: eni-xyzabcdefg)"}
```
```
{"level":"info","ts":"2025-11-25T00:00:25.204Z","caller":"ipamd/rpc_handler.go:232","msg":"Subnet prefix availability: 0 free /28 prefixes in subnet subnet-00cfc91c7b6f8cd14 (ENI: eni-03c6cc3376289525b)"}
```

**Will this break upgrades or downgrades? Has updating a running cluster been tested?**: Yes, updating a running cluster has been tested. **However, my main worry is that this will probably break a large cluster.** Not the implementation itself, but its impact on the EC2 API. In order to calculate free /28 prefixes, we need to retrieve all ENIs in the Subnet, then extract the assigned IPs and prefixes and from there retrieve the free prefixes.

If we have a cluster that is large, say 500 nodes, and there are many ip address assignment failures, this will inevitably lead to EC2 API throttling.
- https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeNetworkInterfaces.html
- https://docs.aws.amazon.com/ec2/latest/devguide/ec2-api-throttling.html#:~:text=DescribeNetworkInterfaces

A better design here would likely be a controller in the Control Plane that is doing these API calls and calculations. That way, the number of API calls is limited and results per subnet can be cached/used across instances, rather than making the same call x number of times for all ENIs in the same subnet (possibly across many instances)...

```release-note
Added additional logging and functions to IPAMd to log /28 prefix availability in subnet when PD is enabled for IPv4 Clusters
```

### DIFF
--- a/pkg/awsutils/awsutils_test.go
+++ b/pkg/awsutils/awsutils_test.go
@@ -2392,9 +2392,9 @@ func TestCheckSubnetPrefixAvailability(t *testing.T) {
 			expectedError: true,
 		},
 		{
-			name:      "describe subnets error",
-			subnetID:  "subnet-12345",
-			subnetErr: errors.New("permission denied"),
+			name:          "describe subnets error",
+			subnetID:      "subnet-12345",
+			subnetErr:     errors.New("permission denied"),
 			expectedError: true,
 		},
 		{
@@ -2419,7 +2419,7 @@ func TestCheckSubnetPrefixAvailability(t *testing.T) {
 				mockEC2.EXPECT().DescribeSubnets(gomock.Any(), gomock.Any()).Return(nil, tt.subnetErr)
 			} else if tt.subnetResp != nil {
 				mockEC2.EXPECT().DescribeSubnets(gomock.Any(), gomock.Any()).Return(tt.subnetResp, nil)
-				
+
 				if tt.eniErr != nil {
 					mockEC2.EXPECT().DescribeNetworkInterfaces(gomock.Any(), gomock.Any()).Return(nil, tt.eniErr)
 				} else if tt.eniResp != nil {
@@ -2511,10 +2511,10 @@ func TestCountFreePrefixes(t *testing.T) {
 		expected    int
 	}{
 		{
-			name:       "/24 subnet with no assigned IPs",
-			subnetCIDR: "10.0.1.0/24",
+			name:        "/24 subnet with no assigned IPs",
+			subnetCIDR:  "10.0.1.0/24",
 			assignedIPs: map[uint32]bool{},
-			expected:   14, // 16 total /28 prefixes minus reserved
+			expected:    14, // 16 total /28 prefixes minus reserved
 		},
 		{
 			name:       "/24 subnet with assigned IPs in first prefix",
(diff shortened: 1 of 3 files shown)


cdirubbio:
Put together what I was talking about ^ The end _flow_ i would want to be something like this:
1. When an ENI attached to Node by VPC CNI, the VPC CNI will check if there is a SubnetInformation object for the subnet for which the ENI was just created in
2a. If yes, do nothing
2b. If no object, create it and leave the IP address/prefix fields empty.
3. The component in control plane (vpc-rc or separate subnet-monitor), is reviewing the `SubnetInformation` custom resources every 30 seconds
4. It then will do calculation on available prefixes, and return available ipv4 addresses from EC2 and write results to the `subnetInformation` Custom resource for each subnet
```
➜ kg subnetinformation.monitoring.aws.k8s.io 
NAME                       AGE
subnet-00cfc91c7b6f8cd14   16m
..
subnet-xyzbasdfhas   16m

➜  kd subnetinformation subnet-00cfc91c7b6f8cd14 
Name:         subnet-00cfc91c7b6f8cd14
Namespace:    
Labels:       <none>
Annotations:  <none>
API Version:  monitoring.aws.k8s.io/v1alpha1
Kind:         SubnetInformation
Metadata:
  Creation Timestamp:  2025-11-25T19:11:58Z
  Generation:          1
  Resource Version:    75599910
  UID:                 5f876f5d-663d-4699-a5f6-54440aa4b090
Spec:
  Availability Zone:  us-east-1c
  Cidr Block:         172.31.254.0/24
  ipv4AddressesFree:  220
  ipv4PrefixesFree:   3
Status:
Events:  <none>
```

jaydeokar:
Have you checked adding this as a part of cni-metrics-helper ? The risk here is that critical functions will be impacted on EC2 API throttling, which you already mention. 
Control plane integration seems like a good idea too, but not sure we want to go that approach right now

cdirubbio:
> Have you checked adding this as a part of cni-metrics-helper ? The risk here is that critical functions will be impacted on EC2 API throttling, which you already mention. Control plane integration seems like a good idea too, but not sure we want to go that approach right now

Hadn't thought of putting it in the `cni-metrics-helper`, but that could be an option... could be easily added as a second container in that deployment too, with an option in the Helm Chart to add the `subnet-monitor` container and CRD or not.

PR #3565, opened 2026-01-08
Created GetIptablesMode to log legacy or nftables backend

cdirubbio:
**What type of PR is this?** improvement

**What does this PR do / Why do we need it?**: `/usr/sbin/iptables-wrapper` picks the iptables backend used based on a few conditions. With this change, a method is introduced to get the iptables mode (either nf_tables or legacy) in-use.

The package `go-iptables` that the vpc cni uses to interact with iptables is no longer being maintained, so this is the current solution. Following this change, we get the following log line during startup, prior to creation and insertion of iptables rules.

```
{"level":"info","ts":"2026-01-08T00:37:15.024Z","caller":"networkutils/network.go:377","msg":"Using iptables mode (nf_tables)"}
```

```release-note
Add GetIptablesMode() function to log iptables backend in-use
```

### DIFF
--- a/pkg/iptableswrapper/iptables.go
+++ b/pkg/iptableswrapper/iptables.go
@@ -115,10 +115,13 @@ func (i ipTables) HasRandomFully() bool {
 	return i.ipt.HasRandomFully()
 }
 
+
+var iptablesModeExecCommand = exec.Command
+
 // Runs "iptables --version" to get the mode  (nf_tables or legacy)
 func GetIptablesMode() (string, error) {
 	path := "iptables"
-	cmd := exec.Command(path, "--version")
+	cmd := iptablesModeExecCommand(path, "--version")
 	var out bytes.Buffer
 	cmd.Stdout = &out
 	err := cmd.Run()
--- a/pkg/iptableswrapper/iptables.go
+++ b/pkg/iptableswrapper/iptables.go
@@ -14,7 +14,13 @@
 // Package iptableswrapper is a wrapper interface for the iptables package
 package iptableswrapper
 
-import "github.com/coreos/go-iptables/iptables"
+import (
+	"bytes"
+	"fmt"
+	"github.com/coreos/go-iptables/iptables"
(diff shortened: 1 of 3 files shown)


yash97:
hey @cdirubbio  can we add any kind of test for this? This will be helpful addition.

yash97:
I would suggest to use this code https://github.com/kubernetes-sigs/iptables-wrappers/blob/master/internal/iptables/detect.go#L48. It is internal so we cannot use this directly.

cdirubbio:
Added unit tests and this method is essentially just running `iptables --version` and extracting the mode. All versions of the CNI have the iptables` binary since `go-iptables` is using the binary under the hood, so we dont need to worry about it not having the binary for now.

https://github.com/coreos/go-iptables/blob/main/iptables/iptables.go#L654

cdirubbio:
Will run go fmt to fix the CI issue. Here's the manual testing:
```
# BottleRocket v1.33 AMI (iptables-legacy)
cat /host/var/log/aws-routed-eni/ipamd.log | grep "mode" 
{"level":"info","ts":"2026-02-18T20:51:57.899Z","caller":"networkutils/network.go:377","msg":"Using iptables mode (legacy)"}

# AL2023 v1.33 AMI (iptables-nft)
cat /host/var/log/aws-routed-eni/ipamd.log | grep "mode"
{"level":"info","ts":"2026-02-18T20:51:52.628Z","caller":"networkutils/network.go:377","msg":"Using iptables mode (nf_tables)"}
```

cdirubbio:
> I would suggest to use this code https://github.com/kubernetes-sigs/iptables-wrappers/blob/master/internal/iptables/detect.go#L48. It is internal so we cannot use this directly.

I chose to use the [code from the private method in `go-iptables`](https://github.com/coreos/go-iptables/blob/main/iptables/iptables.go#L654) since there's no need to run the detection logic again, just need to check what was picked by iptables-wrapper already

PR #3568, opened 2026-01-10
fix: decrement unmanagedENI counter when ENIs are removed

### LINKED ISSUE #3544: Issue with count of unmanaged ENIs
**What happened**:

```
  Warning  FailedCreatePodSandBox  2m6s (x62 over 15m)  kubelet            (combined from similar events): Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "f44b32fc51168ef2352e384a3d853da635c5fa0b469121f1aede7a924004b870": plugin type="aws-cni" name="aws-cni" failed (add): add cmd: failed to assign an IP address to container
```

This code notices unmanaged ENIs and increments the count, but never decrements it: https://github.com/aws/amazon-vpc-cni-k8s/blob/a2c18c1cb0ff32c36157b6dae4aacc5d5f243b3d/pkg/ipamd/ipamd.go#L2104

In our workflow we attach and detach unmanaged ENIs from our nodes for during some operations. Doing so can result in the aws-cni eventually failing to assign IPs to containers.

**Attach logs**

Detaching an unmanaged ENI should decrement the count.

**Environment**:
- Kubernetes version (use `kubectl version`):
- CNI Version
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):

OlTrenin:
  Fixes #3544

  This PR fixes a bug where the `unmanagedENI` counter was incremented when unmanaged ENIs were filtered in `filterUnmanagedENIs()`, but never decremented when those ENIs were removed from the system.

  This caused the counter to grow indefinitely when users attach and detach unmanaged ENIs during node operations. Eventually, the incorrect counter value prevents the IPAMD from correctly calculating available ENI slots, leading to pod sandbox creation failures with error: "failed to assign an IP address to container".

  The fix adds decrement logic in two locations where ENIs are removed:

  1. **`tryFreeENI()` (pkg/ipamd/ipamd.go:865-870)** - Decrements counter when an ENI is freed/detached during IP pool decrease operations
  2. **`nodeIPPoolReconcile()` (pkg/ipamd/ipamd.go:1682-1687)** - Decrements counter when a detached ENI is removed during reconciliation sweep phase

  Both implementations:
  - Check if the ENI is unmanaged (via `IsUnmanagedENI()`) or EFA-only (via `IsEfaOnlyENI()`) before removal
  - Only decrement if counter > 0 to prevent underflow (defensive programming)
  - Add debug logging for tracking counter changes

  - Code compiles successfully with `GOOS=linux GOARCH=amd64 go build ./pkg/ipamd/...`
  - Passes `go vet ./pkg/ipamd/...` checks without errors
  - Verified the decrement logic mirrors the increment logic in `filterUnmanagedENIs()` (lines 2104, 2112, 2119, 2127)
  - Confirmed the fix addresses both code paths where ENIs are removed from the system
  - Code follows existing patterns and style in the codebase

  No. This change only uses existing methods:
  - `awsClient.IsUnmanagedENI(eni)` - already used in filterUnmanagedENIs
  - `awsClient.IsEfaOnlyENI(networkCard, eni)` - already used in filterUnmanagedENIs

  No new EC2/K8s APIs, IMDS calls, or external dependencies are introduced.

  No breaking changes. This is a backward-compatible bug fix that only affects internal counter logic.

  - **Upgrades**: The counter will start being decremented correctly after upgrade. Any accumulated incorrect counter values will naturally correct themselves as ENIs are removed.
  - **Downgrades**: If downgraded, the system returns to the buggy behavior, but no data corruption or crashes will occur.

  The change maintains the same external behavior but fixes the counter accuracy, which improves ENI capacity calculations.

  No. This fix works with a simple "kubectl patch" of the image tag. No configuration changes are required.

  ```release-note
  Fixed a bug where the unmanagedENI counter was not decremented when unmanaged ENIs were removed, causing incorrect ENI capacity calculations and eventual pod creation failures with "failed to assign an IP address to container" error in environments with frequent unmanaged ENI attach/detach operations.

  EOF

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -846,6 +846,11 @@ func (c *IPAMContext) tryFreeENI(networkCard int) {
 		return
 	}
 
+	// Check if this ENI was unmanaged before freeing it
+	isUnmanagedENI := c.awsClient.IsUnmanagedENI(eni)
+	isEfaOnlyENI := c.awsClient.IsEfaOnlyENI(networkCard, eni)
+	wasUnmanaged := isUnmanagedENI || isEfaOnlyENI
+
 	log.Debugf("Start freeing ENI %s from network card %d", eni, networkCard)
 	err := c.awsClient.FreeENI(eni)
 	if err != nil {
@@ -855,6 +860,13 @@ func (c *IPAMContext) tryFreeENI(networkCard int) {
 		if errors.Is(err, awsutils.ErrENIAttachmentIdNotFound) || errors.Is(err, awsutils.ErrUnableToDetachENI) {
 			return
 		}
+	}
+
+	// Decrement the counter if this ENI was unmanaged
+	if wasUnmanaged && c.unmanagedENI[networkCard] > 0 {
+		c.unmanagedENI[networkCard] -= 1
+		log.Debugf("Decremented unmanagedENI counter for network card %d to %d after freeing ENI %s",
+			networkCard, c.unmanagedENI[networkCard], eni)
 	}
 
 	err = c.networkClient.DeleteRulesBySrc(c.primaryIP[eni], c.enableIPv6)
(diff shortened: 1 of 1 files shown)


OlTrenin:
@yash97 hi, what about review?

pkg/ipamd/ipamd.go
```diff
@@ -857,6 +862,13 @@ func (c *IPAMContext) tryFreeENI(networkCard int) {
 		}
 	}
 
+	// Decrement the counter if this ENI was unmanaged
+	if wasUnmanaged && c.unmanagedENI[networkCard] > 0 {
```
jaydeokar, line 866: If the ENI was marked unmanaged, it will not be added to the datastore, so this is dead code and will never be enforced

OlTrenin: @jaydeokar hey, I remove this

jaydeokar:
Closing as this is not required

Issue #3563, opened 2026-01-02
Service CIDR must be manually configured when using kube-proxy in nftables mode with VPC CNI iptables-nft


shaikatzz:
### What happened?

When running kube-proxy in **native nftables mode** while VPC CNI uses **iptables-nft**, pods on secondary ENIs cannot reach Kubernetes services. Traffic gets incorrectly marked with `0x80` and routed via eth0 instead of the correct secondary interface.

### Root Cause Analysis

Using `nft monitor trace`, I identified the issue:

1. **MANGLE PREROUTING** (priority -150) runs **before** kube-proxy's NAT DNAT
2. VPC CNI's CONNMARK `--restore-mark` rule in mangle restores mark `0x80` from conntrack
3. At this point, destination is still the **ClusterIP** (not yet DNAT'd to pod IP)
4. The routing decision uses fwmark `0x80` → routes via main table → eth0
5. kube-proxy DNAT happens **after** the routing decision

**Trace evidence:**
```
ip mangle PREROUTING rule iifname "eni*" xt match "comment" counter packets 3325625 bytes 644878056 xt target "CONNMARK" (verdict continue)
ip mangle PREROUTING policy accept meta mark 0x00000080   ← MARK SET BEFORE DNAT
...
ip mangle FORWARD packet: iif "eni1caf450b99a" oif "eth0" ... ip daddr 10.71.20.123  ← WRONG INTERFACE
```

The packet enters with `dst=172.20.96.189` (ClusterIP), gets marked `0x80` in mangle PREROUTING, routing decides to use eth0, then DNAT changes destination to pod IP.

### Why this doesn't happen with iptables mode kube-proxy

In iptables mode, both VPC CNI and kube-proxy use the same `ip nat` table. The rule ordering within a single table is deterministic - kube-proxy's DNAT rules in PREROUTING run before VPC CNI's CONNMARK chain checks the destination.

In nftables mode, kube-proxy creates rules in the `ip kube-proxy` table while VPC CNI (via iptables-nft) creates rules in the `ip nat` table. The **mangle table** (priority -150) runs before both, and the CONNMARK restore happens there.

### Expected behavior

- Traffic from pods on secondary ENIs to ClusterIP services should be routed via the correct interface (eth1, eth2, etc.)
- Service CIDR should be automatically excluded from CONNMARK, or documentation should clearly state this requirement for nftables mode

### Workaround

Manually set the service CIDR in environment variables:

[...]

PR #3588, opened 2026-02-06
Add prerouting connmark rules through nftable instead of iptables.

yash97:
Solves #3563 

**What type of PR is this?**

improvement

#3563 
### Summary
This PR introduces a Connmark interface that abstracts connmark rule management, with implementations for both iptables-legacy and nftables backends. The VPC CNI now auto-detects the node's iptables mode and uses the appropriate backend.

### Changes

Core Implementation:

- Connmark interface with Setup() and Cleanup() methods

- nftConnmark for nftables backend using new nft.Client wrapper

- iptablesConnmark for iptables-legacy backend

- GetIptablesMode() to auto-detect iptables backend (legacy vs nf_tables)

- fib daddr type local return rule to skip connmark processing for local traffic (fixes IPVS compatibility)

### nftables Implementation:

- Create aws-cni table with nat-prerouting and snat-mark chains

- Idempotent rule reconciliation - adds missing rules, removes stale CIDR rules

- One-time cleanup of legacy iptables connmark rules when migrating to nftables

### Testing:

- Integration test snat_kube_proxy_test.go to verify SNAT works with all kube-proxy modes (iptables, nftables, ipvs)

- PodExecInContainer() to test framework for multi-container pod exec

- Test script run-mixed-os-snat-kube-proxy-test.sh for mixed OS testing

### What integration test does?

After switching kube-proxy mode (iptables/nftables/ipvs), verify connmark rules and pod connectivity (service, API server, external) from both primary and secondary ENI pods. Then restart aws-node,  wait for readiness, and re-verify all connectivity to confirm CNI restart causes no disruption to existing pods. 

#### Sample nft list rules, how does it look on host through nft command

```
nft list ruleset

(irrelevant output ommited)

table ip aws-cni {
        chain snat-mark {
                counter packets 9102174 bytes 836926787 ip daddr 172.31.0.0/20 return
                counter packets 9102174 bytes 836926787 ip daddr 192.168.0.0/16 return
                counter packets 0 bytes 0 ct mark set ct mark ^ 0x00000080
        }

        chain nat-prerouting {
                type nat hook prerouting priority dstnat + 10; policy accept;
                fib daddr type local return
                iifname "eni*" counter packets 0 bytes 0 jump snat-mark
                counter packets 0 bytes 0 meta mark set ct mark & 0x00000080
        }
}
```

  ### Observability

  Two Prometheus metrics added to the existing `awscni_*` surface (no new exporter, no new port):
  
  | Metric | Type | Labels | Purpose |
  |---|---|---|---| 
  | `awscni_connmark_backend` | Gauge | `backend` ∈ `{nftables, iptables}` | Set to `1` for the active backend immediately after auto-detection in `NewConnmark`. Lets dashboards filter the fleet by which backend a node picked, without shelling into the host. |
  | `awscni_connmark_reconcile_total` | Counter | `result` ∈ `{success, error}` | Incremented every time the CNI applies (or removes) connmark rules on the node. Counts errors from rule application, cleanup, and the one-time iptables↔nftables migration in a single signal. Both `success` and `error` series are initialized to 0 at startup so  dashboards work immediately.  |

```
curl -s http://192.168.45.7:61678/metrics | grep connmark
# HELP awscni_connmark_backend The connmark backend in use; the active backend's series is set to 1
# TYPE awscni_connmark_backend gauge
awscni_connmark_backend{backend="nftables"} 1
# HELP awscni_connmark_reconcile_total The number of connmark Setup/Cleanup reconciliations by result
# TYPE awscni_connmark_reconcile_total counter
awscni_connmark_reconcile_total{result="error"} 0
awscni_connmark_reconcile_total{result="success"} 1
```

It will not break pod networking however, it will leave nft rules on host. But they will apply same operations applied by iptables-nft. As nft rules will only be executed if iptables-nft is used by cni and kube-proxy.

```release-note

```


pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,521 @@
...
+	return chain
+}
+
+func (c *nftConnmark) addJumpRule(table *nftables.Table, baseChain, targetChain *nftables.Chain, vethPrefix string) {
+	c.nft.AddRule(&nftables.Rule{
+		Table: table,
+		Chain: baseChain,
+		Exprs: []expr.Any{
+			&expr.Meta{Key: expr.MetaKeyIIFNAME, Register: 1},
+			&expr.Cmp{Op: expr.CmpOpEq, Register: 1, Data: []byte(vethPrefix)},
```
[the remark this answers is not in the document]
yash97: not needed, it will append * by default.

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,521 @@
...
+		existing = nil
+	}
+	if existing == nil {
+		priority := c.getDesiredPriority()
+		chain := c.nft.AddChain(&nftables.Chain{
+			Name:     nftBaseChainName,
+			Table:    table,
+			Type:     nftables.ChainTypeNAT,
+			Hooknum:  nftables.ChainHookPrerouting,
+			Priority: &priority,
```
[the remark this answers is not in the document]
yash97: it does that by default, but will explicity add.

test/integration/cni/snat_kube_proxy_test.go
```diff
@@ -0,0 +1,261 @@
...
+		Expect(string(out)).ToNot(ContainSubstring("AWS-CONNMARK"))
+	} else {
+		out, err := execNodeShell(nodeName, "iptables-legacy -t nat -L PREROUTING -n")
+		fmt.Fprintf(GinkgoWriter, "iptables-legacy:\n%s\n", string(out))
+		Expect(err).ToNot(HaveOccurred())
+		Expect(string(out)).To(ContainSubstring("AWS-CONNMARK"))
+
+		_, err = execNodeShell(nodeName, "nft list table ip aws-cni")
+		Expect(err).To(HaveOccurred())
+	}
```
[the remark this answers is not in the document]
yash97: switching iptables mode backend for node in runtime is not scope of this pr.

pkg/networkutils/iptables_connmark.go
```diff
@@ -0,0 +1,148 @@
...
+	jumpRule := []string{"-i", c.vethPrefix + "+", "-m", "comment", "--comment", "AWS, outbound connections", "-j", "AWS-CONNMARK-CHAIN-0"}
+	_ = ipt.Delete("nat", "PREROUTING", jumpRule...)
+
+	// Delete restore rule
+	restoreRule := []string{"-m", "comment", "--comment", "AWS, CONNMARK", "-j", "CONNMARK", "--restore-mark", "--mask", fmt.Sprintf("%#x", c.mark)}
+	_ = ipt.Delete("nat", "PREROUTING", restoreRule...)
+
+	// Clear and delete chain
+	_ = ipt.ClearChain("nat", "AWS-CONNMARK-CHAIN-0")
+	//_ = ipt.DeleteChain("nat", "AWS-CONNMARK-CHAIN-0")
```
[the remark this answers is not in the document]
yash97: We only clear rules under the chain, not delete the chain, according to this old unit test of expectation. https://github.com/aws/amazon-vpc-cni-k8s/blob/v1.22.0/pkg/networkutils/network_test.go#L725. We can delete it too, but keep behavior same as before. Will remove commented line.

pkg/iptableswrapper/iptables.go
```diff
@@ -108,3 +115,37 @@ func (i ipTables) ChainExists(table, chain string) (bool, error) {
...
+const (
+	IptablesModeLegacy IptablesMode = "legacy"
+	IptablesModeNFT    IptablesMode = "nf_tables"
+)
+
+func (m IptablesMode) IsNFTables() bool {
+	return m == IptablesModeNFT
+}
+
+var iptablesVersionRegex = regexp.MustCompile(`v([0-9]+)\.([0-9]+)\.([0-9]+)(?:\s+\((\w+))?`)
```
Pavani-Panakanti: Looks like this will match (nftables - missing ). Any potential issues in future ? Should we include closing brace as well

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+			Policy:   &policy,
+		})
+		return chain
+	}
+	return existing
+}
+
+func (c *nftConnmark) cleanupIptablesConnmarkRules() error {
```
Pavani-Panakanti, line 207: How about moving this to iptables_connmark file ?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+		if err := c.cleanupIptablesConnmarkRules(); err != nil {
+			return errors.Wrap(err, "failed to cleanup legacy iptables connmark rules")
+		}
+		c.cleanupOnce.Store(true)
+	}
+	return nil
+}
+
+func (c *nftConnmark) ensureBaseChain(table *nftables.Table) *nftables.Chain {
+	existing := c.getBaseChain(table)
```
Pavani-Panakanti: why are we not returning error from getBaseChain ? If chain does exist but listchain failed for some reason, we would treat that as not existing by retuning nil and add duplicate chain ?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+				Register: 1,
+				Data:     binaryutil.NativeEndian.PutUint32(2),
+			},
+			&expr.Verdict{Kind: expr.VerdictReturn},
+		},
+	})
+}
+
+func (c *nftConnmark) ensureConnmarkChain(table *nftables.Table) (*nftables.Chain, error) {
+	existing := c.getConnmarkChain(table)
```
Pavani-Panakanti: same here. we should capture the error from getconnmarkchain and handle it ?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+		return err
+	}
+	if err := ipt.DeleteChain("nat", "AWS-CONNMARK-CHAIN-0"); err != nil && !isNotExistError(err) {
+		return err
+	}
+	return nil
+}
+
+// it has two rules, one to match vethprefix and jump to connmark chain, second is to mark packet with conntrack mark
+func (c *nftConnmark) ensureBaseChainRules(table *nftables.Table, baseChain, targetChain *nftables.Chain, vethPrefix string, mark uint32) error {
```
Pavani-Panakanti: why do we need to pass vethprefix and mark to the func here ? these are already accessible via ntfConnmark struct right

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+		nft:         client,
+		vethPrefix:  vethPrefix,
+		mark:        mark,
+		newIptables: iptableswrapper.NewIPTables,
+	}, nil
+}
+
+func (c *nftConnmark) Setup(exemptCIDRs []string) error {
+
+	// idempotent
```
Pavani-Panakanti: explain and add a clear comment

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+			return errors.Wrap(err, "failed to cleanup legacy iptables connmark rules")
+		}
+		c.cleanupOnce.Store(true)
+	}
+	return nil
+}
+
+func (c *nftConnmark) ensureBaseChain(table *nftables.Table) *nftables.Chain {
+	existing := c.getBaseChain(table)
+	if existing != nil && !isBaseChainConfigCorrect(existing, c.getDesiredPriority()) {
```
Pavani-Panakanti: how about renaming isBaseChainConfigCorrect" to "needsBaseChainRecreate" or "isBaseChainUpToDate" - current name seems a bit confusing to me

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+func (c *nftConnmark) Setup(exemptCIDRs []string) error {
+
+	// idempotent
+	table := c.nft.AddTable(&nftables.Table{
+		Family: nftables.TableFamilyIPv4,
+		Name:   nftTableName,
+	})
+
+	baseChain := c.ensureBaseChain(table)
+	connmarkChain, _ := c.ensureConnmarkChain(table)
```
Pavani-Panakanti: why is this func returning error if not being used ?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+		} else if isSetMarkRule(r, mark) {
+			setMarkRule = r
+		} else {
+			unknownRules = append(unknownRules, r)
+		}
+	}
+
+	// Delete unknown rules
+	for _, r := range unknownRules {
+		if err := c.nft.DelRule(r); err != nil {
```
Pavani-Panakanti, line 396: why are we not returning error here ?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+
+	desiredCIDRs := make(map[string]bool)
+	for _, cidr := range exemptCIDRs {
+		desiredCIDRs[cidr] = true
+	}
+
+	// Delete stale CIDRs
+	for cidr, rule := range currentCIDRs {
+		if !desiredCIDRs[cidr] {
+			if err := c.nft.DelRule(rule); err != nil {
```
Pavani-Panakanti, line 409: why not retuning error here ?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+			}
+		}
+	}
+
+	// Insert missing CIDRs (prepends - order doesn't matter for CIDR rules)
+	for cidrStr := range desiredCIDRs {
+		if _, exists := currentCIDRs[cidrStr]; !exists {
+			_, cidr, err := net.ParseCIDR(cidrStr)
+			if err != nil {
+				log.Errorf("failed to insert cidr %s in nf table: %s", cidrStr, err)
```
Pavani-Panakanti: if we fail parsing cidr then we would not add it to rule - why are we not retuning error here ?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+
+	// Delete unknown rules
+	for _, r := range unknownRules {
+		if err := c.nft.DelRule(r); err != nil {
+			log.Errorf("failed to delete unknown rule: %s, id: %d", err, r.Handle)
+		}
+	}
+
+	desiredCIDRs := make(map[string]bool)
+	for _, cidr := range exemptCIDRs {
```
Pavani-Panakanti, line 402: exemptCIDRs should never be empty here -- lets add a check to make sure of that. If its empty will be marked

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+		log.Error(err.Error())
+		return err
+	}
+	err = c.ensureConnmarkChainRules(table, connmarkChain, exemptCIDRs, c.mark)
+	if err != nil {
+		log.Error(err.Error())
+		return err
+	}
+
+	if err := c.nft.Flush(); err != nil {
```
Pavani-Panakanti, line 157: if nothing changes is it okay to do two flush calls every 30s ? any perf impact

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+	if err := ipt.ClearChain("nat", "AWS-CONNMARK-CHAIN-0"); err != nil && !isNotExistError(err) {
+		return err
+	}
+	if err := ipt.DeleteChain("nat", "AWS-CONNMARK-CHAIN-0"); err != nil && !isNotExistError(err) {
+		return err
+	}
+	return nil
+}
+
+// it has two rules, one to match vethprefix and jump to connmark chain, second is to mark packet with conntrack mark
```
Pavani-Panakanti: three rules right ?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+	rules, err := c.nft.GetRules(table, baseChain)
+	if err != nil {
+		log.Error(err.Error())
+		return err
+	}
+	hasJumpRule := false
+	hasRestoreRule := false
+	hasFiblocalReturnRule := false
+
+	for _, rule := range rules {
```
Pavani-Panakanti, line 262: what about ordering of these rules ? what if these 3 rules are present but in wrong order

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+	hasReturn := false
+
+	for _, e := range rule.Exprs {
+		if fib, ok := e.(*expr.Fib); ok && fib.FlagDADDR && fib.ResultADDRTYPE {
+			hasFibAddrType = true
+		}
+		// RTN_LOCAL = 2
+		if cmp, ok := e.(*expr.Cmp); ok && cmp.Op == expr.CmpOpEq && len(cmp.Data) == 4 {
+			val := binaryutil.NativeEndian.Uint32(cmp.Data)
+			if val == 2 { // RTN_LOCAL
```
Pavani-Panakanti: lets move this to constant

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+}
+
+// isFibLocalReturnRule checks for: fib daddr type local return
+func isFibLocalReturnRule(rule *nftables.Rule) bool {
+	hasFibAddrType := false
+	hasCmpLocal := false
+	hasReturn := false
+
+	for _, e := range rule.Exprs {
+		if fib, ok := e.(*expr.Fib); ok && fib.FlagDADDR && fib.ResultADDRTYPE {
```
Pavani-Panakanti, line 307: add a comment explaining this. Readability will not be easy

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+			hasFibAddrType = true
+		}
+		// RTN_LOCAL = 2
+		if cmp, ok := e.(*expr.Cmp); ok && cmp.Op == expr.CmpOpEq && len(cmp.Data) == 4 {
```
Pavani-Panakanti: lets add a comment explaining this

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+			}
+		}
+		if v, ok := e.(*expr.Verdict); ok && v.Kind == expr.VerdictReturn {
+			hasReturn = true
+		}
+	}
+	return hasFibAddrType && hasCmpLocal && hasReturn
+}
+
+// Add this function
```
Pavani-Panakanti: clean up comment

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+func (c *nftConnmark) addFibLocalReturnRule(table *nftables.Table, chain *nftables.Chain) {
+	c.nft.InsertRule(&nftables.Rule{
```
Pavani-Panakanti, line 325: missing error handling ? what if this fails

pkg/networkutils/nftables_connmark.go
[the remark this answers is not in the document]
Pavani-Panakanti: Can we add this as a comment explaining * will be added by default

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+	existing := c.getConnmarkChain(table)
+	if existing != nil {
+		return existing, nil
+	}
+	return c.nft.AddChain(&nftables.Chain{
+		Name:  nftChainName,
+		Table: table,
+	}), nil
+}
+func (c *nftConnmark) ensureConnmarkChainRules(table *nftables.Table, chain *nftables.Chain, exemptCIDRs []string, mark uint32) error {
```
Pavani-Panakanti: mark is already available via nftConnmark struct..why pass again as argument ?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+func (c *nftConnmark) getConnmarkChain(table *nftables.Table) *nftables.Chain {
+	chain, err := c.nft.ListChain(table, nftChainName)
+	if err != nil {
+		log.Error(err.Error())
+		return nil
+	}
+	return chain
+}
+
+func (c *nftConnmark) addJumpRule(table *nftables.Table, baseChain, targetChain *nftables.Chain, vethPrefix string) {
```
Pavani-Panakanti: vethprefix is already available via nftConnmark struct ?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+			&expr.Counter{},
+			&expr.Verdict{
+				Kind:  expr.VerdictJump,
+				Chain: targetChain.Name,
+			},
+		},
+	})
+}
+
+func (c *nftConnmark) addRestoreRule(table *nftables.Table, chain *nftables.Chain, mark uint32) {
```
Pavani-Panakanti: mark available via nftConnmark struct ?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+		if bw, ok := e.(*expr.Bitwise); ok {
+			if bytes.Equal(bw.Xor, markBytes) && bytes.Equal(bw.Mask, []byte{0xff, 0xff, 0xff, 0xff}) {
+				hasBitwise = true
+			}
+		}
+	}
+	return hasCtLoad && hasBitwise && hasCtStore
+}
+
+func (c *nftConnmark) addSetMarkRule(table *nftables.Table, chain *nftables.Chain, mark uint32) {
```
Pavani-Panakanti: same mark not required as func argument..available via nftconnmark struct

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+	if err != nil {
+		log.Error(err.Error())
+		return nil
+	}
+	return chain
+}
+
+func (c *nftConnmark) getDesiredPriority() nftables.ChainPriority {
+	const priority nftables.ChainPriority = -90
+	return nftables.ChainPriority(priority)
```
Pavani-Panakanti: we are already initializing it as nftables.ChainPriority whats the point of casting it again ?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+	chain, err := c.nft.ListChain(table, nftBaseChainName)
+	if err != nil {
+		log.Error(err.Error())
+		return nil
+	}
+	return chain
+}
+
+func (c *nftConnmark) getDesiredPriority() nftables.ChainPriority {
+	const priority nftables.ChainPriority = -90
```
Pavani-Panakanti: It does not make sense for this to be a func. Lets make it a package level constant ? Lets also add a comment explaining why -90 here (kube proxy explanation)

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+	return nftables.ChainPriority(priority)
+}
+
+func isBaseChainConfigCorrect(chain *nftables.Chain, desiredPriority nftables.ChainPriority) bool {
+	return chain.Hooknum == nftables.ChainHookPrerouting &&
+		chain.Priority != nil && *chain.Priority == desiredPriority &&
+		chain.Policy != nil && *chain.Policy == nftables.ChainPolicyAccept
```
Pavani-Panakanti: should we also check chain type ?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+			&expr.Bitwise{SourceRegister: 1, DestRegister: 1, Len: 4, Mask: markBytes, Xor: []byte{0, 0, 0, 0}}, // AND with mask
+			&expr.Meta{Key: expr.MetaKeyMARK, SourceRegister: true, Register: 1},                                // store to fwmark
+		},
+	})
+}
+
+func isJumpRule(rule *nftables.Rule, targetChain, vethPrefix string) bool {
+	// check if rule is a jump rule to connmark chain
+	// return true if it is
+	// return false if it is not
```
Pavani-Panakanti: clean up comments

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+	hasIFaceMatch := false
+	hasJump := false
+	hasCounter := false
+
+	for _, e := range rule.Exprs {
+		if cmp, ok := e.(*expr.Cmp); ok && bytes.Equal(cmp.Data, []byte(vethPrefix+"*\x00")) {
```
Pavani-Panakanti: when we compare this with addJumpRule we are not validating other fields other than data - register, MetaKeyIFNAME, op ?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+	}
+	return hasIFaceMatch && hasJump && hasCounter
+}
+
+func isRestoreRule(rule *nftables.Rule, mark uint32) bool {
+	hasCounter := false
+	hasCtLoad := false
+	hasBitwise := false
+	hasMetaStore := false
+	markBytes := []byte{byte(mark), byte(mark >> 8), byte(mark >> 16), byte(mark >> 24)}
```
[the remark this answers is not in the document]
Pavani-Panakanti: this should be addressed

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+	for _, e := range rule.Exprs {
+		if _, ok := e.(*expr.Counter); ok {
+			hasCounter = true
+		}
+		if ct, ok := e.(*expr.Ct); ok && ct.Key == expr.CtKeyMARK && !ct.SourceRegister {
+			hasCtLoad = true
+		}
+		if bw, ok := e.(*expr.Bitwise); ok && bytes.Equal(bw.Mask, markBytes) {
```
Pavani-Panakanti: should we validate xor bytes too ?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+
+	for _, e := range rule.Exprs {
+		if ct, ok := e.(*expr.Ct); ok && ct.Key == expr.CtKeyMARK {
+			if ct.SourceRegister {
+				hasCtStore = true
+			} else {
+				hasCtLoad = true
+			}
+		}
+		// ct mark | 0x80 uses Mask=0xFFFFFFFF, Xor=mark
```
Pavani-Panakanti: comment says "or" but code says XOR ?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+
+func (c *nftConnmark) addSetMarkRule(table *nftables.Table, chain *nftables.Chain, mark uint32) {
+	markBytes := binaryutil.NativeEndian.PutUint32(mark)
+	c.nft.AddRule(&nftables.Rule{
+		Table: table,
+		Chain: chain,
+		Exprs: []expr.Any{
+			&expr.Counter{},
+			&expr.Ct{Key: expr.CtKeyMARK, Register: 1},
+			&expr.Bitwise{SourceRegister: 1, DestRegister: 1, Len: 4, Mask: []byte{0xff, 0xff, 0xff, 0xff}, Xor: markBytes},
```
Pavani-Panakanti: XOR would do a toggle unsetting the mark if already set. I see iptables code is doing OR https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/networkutils/network.go#L686. We should be doing OR here ?

pkg/networkutils/iptables_connmark.go
```diff
@@ -0,0 +1,147 @@
+// ===== iptables_connmark.go =====
```
Pavani-Panakanti: shpuld this be same as other files with copyright ?

pkg/networkutils/iptables_connmark.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/coreos/go-iptables/iptables"
+)
+
+type iptablesConnmark struct {
+	vethPrefix  string
+	mark        uint32
+	newIptables func(iptables.Protocol) (iptableswrapper.IPTablesIface, error)
+}
+
+var connmarkChainName string = "AWS-CONNMARK-CHAIN-0"
```
Pavani-Panakanti: should be const

pkg/networkutils/iptables_connmark.go
```diff
@@ -0,0 +1,147 @@
...
+var _ Connmark = (*iptablesConnmark)(nil)
+
+func newIptablesConnmark(vethPrefix string, mark uint32) (*iptablesConnmark, error) {
```
Pavani-Panakanti: in nftconnmark we are returning interface and here we are returning iptablesconnmark..can we keep it consistent ?

pkg/networkutils/iptables_connmark.go
```diff
@@ -0,0 +1,147 @@
...
+		mark:        mark,
+		newIptables: iptableswrapper.NewIPTables,
+	}, nil
+}
+
+func (c *iptablesConnmark) Setup(exemptCIDRs []string) error {
+	if len(exemptCIDRs) == 0 {
+		return fmt.Errorf("exemptCIDRs cannot be empty")
+	}
+	ipt, err := c.newIptables(iptables.ProtocolIPv4)
```
Pavani-Panakanti, line 69: this runs every 30 secs..any perf impact ? can we cache it instead

pkg/networkutils/iptables_connmark.go
```diff
@@ -0,0 +1,147 @@
...
+	restoreRule := []string{"-m", "comment", "--comment", "AWS, CONNMARK", "-j", "CONNMARK", "--restore-mark", "--mask", fmt.Sprintf("%#x", c.mark)}
+	_ = ipt.Delete("nat", "PREROUTING", restoreRule...)
+
+	// Clear and delete chain
+	_ = ipt.ClearChain("nat", "AWS-CONNMARK-CHAIN-0")
+
+	return nil
+}
+
+func (c *iptablesConnmark) buildRules(exemptCIDRs []string, ipt iptableswrapper.IPTablesIface) ([]iptablesRule, error) {
```
Pavani-Panakanti, line 149: wdyt about separating chain creation and rule creation similar to nft code you have ? looks cleaner that way

pkg/networkutils/iptables_connmark.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// Delete jump rule
+	jumpRule := []string{"-i", c.vethPrefix + "+", "-m", "comment", "--comment", "AWS, outbound connections", "-j", "AWS-CONNMARK-CHAIN-0"}
+	_ = ipt.Delete("nat", "PREROUTING", jumpRule...)
+
+	// Delete restore rule
+	restoreRule := []string{"-m", "comment", "--comment", "AWS, CONNMARK", "-j", "CONNMARK", "--restore-mark", "--mask", fmt.Sprintf("%#x", c.mark)}
+	_ = ipt.Delete("nat", "PREROUTING", restoreRule...)
+
+	// Clear and delete chain
```
Pavani-Panakanti: where are we deleting the chain ? seems to be missing

pkg/networkutils/iptables_connmark.go
```diff
@@ -0,0 +1,147 @@
...
+
+func (c *iptablesConnmark) Cleanup() error {
+	ipt, err := c.newIptables(iptables.ProtocolIPv4)
+	if err != nil {
+		return err
+	}
+
+	// Delete jump rule
+	jumpRule := []string{"-i", c.vethPrefix + "+", "-m", "comment", "--comment", "AWS, outbound connections", "-j", "AWS-CONNMARK-CHAIN-0"}
+	_ = ipt.Delete("nat", "PREROUTING", jumpRule...)
```
Pavani-Panakanti: why are all delete and clear call errors ignored ?

pkg/networkutils/iptables_connmark.go
```diff
@@ -0,0 +1,147 @@
...
+	rules = append(rules, iptablesRule{
+		name: "legacy connmark", shouldExist: false, table: "nat", chain: "PREROUTING",
+		rule: []string{"-i", c.vethPrefix + "+", "-m", "comment", "--comment", "AWS, outbound connections",
+			"-m", "state", "--state", "NEW", "-j", chain},
+	})
+
+	// Jump rule
+	rules = append(rules, iptablesRule{
+		name: "connmark jump", shouldExist: true, table: "nat", chain: "PREROUTING",
+		rule: []string{"-i", c.vethPrefix + "+", "-m", "comment", "--comment", "AWS, outbound connections", "-j", chain},
```
Pavani-Panakanti: these rule strings are duplicated at multiple places - here, in above clean up func, in cleanupiptablesconnmark in nft. Can we centralize it so if we have to make any changes it can be just one place ?

pkg/networkutils/iptables_connmark.go
```diff
@@ -0,0 +1,147 @@
...
+	})
+
+	// CIDR return rules
+	for _, cidr := range exemptCIDRs {
+		rules = append(rules, iptablesRule{
+			name: chain, shouldExist: true, table: "nat", chain: chain,
+			rule: []string{"-d", cidr, "-m", "comment", "--comment", "AWS CONNMARK CHAIN", "-j", "RETURN"},
+		})
+	}
+	// we might not need to mark this packet for delete.
```
Pavani-Panakanti: comment is confusing and does not explain the intent here..can you explain and also add more clear comment

pkg/networkutils/iptables_connmark.go
```diff
@@ -0,0 +1,147 @@
...
+			if rule.name == "AWS-CONNMARK-CHAIN-0" { // CIDR rules
+				err = ipt.Insert(rule.table, rule.chain, 1, rule.rule...)
+			} else {
+				err = ipt.Append(rule.table, rule.chain, rule.rule...)
+			}
+			if err != nil {
+				return err
+			}
+		} else if exists && !rule.shouldExist {
+			_ = ipt.Delete(rule.table, rule.chain, rule.rule...)
```
Pavani-Panakanti: delete errors ignored

pkg/networkutils/iptables_connmark.go
```diff
@@ -0,0 +1,147 @@
...
+
+func (c *iptablesConnmark) applyRules(rules []iptablesRule, ipt iptableswrapper.IPTablesIface) error {
+	for _, rule := range rules {
+		exists, err := ipt.Exists(rule.table, rule.chain, rule.rule...)
+		if err != nil {
+			return err
+		}
+
+		if !exists && rule.shouldExist {
+			if rule.name == "AWS-CONNMARK-CHAIN-0" { // CIDR rules
```
Pavani-Panakanti: we should use const name here and not string directly..we declared it but not using it anywhere

pkg/networkutils/iptables_connmark.go
```diff
@@ -0,0 +1,147 @@
...
+	_ = ipt.Delete("nat", "PREROUTING", restoreRule...)
+
+	// Clear and delete chain
+	_ = ipt.ClearChain("nat", "AWS-CONNMARK-CHAIN-0")
+
+	return nil
+}
+
+func (c *iptablesConnmark) buildRules(exemptCIDRs []string, ipt iptableswrapper.IPTablesIface) ([]iptablesRule, error) {
+	chain := "AWS-CONNMARK-CHAIN-0"
```
Pavani-Panakanti: already declared at beginning of file

pkg/networkutils/iptables_connmark.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (c *iptablesConnmark) Cleanup() error {
+	ipt, err := c.newIptables(iptables.ProtocolIPv4)
+	if err != nil {
+		return err
+	}
+
+	// Delete jump rule
+	jumpRule := []string{"-i", c.vethPrefix + "+", "-m", "comment", "--comment", "AWS, outbound connections", "-j", "AWS-CONNMARK-CHAIN-0"}
```
Pavani-Panakanti: can we use const declared at beginning of file for chain name ?

pkg/networkutils/iptables_connmark.go
```diff
@@ -0,0 +1,147 @@
...
+	_ = ipt.Delete("nat", "PREROUTING", jumpRule...)
+
+	// Delete restore rule
+	restoreRule := []string{"-m", "comment", "--comment", "AWS, CONNMARK", "-j", "CONNMARK", "--restore-mark", "--mask", fmt.Sprintf("%#x", c.mark)}
+	_ = ipt.Delete("nat", "PREROUTING", restoreRule...)
+
+	// Clear and delete chain
+	_ = ipt.ClearChain("nat", "AWS-CONNMARK-CHAIN-0")
```
Pavani-Panakanti: use const declared at beginning

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+	}
+	if err := ipt.Delete("nat", "PREROUTING", restoreRule...); err != nil && !isNotExistError(err) {
+		return err
+	}
+
+	// Flush and delete AWS-CONNMARK-CHAIN-0 if it exists
+	if err := ipt.ClearChain("nat", "AWS-CONNMARK-CHAIN-0"); err != nil && !isNotExistError(err) {
+		return err
+	}
+	if err := ipt.DeleteChain("nat", "AWS-CONNMARK-CHAIN-0"); err != nil && !isNotExistError(err) {
```
Pavani-Panakanti: lets move chain name to const and use that in all places

pkg/networkutils/network.go
```diff
@@ -157,7 +157,8 @@ type NetworkAPIs interface {
 	// SetupENINetwork performs ENI level network configuration. Not needed on the primary ENI
 	SetupENINetwork(eniIP string, eniMAC string, networkCard int, eniSubnetCIDR string, maxENIPerNIC int, isTrunkENI bool, routeTableID int, isRuleConfigured bool) error
 	// UpdateHostIptablesRules updates the nat table iptables rules on the host
-	UpdateHostIptablesRules(vpcCIDRs []string, primaryMAC string, primaryAddr *net.IP, v6Enabled bool) error
+	//UpdateHostIptablesRules(vpcCIDRs []string, primaryMAC string, primaryAddr *net.IP, v6Enabled bool) error
```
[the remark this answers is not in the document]
Pavani-Panakanti: address this

pkg/networkutils/network.go
```diff
@@ -373,15 +381,14 @@ func (n *linuxNetwork) SetupHostNetwork(vpcCIDRs []string, primaryMAC string, pr
...
-	return n.updateHostIptablesRules(vpcCIDRs, primaryMAC, primaryAddr, v6Enabled)
+	return n.UpdateHostSNATRules(vpcCIDRs, primaryMAC, primaryAddr, v6Enabled)
 }
 
-// UpdateHostIptablesRules updates the NAT table rules based on the VPC CIDRs configuration
-func (n *linuxNetwork) UpdateHostIptablesRules(vpcCIDRs []string, primaryMAC string, primaryAddr *net.IP,
-	v6Enabled bool) error {
-	return n.updateHostIptablesRules(vpcCIDRs, primaryMAC, primaryAddr, v6Enabled)
-}
+// // UpdateHostIptablesRules updates the NAT table rules based on the VPC CIDRs configuration
```
Pavani-Panakanti: cleanup

pkg/networkutils/network.go
```diff
@@ -428,42 +435,43 @@ func (n *linuxNetwork) CleanUpStaleAWSChains(v4Enabled, v6Enabled bool) error {
...
-	}
-
-	ipProtocol := iptables.ProtocolIPv4
 	if v6Enabled {
 		// For v6, we don't add any SNAT or Connmark rules as traffic enters and exits from the ENI it came from
 		return nil
 	}
+	err := n.updateHostIptablesRules(vpcCIDRs, primaryMAC, primaryAddr)
+	if err != nil {
+		log.Error(fmt.Sprintf("failed to iptable rules for snat %s", err.Error()))
```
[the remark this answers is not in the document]
Pavani-Panakanti: address this. log.Errorf can be used directly

Pavani-Panakanti: you can do log.Errorf

pkg/networkutils/iptables_connmark.go
```diff
@@ -0,0 +1,147 @@
...
+
+func newIptablesConnmark(vethPrefix string, mark uint32) (*iptablesConnmark, error) {
+	return &iptablesConnmark{
+		vethPrefix:  vethPrefix,
+		mark:        mark,
+		newIptables: iptableswrapper.NewIPTables,
+	}, nil
+}
+
+func (c *iptablesConnmark) Setup(exemptCIDRs []string) error {
```
Pavani-Panakanti, line 65: what about rollback scenario - cx moved to latest version which uses nftables, had to rollback and now CNI will use iptables but nftables are not cleaned up ?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,528 @@
...
+			hasJumpRule = true
+			continue
+		}
+		if isRestoreRule(rule, mark) {
+			hasRestoreRule = true
+		}
+	}
+
+	if !hasFiblocalReturnRule {
+		c.addFibLocalReturnRule(table, baseChain)
```
Pavani-Panakanti: if fib rule is only for ipvs mode should we add only incase of ipvs mode ? Probably no harm in adding always but want to evaluate once

pkg/networkutils/nftables_connmark.go
Pavani-Panakanti: We have restore rule in iptables mangle same as existing code. Is this restore rule here redundant. Do we need it ?

pkg/networkutils/nftables_connmark_test.go
```diff
@@ -0,0 +1,500 @@
...
+	// Setup expectations
+	mockNft.EXPECT().AddTable(gomock.Any()).Return(table)
+	mockNft.EXPECT().ListChain(table, nftBaseChainName).Return(baseChain, nil)
+	mockNft.EXPECT().ListChain(table, nftChainName).Return(nil, errors.New("not found"))
+	mockNft.EXPECT().AddChain(gomock.Any()).Return(connmarkChain)
+	mockNft.EXPECT().Flush().Return(nil).Times(2)
+	mockNft.EXPECT().GetRules(table, baseChain).Return([]*nftables.Rule{}, nil)
+	mockNft.EXPECT().GetRules(table, connmarkChain).Return([]*nftables.Rule{}, nil)
+	mockNft.EXPECT().InsertRule(gomock.Any()).Return(&nftables.Rule{}).Times(3) // fib rule + 2 CIDRs
+	mockNft.EXPECT().AddRule(gomock.Any()).Return(&nftables.Rule{}).Times(3)    // jump, restore, set mark
```
Pavani-Panakanti, line 72: can we verify the rules as well along with count ?

pkg/networkutils/iptables_connmark_test.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/aws/amazon-vpc-cni-k8s/pkg/iptableswrapper"
+	mock_iptables "github.com/aws/amazon-vpc-cni-k8s/pkg/iptableswrapper/mocks"
+	"github.com/coreos/go-iptables/iptables"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/golang/mock/gomock"
+)
+
+func TestIptablesConnmarkSetup(t *testing.T) {
```
Pavani-Panakanti, line 29: Can you also add test for cleanup func ? probably that would have caught the missing delete chain ?

Pavani-Panakanti:
@yash97 Can you add details on all test cases validated here or in a doc ? 
what about integration tests for these ?
Can you also add godoc comments for all new files and functions

Tycale:
Thanks for this PR 🙏🏻

pkg/iptableswrapper/iptables.go
yash97: I copy pasted code from from here https://github.com/coreos/go-iptables/blob/main/iptables/iptables.go#L624. But now i realized, we do not need this, just substring match is enough.

pkg/networkutils/iptables_connmark.go
yash97: yeah that addition is pending. adding it.

pkg/networkutils/iptables_connmark.go
yash97: the 30 sync loop was there before too, i kept it as it is. Do you mean cache object which implements iptables?

pkg/networkutils/nftables_connmark.go
yash97: flush just executes write buffer commands, our commands are in single digit, and only executed if they differ from expected commands. So do not expect perf impact here.

pkg/networkutils/nftables_connmark.go
yash97: We do not change behavior on kube-proxy mode, as it adds another dependency. Right now we do not see any issue, as address is destined for local we do not snat.

pkg/networkutils/nftables_connmark.go
yash97: both serve different purpose. mangle is for marking packet, this one marks ct too.

pkg/networkutils/nftables_connmark.go
yash97: yeah i wrote this comment before fixing ipvs

pkg/networkutils/nftables_connmark.go
yash97: Adding a change which will collect all parse cidr error  and return at the end but continue add rules for cidr that are correct.

pkg/networkutils/nftables_connmark.go
yash97: yeah if exemptCidr we should do nothing, So i am keeping it as as error in Setup fn itself.

pkg/networkutils/nftables_connmark.go
yash97: yup added that change.

pkg/networkutils/nftables_connmark.go
yash97: this function does not return error. In nft api, only read function returns, All mutation calls are stored in buffer, and executed on flush. That is why we do one time flush at top of the all this function. So we get atomic behavior of all or nothing.

pkg/networkutils/iptables_connmark.go
yash97: if you are saying we just cache applied, rules, we can but those are stale too, better to check by calling iptables.

pkg/networkutils/iptables_connmark.go
yash97: it does, but I would like to reuse code right now. As PR is getting larger and larger.

pkg/networkutils/nftables_connmark_test.go
yash97: Added this test which executes this command on host, pkg/networkutils/nftables_connmark_kernel_test.go. rather than just mocking it.

pkg/networkutils/iptables_connmark.go
yash97: yeah right now just keeping it as it was before, pr is already pretty large.

pkg/networkutils/nftables_connmark.go
yash97: Yes we are are doing XOR, but it does operation like this
new_mark = (old_mark & mask) ^ xor

So we need mask as ^xor, iptables does the same thing

https://linux.die.net/man/8/iptables-1.4.7

> --set-xmark value[/mask]
    Zero out the bits given by mask and XOR value into the ctmark.

pkg/networkutils/nftables_connmark.go
yash97: We collect all errors, and then return as failure to delete this rules should not affect enforcement of desired rules. they do not block each other.

pkg/networkutils/nftables_connmark.go
yash97: same as before We collect all errors, and then return as failure to delete this rules should not affect enforcement of desired rules. they do not block each other.

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,641 @@
...
+			jumpRule = rule
+		case isRestoreRule(rule, c.mark) && restoreRule == nil:
+			restoreRule = rule
+		default:
+			staleRules = append(staleRules, rule)
+		}
+	}
+	allRulesPresent := fibRule != nil && jumpRule != nil && restoreRule != nil
+	ordered := allRulesPresent && fibRule.Handle < jumpRule.Handle && jumpRule.Handle < restoreRule.Handle
+	if allRulesPresent && ordered && len(staleRules) == 0 {
```
Pavani-Panakanti, line 276: There are 3 cases here - if all good, if all good and some additional ones, missing or out of order case --> can you add some debugging 
case 1 would be the idea1 one. For case 2 and 3 - can you add some debug log indicating the flow so that it will be easier for debugging

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,641 @@
...
+		Name:   nftTableName,
+	})
+	if err := c.nft.Flush(); err != nil && !isNFTNotExistsError(err) {
+		return err
+	}
+	return nil
+}
+
+func (c *nftConnmark) getDesiredPriority() nftables.ChainPriority {
+	// This is -90 right now, as default Kubeproxy priority is -100, and we want to run over rules after Kube-proxy DNAT's packet.
```
Pavani-Panakanti: can we add bit more detailed comment which gives context why we want to run this after kubeproxy. It will help for anyone without this context

pkg/networkutils/iptables_connmark.go
```diff
@@ -0,0 +1,231 @@
...
+		if err := c.cleanupNftables(); err != nil {
+			return fmt.Errorf("failed to cleanup nft stale connmark rules: %w", err)
+		}
+		c.cleanupOnce.Store(true)
+	}
+	return nil
+}
+
+func cleanupNFTStaleRules() error {
+	client, err := nft.New()
```
Pavani-Panakanti, line 94: want to confirm nf_tables module always be present in iptables_legacy ? even with custom AMI's

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,641 @@
...
+	"github.com/google/nftables/binaryutil"
+	"github.com/google/nftables/expr"
+)
+
+const (
+	nftTableName     = "aws-cni"
+	nftBaseChainName = "nat-prerouting"
+	nftChainName     = "snat-mark"
+	// https://github.com/torvalds/linux/blob/v7.0/include/uapi/linux/rtnetlink.h#L264
+	rtn_local = uint32(2)
```
Pavani-Panakanti: can we follow same naming convention ? rtnLocal

pkg/networkutils/network.go
```diff
@@ -198,6 +198,11 @@ const (
 
 // New creates a linuxNetwork object
 func New() NetworkAPIs {
+	connmark, err := NewConnmark(getVethPrefixName(), getConnmark())
+	if err != nil {
+		log.Fatalf("Failed to detect iptable mode vpc-cni is using: %v", err)
+	}
```
jaydeokar, line 205: Can we return err instead of exiting with fatal ?

pkg/iptableswrapper/iptables.go
```diff
@@ -108,3 +115,30 @@ func (i ipTables) ChainExists(table, chain string) (bool, error) {
...
+	IptablesModeLegacy IptablesMode = "legacy"
+	IptablesModeNFT    IptablesMode = "nf_tables"
+)
+
+func (m IptablesMode) IsNFTables() bool {
+	return m == IptablesModeNFT
+}
+
+// GetIptablesMode runs "iptables --version" to detect backend (nf_tables or legacy)
+func GetIptablesMode() (IptablesMode, error) {
```
tzneal, line 131: Lets add unit tests for this with captured output from the real iptables

tzneal: What if kube-proxy is still in iptables mode, any issues?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,641 @@
...
+}
+
+func (c *nftConnmark) getDesiredPriority() nftables.ChainPriority {
+	// This is -90 right now, as default Kubeproxy priority is -100, and we want to run over rules after Kube-proxy DNAT's packet.
+	const priority nftables.ChainPriority = -90
+	return priority
+}
+
+func isBaseChainConfigCorrect(chain *nftables.Chain, desiredPriority nftables.ChainPriority) bool {
+	return chain.Hooknum == nftables.ChainHookPrerouting &&
```
tzneal: ChainHookPrerouting and Hooknum are pointers, shouldn't we deref before comparing?

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,641 @@
...
+	return hasCounter && hasCtLoad && hasBitwise && hasMetaStore
+}
+
+func extractCIDRFromRule(rule *nftables.Rule) string {
+	var ip net.IP
+	var mask net.IPMask
+	hasPayload := false
+	hasReturn := false
+
+	for _, e := range rule.Exprs {
```
tzneal, line 573: Can we tighten up the checking on this?

pkg/networkutils/iptables_connmark.go
yash97: For amazon and ubuntu ami it is present, it is not mutually exclusive or dependent on iptables_legacy. As custom AMI we cannot guarantee what ami has or what it will adhere too.

pkg/iptableswrapper/iptables.go
yash97: no issue, this rules are not dependent on kube-proxy mode. I have specifically added a integration test which tests network connectivity across different kube-proxy modes.

test/integration/cni/snat_kube_proxy_test.go
```diff
@@ -0,0 +1,313 @@
...
+		return err
+	}
+	for _, pod := range pods.Items {
+		err := f.K8sResourceManagers.PodManager().DeleteAndWaitTillPodDeleted(&pod)
+		if err != nil {
+			return err
+		}
+	}
+	time.Sleep(30 * time.Second)
+	return nil
```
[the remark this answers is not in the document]
yash97: want to keep it code simple here.

pkg/networkutils/iptables_connmark.go
yash97: but yeah we will not return error if nft not present during iptable legacy cleanup.

pkg/networkutils/nftables_connmark.go
```diff
@@ -0,0 +1,641 @@
...
+}
+
+// addJumpRule adds: nft add rule ip aws-cni nat-prerouting iifname "eni*" counter jump snat-mark
+func (c *nftConnmark) addJumpRule(table *nftables.Table, baseChain, targetChain *nftables.Chain) {
+	c.nft.AddRule(&nftables.Rule{
+		Table: table,
+		Chain: baseChain,
+		Exprs: []expr.Any{
+			&expr.Meta{Key: expr.MetaKeyIIFNAME, Register: 1},
+			&expr.Cmp{Op: expr.CmpOpEq, Register: 1, Data: []byte(c.vethPrefix)},
```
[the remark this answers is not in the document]
yash97: adding comment explaining why current code is correct.

pkg/iptableswrapper/iptables.go
```diff
@@ -108,3 +115,39 @@ func (i ipTables) ChainExists(table, chain string) (bool, error) {
...
+// eg:
+//
+//	iptables v1.8.10 (nf_tables)
+//	iptables v1.8.7 (legacy)
+func parseIptablesMode(versionOutput string) IptablesMode {
+	if strings.Contains(versionOutput, string(IptablesModeLegacy)) {
+		return IptablesModeLegacy
+	}
+	return IptablesModeNFT
+}
```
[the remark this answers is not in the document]
yash97: all new eks ami's defaults to nftable backend.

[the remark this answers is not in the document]
yash97: iptable CLI prints this version since 2018 https://www.netfilter.org/projects/iptables/files/changes-iptables-1.8.0.txt.

test/integration/cni/snat_kube_proxy_test.go
```diff
@@ -0,0 +1,313 @@
...
+		Expect(err).ToNot(HaveOccurred(), "API server connectivity failed from pod %s", pod.Name)
+	}
+}
+
+func verifyExternalConnectivity(times int, pod v1.Pod) {
+	for i := 0; i < times; i++ {
+		cmd := []string{"curl", "-s", "-o", "/dev/null", "-w", "%{http_code}", "--max-time", "5", "http://checkip.amazonaws.com"}
+		stdout, stderr, err := f.K8sResourceManagers.PodManager().PodExec(pod.Namespace, pod.Name, cmd)
+		fmt.Fprintf(GinkgoWriter, "external [%s]: stdout=%s stderr=%s\n", pod.Name, stdout, stderr)
+		Expect(err).ToNot(HaveOccurred(), "external connectivity failed from pod %s", pod.Name)
```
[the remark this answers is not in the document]
yash97: This test is about reachability, not checking health of api server, 4xx,5xx still confirms reachability.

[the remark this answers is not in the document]
yash97: This test is about reachability, not checking health of api server, 4xx,5xx still confirms reachability.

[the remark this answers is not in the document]
yash97: This test is about reachability, not checking health of api server, 4xx,5xx still confirms reachability.

test/integration/cni/snat_kube_proxy_test.go
```diff
@@ -0,0 +1,313 @@
...
+// verifyConnmarkRules checks that CNI connmark rules exist ONLY in the appropriate backend
+func verifyConnmarkRules(nodeName, backend string) {
+	if backend == "nftables" {
+		out, err := execNodeShell(nodeName, "nft list table ip aws-cni")
+		fmt.Fprintf(GinkgoWriter, "nftables rules:\n%s\n", string(out))
+		Expect(err).ToNot(HaveOccurred())
+		Expect(string(out)).To(ContainSubstring("chain nat-prerouting"))
+		Expect(string(out)).To(ContainSubstring("chain snat-mark"))
+
+		out, _ = execNodeShell(nodeName, "iptables-legacy -t nat -L PREROUTING -n")
```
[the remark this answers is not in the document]
yash97: "The error is intentionally discarded — on nftables nodes the binary may not be present, and that's fine.
  Both 'no binary' and 'binary present with no CNI rules' satisfy the invariant being tested: no legacy connmark rules exist.

pkg/networkutils/iptables_connmark.go
Pavani-Panakanti: Can you evaluate this - "nft.Client interface has no Close(). cleanupNFTStaleRules() in iptables_connmark.go opens a netlink fd via nft.New() and never closes it. Add Close() error to the Client interface, defer it at the call site"

pkg/networkutils/nftables_connmark.go
Pavani-Panakanti: need to move to package level constant ?

pkg/networkutils/network.go
```diff
@@ -428,42 +428,60 @@ func (n *linuxNetwork) CleanUpStaleAWSChains(v4Enabled, v6Enabled bool) error {
...
+		return nil
+	}
+	err := n.updateHostIptablesRules(vpcCIDRs, primaryMAC, primaryAddr)
+	if err != nil {
+		log.Errorf("failed to update iptable rules for snat %s", err.Error())
+		return err
+	}
+	var connmarkErr error
+	if n.useExternalSNAT {
+		connmarkErr = n.connmark.Cleanup()
```
Pavani-Panakanti, line 454: here we are cleaning up only nft table. What if cx upgrades from previous version to this by enabling externalSNAT=true then iptables legacy will not be cleaned up ?

pkg/networkutils/iptables_connmark.go
yash97: for long lasting connection, we need to explicitly pass lasting config for the behavior you described above Code ref https://github.com/google/nftables/blob/1db35da82052/conn.go#L71-L87

Pavani-Panakanti (approved):

Issue #3623, opened 2026-03-17
Network Policy CRDs constantly Diffing in ArgoCD


sidewinder12s:
**What happened**:

The CRD being installed by the helm chart for `policyendpoints` is not what apparently the EKS Control plane now forcefully installs on all customer clusters, leading to a constant diff in gitops tooling like ArgoCD. 

**Attach logs**

What is currently being installed in cluster:

```
metadata:
  annotations:
    controller-gen.kubebuilder.io/version: v0.17.0
```
What is actually in this repo and the eks-charts repo:

```
metadata:
  annotations:
    controller-gen.kubebuilder.io/version: v0.11.3
```

**What you expected to happen**:

To be honest I would not have expected the control plane to be modifying my CRDs with no input from me.

Otherwise I would expect the helm chart CRD being vended by this project to be the same CRD one being installed by the EKS Control Plane.

jaydeokar:
@sidewinder12s thanks for highlighting this. We will be removing it from EKS charts to avoid the confusion.

jaydeokar:
https://github.com/aws/amazon-vpc-cni-k8s/blob/master/charts/aws-vpc-cni/crds/customresourcedefinition.yaml#L56

sidewinder12s:
Thanks for confirming its unintended/being resolved.

sidewinder12s:
@jaydeokar any update here?

jaydeokar:
We've merge the change. It will be rolled out with the next release. (v1.22.x)

jaydeokar:
Released with v1.22.4. Closing

PR #3625, opened 2026-03-18
Remove PE crds from helm charts

jaydeokar:
**What type of PR is this?**
cleanup

**Which issue does this PR fix?**: https://github.com/aws/amazon-vpc-cni-k8s/issues/3623

Removes the PE CRD since we install it via Control Plane and not Managed Addons.  CRDs are present here for reference 
https://github.com/aws/amazon-network-policy-controller-k8s/blob/main/charts/amazon-network-policy-controller-k8s/crds/crds.yaml

**Testing done on this change**: N/A. The CRD should come installed on an EKS cluster. For self managed clusters, customers have to install NPC to work together with NPA. If customers are not using NPA, they can disable it completely by setting --set nodeAgent.enabled=false

N/A

**Does this PR introduce any user-facing change?**: Yes refer to comment in testing section. The change only affects customers using self managed cluster and using VPC CNI

```release-note
chore: Remove the references of PE CRD from helm chart as it is installed by control plane
```


parlakisik:
do we have any data about how many customer using self managed cluster with aws cni ?
they must upgrade the network policy controller. otherwise they will lose crd

sidewinder12s:
@parlakisik We are 1 of those users of the self managed CNI utilizing the aws-vpc-cni helm chart. 

We'd tried migrating to managed addons in the past but ran into performance issues with it and/or we still needed to modify things we couldn't through addons.

haouc:
> @parlakisik We are 1 of those users of the self managed CNI utilizing the aws-vpc-cni helm chart.
> 
> We'd tried migrating to managed addons in the past but ran into performance issues with it and/or we still needed to modify things we couldn't through addons.

I don't think removing the definition from VPC CNI helm charts will impact EKS managed clusters. For self managed clusters, this might be an issue. But I don't really know if any customer is running the network policy controller in their self managed cluster and need to rely on this charts to deploy CRDs.

@sidewinder12s if you don't mind, is there any improvement we could do to help your teams onboarding to EKS addons?

sidewinder12s:
> > @parlakisik We are 1 of those users of the self managed CNI utilizing the aws-vpc-cni helm chart.
> > We'd tried migrating to managed addons in the past but ran into performance issues with it and/or we still needed to modify things we couldn't through addons.
> 
> I don't think removing the definition from VPC CNI helm charts will impact EKS managed clusters. For self managed clusters, this might be an issue. But I don't really know if any customer is running the network policy controller in their self managed cluster and need to rely on this charts to deploy CRDs.
> 
> @sidewinder12s if you don't mind, is there any improvement we could do to help your teams onboarding to EKS addons?

The 2 biggest issues were we could not modify all fields in an addons configuration & deployments would take 1+ hours on large constantly scaling clusters due to the way that the addon checked if it was done. If DS pods were constantly starting and stopping, it would never finish the rollout from the addon APIs perspective, so it would hold a Terraform apply until our credentials timed out and failed the apply. Other application methods like ArgoCD do not have this issue.

viveksb007 (approved):
lgtm, PE will be installed by NPC https://github.com/aws/amazon-network-policy-controller-k8s/blob/main/charts/amazon-network-policy-controller-k8s/crds/crds.yaml

Issue #3634, opened 2026-03-28
Silent wrong-service communication when IP_COOLDOWN_PERIOD < nf_conntrack_tcp_timeout_syn_sent causes stale conntrack DNAT to route SYN to recycled pod IP


Mohijeet:
**What happened**:

<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]"><strong>What happened:</strong></p>
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">A silent wrong-service communication occurs when all of the following align:</p>
<ol class="[li_&amp;]:mb-0 [li_&amp;]:mt-1 [li_&amp;]:gap-1 [&amp;:not(:last-child)_ul]:pb-1 [&amp;:not(:last-child)_ol]:pb-1 list-decimal flex flex-col gap-1 pl-8 mb-3">
<li class="whitespace-normal break-words pl-2">Client pod initiates a TCP dial to a Service VIP</li>
<li class="whitespace-normal break-words pl-2">Backend pod dies abruptly mid-handshake (OOMKill / crash / forced eviction)</li>
<li class="whitespace-normal break-words pl-2">Client app times out (e.g. 5s dial timeout) and closes the socket</li>
<li class="whitespace-normal break-words pl-2">OS reuses the same source port on retry (valid — SYN_SENT → CLOSED has no TIME_WAIT)</li>
<li class="whitespace-normal break-words pl-2">Pod IP is recycled within <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">IP_COOLDOWN_PERIOD</code> window</li>
<li class="whitespace-normal break-words pl-2">New pod starts with the same recycled IP</li>
<li class="whitespace-normal break-words pl-2">Stale conntrack DNAT routes new SYN to new pod → TCP handshake <strong>succeeds silently</strong></li>
<li class="whitespace-normal break-words pl-2">Client is now talking to the <strong>wrong service with no error, no RST, no log</strong></li>
</ol>
<hr class="border-border-200 border-t-0.5 my-3 mx-1.5">
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]"><strong>Root cause — timer mismatch between two independent subsystems:</strong></p>
[...]

PR #3636, opened 2026-03-31
updating IP_COOLDOWN_PERIOD from 30 seconds to 130 seconds to ensure SYN_SENT (120 seconds) conntrack timer expires 

Mohijeet:
What type of PR is this?
improvement

**Which issue does this PR fix?:**
In-flight TCP connections to a deleted pod's IP can be silently hijacked if a new pod is assigned the same IP within the conntrack SYN_SENT expiry window (120 seconds on most Linux kernels).
[issue](https://github.com/aws/amazon-vpc-cni-k8s/issues/3634)

What does this PR do / Why do we need it?:

**Problem:**

- When a pod is deleted, any in-flight TCP requests that were initiated just before deletion will create a conntrack entry in SYN_SENT state for the now-gone pod IP. On most Linux kernels, SYN_SENT conntrack entries expire after 120 seconds.
- If a new pod is assigned the same IP address within this 120-second window, the stale conntrack entry becomes a valid TCP flow reference.

 **This is dangerous because:**

- When a TCP SYN retry arrives with the same srcIP:srcPort → dstIP:dstPort tuple, the kernel looks up the existing conntrack entry instead of creating a new one
- The source IP here is the originating service and the destination IP is the virtual Kubernetes service IP — there is no guarantee that the source port will differ across retries
- Even though kube-proxy may have already updated iptables/ipvs rules for the new pod, conntrack takes precedence over iptables for established/tracked flows
- The new (wrong) pod will receive and reply to the SYN, completing a TCP handshake silently, establishing a valid connection to the wrong pod
- tcp_syn_retries can extend this exposure window up to 130 seconds

**Fix:**
Increase the default IP cooldown period from 30 seconds to 130 seconds. This ensures:

All SYN_SENT conntrack entries from the deleted pod's IP have fully expired (120 seconds on most Linux OS)
All tcp_syn_retries for in-flight connections have been exhausted (up to 130 seconds)

Only two code changes are made:

Default value in getCooldownPeriod() changed from 30 to 130
Updated the envIPCooldownPeriod constant comment to document the new default and reasoning

The IP_COOLDOWN_PERIOD environment variable continues to work as before, allowing operators to override this value if needed.

Will this PR introduce any new dependencies?:
No.

Will this break upgrades or downgrades? Has updating a running cluster been tested?:
No breaking changes. The cooldown period increase is a conservative safety improvement. On upgrade, existing clusters will automatically benefit from the longer cooldown. Operators who need a shorter cooldown can override via the IP_COOLDOWN_PERIOD environment variable.

No. Works with a kubectl patch of the image tag. No config changes required.

Yes. The default IP cooldown period after pod deletion is increased from 30 seconds to 130 seconds. This may slightly reduce IP churn rate in high-pod-turnover clusters but prevents silent TCP connection hijacking via stale conntrack entries.

### DIFF
--- a/pkg/ipamd/datastore/data_store.go
+++ b/pkg/ipamd/datastore/data_store.go
@@ -37,7 +37,8 @@ const (
 	// minENILifeTime is the shortest time before we consider deleting a newly created ENI
 	minENILifeTime = 1 * time.Minute
 
-	// envIPCooldownPeriod (default 30 seconds) specifies the time after pod deletion before an IP can be assigned to a new pod
+	// envIPCooldownPeriod (default 130 seconds) specifies the time after pod deletion before an IP can be assigned to a new pod
+	// ensuring SYN_SENT conntrack entries are expired (120 seconds on most OS) and tcp_syn_retries are fully completed (which takes up to 130 seconds)
 	envIPCooldownPeriod = "IP_COOLDOWN_PERIOD"
 
 	// DuplicatedENIError is an error when caller tries to add an duplicate ENI to data store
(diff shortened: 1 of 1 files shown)


oliviassss:
@Mohijeet, thanks for the contribution. But increasing the default to 130s would significantly reduce IP churn rate in high-pod-turnover clusters, potentially causing IP starvation issues. I don't think we should set 130s as default value.

slice-mohijeet:
What is the recommended fix for this behavior? Should the VPC CNI explicitly configure the SYN_SENT timeout or retry parameters during startup to override the default Linux kernel settings?

PR #3659, opened 2026-04-15
updating go version

yash97:
**What type of PR is this?**

```release-note

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/aws/amazon-vpc-cni-k8s
 
-go 1.25.7
+go 1.25.9
 
 require (
 	github.com/apparentlymart/go-cidr v1.1.0


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/aws/amazon-vpc-cni-k8s
 
-go 1.25.7
+go 1.25.9
```
viveksb007, line 3: can we move to 1.26

jaydeokar: We shouldn't need this change to begin with. Docker image and any other tools should fetch the latest go patch version

viveksb007: yes but if someone build directly from this source code, so we need to keep it updated

jaydeokar: Good point on direct source builds. From what I can tell, this is already handled via .go-version and the 
Makefile:

https://github.com/aws/amazon-vpc-cni-k8s/blob/master/.go-version#L1
https://github.com/aws/amazon-vpc-cni-k8s/blob/master/Makefile#L30-L32

Does that cover the concern, or is there a case where it wouldn’t?

PR #3660, opened 2026-04-16
Fix unmanaged eni

jaydeokar:
**What type of PR is this?**
bug

Fixes counter values when there's a unmanaged ENI

https://github.com/aws/amazon-vpc-cni-k8s/issues/3544

N/A

**Does this PR introduce any user-facing change?**: N/A


Pavani-Panakanti (approved):

autumn-traveller:
Hello, I have observed the unmanaged ENI issue that this PR will fix by using the local var instead of the global one.

Are there any plans to release it in the near future or to backport it to previous releases e.g. I have observed it on both release 1.20.5 and 1.21.2 , and I assume that based on the code it is basically present in every single release since the counter for umanaged ENIs was introduced.

I hope I am not adding unnecessary pressure, but I would appreciate this fix appearing in a release soon.

The behavior I observe is that one nodes with a managed ENI the MAX_ENI calculation is wrong so no new ENIs are attached and Pods become unschedulable due to IP address exhaustion on the nodes ENI.

Issue #2677, opened 2023-11-22
Allow setting of EC2 Security group connection tracking  configurable idle timeouts in AWS VPC CNI


youwalther65:
**What would you like to be added**:
AWS just released [EC2 Security group connection tracking adds support for configurable idle timeouts](https://aws.amazon.com/about-aws/whats-new/2023/11/ec2-security-connection-tracking-idle-timeouts/).

Modifying these parameters requires EC2 API calls. It would be great if AWS VPC CNI can automatically implement custom configuration of these idle timeouts for newly provisioned ENI, it manages.

**Why is this needed**:
Avoid conntrack (connection tracking) issues leading to paket loss etc.

jdn5126:
For whomever works on this, ENI options are specified on create here: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/awsutils/awsutils.go#L786

The data-structure chain from `aws-sdk-go`(https://raw.githubusercontent.com/aws/aws-sdk-go/main/service/ec2/api.go) is:
```
CreateNetworkInterfaceInput -> ConnectionTrackingSpecification -> ConnectionTrackingSpecificationRequest
```

bawejahritik:
I would like to try and implement this, any guidance is highly appreciated as this is my first issue

jdn5126:
@bawejahritik thank you for the offer! We are currently discussing internally when to pick this up, as we want to limit the number of new environment variables that we introduce until we have a chance to clean existing ones up.

bawejahritik:
Thank you for your response, is there anything else I can work on which is a good first issue?

youwalther65:
@jdn5126 Any news regarding AWS internal decision?

jdn5126:
@youwalther65 I no longer work for AWS, so I cannot answer this

jayanthvn:
cc: @orsenthil

orsenthil:
> Any news regarding AWS internal decision?
No news yet. this is desirable feature that we will bring it up for prioritization.

OverStruck:
bump, please add

jaydeokar:
released with v1.21.1

PR #3666, opened 2026-04-23
Getting ENI settings from primary ENI and reusing it for secondary ENIs

jaydeokar:
**What type of PR is this?**
feature

https://github.com/aws/amazon-vpc-cni-k8s/issues/2677
CNI discovery primary ENI settings during startup and uses these same settings during secondary ENI creation. This helps replicate the eni settings across the node. eg: connection timeout settings on ENI 
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/security-group-connection-tracking.html

Replicate the EC2 ENI settings across secondary ENIs

**Testing done on this change**: Yes, tested with multiple cases. 
1. Primary ENI is created without any settings - Secondary ENI creation also doesn't pass any of the values
2. Primary ENI is created with atleast one setting - Secondary ENI creation passes these values during ENI creation. 

**Will this PR introduce any new dependencies?**: No. Existing calls only no new API calls

No

Yes

```release-note
CNI now discovers ENI settings from primary ENI and uses it while creating secondary ENIs
```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -266,6 +266,7 @@ type EC2InstanceMetadataCache struct {
 	additionalENITags        map[string]string
 	imds                     TypedIMDS
 	ec2SVC                   ec2wrapper.EC2
+	connectionTrackingSpec   *ec2types.ConnectionTrackingSpecificationRequest
 }
 
 // ENIMetadata contains information about an ENI
@@ -1152,6 +1153,11 @@ func (cache *EC2InstanceMetadataCache) createENIInput(eniDescription string, tag
 		SubnetId:          aws.String(cache.subnetID),
 		TagSpecifications: tags,
 	}
+
+	if cache.connectionTrackingSpec != nil {
+		input.ConnectionTrackingSpecification = cache.connectionTrackingSpec
+	}
+
 	// Even though IPv6 PD is enabled, we require a Primary IP for the ENI.
 	// This always creates an ENI which has 1 Primary IPv6 address
 	// We use assignIPv6Prefix to assign a prefix during setupENI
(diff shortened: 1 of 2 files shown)


pkg/awsutils/awsutils.go
```diff
@@ -243,29 +243,30 @@ type APIs interface {
...
+	useCustomNetworking       bool
+	unmanagedNICs             []bool
+	efaOnlyENIsByNetworkCard  []string
+	useSubnetDiscovery        bool
+	enablePrefixDelegation    bool
+	clusterName               string
+	additionalENITags         map[string]string
+	imds                      TypedIMDS
+	ec2SVC                    ec2wrapper.EC2
+	connectionTimeoutSettings *ec2types.ConnectionTrackingSpecificationRequest
```
parlakisik: there is only connectionTimeoutSettings added right ? 
is there any tab or space change in the code ?

pkg/awsutils/awsutils.go
```diff
@@ -1145,12 +1146,18 @@ func (cache *EC2InstanceMetadataCache) createENITags() []ec2types.TagSpecificati
 }
 
 func (cache *EC2InstanceMetadataCache) createENIInput(eniDescription string, tags []ec2types.TagSpecification, needIPs int) *ec2.CreateNetworkInterfaceInput {
+
```
parlakisik: nit: space

pkg/awsutils/awsutils.go
```diff
@@ -1145,12 +1146,18 @@ func (cache *EC2InstanceMetadataCache) createENITags() []ec2types.TagSpecificati
...
 	input := &ec2.CreateNetworkInterfaceInput{
 		Description:       aws.String(eniDescription),
 		Groups:            cache.securityGroups.SortedList(),
 		SubnetId:          aws.String(cache.subnetID),
 		TagSpecifications: tags,
 	}
+
+	if cache.connectionTimeoutSettings != nil {
+		input.ConnectionTrackingSpecification = cache.connectionTimeoutSettings
```
parlakisik: why do we change connectionTimeoutSettings to ConnectionTrackingSpecification?

pkg/awsutils/awsutils.go
jaydeokar: This is because of autoIndent

pkg/awsutils/awsutils.go
jaydeokar: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ConnectionTrackingSpecificationRequest.html
I can change the variable name if that makes it clear - ConnectionTrackingSpecification requires ConnectionTrackingSpecificationRequest Type

jaydeokar: Updated

pkg/awsutils/awsutils.go
```diff
@@ -1168,6 +1174,32 @@ func (cache *EC2InstanceMetadataCache) createENIInput(eniDescription string, tag
...
+	}
+	if config.UdpStreamTimeout != nil {
+		settings.UdpStreamTimeout = config.UdpStreamTimeout
+		msg += fmt.Sprintf(" udpStreamTimeout=%d", *config.UdpStreamTimeout)
+	}
+	if config.UdpTimeout != nil {
+		settings.UdpTimeout = config.UdpTimeout
+		msg += fmt.Sprintf(" udpTimeout=%d", *config.UdpTimeout)
+	}
+	cache.connectionTrackingSpec = settings
```
yash97, line 1200: nit but printing log, we can just do this 
```
log.Debugf("Connection tracking from primary ENI: tcp=%v            
  udpStream=%v udp=%v", config.TcpEstablishedTimeout, config.UdpStreamTimeout, config.UdpTimeout).
```
It will print nil, if value is nil. no need to skip it

jaydeokar: The reason I conditionally build the log message is to avoid any ambiguity of what's present in the DescribeNetworkInterface response. When we log only non-nil values, it makes it more clear what was explicity returned by the API instead of us logging explicity the values. lmk if that makes sense

yash97: i see. yeah i am fine with it. not a hard blocker.

pkg/awsutils/awsutils.go
```diff
@@ -1152,6 +1153,11 @@ func (cache *EC2InstanceMetadataCache) createENIInput(eniDescription string, tag
 		SubnetId:          aws.String(cache.subnetID),
 		TagSpecifications: tags,
 	}
+
+	if cache.connectionTrackingSpec != nil {
+		input.ConnectionTrackingSpecification = cache.connectionTrackingSpec
```
haouc, line 1158: nit: we may want to log out the ENI being set with configuration?

haouc (approved):
lgtm

pkg/awsutils/awsutils.go
jaydeokar: Sure let me add it as a follow up.

PR #3670, opened 2026-04-28
test: improve TestNew_SetsHTTPClientTimeout to actually assert timeout

haouc:
**What type of PR is this?**

Test improvement

Address Copilot review comment: TestNew_SetsHTTPClientTimeout was only checking that cfg.HTTPClient is non-nil, but not verifying the timeout is actually configured.

Enhanced test to assert:
1. HTTPClient is not nil (existing check)
2. HTTPClient is a BuildableClient instance (new assertion)
3. Validates that the HTTP client type is correct for timeout configuration

This ensures the test actually verifies the timeout configuration behavior instead of just checking for null values.

```release-note

```

### DIFF
--- a/pkg/awsutils/awssession/session_test.go
+++ b/pkg/awsutils/awssession/session_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aws/aws-sdk-go-v2/aws/transport/http"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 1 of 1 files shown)


yash97 (dismissed):

pkg/awsutils/awssession/session_test.go
```diff
@@ -45,6 +46,10 @@ func TestNew_SetsHTTPClientTimeout(t *testing.T) {
 	cfg, err := New(context.Background())
 	assert.NoError(t, err)
 	assert.NotNil(t, cfg.HTTPClient)
+	// Verify the HTTP client is the buildable client type with timeout configured
+	httpClient, ok := cfg.HTTPClient.(*http.BuildableClient)
+	assert.True(t, ok, "HTTPClient should be BuildableClient")
+	assert.NotNil(t, httpClient)
```
jaydeokar: Instead of not nil, can we verify the timeout value set here ? We already verify L48 that the object is not null

haouc: makes sense. updated the test.

jaydeokar (approved):

PR #3675, opened 2026-05-04
fix: assign conntrack zone 1 to Branch ENI traffic in SGP strict mode

Supreeth095:
kube-proxy DNAT and kubelet probes independently pick ephemeral source ports from the same range. In SGP strict mode they can collide on the same 5-tuple in zone 0, causing __nf_conntrack_confirm to drop the packet and corrupt the existing entry's TCP state. Clients see 60s timeouts / ALB 504s.

Install two raw-table rules (gated on POD_SECURITY_GROUP_ENFORCING_MODE=strict) that place vlan+ traffic in zone 1, making DNAT (zone 0) and node-to-pod-direct (zone 1) entries unable to collide.

Verified: 43,730 requests / 30 min / zero timeouts post-fix vs. reproduction within ~1 min pre-fix.

Original patch by Hiroki Tateno <htateno@amazon.com>.

**What type of PR is this?**

```release-note

```

### DIFF
--- a/pkg/networkutils/network.go
+++ b/pkg/networkutils/network.go
@@ -462,9 +462,48 @@ func (n *linuxNetwork) updateHostIptablesRules(vpcCIDRs []string, primaryMAC str
 		if err := n.updateIptablesRules(iptablesConnmarkRules, ipt); err != nil {
 			return err
 		}
+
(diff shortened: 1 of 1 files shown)


jaydeokar:
closing this PR as it's not something we are considering at the moment

PR #3676, opened 2026-05-05
adding liveness and readiness probe for npa

yash97:
**What type of PR is this?**

Should leverage probes added here https://github.com/aws/aws-network-policy-agent/pull/565#issuecomment-4381746715.

testing 

```
helm template
```

```release-note

```


charts/aws-vpc-cni/values.yaml
```diff
@@ -55,6 +55,22 @@ nodeAgent:
   conntrackCacheTableSize: 524288
   logLevel: "debug"
   resources: {}
+  livenessProbe:
+    httpGet:
+      path: /healthz
+      port: health
+    initialDelaySeconds: 60
```
viveksb007, line 62: `initialDelaySeconds` 60 seems high, will this result in atleast 60s to mark this container healthy?

yash97: It is same as CNI, container readiness is decided by readiness probe where initialDelaySeconds is `1`

Issue #3615, opened 2026-03-02
Allow Dualstack Networking (ENABLE_IPv6 AND ENABLE_IPv4)


DonavanFritz:
**What would you like to be added**:

Support for running `ENABLE_IPv6=true` and `ENABLE_IPv4=true` simultaneously to enable dual-stack pod networking on the same ENI. 

Specifically:                                                                                                                                                                                                                                                                                                                                                            
- Allow both IPv6 /80 and IPv4 /28 prefix delegation on the same ENI
- Assign both a /128 IPv6 address and a /32 IPv4 address to pod eth0
- Disable the egress plugin automatically in dual-stack mode (no v4if0/v6if0 needed when both families are native)
- Skip SNAT/CONNMARK iptables rules (pods use their own source IPs from delegated prefixes)

**Why is this needed**:

The current CNI forces a choice between IPv6-only or IPv4-only networking, with the other family available only through NAT egress interfaces (v4if0/v6if0). This means pods cannot be true dual-stack citizens — they always rely on address translation for one family.

Dual-stack on the same ENI would enable:
- Native IPv4 and IPv6 connectivity without NAT, reducing latency and complexity
- Simpler migration paths for workloads transitioning between address families
- Compatibility with services and endpoints that are only reachable over one family
- Elimination of egress SNAT bottlenecks and the associated conntrack state

PR #3688, opened 2026-05-19
feat: dual-stack same-ENI for IPAMD pods (IPv6 + IPv4)

DonavanFritz:
> **Note**: Refresh of #3616 (auto-closed by stale-bot on 2026-05-16 — no human review). No content changes vs. the closed PR; mechanically replayed on current master with no conflicts. Patch continues to ship in our internal fork against upstream v1.22.0 and runs in production.

**What type of PR is this?**
feature

#3615

Enables dual-stack pod networking by allowing `ENABLE_IPv4=true` and `ENABLE_IPv6=true` simultaneously. Both address families are assigned to pod `eth0` from prefix-delegated addresses on the same ENI. No NAT/SNAT, no egress interfaces.

Currently the CNI rejects this combination in `isConfigValid()`. Pods must choose IPv4-only or IPv6-only, relying on NAT egress (`v4if0`/`v6if0`) for the other family. This change removes that restriction when `ENABLE_PREFIX_DELEGATION=true`.

Key behaviors:
- IPv6 `/80` and IPv4 `/28` prefixes are allocated on the same ENI (primary first, secondary for overflow)
- Both addresses assigned directly to pod `eth0` — no egress interfaces created
- No SNAT/CONNMARK iptables rules (upstream `updateHostIptablesRules()` returns nil when `enableIPv6=true`)
- IPv4 warm pool manages prefix capacity; IPv6 is abundant (2^48 addresses per /80)
- Secondary ENIs get source-based routing for both families via per-ENI route tables
- Egress plugin is automatically disabled in dual-stack mode regardless of `ENABLE_V4_EGRESS`/`ENABLE_V6_EGRESS` values
- Single-stack behavior is completely unchanged

Deployed to a dev EKS cluster (5 nodes, v1.21.0 base).

Sample pod verification:
$ ip addr show eth0
    inet 100.89.89.128/32 scope global eth0
    inet6 2a03:5640:f13c:1:7a::4/128 scope global

$ ip link show v4if0
Device "v4if0" does not exist.

$ ip link show v6if0
Device "v6if0" does not exist.

IPAMD logs confirm: `"Dual-stack mode: IPv6 + IPv4 prefixes on same ENI (primary first)"`
Entrypoint logs confirm: `"Dual-stack mode: egress plugin disabled (both IPv4 and IPv6 on eth0)"`

Patch has been running in production on Netflix's internal fork against upstream v1.22.0 since the original PR was opened.

No. No new APIs, IMDS calls, kernel modules, or binary dependencies. Uses existing EC2 prefix delegation APIs already called in single-stack modes.

No. Dual-stack only activates when all three flags are explicitly true (`ENABLE_IPv4`, `ENABLE_IPv6`, `ENABLE_PREFIX_DELEGATION`). Existing single-stack configurations are unchanged. Tested upgrade from single-stack to dual-stack via `kubectl set env` on a running cluster — pods created after the rollout get dual-stack, existing pods continue working until recycled.

Yes — set `ENABLE_IPv6=true` on a cluster already running with `ENABLE_IPv4=true` and `ENABLE_PREFIX_DELEGATION=true`. Works with `kubectl set env daemonset/aws-node -n kube-system ENABLE_IPv6=true`. No other config changes required.

Yes. Setting `ENABLE_IPv4=true` together with `ENABLE_IPv6=true` (with `ENABLE_PREFIX_DELEGATION=true`) is now a valid configuration that enables same-ENI dual-stack pod networking. Previously this combination was rejected at startup by `isConfigValid()`. No action required for users who don't opt into dual-stack; single-stack behavior is unchanged.

```release-note
Add dual-stack support for IPAMD pods. Setting ENABLE_IPv4=true, ENABLE_IPv6=true, and ENABLE_PREFIX_DELEGATION=true simultaneously now assigns both IPv4 and IPv6 addresses to pod eth0 from prefix-delegated addresses on the same ENI. No NAT/SNAT is applied. The egress plugin (v4if0/v6if0) is automatically disabled in dual-stack mode. Single-stack behavior is unchanged.
```

### DIFF
--- a/cmd/aws-vpc-cni/main.go
+++ b/cmd/aws-vpc-cni/main.go
@@ -84,6 +84,7 @@ const (
 	defaultIPCooldownPeriod      = 30
 	defaultDisablePodV6          = false
 	defaultEnableMultiNICSupport = false
+	defaultEnableIPv4           = true
 
 	envHostCniBinPath        = "HOST_CNI_BIN_PATH"
 	envHostCniConfDirPath    = "HOST_CNI_CONFDIR_PATH"
@@ -101,6 +102,7 @@ const (
 	envMinIPTarget           = "MINIMUM_IP_TARGET"
 	envWarmPrefixTarget      = "WARM_PREFIX_TARGET"
 	envEnBandwidthPlugin     = "ENABLE_BANDWIDTH_PLUGIN"
+	envEnIPv4                = "ENABLE_IPv4"
 	envEnIPv6                = "ENABLE_IPv6"
 	envEnIPv6Egress          = "ENABLE_V6_EGRESS"
 	envEnIPv4Egress          = "ENABLE_V4_EGRESS"
@@ -218,9 +220,7 @@ func generateJSON(jsonFile string, outFile string, getPrimaryIP func(ipv4 bool) 
 		return err
 	}
 
-	// enabledIPv6 is to determine if EKS cluster is IPv4 or IPv6 cluster
-	// if this EKS cluster is IPv6 cluster, egress-cni-plugin will enable IPv4 egress by default
-	// if this EKS cluster is IPv4 cluster, egress-cni-plugin will only enable IPv6 egress if env var "ENABLE_V6_EGRESS" is "true"
+	enabledIPv4 := utils.GetBoolAsStringEnvVar(envEnIPv4, defaultEnableIPv4)
 	enabledIPv6 := utils.GetBoolAsStringEnvVar(envEnIPv6, defaultEnableIPv6)
 	var egressIPAMSubnet string
 	var egressIPAMDst string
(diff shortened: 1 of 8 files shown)


parlakisik:
we need to move this one to the `resource-mgmt-dev` branch

mchtech:
Adding another production use case for this PR.

Kubernetes supports dual-stack networking, but Amazon EKS currently requires choosing either IPv4 or IPv6 as the cluster IP family. It does not support a dual-stack cluster where pods receive both address families, and the selected IP family cannot be changed after cluster creation.

Recreating an existing production IPv4 cluster solely to migrate its IP family is not a practical path. Many of our external destinations already support IPv6, so we want eligible outbound traffic to use IPv6 while retaining IPv4 compatibility, reducing traffic and pressure on NAT Gateway.

`ENABLE_V6_EGRESS` provides IPv6 reachability through a ULA-backed `v6if`, but dual-stack destinations still prefer IPv4 over ULA under RFC 6724. Native IPv4 and IPv6 on the same pod `eth0`, without the additional egress interface and SNAT, addresses this limitation directly.

Could the maintainers clarify the remaining blockers for review and whether same-ENI dual-stack is currently on the VPC CNI roadmap?

Issue #3719, opened 2026-06-08
Enhanced Subnet Discovery (v1.22.1) causes excessive DescribeSecurityGroups API calls leading to EC2 throttling


beatrizdemiguelperez:
**What happened**:
After upgrading to v1.22.1, the Enhanced Subnet Discovery feature (enabled by default via 
ENABLE_SUBNET_DISCOVERY=true) generates 10,000+ DescribeSecurityGroups EC2 API calls within 
~15 minutes, causing severe throttling even with an already-increased quota.

AWS Support confirmed via CloudTrail analysis that the vast majority of calls originate 
from the VPC CNI itself (900+ in a single burst), with a smaller number from Karpenter.

Throttling quota in place (already increased by AWS):
- Bucket Maximum Capacity: 120
- Bucket Refill Rate: 20
Still insufficient to absorb the call volume generated by this feature.

**Attach logs**

Logs shared privately with AWS Support (case #178064739700108)

**What you expected to happen**:
Enhanced Subnet Discovery should not generate DescribeSecurityGroups calls at a rate that 
exceeds standard EC2 API limits. The feature should include caching and/or exponential 
backoff to avoid thundering-herd API patterns.

**How to reproduce it (as minimally and precisely as possible)**:
1. Enable Security Groups for Pods (Pod ENIs) on your cluster
2. Upgrade amazon-vpc-cni-k8s to v1.22.1 (ENABLE_SUBNET_DISCOVERY defaults to true)
3. Monitor CloudTrail for DescribeSecurityGroups API calls
4. Observe 10,000+ calls within ~15 minutes → throttling errors in CNI logs

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): 1.33
- CNI Version: v1.22.1 
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):

cdirubbio:
Looking into this @beatrizdemiguelperez

cdirubbio:
We're in the process of releasing a new version (v1.22.2) which has reverted this.

cdirubbio:
Going to close this issue as we have released version `v1.22.2` of the VPC CNI via Public ECR, Managed addons, and eks-charts. As mentioned above, this new version has reverted the feature and so the `DescribeSecurityGroups` API call is no longer present

https://github.com/aws/amazon-vpc-cni-k8s/releases/tag/v1.22.2
[...]

PR #3720, opened 2026-06-08
Disabling SG discovery for ESD feature

cdirubbio:
**What type of PR is this?**

**Which issue does this PR fix?**: #3719 

**What does this PR do / Why do we need it?**: [Disabling SG discovery for ESD feature](https://github.com/aws/amazon-vpc-cni-k8s/commit/742318987da5668696f8ac03b1f0170606cc90bf) to lower EC2 `DescribeSecurityGroups` call volume. We are going to reevaluate implementation.

In larger clusters, the `DescribeSecurityGroups` 30s call per-node can lead to API throttling. For example, in a cluster with 300 nodes, that is 600 calls per minute. 

The refreshSGIDs function returns to its former state, where SGs applied to the primary ENI are refreshed on a 30s basis, and applied to the other ENIs when changes are made.

**Testing done on this change**: Checking EC2 API call volume before and after revert. Before, 600 calls per minute in 300 node cluster, after, near 0.

```release-note
Disable Security Group Discovery feature introduced via Enhanced Subnet Discovery
```


haouc:
Let's also comment out the call to `c.awsClient.RefreshCustomSGIDs()` in nodeInit and reconciliation at https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/ipamd.go#L611-L617

Empty list will still trigger modify call to ENIs repeatedly.

cdirubbio:
@haouc reverted based on our previous discussion to return to original `refreshSGIDs` behavior. The function `refreshCustomSGIDs` was the problematic one, but we still needed to use this original function to maintain previous behavior.

haouc (commented):
Looks good to me. I think you can just delete those disabled code and if needed they can always be recovered from git. In the PR description, please clarify:
1. Document the reason for the revert in the PR description
2. Clarify whether applying primary SGs to all ENIs (including secondary subnet ones) is intentional

cdirubbio:
Added to the PR description, will make follow-up PR deleting dead code

haouc (approved):

PR #3727, opened 2026-06-11
Clean up primaryIP map entry when ENI is freed in tryFreeENI

jaydeokar:
**What type of PR is this?**
bug, cleanup

We store the ENI -> Primary IP data which is used while configuring the ENI network and verifying this primary IP is not assigned to the Pod. However, when the ENI is freed, it is not deleted from this store. 
This is a harmless as we do not check this store when this ENI appears as a secondary IP on a different ENI, so overall there is no impact, just cleans up the logs

No-op mostly, but just removes all the entry so that it doesn't keep listing

N/A

```release-note

```


cdirubbio (approved):
lgtm

PR #3739, opened 2026-06-23
Move gRPC server to Unix socket and scrub introspection data

oliviassss:
**What type of PR is this?**

bug

IPAMD IP Hijack issue, more detail in internal docs.

  1. Moves gRPC server to Unix domain socket (/var/run/aws-node/ipamd.sock, permissions 0660). The CNI plugin runs as root (invoked by kubelet) and can connect; unprivileged
  hostNetwork pods cannot access the socket due to filesystem permissions and container mount namespace isolation.
  2. Scrubs ContainerID and IfName from the /v1/enis introspection response to block the reconnaissance step. The endpoint remains functional for debugging (ENI IDs, IP
  addresses, pod names) but no longer exposes the fields required to construct a valid DelNetwork call.

  **Unit tests**
```
  $ go test ./pkg/ipamd/... ./cmd/routed-eni-cni-plugin/... -count=1
  ok   github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd          0.332s
  ok   github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd/datastore    32.032s
  ok   github.com/aws/amazon-vpc-cni-k8s/cmd/routed-eni-cni-plugin  0.017s
  ok   github.com/aws/amazon-vpc-cni-k8s/cmd/routed-eni-cni-plugin/driver  0.019s
```
  **New tests added**
```
  - TestRunRPCHandler_UnixSocket — verifies socket creation and 0660 permissions
  - TestRunRPCHandler_RemovesStaleSocket — verifies stale socket file is cleaned up before binding
  - TestRunRPCHandler_CreatesSocketDirectory — verifies parent directory is created if missing
  - TestEniV1RequestHandler_ScrubsSensitiveFields — verifies ContainerID/IfName removed from IPv4 introspection response
  - TestEniV1RequestHandler_ScrubsIPv6Fields — verifies ContainerID/IfName removed from IPv6 introspection response
  - TestDialIPAMD_ConnectsViaUnixSocket — verifies client connects to IPAMD via Unix socket
  - TestDialIPAMD_ReturnsErrorWhenDialFails — verifies error is propagated when socket dial fails
```
  **Integration tests (IPAMD suite, EKS v1.35.5, m5.xlarge nodes)**
```
  Ran 26 of 26 Specs in 2491.022 seconds
  26 Passed | 0 Failed | 0 Pending | 0 Skipped
```
  **Manual verification on live cluster (VPC CNI sec-fix.v0.2)**
 1. Unix socket health probe passes:
 ```
  $ kubectl exec -n kube-system aws-node-hzcgl -c aws-node -- /app/grpc-health-probe -addr=unix:///var/run/aws-node/ipamd.sock
  {"level":"info","msg":"status: SERVING"}
```

  2. TCP port 50051 is no longer listening:
 ```
  $ kubectl exec -n kube-system aws-node-hzcgl -c aws-node -- /app/grpc-health-probe -addr=:50051
  {"level":"info","msg":"timeout: failed to connect service \":50051\" within 1s"}
  (exit code 2)
```
  3. Pods healthy with Unix socket probes (0 restarts):
  ```
  $ kubectl get pods -n kube-system -l k8s-app=aws-node
  NAME             READY   STATUS    RESTARTS   AGE
  aws-node-hzcgl   2/2     Running   0          3m12s
  aws-node-wtw44   2/2     Running   0          3m8s
```
  4. IPAMD logs confirm socket-only startup:
  ```
  {"level":"info","caller":"ipamd/rpc_handler.go:473","msg":"Serving RPC Handler version  on unix:/var/run/aws-node/ipamd.sock"}
```

5.DaemonSet rollout completed successfully with no pod disruption.

No
Not for cx inheriting eks addon/charts default values, for cx who supplying their own default values, they should update livenessProbe and readinessProbe config to use the unix socket path `-addr=unix:///var/run/aws-node/ipamd.sock`

Yes

Yes, daemonset config change to use `-addr=unix:///var/run/aws-node/ipamd.sock`
```release-note

```

### DIFF
--- a/charts/aws-vpc-cni/values.yaml
+++ b/charts/aws-vpc-cni/values.yaml
@@ -169,7 +169,7 @@ livenessProbe:
   exec:
     command:
       - /app/grpc-health-probe
-      - '-addr=:50051'
+      - '-addr=unix:///var/run/aws-node/ipamd.sock'
       - '-connect-timeout=5s'
       - '-rpc-timeout=5s'
   initialDelaySeconds: 60
@@ -180,7 +180,7 @@ readinessProbe:
   exec:
     command:
       - /app/grpc-health-probe
-      - '-addr=:50051'
+      - '-addr=unix:///var/run/aws-node/ipamd.sock'
       - '-connect-timeout=5s'
       - '-rpc-timeout=5s'
   initialDelaySeconds: 1
--- a/cmd/routed-eni-cni-plugin/cni.go
+++ b/cmd/routed-eni-cni-plugin/cni.go
@@ -46,7 +46,7 @@ import (
 )
 
 const (
-	ipamdAddress            = "127.0.0.1:50051"
+	ipamdSocketPath         = "/var/run/aws-node/ipamd.sock"
 	dummyInterfacePrefix    = "dummy"
 	npAgentConnTimeout      = 2
 	npaSocketPath           = "/var/run/aws-node/npa.sock"
@@ -136,6 +136,15 @@ func LoadNetConf(bytes []byte) (*NetConf, logger.Logger, error) {
 	return &conf, log, nil
 }
 
+func dialIPAMD(grpcClient grpcwrapper.GRPC, log logger.Logger, socketPath string) (*grpc.ClientConn, error) {
+	conn, err := grpcClient.Dial("unix://"+socketPath, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		return nil, err
+	}
+	log.Debugf("Connected to IPAMD via Unix socket: %s", socketPath)
+	return conn, nil
+}
+
 func cmdAdd(args *skel.CmdArgs) error {
 	return add(args, typeswrapper.New(), grpcwrapper.New(), rpcwrapper.New(), driver.New())
 }
@@ -177,8 +186,8 @@ func add(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 
 	log.Debugf("pod requires multi-nic attachment: %t", requiresMultiNICAttachment)
 
-	// Set up a connection to the ipamD server.
-	conn, err := grpcClient.Dial(ipamdAddress, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	// Set up a connection to the ipamD server via Unix socket.
+	conn, err := dialIPAMD(grpcClient, log, ipamdSocketPath)
 	if err != nil {
 		log.Errorf("Failed to connect to backend server for container %s: %v",
 			args.ContainerID, err)
@@ -395,8 +404,8 @@ func del(args *skel.CmdArgs, cniTypes typeswrapper.CNITYPES, grpcClient grpcwrap
 	}
 
 	// notify local IP address manager to free secondary IP
-	// Set up a connection to the server.
-	conn, err := grpcClient.Dial(ipamdAddress, grpc.WithInsecure())
+	// Set up a connection to the server via Unix socket.
+	conn, err := dialIPAMD(grpcClient, log, ipamdSocketPath)
 	if err != nil {
 		log.Errorf("Failed to connect to backend server for container %s: %v",
 			args.ContainerID, err)
--- a/cmd/routed-eni-cni-plugin/cni_test.go
+++ b/cmd/routed-eni-cni-plugin/cni_test.go
@@ -99,7 +99,7 @@ func TestCmdAdd(t *testing.T) {
 
 	mocksTypes.EXPECT().LoadArgs(gomock.Any(), gomock.Any()).Return(nil)
 
-	conn, _ := grpc.Dial(ipamdAddress, grpc.WithInsecure())
+	conn, _ := grpc.Dial("unix://"+ipamdSocketPath, grpc.WithInsecure())
 
 	mocksGRPC.EXPECT().Dial(gomock.Any(), gomock.Any()).Return(conn, nil)
 	mockC := mock_rpc.NewMockCNIBackendClient(ctrl)
@@ -144,7 +144,7 @@ func TestCmdAddWithNPenabled(t *testing.T) {
 
 	mocksTypes.EXPECT().LoadArgs(gomock.Any(), gomock.Any()).Return(nil)
 
-	conn, _ := grpc.Dial(ipamdAddress, grpc.WithInsecure())
+	conn, _ := grpc.Dial("unix://"+ipamdSocketPath, grpc.WithInsecure())
 
 	mocksGRPC.EXPECT().Dial(gomock.Any(), gomock.Any()).Return(conn, nil)
 	mockC := mock_rpc.NewMockCNIBackendClient(ctrl)
@@ -189,7 +189,7 @@ func TestCmdAddWithNPenabledWithErr(t *testing.T) {
 
 	mocksTypes.EXPECT().LoadArgs(gomock.Any(), gomock.Any()).Return(nil)
 
-	conn, _ := grpc.Dial(ipamdAddress, grpc.WithInsecure())
+	conn, _ := grpc.Dial("unix://"+ipamdSocketPath, grpc.WithInsecure())
 
 	mocksGRPC.EXPECT().Dial(gomock.Any(), gomock.Any()).Return(conn, nil)
 	mockC := mock_rpc.NewMockCNIBackendClient(ctrl)
@@ -232,7 +232,7 @@ func TestCmdAddNetworkErr(t *testing.T) {
 
 	mocksTypes.EXPECT().LoadArgs(gomock.Any(), gomock.Any()).Return(nil)
 
-	conn, _ := grpc.Dial(ipamdAddress, grpc.WithInsecure())
+	conn, _ := grpc.Dial("unix://"+ipamdSocketPath, grpc.WithInsecure())
 
 	mocksGRPC.EXPECT().Dial(gomock.Any(), gomock.Any()).Return(conn, nil)
 	mockC := mock_rpc.NewMockCNIBackendClient(ctrl)
@@ -265,7 +265,7 @@ func TestCmdAddErrSetupPodNetwork(t *testing.T) {
 
 	mocksTypes.EXPECT().LoadArgs(gomock.Any(), gomock.Any()).Return(nil)
 
-	conn, _ := grpc.Dial(ipamdAddress, grpc.WithInsecure())
+	conn, _ := grpc.Dial("unix://"+ipamdSocketPath, grpc.WithInsecure())
 
 	mocksGRPC.EXPECT().Dial(gomock.Any(), gomock.Any()).Return(conn, nil)
 	mockC := mock_rpc.NewMockCNIBackendClient(ctrl)
@@ -303,7 +303,7 @@ func TestCmdAddWithNetworkPolicyModeUnset(t *testing.T) {
 
 	mocksTypes.EXPECT().LoadArgs(gomock.Any(), gomock.Any()).Return(nil)
 
-	conn, _ := grpc.Dial(ipamdAddress, grpc.WithInsecure())
+	conn, _ := grpc.Dial("unix://"+ipamdSocketPath, grpc.WithInsecure())
 
 	mocksGRPC.EXPECT().Dial(gomock.Any(), gomock.Any()).Return(conn, nil)
 	mockC := mock_rpc.NewMockCNIBackendClient(ctrl)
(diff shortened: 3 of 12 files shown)


cmd/routed-eni-cni-plugin/cni_test.go
```diff
@@ -1833,3 +1834,41 @@ func TestLoadNetConf(t *testing.T) {
...
+	ctrl, _, mocksGRPC, _, _ := setup(t)
+	defer ctrl.Finish()
+
+	log := logger.New(&logger.Configuration{LogLevel: "DEBUG"})
+
+	// Create a file at the socket path so os.Stat succeeds, but it's not a real socket
+	// so grpcClient.Dial("unix://...") will fail, triggering TCP fallback.
+	tmpDir := t.TempDir()
+	socketPath := tmpDir + "/ipamd.sock"
+	os.WriteFile(socketPath, []byte("not a socket"), 0600)
```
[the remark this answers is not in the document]
oliviassss: Over-engineering. t.TempDir() guarantees a writable directory that's cleaned up automatically. If WriteFile
fails there, CI has bigger problems.

cmd/routed-eni-cni-plugin/cni_test.go
```diff
@@ -1833,3 +1834,41 @@ func TestLoadNetConf(t *testing.T) {
...
+
+	// Since ipamdSocketPath is a const pointing to /var/run/aws-node/ipamd.sock (doesn't exist in CI),
+	// dialIPAMD will skip the socket and go straight to TCP fallback.
+	conn, _ := grpc.Dial(ipamdAddress, grpc.WithInsecure())
```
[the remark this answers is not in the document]
oliviassss: fixed

pkg/ipamd/rpc_handler_test.go
```diff
@@ -576,3 +578,45 @@ func TestServer_GetNetworkPolicyConfigs(t *testing.T) {
...
+	mockContext := &IPAMContext{
+		awsClient:       m.awsutils,
+		networkClient:   m.network,
+		dataStoreAccess: datastore.InitializeDataStores([]bool{false}, "test", false, log),
+	}
+
+	errCh := make(chan error, 1)
+	go func() {
+		errCh <- mockContext.runRPCHandlerWithSocketPath("1.0.0", socketPath)
+	}()
```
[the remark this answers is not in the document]
oliviassss: The fix suggested is heavy for a UT. for a unit test that runs in <100ms and exits, this is low priority. if we see flakiness in UT later we can revisit this.

pkg/ipamd/rpc_handler.go
```diff
@@ -469,32 +471,75 @@ func (s *server) GetNetworkPolicyConfigs(ctx context.Context, e *emptypb.Empty)
...
+	// Remove stale socket file from a previous run
+	if err := os.Remove(socketPath); err != nil && !os.IsNotExist(err) {
+		return errors.Wrap(err, "ipamd: failed to remove stale socket")
+	}
+
+	// Set umask before creating the socket to avoid a TOCTOU window where the socket
+	// exists with default permissions before chmod is applied.
+	oldMask := syscall.Umask(0117)
+	listener, err := net.Listen("unix", socketPath)
+	syscall.Umask(oldMask)
```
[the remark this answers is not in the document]
oliviassss: fixed

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -136,6 +137,32 @@ func LoadNetConf(bytes []byte) (*NetConf, logger.Logger, error) {
...
+		if err == nil {
+			log.Debugf("Connected to IPAMD via Unix socket: %s", socketPath)
+			return conn, nil
+		}
+		log.Warnf("Unix socket exists but dial failed (%v), trying TCP fallback", err)
+	}
+
+	// TODO: Remove TCP fallback once all nodes run the socket-based IPAMD.
+	log.Debugf("Falling back to TCP connection: %s", ipamdAddress)
+	conn, err := grpcClient.Dial(ipamdAddress, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
viveksb007: CNI to IPAMD is node local communication, why do we need this fallback?

pkg/ipamd/introspect.go
```diff
@@ -128,6 +128,24 @@ func eniV1RequestHandler(ipam *IPAMContext) func(http.ResponseWriter, *http.Requ
 		for _, ds := range ipam.dataStoreAccess.DataStores {
 			eniInfos[ds.GetNetworkCard()] = ds.GetENIInfos()
 		}
+		// Scrub sensitive fields that could be used to construct malicious gRPC calls.
```
viveksb007, line 131: q - why do we have this introspection endpoint? i wonder if eni.json or some datastore file that we have has similar info and we can get rid of this introspection endpoint.

don't need to remove as part of this PR

viveksb007 (commented):
we should get rid of the TCP fallback which might hide the issues if there is anything with unix socket based implementation.

cmd/routed-eni-cni-plugin/cni.go
oliviassss: This is majorly for upgrade safety.
During a DS rolling update, kubelet may still invoke the old CNI binary while the new IPAMD has started with the Unix socket. TCP fallback ensures those in-flight calls succeed. 
Additionally, health probe still use TCP :50051 so the server need to bind to TCP anyway. The fallback is a cheap safeguard for backward compatibility.

pkg/ipamd/introspect.go
oliviassss: yeah we still need the introspection endpoint, The datastore file is a minimal recovery journal used only by IPAMD on restart vs the introspection endpoint provides a rich live view.

cmd/routed-eni-cni-plugin/cni.go
viveksb007: i don't think `the old CNI binary while the new IPAMD has started` will ever happen on a Node. check this out https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set/

some old ref about DS update mechanism of `delete then create`, so DS poses some un-availability for short window as the DS pod is always deleted first and then created with updated manifest. (https://github.com/kubernetes/kubernetes/issues/48841)

when a new pod will come, IPAMD won't start unless old CNI binary isn't replaced as init-container will run first. 

Nice catch on health probe, we need to change that also to dial on unix socket that IPAMD is creating.

pkg/ipamd/introspect.go
viveksb007: can you update PR details with what do we get from introspection endpoint before and after this PR change?

pkg/ipamd/rpc_handler_test.go
```diff
@@ -576,3 +578,45 @@ func TestServer_GetNetworkPolicyConfigs(t *testing.T) {
 	assert.Equal(t, "standard", resp.NetworkPolicyMode)
 	assert.True(t, resp.MultiNICEnabled)
 }
+
+func TestRunRPCHandler_UnixSocket(t *testing.T) {
```
jaydeokar, line 583: Can you add other test cases here ? Removing existing sockets, change permission failing

pkg/ipamd/rpc_handler.go
```diff
@@ -469,28 +471,65 @@ func (s *server) GetNetworkPolicyConfigs(ctx context.Context, e *emptypb.Empty)
...
 	healthpb.RegisterHealthServer(grpcServer, healthServer)
 
 	// Register reflection service on gRPC server.
 	reflection.Register(grpcServer)
+
+	// TODO: Remove TCP fallback once all nodes run the socket-based IPAMD.
+	// TCP must bind successfully — health probes and waitForIPAM depend on it.
+	tcpListener, err := net.Listen("tcp", ipamdgRPCaddress)
+	if err != nil {
+		listener.Close()
```
jaydeokar: Why do we have to close the socket connection if tcp one fails ?

pkg/ipamd/introspect.go
oliviassss: only ifname and containerID are scrubbed from introspection endpoint as listed in PR description, the code change is pretty straightforward.

pkg/ipamd/rpc_handler.go
oliviassss: TCP fail -> health probe fail -> container restart. just to handle the restart gracefully.

cmd/routed-eni-cni-plugin/cni.go
oliviassss: You're right that ds updates and the upgrade race shouldn't happen in practice. But I'd prefer to keep the TCP fallback just in case the socket is unavailable for reasons like file deletion or permission issues. We can evaluate removal after 1-2 versions.
For health probes - agree we should ideally migrate to the Unix socket, but it requires Helm chart and addon config changes plus evaluation on the rollout/rollback strategy. That's irrelevant to the security fix so better to decouple it.

pkg/ipamd/rpc_handler_test.go
oliviassss: Added tests for stale socket removal and directory creation. The os perm failure need an OS abstraction layer (not used elsewhere in the codebase), may not worth adding it as the failure handling is straightforward: close listener -> remove socket -> return error.

cmd/routed-eni-cni-plugin/cni.go
viveksb007: > We can evaluate removal after 1-2 versions.

what data will support this? 

how would we know if Unix socket is being exercised or TCP fallback is being used by CNI binary?


For health probes, I think we just need to change our [grpc health probe](https://github.com/aws/amazon-vpc-cni-k8s/tree/master/cmd/grpc-health-probe), can you double check what changes are needed to release this change via AddOns?

oliviassss: discussed offline, we will remove the TCP fallback. This won't affect cx using default values supplied by eks addon or helm chart. but for cx supplies their own chart default values, they need to explicitly change to use unix socket path for liveness and readiness probe. will call out in release note.

pkg/ipamd/rpc_handler.go
```diff
@@ -469,28 +470,50 @@ func (s *server) GetNetworkPolicyConfigs(ctx context.Context, e *emptypb.Empty)
...
+	return c.runRPCHandlerWithSocketPath(version, ipamdGRPCSocketPath)
+}
+
+func (c *IPAMContext) runRPCHandlerWithSocketPath(version string, socketPath string) error {
+	log.Infof("Serving RPC Handler version %s on unix:%s", version, socketPath)
+
+	// Ensure the parent directory exists
+	if err := os.MkdirAll(filepath.Dir(socketPath), 0755); err != nil {
+		return errors.Wrap(err, "ipamd: failed to create socket directory")
+	}
```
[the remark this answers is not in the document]
oliviassss: over-stated. /var/run/aws-node is created by kubelet via hostPath: DirectoryOrCreate before our code runs

pkg/ipamd/rpc_handler_test.go
```diff
@@ -576,3 +579,132 @@ func TestServer_GetNetworkPolicyConfigs(t *testing.T) {
...
+		errCh <- mockContext.runRPCHandlerWithSocketPath("1.0.0", socketPath)
+	}()
+
+	// Wait for the socket to appear
+	timeout := time.After(5 * time.Second)
+	for {
+		info, err := os.Stat(socketPath)
+		if err == nil {
+			assert.Equal(t, os.FileMode(0660), info.Mode().Perm(), "Socket should have 0660 permissions")
+
```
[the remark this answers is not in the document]
oliviassss: low-risk. Chmod executes synchronously in the same goroutine immediately after Listen

cmd/routed-eni-cni-plugin/cni.go
```diff
@@ -136,6 +136,20 @@ func LoadNetConf(bytes []byte) (*NetConf, logger.Logger, error) {
 	return &conf, log, nil
 }
 
+// dialIPAMD connects to the IPAMD gRPC server via Unix socket.
+func dialIPAMD(grpcClient grpcwrapper.GRPC, log logger.Logger) (*grpc.ClientConn, error) {
```
viveksb007: nit - this fn isn't doing anything, just calling another fn, we can remove this and keep only 1 fn

viveksb007 (commented):
overall change lgtm

cmd/routed-eni-cni-plugin/cni.go
oliviassss: addressed

charts/aws-vpc-cni/values.yaml
```diff
@@ -169,7 +169,7 @@ livenessProbe:
   exec:
     command:
       - /app/grpc-health-probe
-      - '-addr=:50051'
+      - '-addr=unix:///var/run/aws-node/ipamd.sock'
```
jaydeokar, line 172: Do we allow this to be configurable ?

oliviassss: for helm chart yes it's configurable. eks addon api does not expose this config to customer.

viveksb007 (approved):

Issue #3753, opened 2026-07-02
v1.22.2 assigns the node's primary IP to pods on SageMaker HyperPod EKS nodes (subnet-discovery check failure leaves c.primaryIP unset)


kcfigaro:
## What happened

On SageMaker HyperPod EKS nodes, VPC CNI **v1.22.2** adds the **primary ENI's primary IP** to the pod IP pool and eventually assigns it to a pod. The victim pod's IP is then identical to the node's InternalIP, so all return traffic routes to the host network namespace: in-pod DNS fails (`Temporary failure in name resolution` to kube-dns), kubelet HTTP probes to `<podIP>:<port>` get `connection refused`, and the pod crashloops indefinitely (one workload accumulated 131 restarts in 8h). Which pod breaks is random — whichever draw pulls the primary IP from the pool.

**v1.21.2 on the same node handles the same conditions correctly.** We verified with an in-place A/B/A test (downgrade → upgrade on one node, no other changes):

| CNI on same node | ipamd datastore (`/v1/enis` introspection) |
|---|---|
| v1.22.2-eksbuild.1 | `TotalIPs=50`, node primary IP **in pool**, assigned to pods |
| v1.21.2-eksbuild.2 | `TotalIPs=49`, node primary IP **correctly excluded** |
| v1.22.2-eksbuild.1 (again) | `TotalIPs=50`, node primary IP **back in pool** |

## Root cause analysis

HyperPod instances have two relevant quirks: the primary ENI is attached at **device index 1** (both versions log `Device number of primary ENI is 1! Forcing it to be 0 as expected`), and the ENI/subnet topology is SageMaker-managed (cross-account), so `DescribeSubnets` from the node does not return the subnet that IMDS reports for the ENI.

v1.22.2 inserted the subnet-discovery exclusion check into `setupENI` **between** `AddENI` and the `c.primaryIP[eni]` assignment ([`pkg/ipamd/ipamd.go#L1277-L1294` @ v1.22.2](https://github.com/aws/amazon-vpc-cni-k8s/blob/v1.22.2/pkg/ipamd/ipamd.go#L1277-L1294)):

```go
err = c.dataStoreAccess.GetDataStore(...).AddENI(eni, ...)           // ENI registered in datastore
...
if c.useSubnetDiscovery {                                            // ENABLE_SUBNET_DISCOVERY defaults true
[...]

PR #3754, opened 2026-07-02
fix: hyperpod ESD issue

cdirubbio:
**What type of PR is this?** bug

**Which issue does this PR fix?**: #3753 

**What does this PR do / Why do we need it?**: Hyperpod has ENI with subnet not in customer account, breaking assumption made in ESD flow for subnet exclusion.

**Testing done on this change**: Spun up hyperpod cluster, using VPC CNI `v1.22.2` then test image for comparison:
#### v1.22.2 (10.2.101.235 is ENI primary IP)
```

    "TotalIPs": 15,
...        "AvailableIPv4Cidrs": {
          "10.2.101.235/32": {
            "Cidr": { "IP": "10.2.101.235", "Mask": "/////w==" },
```
#### fixed image
```
{
  "0": {
    "TotalIPs": 14,
    "AssignedIPs": 3,
          "10.2.144.185/32": {
            "Cidr": { "IP": "10.2.144.185", "Mask": "/////w==" },
            "IPAddresses": {},
            "IsPrefix": false,
            "AddressFamily": ""
```

**Will this PR introduce any new dependencies?**: no

**Will this break upgrades or downgrades? Has updating a running cluster been tested?**: no/yes

**Does this change require updates to the CNI daemonset config files to work?**: no

```release-note
fix: hyperpod compatibility for esd exclusion
```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -2591,7 +2591,7 @@ func (cache *EC2InstanceMetadataCache) IsSubnetExcluded(ctx context.Context, sub
 		}
 	}
 
-	// Subnet not found in VPC
-	log.Warnf("IsSubnetExcluded: subnet %s not found in VPC", subnetID)
-	return false, fmt.Errorf("subnet %s not found in VPC", subnetID)
+	// Subnet not found in VPC — treat as not excluded (e.g. HyperPod cross-account subnets)
+	log.Warnf("IsSubnetExcluded: subnet %s not found in VPC, treating as not excluded", subnetID)
+	return false, nil
 }
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -1291,7 +1291,6 @@ func (c *IPAMContext) setupENI(ctx context.Context, eni string, eniMetadata awsu
 		c.primaryIP[eni] = eniMetadata.PrimaryIPv4Address()
 	}
 
-
 	// Check if this ENI (primary or secondary) is in an excluded subnet and mark it for exclusion
 	if c.useSubnetDiscovery {
 		if _, err := c.excludedENIBasedOnSubnetTags(ctx, eni, eniMetadata); err != nil {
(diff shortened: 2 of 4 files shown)


pkg/awsutils/awsutils.go
haouc, line 2570: Shouldn't be handled here instead of the final return?

pkg/awsutils/awsutils.go
```diff
@@ -2591,7 +2591,7 @@ func (cache *EC2InstanceMetadataCache) IsSubnetExcluded(ctx context.Context, sub
 		}
 	}
 
-	// Subnet not found in VPC
-	log.Warnf("IsSubnetExcluded: subnet %s not found in VPC", subnetID)
-	return false, fmt.Errorf("subnet %s not found in VPC", subnetID)
+	// Subnet not found in VPC — treat as not excluded (e.g. HyperPod cross-account subnets)
+	log.Warnf("IsSubnetExcluded: subnet %s not found in VPC, treating as not excluded", subnetID)
+	return false, nil
```
haouc, line 2596: Logging at return doesn't make sense... API error was handled earlier and the rest of this func is processing tags locally.

pkg/awsutils/awsutils.go
haouc: Also, should we filter on specific error types instead of returning `nil` on all?

cdirubbio: That function GetVPCSubnets grabs all VPCs, it doesnt error there. We just loop through all subnets checking if it matches the subnet in arg.
```go
// Get all VPC subnets with their tags
	subnets, err := cache.GetVpcSubnets(ctx)
	if err != nil {
		return false, fmt.Errorf("failed to get VPC subnets: %v", err)
```
This err is nil, this function just grabs all subnets in VPC, doesnt do any filtering on the subnet passed in arg to function.

haouc (approved):

pkg/ipamd/ipamd.go
```diff
@@ -1284,20 +1284,20 @@ func (c *IPAMContext) setupENI(ctx context.Context, eni string, eniMetadata awsu
...
-
 	// Store the addressable IP for the ENI
 	if c.enableIPv6 {
 		c.primaryIP[eni] = eniMetadata.PrimaryIPv6Address()
 	} else {
 		c.primaryIP[eni] = eniMetadata.PrimaryIPv4Address()
 	}
 
+	// Check if this ENI (primary or secondary) is in an excluded subnet and mark it for exclusion
+	if c.useSubnetDiscovery {
```
haouc, line 1295: Looked this again, should we try this before addENI()? What if fetching subnets from VPC fails?

PR #3757, opened 2026-07-02
tests: update eksctl for CI to not use default addon PDBs

cdirubbio:
**What type of PR is this?** testing

**Which issue does this PR fix?**: n/a/

**What does this PR do / Why do we need it?**: Remove CoreDNS, metrics-server PDB so that custom networking tests don't timeout in CI. The `autoModeConfig` set to `enabled: false` was going to be needed in the future anyway as seen in CI outputs, so may as well do it now

**Testing done on this change**: Local cluster created with `eksctl create cluster -f test-cluster.yaml`, confirmed PDBs are disabled

```release-note

```

### DIFF
--- a/scripts/test/config/bottlerocket.yaml
+++ b/scripts/test/config/bottlerocket.yaml
@@ -6,6 +6,17 @@ metadata:
   name: CLUSTER_NAME_PLACEHOLDER
   region: us-west-2
 
+autoModeConfig:
+  enabled: false
+
+addons:
+  - name: vpc-cni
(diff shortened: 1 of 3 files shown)


haouc (approved):
lgtm. Makes sense disabling them. Although PDB is not in networking tests' scope, I wonder if any use cases we should test them as well?

cdirubbio:
Probably good to do some internal tests but for github CI, no need to have, just slows down the tests and fails here

PR #3764, opened 2026-07-07
Create the host-side veth directly in the host namespace via the netlink PeerNamespace attribute

amrmahdi:
**What type of PR is this?**

improvement

Not filing a separate issue; repro steps and telemetry are in the Testing section below.

The plugin currently creates the veth pair inside the pod network namespace and then moves the host-side end to the host namespace with `LinkSetNsFd`. The kernel implements this move as a device unregister plus re-register under `rtnl_lock`, and the calling thread blocks in `rcu_barrier()` inside `__dev_change_net_namespace` for nearly the entire move. With many concurrent CNI ADDs on one node these serialize and dominate pod network setup latency. On a 30-pod concurrent create burst, CNI ADD p50 is ~5s, and a kernel probe on `__dev_change_net_namespace` attributes exactly one namespace move per pod to aws-cni. The cost is worst on the EKS AMIs because their kernels build with `CONFIG_HZ=100` (see PR discussion).

This PR creates each veth end directly in its final namespace using the `PeerNamespace` attribute on the veth create, the same approach the upstream containernetworking reference plugins use. Each device is registered exactly once and the move disappears. The same burst: CNI ADD p50 0.75s, zero namespace moves on the same kernel probe. Per review, this is the default and only path (the change was initially gated behind an opt-in env var, since removed), and it covers the branch-ENI (Security Groups for Pods) veth as well.

Interface naming is unaffected since the container end is still created in the pod namespace with its final name. Crash cleanup is unaffected because deleting either end of a veth pair removes both, so pod namespace teardown still removes the host end. Two orderings from review are preserved: the host end is brought up before the IPv6 DAD wait (the container end needs carrier for DAD), and the host-side IPv6 sysctls (`accept_ra=0`, `accept_redirects=1`, `forwarding=0`) are applied before the host end comes up, so it is never up while still accepting router advertisements.

Unit tests cover the creation path, the sysctls-before-up ordering, the host-end-up-before-DAD ordering, sysctl failure propagation, and ENOENT tolerance on kernels without IPv6, in `cmd/routed-eni-cni-plugin/driver`.

Manual testing on an EKS node, comparing stock and patched builds. Burst harness (30 pods pinned to one node, created concurrently):

```bash
NODE=<node-name>
for i in $(seq -w 1 30); do
cat <<POD
apiVersion: v1
kind: Pod
metadata: {name: burst-$i}
spec:
  nodeName: $NODE
  restartPolicy: Never
  terminationGracePeriodSeconds: 0
  containers:
  - name: c
    image: registry.k8s.io/pause:3.9
POD
echo ---
done | kubectl apply -f -
```

CNI ADD latency from the containerd journal (RunPodSandbox request line to "connecting to shim" brackets network setup):

```bash
journalctl -u containerd -o cat --since '-5 min' | grep -E 'RunPodSandbox for|connecting to shim'
```

Namespace-move count per process during the burst:

```bash
bpftrace -e 'kprobe:__dev_change_net_namespace { @moves[comm] = count(); } interval:s:60 { exit(); }'
```

Results:

| | CNI ADD p50 (30-pod burst) | `@moves[aws-cni]` |
|---|---|---|
| create-then-move (before) | ~5s | 30 (one per pod) |
| PeerNamespace (this PR) | 0.75s | absent (zero) |

Probe sanity check: a manual `ip link set <veth> netns <ns>` fires `__dev_change_net_namespace` once. Connectivity and policy with the patched build: pod-to-pod TCP, DNS, external egress, NetworkPolicy enforcement (deny applied and traffic blocked, removed and traffic restored), and DEL cleanup with no orphaned host-side veths.

IPv6: validated end to end on an IPv6-native EKS 1.33 cluster. The initial version failed there (host end down during the DAD wait, every ADD timing out after 10s); with the fix, pods are Ready in seconds with no tentative addresses, and pod-to-pod v6 (cross- and same-node), DNS, and both v6 and v4 egress pass. Flag-off/on equivalence was probe-verified while the gate existed.

No. Uses the `PeerNamespace` attribute from vishvananda/netlink, already vendored. No new EC2, IMDS, or kernel-module dependency.

No. The creation path changes for new pods after upgrade, but the end state is identical: same interface names, same explicitly-assigned MACs, same routes and sysctls. Existing pods are untouched. Downgrade restores the old path; there is no persistent state tied to how a veth was created.

No. There is no configuration surface; manifests and Helm values are unchanged.

Yes. Pod veth creation changes from create-then-move to creating each end directly in its final namespace, substantially reducing pod network setup latency under concurrent pod creation. No configuration change is required.


viveksb007:
I isolated this rtnl_lock contention using following mechanism and can validate that concurrent CNI ADD do serialize on this lock. 

Isolation methodology and results

https://gist.github.com/viveksb007/18430a4230a2909ce42f74867c48417a

Screenshot 2026-07-13 at 12 22 22 PM


PR lgtm, there is no new concern related to MAC addrs as we are randomly generating a MAC Addr and providing that explicitly while Veth creation, so behavior is similar to old mechanism.

Let me do some more testing on IPv6 and IP family combination cases just for sanity testing. 

Another orthogonal decision that we need to make is whether 
- to directly make this default OR
- put this behind flag for few VPC-CNI versions and then make it default

amrmahdi:
> I isolated this rtnl_lock contention using following mechanism and can validate that concurrent CNI ADD do serialize on this lock.
> 
> Isolation methodology and results
> 
> https://gist.github.com/viveksb007/18430a4230a2909ce42f74867c48417a
> 
> Screenshot 2026-07-13 at 12 22 22 PM
> PR lgtm, there is no new concern related to MAC addrs as we are randomly generating a MAC Addr and providing that explicitly while Veth creation, so behavior is similar to old mechanism.
> 
> Let me do some more testing on IPv6 and IP family combination cases just for sanity testing.
> 
> Another orthogonal decision that we need to make is whether
> 
> * to directly make this default OR
> * put this behind flag for few VPC-CNI versions and then make it default

  @viveksb007 thanks for confirming. I added the flag to give you the optionality of going either path.

Adding additional context for you and other readers who land here, on why `__dev_change_net_namespace` is specifically slow on the AL2023 kernel. I did an isolated experiment with just the veth cross-namespace move with plain ip commands with 200 pairs at concurrency 30, same instance type (r5d.8xlarge), only the node OS changed:
  - AL2023 6.12 (CONFIG_HZ=100): ~33s
  - Bottlerocket 6.12 (CONFIG_HZ=100): ~33s
  - Ubuntu linux-aws 6.17 (CONFIG_HZ=1000): ~5.3s
  - GKE COS 6.12 (CONFIG_HZ=1000): ~4.7s

The move blocks in `rcu_barrier()` inside `__dev_change_net_namespace`, and that barrier drains on a clock set by `CONFIG_HZ`. Both EKS AMIs build with HZ=100, so each move costs about 7x more than on an HZ=1000 kernel. Quick check on any node:  `grep 'CONFIG_HZ=' /boot/config-$(uname -r)`.

This change removes the move entirely, so it helps at any HZ.

viveksb007:
Thanks @amrmahdi for the info, its very helpful. i can see upstream linux has default as `CONFIG_HZ_1000=y` (https://github.com/amazonlinux/linux/blob/a13c140cc289c0b7b3770bce5b3ad42ab35074aa/arch/x86/configs/x86_64_defconfig#L43), so it seems AL2023 and Bottlerocket have overriden that config internally, will update once I know the reasoning behind that override.

On side node, we create another interface as part of egress-cni-plugin which calls upstream containernetworking code (https://github.com/containernetworking/plugins/blob/9b3772e1a7abf93cbb7c6526a28bc0d27b830e02/pkg/ip/link_linux.go#L40) which by default creates the veth end on host side. So main plugin is the case doing "create-then-move".

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -124,20 +126,29 @@ func (createVethContext *createVethPairContext) run(hostNS ns.NetNS) error {
...
 
-	hostVeth, err := createVethContext.netLink.LinkByName(createVethContext.hostVethName)
-	if err != nil {
-		return errors.Wrapf(err, "setup NS network: failed to find link %q", createVethContext.hostVethName)
-	}
+	var hostVeth netlink.Link
+	hostHardwareAddr := createVethContext.hostMACAddr
+	if !createVethContext.usePeerNS {
+		var err error
+		hostVeth, err = createVethContext.netLink.LinkByName(createVethContext.hostVethName)
```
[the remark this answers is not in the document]
viveksb007: this is actually pointing to a bug, this PR changes are breaking Pod setup on IPv6 clusters.

----

## Summary

Enabling `AWS_VPC_K8S_CNI_VETH_PEER_NAMESPACE=true` (added by PR #3764) prevents
**all** pods from starting on IPv6 EKS clusters. Every pod is stuck in
`ContainerCreating`. IPv4 clusters are unaffected. The failure is 100%
reproducible and is isolated entirely to the flag — the same image with the flag
off works fine.

## Impact

| cluster | flag off | flag on |
|---|---|---|
| IPv4 (`eth0` is IPv4) | works | works |
| IPv6 (`eth0` is IPv6) | works | **all pods fail to start** |

Severity: high for IPv6 clusters — the feature is completely unusable there and
takes the whole node's pod scheduling down with it (no pod can pass CNI ADD).

## Symptom

Pod events on an IPv6 cluster with the flag enabled:

```
failed to setup network for sandbox "...": plugin type="aws-cni" name="aws-cni" failed (add):
add command: failed to setup network: SetupPodNetwork: failed to setup veth pair:
failed to setup veth network: setup NS network:
failed while waiting for v6 addresses to be stable:
link eth0 still has tentative addresses after 10000000000 seconds
```

## Evidence

Captured on the node while a pod was stuck (peer-namespace path enabled):

```
# container side
eth0@ifNN: <NO-CARRIER,BROADCAST,MULTICAST,UP>  state DOWN
    inet6 2600:1f13:...:87b5::1/128 scope global tentative      # never leaves tentative
    accept_dad = 1   dad_transmits = 1

# host side (peer)
eniXXXX@if2: <BROADCAST,MULTICAST>  state DOWN  qdisc noop       # never brought UP
    operstate = down
```

The container interface is admin-UP but reports **NO-CARRIER**; the host-side
peer is **DOWN**. The pod's IPv6 address is stuck **tentative**.

## Root cause

### Background: veth carrier and IPv6 DAD

- A veth pair has link carrier (`LOWER_UP`) only when **both** ends are UP. If one
  end is down, the other shows `NO-CARRIER`.
- IPv6 runs **Duplicate Address Detection (DAD)** on every tentative global
  address before it becomes usable. DAD sends Neighbor Solicitation probes on the
  link and waits for the RetransTimer. This requires **carrier** — no carrier, no
  probe, DAD never completes, the address stays `tentative`.
- The plugin explicitly waits for the container address to leave `tentative`
  (`WaitForAddressesToBeStable`, 10s timeout) so it returns a ready interface.
  This wait is **IPv6-only** — IPv4 has no DAD / tentative phase.

### The bug: the host-side veth is down during the DAD wait (ordering issue)

The veth setup has **two** `LinkSetUp(hostVeth)` calls:

1. **Inside the per-namespace `run()` closure, before the DAD wait** — this call
   is gated behind `if !usePeerNS`. In the move (default) path it brings the host
   end UP *before* the IPv6 DAD wait. **With the flag on, this branch is skipped.**

2. **In `setupVeth()`, after `run()` returns** — this call is unconditional and
   runs in both modes. It is why IPv4 works with the flag on. But it executes
   **after** the DAD wait (the wait lives inside `run()`).

Consequences by path:

- **Move mode (flag off):** host end brought up (call 1) → veth has carrier →
  IPv6 DAD completes → host end moved to host ns. Works.
- **Peer mode (flag on):** call 1 is skipped. The host end is created directly in
  the host ns via the `PeerNamespace` attribute but is **not brought up** — only
  `net.FlagUp` was requested on the veth create, which the code's own comments
  note is unreliable across platforms (that unreliability is exactly why the
  explicit `LinkSetUp` calls exist). At the DAD wait the host end is still DOWN →
  container end has NO-CARRIER → DAD cannot complete → address stays tentative →
  `WaitForAddressesToBeStable` times out after 10s → `run()` returns an error →
  the unconditional `LinkSetUp` (call 2) is **never reached**.

So the bring-up that would give the veth carrier does exist, but for the peer
path it is sequenced **after** the DAD gate that depends on it.

### Why IPv4 did not catch it

The DAD wait is guarded by an IPv6-only condition. IPv4 has no DAD, no tentative
state, and nothing blocks on carrier at ADD time, so the flag "passes" on IPv4
even though the host veth is only brought up later. On an IPv6 cluster `eth0`
itself is IPv6, so the DAD gate is on the pod's primary interface and the failure
is immediate and total.

viveksb007 (changes requested):
Need to fix logic for IPv6 cluster.

cmd/routed-eni-cni-plugin/driver/driver.go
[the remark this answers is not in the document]
amrmahdi: Thanks @viveksb007 for the testing and the catch. I pushed a fix based on your feedback, can you run your IPv6 test and confirm it addresses the issue?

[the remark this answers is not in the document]
viveksb007: @amrmahdi thanks for the changes, this looks good now. Would you able to remove the flag `AWS_VPC_K8S_CNI_VETH_PEER_NAMESPACE ` and make this default behavior of CNI?

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -124,20 +126,45 @@ func (createVethContext *createVethPairContext) run(hostNS ns.NetNS) error {
...
+	} else {
+		// The container end only has carrier when both veth ends are up, and the
+		// IPv6 DAD wait below needs carrier. Bring the host end up now; net.FlagUp
+		// at create time is not reliably applied on all platforms.
+		if err := hostNS.Do(func(ns.NetNS) error {
+			hv, err := createVethContext.netLink.LinkByName(createVethContext.hostVethName)
+			if err != nil {
+				return errors.Wrapf(err, "setup NS network: failed to find host link %q", createVethContext.hostVethName)
+			}
+			if err := createVethContext.netLink.LinkSetUp(hv); err != nil {
```
Pavani-Panakanti: There is a behavior change in ordering here -  In peer mode we bring the host veth up inside this hostNS.Do block, but the host-side sysctls (accept_ra=0, accept_redirects=1, forwarding=0) still gets applied later in setupVeth after run() returns. So there's a small window where the host end is already up but accept_ra is still at the node default

Did some digging into these accept_ra default 1 and why cni sets to 0 -- 
```
the host veth's only neighbor on that link is the pod's eth0, so the only thing that can send a Router Advertisement to it is the pod itself. If the host end accepts an RA from the pod while it's still at 1, the impact would be like — the pod installing a default IPv6 route on the host pointing at itself (traffic redirection / MITM), poisoning the host's gateway/prefix, or changing the interface MTU. Basically a pod influencing host routing, which is exactly what accept_ra=0 is there to prevent. 
```
In the move path this can't happen. The host end gets moved and move clears IFF_UP, so it arrives DOWN. CNI sets the sysctls (accept_ra=0, accept_redirects=1, forwarding=0) and then LinkSetUp

Practically nothing in pod runs before CNI add is done so not an issue but since we are planning to make this default its better to keep the earlier ordering. Moving those three sysctl writes into this block before LinkSetUp(hv) would match the move path.

amrmahdi:
@viveksb007 @Pavani-Panakanti done. Two commits: first the sysctl ordering fix (good catch, sysctls now run before the link comes up, same as the old path), then removed the flag and the old create-then-move path. PeerNamespace is now the only path. One note: this also covers the branch-ENI veth, which the flag had excluded. Same code path, but calling it out so your testing covers it.

cmd/routed-eni-cni-plugin/cni_test.go
viveksb007, line 1740: the new test added below `custom veth peer namespace` is exactly similar to this one. we can omit remove new one

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -407,40 +447,12 @@ func (n *linuxNetwork) setupVeth(hostVethName string, contVethName string, netns
 		return nil, errors.Wrap(err, "failed to setup veth network")
 	}
 
+	// The IPv6 sysctls and bring-up of the host end happen inside the pod
```
viveksb007: minor - this reads like the sysctl and host end is coming up in pod ns which is what we are modifying. something like below 
```// The host end was created directly in the host namespace by run(); its
	// IPv6 sysctls and bring-up were applied there (via hostNS.Do) before the
	// link came up.
```

cmd/routed-eni-cni-plugin/driver/driver.go
```diff
@@ -123,21 +123,40 @@ func (createVethContext *createVethPairContext) run(hostNS ns.NetNS) error {
...
-	if err != nil {
-		return errors.Wrapf(err, "setup NS network: failed to find link %q", createVethContext.hostVethName)
-	}
-
-	// Explicitly set the veth to UP state, because netlink doesn't always do that on all the platforms with net.FlagUp.
-	// veth won't get a link local address unless it's set to UP state.
-	if err = createVethContext.netLink.LinkSetUp(hostVeth); err != nil {
-		return errors.Wrapf(err, "setup NS network: failed to set link %q up", createVethContext.hostVethName)
+	hostHardwareAddr := createVethContext.hostMACAddr
+	{
```
viveksb007: this bracket at open 138- close 160 seems redundant, can remove it

cmd/routed-eni-cni-plugin/driver/driver_test.go
```diff
@@ -3522,6 +3189,304 @@ func Test_createVethPairContext_run(t *testing.T) {
...
+			Name:  "eni8ea2c11fe35",
+			Index: 9,
+		},
+	}
+	netLink := mock_netlinkwrapper.NewMockNetLink(ctrl)
+	netLink.EXPECT().LinkAdd(gomock.Any()).Return(nil)
+	netLink.EXPECT().LinkByName("eni8ea2c11fe35").Return(hostVeth, nil)
+
+	procSys := mock_procsyswrapper.NewMockProcSys(ctrl)
+	procSys.EXPECT().Set("net/ipv6/conf/eni8ea2c11fe35/accept_ra", "0").Return(errors.New("some error"))
```
viveksb007: can we cover all failure branches in function `setHostVethV6Sysctls ` via UT (some table driven test works)?

currently the function has 3 failure branch, in this test we are hitting 1 of those.

cmd/routed-eni-cni-plugin/driver/driver_test.go
```diff
@@ -42,9 +44,11 @@ var testLogCfg = logger.Configuration{
 var testLogger = logger.New(&testLogCfg)
 
 type vethMatcher struct {
-	contVethName string
-	flags        net.Flags
-	mtu          int
+	contVethName       string
+	flags              net.Flags
+	mtu                int
+	checkPeerNamespace bool
```
viveksb007: don't think we need this `checkPeerNamespace` flag in test now as we are getting rid of branching behavior, can you double check and remove it

amrmahdi:
@viveksb007 all five addressed, thanks for the review.

viveksb007 (approved):
lgtm, thanks @amrmahdi for these changes

PR #3773, opened 2026-07-09
test/eni-subnet-discovery: bound subnet teardown to 5m poll and fail loud on leak

yash97:
**What type of PR is this?**

testing

N/A. Hardens teardown of the `eni-subnet-discovery` integration suite so it does not leak the secondary subnet it creates.

The suite's `AfterSuite` creates a subnet in the secondary CIDR (`100.64.0.0/16`), then deletes it after a fixed 90s sleep. When the CNI has not finished draining the secondary ENIs within that window, `DeleteSubnet` returns `DependencyViolation`. The error was appended to a `MultiError` that was only logged as a warning, so the suite reported success while leaking the subnet. A leaked subnet in the secondary CIDR then blocks teardown of the cluster VPC.

Observed teardown output (the suite passes while the subnet leak and the follow-on CIDR-disassociation both fail):

```
WARNING: Some cleanup operations failed: 2 error(s) occurred:
 * operation error EC2: DeleteSubnet, https response error StatusCode: 400,
   api error DependencyViolation: The subnet 'subnet-xxxxxxxxxxxxxxxxx' has dependencies and cannot be deleted.
 * operation error EC2: DisassociateVpcCidrBlock, https response error StatusCode: 400,
   api error InvalidCidrBlock.InUse: The vpc vpc-xxxxxxxxxxxxxxxxx currently has a subnet within CIDR block 100.64.0.0/16
[AfterSuite] PASSED
Ran 15 of 15 Specs
SUCCESS! -- 15 Passed | 0 Failed | 0 Pending | 0 Skipped
```

This change:

1. Replaces the fixed 90s sleep + single `DeleteSubnet` with a bounded `Eventually` poll (max 5m, 15s interval). Teardown waits for the CNI to drain the secondary ENIs so `DeleteSubnet` can succeed, but is time-bounded so it can never hang.
2. Bounds the CIDR-disassociation retry the same way (90s).
3. Asserts on the cleanup `MultiError` (`Expect(...).ToNot(HaveOccurred())`) so a residual leak fails the suite instead of being logged and swallowed.

**AWS APIs invoked**: no new APIs. Same `DeleteSubnet` / `DisassociateVpcCidrBlock` / `DisassociateRouteTable` calls as before, now issued on a 15s poll interval within the bounded windows above (worst case ~20 `DeleteSubnet` calls over 5m).

- `go vet ./test/integration/eni-subnet-discovery/` passes.
- Package compiles.

No. Uses `Eventually` from the Gomega matcher library already vendored in the repo.

No. Test-only change.

No.

No.

```release-note
NONE
```


haouc:
LGTM. Nice root-cause fix.
One non-blocking ask: can you confirm it's been exercised in a full integration run against a slow-drain scenario (not just compile)? That's the only real validation of the poll logic. Happy to approve on confirmation.

haouc (approved):

Issue #150, opened 2018-08-07
vpc_ip_resource_limit is missing r5 instances


mejran:
This is causing our r5 instances to get no additional IPs assigned to them since it defaults to 0 for unknown instance types.

seb-steuer:
+1 I just found out the hard way ;) and spent a lot of time debugging my EKS cluster

oded-dd:
+1

seb-steuer:
there is already a pull request for that :)
https://github.com/aws/amazon-vpc-cni-k8s/pull/145

liwenwu-amazon:
This should be resolved with https://github.com/aws/amazon-vpc-cni-k8s/pull/145.  Please reopen it if this is still an issue to you.

PR #3782, opened 2026-07-16
fix: tolerate concurrent NewChain race in egress-cni snat.Add

imuni4fun:
## Problem

When many pods start simultaneously on the same node (e.g. at a 5-minute cron tick on a jobs node), multiple concurrent invocations of `egress-cni` race to create the same iptables chain. Each invocation snapshots existing chains via `ListChains`, sees the chain absent, then calls `NewChain` — the second concurrent invocation fails with:

```
iptables -t nat -N POSTROUTING --wait: exit status 1: iptables: Chain already exists.
```

This causes the pod sandbox setup to fail entirely, leaving the pod stuck in `Init:0/1`. On busy nodes receiving a burst of simultaneous pod starts (observed with 10+ pods scheduled at the same cron tick), this failure is consistent and reproducible.

The error propagates back to the container runtime as:
```
Failed to create pod sandbox: plugin type=&#34;egress-cni&#34; name=&#34;egress-cni&#34; failed (add):
running [/usr/sbin/iptables -t nat -N POSTROUTING --wait]: exit status 1: iptables: Chain already exists.
```

## Root Cause

In `snat.Add()`, the existing chains check is a non-atomic snapshot+check+create sequence:

```go
// Step 1: snapshot (non-atomic)
chains, _ := ipt.ListChains(&#34;nat&#34;)
existingChains := map[string]bool{...}

// Step 2: create if absent — RACE WINDOW
if !existingChains[_chain] {
    if err = ipt.NewChain(&#34;nat&#34;, _chain); err != nil {
        return err  // ← fails on concurrent second call
    }
}
```

Note: the existing `pkg/networkutils` package already handles this exact race with a `containChainExistErr` helper. This fix applies the same pattern to the egress-cni snat package.

## Fix

Tolerate the `&#34;Chain already exists&#34;` error from `NewChain` — it is benign when caused by a concurrent invocation that won the race and successfully created the chain first. A chain that already exists is exactly what we want.

The chain-exists check is extracted to a new exported helper `IsChainExistErr()` in `pkg/utils/cniutils` (alongside the existing `IsIptableTargetNotExist()`), using a typed `*iptables.Error` check (`ExitStatus() == 1`) — the same sentinel used by `ClearChain` in `github.com/coreos/go-iptables`:

```go
// Tolerate "Chain already exists" from a concurrent invocation that won the race.
if err = ipt.NewChain("nat", _chain); err != nil && !cniutils.IsChainExistErr(err) {
    return err
}
```

The subsequent `AppendUnique` call is already idempotent, so no further changes are needed.

## Testing

This race is observable in production on IPv6 EKS clusters with `ENABLE_IPv4=false` + `egress-cni` enabled, when 10+ pods start simultaneously on the same node. The fix makes `NewChain` idempotent under concurrent execution.

New unit tests added:
- `TestAddV4_ChainAlreadyExists` — `Add()` must succeed when `NewChain` returns `"Chain already exists"` (the exact production race scenario)
- `TestIsChainExistErr` — unit-tests the new `cniutils.IsChainExistErr` helper

## Production Validation

This fix has been applied to an internal fork (`acquia/kaas-aws-vpc-cni`, [PR #150](https://github.com/acquia/kaas-aws-vpc-cni/pull/150)) and is pending production rollout. The race was confirmed via incident prod-saturnc368 (2026-07-15): a kaas-workflow rollback that reduced the per-node concurrency below the race threshold immediately stopped the `FailedCreatePodSandBox` failures, directly confirming the race condition theory. The internal fork PR is open and will adopt the upstream merge once this PR lands.

### DIFF
--- a/cmd/egress-cni-plugin/snat/snat.go
+++ b/cmd/egress-cni-plugin/snat/snat.go
@@ -15,6 +15,7 @@ package snat
 
 import (
 	"net"
+	"strings"
 
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/iptableswrapper"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/cniutils"
@@ -72,7 +73,7 @@ func Add(ipt iptableswrapper.IPTablesIface, nodeIP, src net.IP, multicastRange, 
 	for _, rule := range rules {
 		_chain := rule[0]
 		if !existingChains[_chain] {
-			if err = ipt.NewChain("nat", _chain); err != nil {
+			if err = ipt.NewChain("nat", _chain); err != nil && !chainExistErr(err) {
 				return err
 			}
 			existingChains[_chain] = true
@@ -87,6 +88,14 @@ func Add(ipt iptableswrapper.IPTablesIface, nodeIP, src net.IP, multicastRange, 
 	}
 
 	return nil
+}
+
+// chainExistErr returns true if the error indicates an iptables chain already exists.
+// This can occur as a benign race when multiple CNI invocations run concurrently on the
+// same node (e.g. many pods starting simultaneously at a cron tick) and both attempt to
+// create the same chain at the same instant.
+func chainExistErr(err error) bool {
+	return strings.Contains(err.Error(), "Chain already exists")
 }
 
 // Del removes rules added by snat
--- a/cmd/egress-cni-plugin/snat/snat.go
+++ b/cmd/egress-cni-plugin/snat/snat.go
@@ -15,7 +15,6 @@ package snat
 
 import (
 	"net"
-	"strings"
 
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/iptableswrapper"
 	"github.com/aws/amazon-vpc-cni-k8s/pkg/utils/cniutils"
@@ -73,7 +72,13 @@ func Add(ipt iptableswrapper.IPTablesIface, nodeIP, src net.IP, multicastRange, 
 	for _, rule := range rules {
 		_chain := rule[0]
 		if !existingChains[_chain] {
-			if err = ipt.NewChain("nat", _chain); err != nil && !chainExistErr(err) {
+			// Tolerate "Chain already exists" from a concurrent invocation that won
+			// the race. On a node with many pods starting simultaneously (e.g. at a
+			// cron tick), multiple egress-cni processes snapshot ListChains before
+			// any has called NewChain and all race to create the same chain. The
+			// second caller gets this error, but the chain is already in the desired
+			// state — it is safe to proceed. See https://github.com/aws/amazon-vpc-cni-k8s/pull/3782.
+			if err = ipt.NewChain("nat", _chain); err != nil && !cniutils.IsChainExistErr(err) {
 				return err
 			}
 			existingChains[_chain] = true
@@ -88,14 +93,6 @@ func Add(ipt iptableswrapper.IPTablesIface, nodeIP, src net.IP, multicastRange, 
 	}
 
 	return nil
-}
-
-// chainExistErr returns true if the error indicates an iptables chain already exists.
-// This can occur as a benign race when multiple CNI invocations run concurrently on the
-// same node (e.g. many pods starting simultaneously at a cron tick) and both attempt to
-// create the same chain at the same instant.
-func chainExistErr(err error) bool {
-	return strings.Contains(err.Error(), "Chain already exists")
 }
 
 // Del removes rules added by snat
--- a/cmd/egress-cni-plugin/snat/snat.go
+++ b/cmd/egress-cni-plugin/snat/snat.go
@@ -97,18 +97,29 @@ func Add(ipt iptableswrapper.IPTablesIface, nodeIP, src net.IP, multicastRange, 
 
 // Del removes rules added by snat
 func Del(ipt iptableswrapper.IPTablesIface, src net.IP, chain, comment string) (err error) {
+	// Tolerate "rule does not exist" — a concurrent DEL invocation may have
+	// already removed this POSTROUTING jump rule before we reached this point.
+	// The desired end state (rule absent) is already achieved; proceed.
 	err = ipt.Delete("nat", "POSTROUTING", "-s", src.String(), "-j", chain, "-m", "comment", "--comment", comment)
-	if err != nil && !cniutils.IsIptableTargetNotExist(err) {
+	if err != nil && !cniutils.IsChainNotExistErr(err) {
 		return err
 	}
 
+	// Tolerate "No chain/target/match" — a concurrent DEL invocation may have
+	// already cleared and deleted this chain. ClearChain internally calls
+	// NewChain; if the chain was already removed it will be re-created as an
+	// empty chain, so a genuine "not exist" return here would be unexpected.
+	// Guard defensively for any implementation that does return it.
 	err = ipt.ClearChain("nat", chain)
-	if err != nil && !cniutils.IsIptableTargetNotExist(err) {
+	if err != nil && !cniutils.IsChainNotExistErr(err) {
 		return err
 	}
 
+	// Tolerate "No chain/target/match" — a concurrent DEL invocation may have
+	// already deleted the chain (after we cleared it or before we got here).
+	// Either way the chain is gone, which is the desired post-DEL state.
 	err = ipt.DeleteChain("nat", chain)
-	if err != nil && !cniutils.IsIptableTargetNotExist(err) {
+	if err != nil && !cniutils.IsChainNotExistErr(err) {
 		return err
 	}
 
--- a/cmd/egress-cni-plugin/snat/snat_test.go
+++ b/cmd/egress-cni-plugin/snat/snat_test.go
@@ -170,3 +170,61 @@ func setupDelExpect(ipt iptableswrapper.IPTablesIface, actualClearChain, actualD
 		*actualRule = append(*actualRule, rule)
 	}).Return(nil).AnyTimes()
 }
+
+// TestDelV4_ClearChainNotExist verifies the teardown DEL+DEL race fix: when
+// ClearChain returns "No chain/target/match by that name" (because a concurrent
+// DEL invocation already removed the chain), Del() must succeed rather than
+// propagating the error.
+func TestDelV4_ClearChainNotExist(t *testing.T) {
+	ctrl := gomock.NewController(t)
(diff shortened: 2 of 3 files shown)


cmd/egress-cni-plugin/snat/snat.go
```diff
@@ -89,6 +90,14 @@ func Add(ipt iptableswrapper.IPTablesIface, nodeIP, src net.IP, multicastRange,
 	return nil
 }
 
+// chainExistErr returns true if the error indicates an iptables chain already exists.
+// This can occur as a benign race when multiple CNI invocations run concurrently on the
+// same node (e.g. many pods starting simultaneously at a cron tick) and both attempt to
+// create the same chain at the same instant.
+func chainExistErr(err error) bool {
+	return strings.Contains(err.Error(), "Chain already exists")
```
haouc: Relying on the exact error string is brittle. I would recommend to replace the fragile strings.Contains(err.Error(), "Chain already exists") check with a more robust, typed approach.

cmd/egress-cni-plugin/snat/snat.go
```diff
@@ -72,7 +73,7 @@ func Add(ipt iptableswrapper.IPTablesIface, nodeIP, src net.IP, multicastRange,
 	for _, rule := range rules {
 		_chain := rule[0]
 		if !existingChains[_chain] {
-			if err = ipt.NewChain("nat", _chain); err != nil {
+			if err = ipt.NewChain("nat", _chain); err != nil && !chainExistErr(err) {
```
haouc: I think the comment added to the helper function should be added here. Can you add a short comment referencing why ignoring this specific error is safe and desirable (race condition when multiple CNIs run) and possibly link to an issue/commit where this was observed.

haouc:
Thanks for the PR. Can you also confirm if the fix worked in your cluster.

cmd/egress-cni-plugin/snat/snat.go
```diff
@@ -89,6 +90,14 @@ func Add(ipt iptableswrapper.IPTablesIface, nodeIP, src net.IP, multicastRange,
 	return nil
 }
 
+// chainExistErr returns true if the error indicates an iptables chain already exists.
+// This can occur as a benign race when multiple CNI invocations run concurrently on the
+// same node (e.g. many pods starting simultaneously at a cron tick) and both attempt to
+// create the same chain at the same instant.
+func chainExistErr(err error) bool {
```
jaydeokar: Can we reuse the existing functions instead of defining here ? 
Also do add a UT for this behavior

yash97:
Following up on this PR, any update? I noticed it is still in draft mode.

imuni4fun:
**Follow-up: teardown DEL+DEL race also identified and fixed in this PR branch**

While reviewing the Add() race fix, I also identified a symmetric race in the `Del()` teardown path and have added the fix to this branch (commit `eeab869`).

**The teardown race:**

When two concurrent `DEL` invocations tear down the same pod's SNAT rules simultaneously, the second invocation can arrive at any of these three steps after the first has already completed them:

- `Delete("POSTROUTING" jump rule)` → `"No chain/target/match by that name"` — rule already removed
- `ClearChain(chain)` → same — chain already cleared+deleted  
- `DeleteChain(chain)` → same — chain deleted between our `ClearChain` and `DeleteChain` calls

All three are benign: the desired post-DEL state (chain and rules absent) is already achieved. The previous code used `IsIptableTargetNotExist` which requires a typed `*iptables.Error`; the fix uses the new `IsChainNotExistErr` (added to `pkg/utils/cniutils` alongside `IsChainExistErr`) which adds a string fallback and has more precise semantics at the call site.

**What changed:**
- `pkg/utils/cniutils/cni_utils.go`: added `IsChainNotExistErr` (and moved `IsChainExistErr` here from the local `chainExistErr` in `snat.go`)
- `cmd/egress-cni-plugin/snat/snat.go`: `Del()` now uses `IsChainNotExistErr` at all three call sites with inline comments explaining the race; `Add()` now uses `cniutils.IsChainExistErr` (the local `chainExistErr` helper is removed)
- `cmd/egress-cni-plugin/snat/snat_test.go`: added `TestDelV4_ClearChainNotExist`, `TestDelV4_DeleteChainNotExist`, `TestIsChainNotExistErr`

yash97:
Could we get an integration test here, or at least a repro we can point to that fails on master and passes with this change? The unit test mocks the "Chain already exists" error, so it shows we handle that error but doesn't actually trigger the race itself.

yash97:
Also what is your iptable backend?

pkg/utils/cniutils/cni_utils.go
```diff
@@ -147,6 +147,36 @@ func IsIptableTargetNotExist(err error) bool {
...
+// IsChainNotExistErr returns true if err indicates that an iptables chain or
+// rule does not exist. This is the expected error when a concurrent DEL
+// invocation has already removed the chain or rule before this invocation
+// reaches the same operation — the desired end state (chain gone) is already
+// achieved and it is safe to proceed.
+//
+// The primary check uses the typed *iptables.Error.IsNotExist() which matches
+// "No chain/target/match by that name" and related patterns. A string fallback
+// covers test doubles that return plain errors rather than *iptables.Error.
+func IsChainNotExistErr(err error) bool {
```
yash97, line 174: IsChainExistErr(nil) panics at err.Error() in the fallback. Its sibling IsIptableTargetNotExist(nil) safely returns false. The call site guards with err != nil &&, but an exported helper
   shouldn't be less nil-safe than the function directly above it — add if err == nil { return false } (and a test case).

pkg/utils/cniutils/cni_utils.go
```diff
@@ -147,6 +147,36 @@ func IsIptableTargetNotExist(err error) bool {
...
+// sentinel used by ClearChain in github.com/coreos/go-iptables). A string
+// fallback covers test doubles that return plain errors rather than *iptables.Error.
+func IsChainExistErr(err error) bool {
+	if e, ok := err.(*iptables.Error); ok {
+		return e.ExitStatus() == 1
+	}
+	return strings.Contains(err.Error(), "Chain already exists")
+}
+
+// IsChainNotExistErr returns true if err indicates that an iptables chain or
```
yash97, line 165: Also lets, keep comment concise and keep it if they provide some value.

cmd/egress-cni-plugin/snat/snat_test.go
```diff
@@ -89,6 +90,51 @@ func TestDelV4(t *testing.T) {
 	assert.EqualValuesf(t, expectRule, actualRule, "iptables rule is expected to be removed")
 }
 
+// TestAddV4_ChainAlreadyExists verifies that Add() succeeds when NewChain
```
yash97, line 93: Test assertions don't enforce what the comments claim

   TestAddV4_ChainAlreadyExists uses .AnyTimes() on both NewChain and AppendUnique — the comment says "AppendUnique must still be called for all rules," but AnyTimes() passes even if it's
   never called. Use NewChain(...).Times(2) and AppendUnique(...).Times(3) so the "tolerate and proceed" behavior is actually pinned.

   Also: every test error is a plain fmt.Errorf, so the typed *iptables.Error branch — the actual production path — has zero coverage; all four TestIsChainExistErr cases go through the
   string fallback (the "exit status 2" case reads as if exit-status logic is tested, but it isn't). I realize iptables.Error can't be constructed outside its package; if the typed branch is
   untestable, a comment acknowledging that would be honest. Relatedly, the doc comment justifies the string fallback as "covers test doubles" — production behavior shouldn't be justified by
   mock needs; defense-in-depth for wrapped errors is the honest rationale.

imuni4fun: i agree with all of this. see next general comment (not code specific).

imuni4fun:
Hello!

I apologize for the late communication and agree with the review comments.

This was actually drafted by Claude Sonnet 4.6.

I am not familiar with this codebase directly and I guided Claude with concerns about cleanup (DEL) and the test coverage concerns that you also noted. This includes pointing out that this is just testing the same things and making them "typed" but still string tests that just follow the norms in the repo.

As I am not deeply familiar with this repo's inner workings, I would appreciate any architectural guidance or to have you take over this if you have test harnesses and familiarity already. We believe we uncovered this problem but I am trying to guide AI tooling to build out the solution and I often don't trust it in our code bases where I AM very familiar.

Any help would be appreciated.

The main problem was described reasonably well but the implications on cleanup, etc are an area where my knowledge is lacking.

Personally, I am surprised there is so much optimistic concurrency in Linux. I generally prefer to do pessimistic locking to be SURE operations are atomic. I don't know enough to project where it is safe to use optimistic concurrency here and rely on retries from the kubelet in subsequent CNI calls. Our team has run into many complex issues relating to race conditions or corrupted state in Linux/Kubernetes which tie back to non-atomic actions.

Please let me know how I can help move the ultimate resolution along and we can figure out what role is appropriate and what test harnesses you may have available that you or I may utilize.

Thank you,
Jason

imuni4fun:
> Also what is your iptable backend?

do you mean `iptables` versus `nf tables`? if so, `iptables`

imuni4fun:
i'd be much more comfortable if a familiar human wrote this fix and tests versus an unfamilair human or an LLM 😅

yash97:
> do you mean iptables versus nf tables? if so, iptables

No iptables has legacy backend and nftable backend. Curious you see chain absent, is triggering this so nftable does not have this chain by default while legacy does have this chain, so wanted to get confirmation.

imuni4fun:
> No iptables has legacy backend and nftable backend. Curious you see chain absent, is triggering this so nftable does not have this chain by default while legacy does have this chain, so wanted to get confirmation

```
kubectl exec -n kube-system aws-node-cjcv9 -c aws-node -- iptables --version 2>&1
iptables v1.8.8 (nf_tables)
```

imuni4fun:
Hello,

Just checking in to see where we are on this race condition issue.

Please let me know if there is more that's needed from me or my team.

I think it's best that this is fixed by someone with direct working knowledge of the lifecycle and that's not us.

Thank you,
Jason

Issacwww:
Thanks, Jason. I have taken this over in #3826 because maintainer edits are not enabled on this branch. Please continue review there; implementation and validation details are in the new PR.

imuni4fun:
feel free to close this PR since the new one is where active development is occurring

cdirubbio:
Closing in favor of https://github.com/aws/amazon-vpc-cni-k8s/pull/3826

PR #3797, opened 2026-07-23
test: replace fixed sleeps with readiness polls in pod-eni and eni-subnet-discovery suites

yash97:
## What

Two integration-test suites use fixed sleeps where they should observe completion. Both cause flaky BeforeSuite failures that red the networking canaries. This replaces the guesses with bounded polls.

### 1. pod-eni (SGP) BeforeSuite — panic after node recycle

`security_group_per_pod_suite_test.go` calls `TerminateInstances(f)` (recycles the nodegroup), then immediately `GetNodes(...)` + `targetNode = nodeList.Items[0]`. `TerminateInstances` waits only for the ASG at the **EC2** level, not for replacement **Kubernetes Nodes** to re-register, so `GetNodes` can return an empty list and `Items[0]` panics:

```
[PANICKED] runtime error: index out of range [0] with length 0
  test/integration/pod-eni/security_group_per_pod_suite_test.go:126
FAIL! -- A BeforeSuite node failed so all tests were skipped.
```

Fix: `WaitTillNodesReady(...)` (existing framework primitive, bounded 15m) before listing, plus a non-empty guard before indexing. This mirrors the fix already shipped for the sibling `custom-networking-sgpp` suite in #3787, which missed the pod-eni suite.

### 2. eni-subnet-discovery — fixed 90s ENI-drain sleep

The suite sets `WARM_ENI_TARGET=0` then sleeps a fixed 90s hoping the CNI has detached warm ENIs — in BeforeSuite and again in the per-test JustAfterEach. Flaky (slow drain races the next assertion) and wasteful (always waits 90s).

Fix: `waitForSecondaryENIsDrained()` polls the instance's ENIs (`DescribeInstance` + `common.IsPrimaryENI`) until no secondary ENIs remain (bounded 5m), returning as soon as the drain completes.

## Scope note

Sibling scenario files in the eni-subnet-discovery suite (`enhanced` / `primary-exclusion` / `secondary-exclusion`) still contain fixed `sleep 30/45/60/90` calls — same anti-pattern, but converting them is left to a follow-up to keep this change reviewable. The two sleeps fixed here are in the core flow run by every scenario.

## Testing

`go vet ./test/integration/pod-eni/ ./test/integration/eni-subnet-discovery/` passes; `gofmt` clean. Test-only change; no product code touched.

### DIFF
--- a/test/integration/eni-subnet-discovery/eni_subnet_discovery_suite_test.go
+++ b/test/integration/eni-subnet-discovery/eni_subnet_discovery_suite_test.go
@@ -214,8 +214,8 @@ var _ = BeforeSuite(func() {
 	k8sUtils.AddEnvVarToDaemonSetAndWaitTillUpdated(f, utils.AwsNodeName, utils.AwsNodeNamespace,
 		utils.AwsNodeName, map[string]string{"WARM_ENI_TARGET": "0"})
 
-	By("sleeping to allow CNI Plugin to delete unused ENIs")
-	time.Sleep(time.Second * 90)
+	By("waiting for the CNI to delete unused ENIs")
+	waitForSecondaryENIsDrained()
 
 	createdSubnet = subnetID
 })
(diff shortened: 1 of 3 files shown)


parlakisik (approved):
lgtm

PR #3802, opened 2026-07-24
(bug) Return error if the parsing fails and UT

jaydeokar:
**What type of PR is this?** 
bug

**Which issue does this PR fix?**: If annotation is malformed, we do not handle the failure and return. 

SGP pod deletion is always handled by Prev Result stored. However in the rare cases where the Prev Result is not present, CNI falls back to IPAMD to get the VlanID and other metadata about SGP pod on delete flow. 
If the annotation present on the SGP pod is invalid or malformed, we are not returning an error which can cause other issues

Fixed similar issues in other parts of the code where the result gets used before checking the error

**Testing done on this change**: N/A. This is rare condition to get triggered as we block that annotation to get created from webhook and Prev Result deletion has to fail due to some reason. 

No

```release-note

```


parlakisik (dismissed):
lgtm

parlakisik (approved):
lgtm

PR #3810, opened 2026-07-29
test: fix flaky ENI-placement precondition in cni integration suite

yash97:
## Problem

Several cni integration specs (snat_kube_proxy, pod_traffic, soak, host_networking) intermittently fail their shared precondition:

```
[FAIL] test SNAT with kube-proxy modes [BeforeEach]
Expected <int>: 0 to be >= <int>: 1
```

Each spec creates a deployment pinned to one node and asserts pods land on both the primary and a secondary ENI. The primary bucket sometimes comes back empty.

## Failure modes found (validated on live clusters, c5.xlarge + m5.large)

**1. Non-deterministic ENI placement + IP cooldown carryover (the original flake).**
ipamd assigns pod IPs by ranging over Go maps (`AssignPodIPv4Address`) — no primary-first order. A prior spec teardown puts freed IPs into a 30s cooldown, leaving an already-attached secondary ENI as the only source of eligible IPs; all pods can land there. The old `maxIPPerInterface*2` replica count (26 on c5.xlarge) is below secondary-ENI capacity (42), so nothing forces primary-ENI placement.
*Fix:* leave `IP_COOLDOWN_PERIOD` unset so ipamd supplies its 30-second product default, and size the deployment by pigeonhole: `(maxENIs-1)*(ipsPerENI-1)+2` pods — more than all secondary ENIs can hold, so >=2 land on the primary for any placement order. Preconditions assert the assumptions (>=2 ENIs, no prefix delegation, no custom networking, no trunk ENI) per node and diagnostics dump ENI/IP layout and pod phases on failure. The fix is deterministic under cooldown carryover rather than disabling that production safeguard.

**2. IP pool growth too slow for the burst (exposed by fix 1).**
With `WARM_IP_TARGET=3`/`WARM_ENI_TARGET=0`, ipamd grows the pool a few IPs per ~45s reconcile round; a 44-pod burst cannot converge inside the 300s deployment wait (observed: `FailedCreatePodSandBox: failed to assign an IP address`, pool at 3->6->9 across rounds). A diagnostic cooldown-zero run also released the pool immediately after teardown and re-paid this warmup; that diagnostic setting is not part of the final design.
*Fix:* `WARM_ENI_TARGET=1` — ENI-granular growth, one spare ENI persists across teardowns. `WARM_IP_TARGET` must be removed (ipamd ignores `WARM_ENI_TARGET` while it is set), including host_networking's per-spec bump. `IP_COOLDOWN_PERIOD` remains unset so the product default continues protecting kube-proxy/network-proxy convergence.

**3. Daemonset-ready wait shorter than aws-node startup (pre-existing, amplified by 44-pod nodes).**
`CheckIfDaemonSetIsReady` gave up after 4x5s=20s. aws-node startup includes a 30s background wait for API-server connectivity, and the SNAT specs restart kube-proxy right before restarting aws-node, extending that window (measured ~56s to Ready; checkpoint recovery itself is ~0.3s). Any spec restarting aws-node under load flaked.
*Fix:* 2-minute poll budget (returns immediately when ready) — a calibration fix, the wait was shorter than the healthy operation it waits for. Follow-up worth filing: aws-node restart during kube-proxy reprogramming rides a 30s+ API blackout.

**4. service_connectivity ENI-count snapshot races ipamd reconciliation (found while validating fix 2).**
The <=3-ENI leak check asserts a snapshot, but with a warm ENI a prior spanning spec can leave 4 attached and ipamd detaches surplus over ~45s cycles; the spec only passed via its FlakeAttempts retry.
*Fix:* poll for convergence (3m) — a genuine leak never converges, detection intent unchanged, retry no longer needed.

## Constraints held

No retries or skips added, no assertion weakening (`>=1`, `>1`, `>0` intact — zero assertion lines changed in the diff), and no product runtime-code changes. Changes are limited to tests, test framework code, and build-test wiring.

## Validation (live clusters, us-west-2)

- Focused matrix (SNAT + pod networking + host networking): 8/8 PASS across c5.xlarge (4 ENI/15 IP) and m5.large (3 ENI/10 IP), including fresh (cold-start) clusters and freshly recycled nodes.
- Full unfocused cni suite (120m): 29 passed, 0 assertion failures; 1 spec interrupted by the suite deadline + 3 skipped behind it (all host_networking specs already covered 8x in the focused matrix); 1 flaked-then-passed spec was failure mode 4, now fixed.
- Unit test locks the pigeonhole algebra including the fits-max-pods-by-construction invariant (`TestSpanningENIsReplicaCount`); `go test ./test/integration/common/...`, `go vet ./test/...`, and `go build ./test/integration/...` pass. GitHub unit, image-build, dependency-review, govulncheck, and Codecov checks are green.
- Full end-to-end `scripts/run-cni-release-tests.sh` (unmodified) PASS on an EKS 1.33 c5.xlarge cluster in us-east-1, with the product-default cooldown and custom-built CNI/init images: ipamd 25 passed / 0 failed (1 skipped), cni 21 passed / 0 failed (12 skipped, 2202s — well under the 120m budget), cni-metrics-helper passed, core plugins passed, exit code 0, zero retries or flaked specs. Parallel runs on a second c5.xlarge and an m5.large / Kubernetes 1.36 cluster reproduced clean ipamd and cni suites; their remaining stages and a confirmation run at the final HEAD are in progress.


test/integration/common/util.go
```diff
@@ -160,6 +163,168 @@ func GetPodsOnPrimaryAndSecondaryInterface(node coreV1.Node,
...
+// SpanningENIsReplicaCount returns the number of pods needed so that placement
+// must occupy the primary ENI and at least one secondary ENI. ipamd assigns pod
+// IPs by ranging over a Go map of ENIs, so placement order is not deterministic;
+// requesting two more pods than every secondary ENI can hold forces at least two
+// onto the primary ENI regardless of iteration order.
+func SpanningENIsReplicaCount(netInfo ec2types.NetworkInfo) int {
+	maxENIs := int(*netInfo.MaximumNetworkInterfaces)
+	ipsPerENI := int(*netInfo.Ipv4AddressesPerInterface)
+	return (maxENIs-1)*(ipsPerENI-1) + 2
+}
```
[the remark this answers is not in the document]
yash97: Addressed in 414dbc28 and tightened in 2c1bd3ad. This was legitimate: the SDK fields are pointers, and very small per-ENI capacity invalidates the spanning guarantee. SpanningENIsReplicaCount now returns explicit errors for missing limits, fewer than 2 ENIs, or fewer than 3 addresses per ENI (one primary address plus two pod-IP slots); AssertSpanningENIsPreconditions owns validation before returning the replica count. Invalid-NetworkInfo unit cases cover each boundary and now run in presubmit. I also handled the related suppressed concern by asserting NetworkInfo is non-nil before dereferencing it.

test/integration/common/util.go
```diff
@@ -160,6 +163,168 @@ func GetPodsOnPrimaryAndSecondaryInterface(node coreV1.Node,
...
+	env := map[string]string{}
+	for _, container := range ds.Spec.Template.Spec.Containers {
+		if container.Name != utils.AwsNodeName {
+			continue
+		}
+		for _, e := range container.Env {
+			env[e.Name] = e.Value
+		}
+	}
+	return env
```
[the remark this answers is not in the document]
yash97: Addressed in 414dbc28. This was also legitimate: an absent aws-node container previously produced an empty map, making both boolean preconditions look disabled. getAWSNodeEnv now tracks whether the named container was found and fails with the daemonset/container identity when it is absent.

test/framework/resources/agent/traffic_tester.go
```diff
@@ -62,9 +62,37 @@ type TrafficTest struct {
 
 // Tests traffic by creating multiple server pods using a deployment and multiple client pods
 // using a Job. Each client Pod tests connectivity to each Server Pod.
-func (t *TrafficTest) TestTraffic() (float64, error) {
+func (t *TrafficTest) TestTraffic() (successRate float64, err error) {
```
haouc, line 67: Please aggregate cleanup errors and enforce bounded timeouts in the deferred teardown.

Suggested changes (high-level):
- Collect all deletion errors instead of keeping only the first. On Go 1.20+ you can use errors.Join; otherwise append errors and return a combined message.
- Make DeleteAndWaitTill* calls in the defer use a context with a bounded timeout so a single stuck delete cannot hang test teardown indefinitely.

Example pattern to adopt inside the defer:
```
    // collect deletion failures
    var cleanupErrs []error
    cleanup := func(name string, del func() error) {
        // use a timeout so teardown cannot hang forever
        ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
        defer cancel()

        // If the manager API supports context, prefer passing ctx; otherwise run the blocking call in a goroutine and select on ctx.Done()
        if delErr := del(); delErr != nil {
            cleanupErrs = append(cleanupErrs, fmt.Errorf("%s deletion failed: %w", name, delErr))
        }
    }

    // after running all cleanup attempts:
    if err == nil && len(cleanupErrs) > 0 {
        // prefer errors.Join(cleanupErrs...) on Go 1.20+ so all errors are preserved
        err = fmt.Errorf("teardown failures: %v", cleanupErrs)
        successRate = 0
    }
```
If the underlying DeleteAndWaitTill* helpers accept a context, pass ctx directly to them; otherwise wrap the call and enforce the timeout via goroutine + select.

Please add a short comment explaining the intent (we want teardown failures surfaced and teardown bounded by a sane timeout).

test/integration/cni/snat_kube_proxy_test.go
```diff
@@ -248,6 +241,26 @@ func setKubeProxyMode(mode string) error {
...
+// Ref: https://github.com/kubernetes/kubeadm/issues/3133
+func cleanupIPVSLeftovers() error {
+	nodes, err := f.K8sResourceManagers.NodeManager().GetNodes(f.Options.NgNameLabelKey, f.Options.NgNameLabelVal)
+	if err != nil {
+		return err
+	}
+	cleanup := "ipvsadm --clear 2>/dev/null; ip link del kube-ipvs0 2>/dev/null; " +
+		"[ ! -e /sys/class/net/kube-ipvs0 ] && ! grep -qE \"^(TCP|UDP|SCTP)\" /proc/net/ip_vs 2>/dev/null"
+	for _, node := range nodes.Items {
+		if _, err := execNodeShell(node.Name, cleanup); err != nil {
```
haouc: Should we make the node-level ipvs cleanup more robust: enforce a timeout for the remote exec and include stdout/stderr in errors.
```
out, err := execNodeShellWithTimeout(node.Name, cleanupCmd, 30*time.Second)
```

test/framework/resources/agent/traffic_tester.go
yash97: Addressed in a8c11d75. Confirmed the concern is real: the job and deployment delete-and-wait helpers poll on context.Background() with no deadline, so teardown could hang indefinitely. Each deletion now runs under a 2-minute bound (the managers do not accept a context, so goroutine + select), all failures are collected and surfaced with errors.Join, and successRate is zeroed only when the test had not already failed. Comment explains the intent.

test/integration/cni/snat_kube_proxy_test.go
yash97: Addressed in a8c11d75. Added execNodeShellWithTimeout (exec.CommandContext + CombinedOutput) and the ipvs cleanup now uses it with a 30s bound, including the command output in the error message. Follow-up 6a8dcea0 fixes a missing import caught by go vet.

haouc (approved):

PR #3820, opened 2026-08-04
test: pin busybox and curl image

Issacwww:
**What type of PR is this?**
testing

N/A

`busybox:latest` and `curlimages/curl:latest` in the test registry were last pushed years ago and have accumulated critical CVEs — nothing referenced a version, so nothing ever refreshed them. Pins both to current versions (`busybox:1.38.0`, `curlimages/curl:8.21.0`), matching how every other image in this file is already pinned (`nginx:1.31.2`, `netcat-openbsd:v1.0`). Both tags are published as multi-arch image indexes (linux/amd64 + linux/arm64).

- `docker buildx imagetools inspect` confirms both tags are image indexes with linux/amd64 + linux/arm64

**Will this PR introduce any new dependencies?**: No.

**Will this break upgrades or downgrades? Has updating a running cluster been tested?**: No — test-only.

**Does this change require updates to the CNI daemonset config files to work?**: No.

**Does this PR introduce any user-facing change?**: No.

```release-note
NONE

### DIFF
--- a/test/framework/utils/const.go
+++ b/test/framework/utils/const.go
@@ -25,10 +25,10 @@ const (
 
 	// See https://gallery.ecr.aws/eks/aws-vpc-cni-test-helper
 	TestAgentImage = "networking-e2e-test-images/aws-vpc-cni-test-helper:20231212"
-	BusyBoxImage   = "networking-e2e-test-images/busybox:latest"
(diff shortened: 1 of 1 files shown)


jaydeokar (approved):
lgtm

PR #3825, opened 2026-08-09
fix(ipamd): retain prefix tracking in seenIPs and defer cooldown evic…

### LINKED ISSUE #3824: ipamd releases an in-use /28 prefix after 46h of incomplete datastore adds and a false "stale/dead" sandbox verdict (prefix delegation; recurring across multiple versions)
**What happened**:

On a single-ENI node running VPC CNI `v1.22.4-eksbuild.3` with `ENABLE_PREFIX_DELEGATION=true` and `WARM_PREFIX_TARGET=1`, ipamd unassigned the prefix `10.205.102.160/28` from the node's primary ENI while a running pod still held `10.205.102.160` from it. The pod kept its veth, route and `ip rule`, but its address was no longer assigned to any ENI, so every packet it sent was silently discarded by the VPC fabric. Pods on the surviving prefix were unaffected.

`aws-node` had been running continuously since `2026-08-05T12:01` with no restarts. The release happened ~46h later, at `2026-08-07T09:56`.

**This is not a regression in `v1.22.4`.** The same failure — pods on one node losing all network connectivity, unrecoverable without deleting the pods or replacing the node — has recurred several times in this cluster across multiple VPC CNI versions, including `v1.22.2-eksbuild.1`. The logs below are the first capture with the prefix-release sequence recorded end to end, so the version in the environment section is where it was *diagnosed*, not where it was introduced.

The distinguishing detail is that the prefix appears to have been stuck in an **incomplete add loop for the entire 46 hours**, on a different code path from the one that registered it successfully at startup.

At startup the add completes normally — `Adding …` is immediately followed by the `Added …to datastore` confirmation, both from `ipamd/ipamd.go:1366`:

```
12:01:22.570  ipamd/ipamd.go:1366  Adding 10.205.102.160/28 to DS for eni-06947d98fad1f7223
12:01:22.570  ipamd/ipamd.go:1366  Added ENI(eni-06947d98fad1f7223)'s IP/Prefix 10.205.102.160/28 to datastore
```

An IP from that prefix is then assigned to a pod:

```
2026-08-05T13:53:15.838  datastore/data_store.go:838  assignPodIPAddressUnsafe: Assign IP 10.205.102.160 to sandbox
                         aws-cni/9d4dcdd7f215628c0b13104f219b963d8a5dbdbdb747a6769fa9ef0ed47b610b/eth0
                         with metadata {K8SPodNamespace:prod K8SPodName:app-a-7756f4bb55-9zlbw InterfacesCount:1}
```

From `12:02:25` onward, however, this pair repeats roughly **2,700 times over 46 hours** — from `ipamd.go:1920`, and **never once followed by the `Added …to datastore` confirmation**:

```
2026-08-05T12:02:25.236  ipamd/ipamd.go:1766  Check in coolddown Found prefix 10.205.102.160/28
2026-08-05T12:02:25.236  ipamd/ipamd.go:1920  Adding 10.205.102.160/28 to DS for eni-06947d98fad1f7223
2026-08-05T12:03:25.501  ipamd/ipamd.go:1766  Check in coolddown Found prefix 10.205.102.160/28
2026-08-05T12:03:25.501  ipamd/ipamd.go:1920  Adding 10.205.102.160/28 to DS for eni-06947d98fad1f7223
                                              … (repeats every ~60s until 2026-08-07T09:54) …
```

Then, within 64 seconds, the state resolves and the prefix is released:

```
2026-08-07T09:55:17.360  datastore/data_store.go:1767  datastore: Sandbox aws-cni/9d4dcdd7f215628c0b13104f219b963d8a5dbdbdb747a6769fa9ef0ed47b610b/eth0
                                                       uses unknown IP Address 10.205.102.160 - presuming stale/dead
2026-08-07T09:55:17.914  ipamd/ipamd.go:2709           Prefix target is 1, short of 1 prefixes, free 0 prefixes

2026-08-07T09:56:18.563  ipamd/ipamd.go:1920           Adding 10.205.102.160/28 to DS for eni-06947d98fad1f7223
2026-08-07T09:56:18.563  ipamd/ipamd.go:1920           Added ENI(eni-06947d98fad1f7223)'s IP/Prefix 10.205.102.160/28 to datastore
                                                       ^^^ the add finally completes, 46h later

2026-08-07T09:56:21.065  ipamd/ipamd.go:804            Prefix pool is high so might be able to deallocate -
                                                       free prefixes: 3, warm prefix target: 1 from NetworkCard 0
2026-08-07T09:56:21.065  ipamd/ipamd.go:917            Deleted ENI(eni-06947d98fad1f7223)'s IP/Prefix 10.205.102.160/28 from datastore
2026-08-07T09:56:21.065  ipamd/ipamd.go:2692           Trying to unassign the following Prefixes
                                                       [10.205.102.64/28 10.205.102.160/28] from ENI eni-06947d98fad1f7223
2026-08-07T09:56:21.448  ipamd/ipamd.go:2692           Successfully freed IPv4 Prefixes
                                                       [10.205.102.64/28 10.205.102.160/28] from ENI eni-06947d98fad1f7223
```

Reading this as a chain: because the add never completed, the IPs inside `10.205.102.160/28` were not properly present in the datastore. The reconciler therefore found a sandbox holding `10.205.102.160` that the datastore did not recognise, and declared it stale/dead. Once the add did complete, all 16 addresses in the prefix counted as free, the free-prefix count reached 3 against `WARM_PREFIX_TARGET=1`, and the prefix was unassigned three seconds later.

**No CNI DEL was ever received for sandbox `9d4dcdd7f2…`.** The log search that produced these excerpts matched on the literal string `10.205.102.160`, so a teardown (`unassignPodIPAddressUnsafe: Unassign IP 10.205.102.160 …`) would have been caught — one *is* present for an earlier, unrelated sandbox on the same IP. ipamd reached the stale/dead verdict on its own.

Post-release, `describe-network-interfaces` confirms the prefix is gone from the ENI:

```
eni-06947d98fad1f7223  idx 0  Ipv4Prefixes: [10.205.100.176/28, 10.205.102.240/28]
```

Observed symptoms in the affected pods were `i/o timeout` to every destination, TCP and UDP alike — e.g. CoreDNS to the VPC resolver (`read udp 10.205.10x.x:xxxxx->10.205.96.2:53: i/o timeout`) and applications to the API server ClusterIP (`dial tcp 172.20.0.1:443: i/o timeout`). `hostNetwork` pods on the same node were completely unaffected, which is consistent with the damage being confined to addresses the fabric no longer routes.

1. ipamd should not unassign a prefix from an ENI while a live pod holds an address from it.
2. A prefix add to the datastore should not silently fail to complete for 46 hours without surfacing an error, a retry limit, or a warning.
3. A sandbox that has received a successful CNI ADD and no CNI DEL should not be classified as `stale/dead` purely because the datastore has lost track of its address.

I do not have a deterministic reproduction, and I want to be precise about the pattern because my initial reading of it was wrong.

The failure has recurred several times in this cluster over multiple months and across at least `v1.22.2-eksbuild.1` and `v1.22.4-eksbuild.3`. I originally associated it with EKS managed add-on upgrades, because that was when I noticed it and because rotating nodes afterwards resolved it. This capture shows that association is at best incomplete: `aws-node` had been running for ~46h with no restart when the prefix was released, and the stuck-add loop began about 60 seconds after ipamd startup — long before any upgrade activity. Whatever the true trigger is, it is not the add-on restart itself.

Conditions present on the affected node:

- Single-ENI node, prefix delegation enabled, `WARM_PREFIX_TARGET=1`.
- The affected prefix entered the `Check in coolddown Found prefix` path about 60 seconds after ipamd startup and never left it.
- Moderate pod churn on the node throughout the window, with IPs from both prefixes cycling through the 30s cooldown normally.

The `ipamd.go:1366` (completes) versus `ipamd.go:1920` (does not complete) split seems like the most direct place to start.

Configuration (managed via Terraform `configuration_values`, unchanged across the upgrade; everything else at defaults):

```json
{
  "env": {
    "AWS_VPC_K8S_CNI_EXTERNALSNAT": "true",
    "ENABLE_PREFIX_DELEGATION":     "true",
    "WARM_PREFIX_TARGET":           "1",
    "ANNOTATE_POD_IP":              "true"
  }
}
```

Per `docs/prefix-and-ip-target.md`, `WARM_PREFIX_TARGET=1` is the minimum supported value (0 is explicitly unsupported with prefix delegation) and is the value used in every example in that document, so I don't believe this is a misconfiguration. It does mean the free-prefix threshold is as tight as it can be, which is what turned the stuck add into a release rather than merely wasted IPs.

Ruled out:

- **No ENI was detached or deleted.** CloudTrail shows no `DetachNetworkInterface` / `DeleteNetworkInterface` for this instance, and the node only ever had the primary ENI at device index 0. The release was an `UnassignPrivateIpAddresses` against the primary ENI.
- **No `aws-node` restart.** `ipamd.log` begins at the process start on Aug 5 12:01 and runs unbroken to the release; there is no previous container log.
- **Not custom networking or subnet discovery.** The node and all three prefixes live in the same subnet.
- **Not a security group mismatch.** There are no secondary ENIs.

Two things I want to be upfront about rather than overstate:

1. I cannot independently prove the pod holding `10.205.102.160` was healthy at `09:55:17`. What I can show is that it received a successful CNI ADD and no DEL.
2. Application pods on this node were already restarting before `09:56` on Aug 7, so the prefix release is not necessarily the *initial* trigger of the outage — but it definitively strands the addresses, and the affected pods' IPs cluster exactly on the released `/28`.

One operational note that may be worth a docs mention: because CrashLoopBackOff restarts the container and not the sandbox, affected pods keep their now-unroutable IP indefinitely and never self-heal. Deleting the pods (or replacing the node) is the only recovery.

**Environment**:

- Kubernetes version: EKS `v1.36` (`kube-proxy` add-on `v1.36.0-eksbuild.14`)
- CNI Version: `v1.22.4-eksbuild.3` (EKS managed add-on), upgraded from `v1.22.2-eksbuild.1`
- OS: Bottlerocket OS 1.64.0 (aws-k8s-1.36)
- Kernel: 6.18.38
- Region: `eu-west-1`

tirthpatel90:
**What type of PR is this?**
bug

Fixes #3824

This PR resolves an issue in prefix reconciliation where active `/28` prefixes were being evicted from the cooldown cache prematurely and lost from tracking in `seenIPs`, causing `ipamd` to classify live sandboxes as `stale/dead` and unassign in-use prefixes from the primary ENI.

1. **Deferred Cooldown Eviction:** In `verifyAndAddPrefixesToDatastore`, `c.reconcileCooldownCache.Remove(strPrivateIPv4Cidr)` was previously called before verifying/persisting the prefix to the datastore. If `AddIPv4CidrToStore` encountered errors or non-fatal states, the cache was already cleared, leading to missing reconciliation states. The eviction call is now deferred until after the prefix is successfully processed and saved.
2. **Preserved Tracking in `seenIPs`:** When datastore errors or retry loops occur during prefix reconciliation, the prefix is now explicitly tracked in `seenIPs[strPrivateIPv4Cidr] = true` to prevent the `ipamd` reconciler from falsely assuming the prefix is missing or unassigned, which previously led to releasing in-use ENI prefixes.

- Verified logic flow against `verifyAndAddIPsToDatastore` reference pattern.
- Executed `go vet ./pkg/ipamd/...` across the package with zero compilation or static analysis errors.

No.

No breaking changes. This only corrects the internal IPAM datastore state reconciliation for prefix delegation.

No.

```release-note
fix(ipamd): retain prefix tracking in seenIPs and defer cooldown eviction until datastore write completes (#3824)

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -1888,8 +1888,6 @@ func (c *IPAMContext) verifyAndAddPrefixesToDatastore(ctx context.Context, eni s
 				if needEC2Reconcile {
 					// IMDS data might be stale
 					log.Debugf("This IP was recently freed, but is now out of cooldown. We need to verify with EC2 control plane.")
-					// Only call EC2 once for this ENI and post GA fix this logic for both prefixes
-					// and secondary IPs as per "split the loop" comment
 					if ec2VerifiedAddresses == nil {
 						var err error
 						// Call EC2 to verify Prefixes on this ENI
@@ -1915,9 +1913,6 @@ func (c *IPAMContext) verifyAndAddPrefixesToDatastore(ctx context.Context, eni s
 						continue
 					}
 				}
-				// The IP can be removed from the cooldown cache
-				// TODO: Here we could check if the Prefix is still used by a pod stuck in Terminating state. (Issue #1091)
-				c.reconcileCooldownCache.Remove(strPrivateIPv4Cidr)
 			}
 		}
 
@@ -1925,10 +1920,13 @@ func (c *IPAMContext) verifyAndAddPrefixesToDatastore(ctx context.Context, eni s
 		if err != nil && err.Error() != datastore.IPAlreadyInStoreError {
 			log.Errorf("Failed to reconcile Prefix %s on ENI %s", strPrivateIPv4Cidr, eni)
 			ipamdErrInc("prefixReconcileAdd")
-			// Continue to check the other Prefixs instead of bailout due to one wrong IP
+			// Preserve in seenIPs on datastore error to prevent accidental deallocation
+			seenIPs[strPrivateIPv4Cidr] = true
 			continue
-
 		}
+
+		// Remove from cooldown cache ONLY after successful verification & datastore write
+		c.reconcileCooldownCache.Remove(strPrivateIPv4Cidr)
 		// Mark action
 		seenIPs[strPrivateIPv4Cidr] = true
 		prometheusmetrics.ReconcileCnt.With(prometheus.Labels{"fn": "eniDataStorePoolReconcileAdd"}).Inc()


robertkubik-secro:
@cdirubbio @zac-nixon @jaydeokar any chance to look into this?

PR #3826, opened 2026-08-09
fix: tolerate concurrent egress SNAT chain creation

Issacwww:
## What this changes

When concurrent egress-cni ADD calls both see a chain as absent, one may create it while the other receives a `NewChain` error. On that error, the caller does the following:

```text
NewChain fails
-> ChainExists(nat, chain)
-> chain exists: continue with AppendUnique; Pod ADD succeeds
-> chain does not exist: return the original NewChain error
-> state cannot be checked: return both the create and state-check errors
```

This validates the resulting iptables state rather than matching backend-specific error text.

## Scope

This handles the reported `Chain already exists` race. It does not explain why the customer's `ListChains` snapshot omitted normally built-in `POSTROUTING`, add a global lock, or change DEL behavior.

## Validation

- Deterministic concurrent unit test verifies both Pod ADDs succeed with complete, non-duplicate rules when `POSTROUTING` is absent from the initial snapshot.
- Linux race test: `go test -race -count=100 ./cmd/egress-cni-plugin/snat`.
- `hack/egress-snat-raceprobe` demonstrates the old failure and fixed recovery against a real `iptables-nft` backend when the reported missing-chain condition is supplied.

Natural burst testing, including the FIPS environment, did not reproduce the missing `POSTROUTING` snapshot; therefore it cannot exercise this recovery path directly.


cmd/egress-cni-plugin/snat/snat.go
```diff
@@ -73,7 +74,16 @@ func Add(ipt iptableswrapper.IPTablesIface, nodeIP, src net.IP, multicastRange,
 		_chain := rule[0]
 		if !existingChains[_chain] {
 			if err = ipt.NewChain("nat", _chain); err != nil {
-				return err
+				// A concurrent ADD may have created the chain after ListChains.
+				// Verify the resulting state instead of classifying error text.
+				createErr := err
```
jaydeokar, line 79: You can create this as a helper function something like `CreateChainIfNotExist` ? 
Do we have other places where we have similar checks ?

PR #3827, opened 2026-08-11
Serve node IP and instance ID to the network policy agent via gRPC

viveksb007:
**What type of PR is this?**

feature

N/A

The `aws-network-policy-agent` (NPA) currently queries the EC2 Instance
Metadata Service (IMDS) at startup to learn the node IP and (for the
CloudWatch-logs path) the instance ID. We want to remove IMDS dep from NPA, since it can get the data from IPAMD which already depends on IMDS (to minimize external deps)

NPA already talks to ipamd over the local gRPC unix socket
(`ConfigServerBackend.GetNetworkPolicyConfigs`) to fetch the network policy
mode. ipamd already caches the node IP and instance ID, so this PR serves
those values over that existing reply. A follow-up change in the
`aws-network-policy-agent` repo will consume these fields and drop its IMDS
calls.

This PR (ipamd side only) extends `NetworkPolicyAgentConfigReply` with:

- `NodeIPv4` (field 3) — primary IPv4 address, from the cached metadata
- `NodeIPv6` (field 4) — primary IPv6 address, populated only in IPv6 mode
- `InstanceID` (field 5) — EC2 instance ID, from the cached metadata

To keep this a pure cache read, `GetLocalIPv6()` is changed to return a value
cached once during `initWithEC2Metadata` (gated on IPv6 mode) instead of
issuing a live IMDS request on every call, matching the existing behavior of
`GetLocalIPv4()` / `GetInstanceID()`.

Built the image and patched the `aws-node` DaemonSet on two clusters, leaving
the upstream (unchanged) network policy agent in place to confirm backward
compatibility.

IPv4 cluster (K8s v1.35):
```
ipamd: Discovered the instance primary IPv4 address: 192.168.95.5
ipamd: Send NetworkPolicyAgentConfigReply: NetworkPolicyMode: standard, MultiNICEnabled: false, NodeIPv4: 192.168.95.5, NodeIPv6: , InstanceID: i-0120354ae03d7f5d8
NPA : Connected to ipamd at unix:///var/run/aws-node/ipamd.sock. NetworkPolicyMode: standard MultiNICEnabled: false
```

IPv6 cluster (K8s v1.35):
```
ipamd: Discovered the instance primary IPv6 address: 2600:1f14:3c59:2a00:fc47:9631:bc12:dc4c
ipamd: Send NetworkPolicyAgentConfigReply: NetworkPolicyMode: standard, MultiNICEnabled: false, NodeIPv4: , NodeIPv6: 2600:1f14:3c59:2a00:fc47:9631:bc12:dc4c, InstanceID: i-048a9c9d5355732c7
NPA : Connected to ipamd at unix:///var/run/aws-node/ipamd.sock. NetworkPolicyMode: standard MultiNICEnabled: false
```

Unit tests: `TestServer_GetNetworkPolicyConfigs` is table-driven, covering the
IPv4 branch, the IPv6 branch, and nil local IP for both. `go build ./...`,
`go test ./pkg/ipamd/...` and `go test ./pkg/awsutils/...` pass.

No.

No. The new fields use additive proto3 field numbers (3, 4, 5); existing
fields (1, 2) are unchanged. An older network policy agent that does not know
these fields ignores them and continues to work — verified on both clusters
above (unchanged NPA, 0 restarts). This PR does not reduce the number of IMDS
calls ipamd makes.

No.

No.

```release-note
NONE
```


pkg/awsutils/awsutils.go
```diff
@@ -486,6 +487,16 @@ func (cache *EC2InstanceMetadataCache) initWithEC2Metadata(ctx context.Context)
 	}
 	log.Debugf("Discovered the instance primary IPv4 address: %s", cache.localIPv4)
 
+	// Cache IPv6 so reads don't trigger a live IMDS request per call.
+	if cache.v6Enabled {
+		cache.localIPv6, err = cache.imds.GetLocalIPv6(ctx)
+		if err != nil {
+			awsAPIErrInc("GetLocalIPv6", err)
```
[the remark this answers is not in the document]
viveksb007: done

pkg/ipamd/rpc_handler.go
```diff
@@ -468,9 +468,22 @@ func (s *server) GetNetworkPolicyConfigs(ctx context.Context, e *emptypb.Empty)
 	resp := &rpc.NetworkPolicyAgentConfigReply{
 		NetworkPolicyMode: s.ipamContext.networkPolicyMode,
 		MultiNICEnabled:   s.ipamContext.enableMultiNICSupport,
+		InstanceID:        s.ipamContext.awsClient.GetInstanceID(),
 	}
 
-	log.Infof("Send NetworkPolicyAgentConfigReply: NetworkPolicyMode: %v, MultiNICEnabled: %v", resp.NetworkPolicyMode, resp.MultiNICEnabled)
+	// Serve the node IP so the network policy agent doesn't query IMDS itself.
```
zac-nixon, line 474: if ipv6 is enabled but ipv6 is null; we would return empty ips. do we need to be more defensive here?

zac-nixon (approved):

pkg/ipamd/rpc_handler.go
viveksb007: yeah, i was thinking about this, i don't think `if ipv6 is enabled but ipv6 is null` is possible though. do you foresee this happening where we have a IPv6 cluster but primary ENI doesn't have IPv6 addr?

pkg/ipamd/rpc_handler.go
```diff
@@ -468,9 +468,22 @@ func (s *server) GetNetworkPolicyConfigs(ctx context.Context, e *emptypb.Empty)
...
+	if s.ipamContext.enableIPv6 {
+		if ipv6 := s.ipamContext.awsClient.GetLocalIPv6(); ipv6 != nil {
+			resp.NodeIPv6 = ipv6.String()
```
jaydeokar, line 477: nit - can we just have NodeIP instead of field for v4 and v6

pkg/ipamd/rpc_handler_test.go
```diff
@@ -565,22 +565,76 @@ func TestServer_AddNetwork(t *testing.T) {
 }
 
 func TestServer_GetNetworkPolicyConfigs(t *testing.T) {
```
jaydeokar, line 567: One more case - 
IPv6 node in dual stack subnet will have a v6 and v4 primary address, but it should only return v6

pkg/ipamd/rpc_handler.go
viveksb007: wanted to keep it explicit (if we need to do some validation or anything), this can also be useful when both IP families are supported in NP

pkg/ipamd/rpc_handler_test.go
viveksb007: it is already covered by tc `ipv6 cluster populates NodeIPv6 only` as we are asserting on both V4 and V6

jaydeokar: Should be fine if the plan is to keep two different fields for IPv4 and IPv6

jaydeokar (approved):

PR #3829, opened 2026-08-13
Add Region to NetworkPolicyAgentConfigReply

viveksb007:
**What type of PR is this?**

  feature

**Which issue does this PR fix?:**

  N/A

 **What does this PR do / Why do we need it?:**

  Adds a Region field to the NetworkPolicyAgentConfigReply gRPC message so ipamd serves the node's AWS region to the network policy agent over the existing local gRPC socket.

  This follows the earlier change that added NodeIPv4, NodeIPv6, and InstanceID to the same reply so the agent can stop querying IMDS. Region was the remaining value the agent needed for its CloudWatch-logs path (region for the AWS SDK client). 

**Testing done on this change:**

  Built the image and patched the aws-node DaemonSet on an IPv4 cluster and an IPv6 cluster (K8s v1.35)

```
  ipamd now returns the region in the reply:
  IPv4: Send NetworkPolicyAgentConfigReply: NetworkPolicyMode: standard, MultiNICEnabled: false, NodeIPv4: 192.168.95.5, NodeIPv6: , InstanceID: i-0120354ae03d7f5d8, Region: us-west-2
  IPv6: Send NetworkPolicyAgentConfigReply: NetworkPolicyMode: standard, MultiNICEnabled: false, NodeIPv4: , NodeIPv6: 2600:1f14:3c59:2a00:fc47:9631:bc12:dc4c, InstanceID: i-048a9c9d5355732c7, Region: us-west-2
```

Also testing NPA CW path taking these CNI changes, will be raising separate PR for it.

**Will this PR introduce any new dependencies?:**

  No.

 **Will this break upgrades or downgrades? Has updating a running cluster been tested?:**

  No. Region uses an additive proto3 field number (6); existing fields (1–5) are unchanged, so an older network policy agent that does not know the field ignores it and continues to work. Verified by patching the DaemonSet on running IPv4 and IPv6 clusters — pods rolled cleanly, 0
  restarts. This PR adds no IMDS or AWS API calls (region is read from ipamd's existing cache).

**Does this change require updates to the CNI daemonset config files to work?:**

  No.


cdirubbio (approved):
Lgtm

Issue #3777, opened 2026-07-14
Improper error check in IPAMD for already removed pods


aagarwalla-fx:
**What happened**:
We see the following log generated (**Pod not found**) by the code path - https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/rpc_handler.go?plain=1#L393
```json
{"level":"warn","ts":"2026-07-14T12:21:26.459Z","caller":"rpc/rpc_grpc.pb.go:135","msg":"Send DelNetworkReply: Failed to get pod spec: error while trying to retrieve pod info: Pod \"test-deployment-84874686cb-k4jqa\" not found"}
```
Looks like there is a check before this exact same reason, **but it does not work**.
https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/ipamd/rpc_handler.go?plain=1#L389-L392

Caught this minor issue while debugging something else. Because of this, the deleted pod does not cleanup the resources properly until a long time, thereby holding up the branch ENI, and thus it's unavailable for the new pod. So if the node is running at the full capacity, and a pod is deleted and a new pod is scheduled on that node, it runs into the below error for 2 mins

```
Failed to create pod sandbox: rpc error: code = Unknown desc = failed to setup network for sandbox "2eeb7e737366bf0bc3b0fadc3fe2bac25c6504d832bdba38a6daf3e8b68b2293": plugin type="aws-cni" name="aws-cni" failed (add): add cmd: failed to assign an IP address to container
```

**What you expected to happen**:
The `DelNetwork()` should have gracefully returned since the pod is not found.

**How to reproduce it (as minimally and precisely as possible)**:
This happens when the below VPC CNI feature is enabled. When the pod gets deleted, we see these logs.
- ENABLE_POD_ENI: "true"

**Environment**:

- Kubernetes version (use `kubectl version`): 1.34
- CNI Version: v1.22.3-eksbuild.1
- OS (e.g: `cat /etc/os-release`): 
```
NAME=Bottlerocket
ID=bottlerocket
VERSION="1.60.0 (aws-k8s-1.34)"
PRETTY_NAME="Bottlerocket OS 1.60.0 (aws-k8s-1.34)"
VARIANT_ID=aws-k8s-1.34
VERSION_ID=1.60.0
BUILD_ID=c1f9ba0c
VENDOR_NAME=Bottlerocket
HOME_URL="https://github.com/bottlerocket-os/bottlerocket"
[...]

PR #3831, opened 2026-08-15
fix: handle pod-not-found error in DelNetwork

Rajkaran-122:
**What type of PR is this?**:
bug

#3777

When a pod has already been deleted from the cluster, the `DelNetwork()` flow does not correctly recognize the Kubernetes `NotFound` error.

`GetPod()` was converting the original Kubernetes error into a formatted string, which prevented `k8serror.IsNotFound()` from identifying the error correctly.

This change preserves the original Kubernetes error and updates the related pod annotation handling to use `k8serror.IsNotFound()`.

A dedicated unit test is also added to verify that `DelNetwork()` gracefully succeeds when the requested pod is no longer present.

- Added a unit test covering the `DelNetwork()` pod-not-found scenario.
- Ran `gofmt`.
- Ran `git diff --check`.
- Verified the changes are limited to the required production code and unit test.

No.

No. The change only improves error handling and does not change configuration, APIs, or upgrade behavior.

No.

No.

### DIFF
--- a/pkg/ipamd/ipamd.go
+++ b/pkg/ipamd/ipamd.go
@@ -36,6 +36,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/samber/lo"
 	corev1 "k8s.io/api/core/v1"
+	k8serror "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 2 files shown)


Rajkaran-122:
@aagarwalla-fx I’ve opened an independent PR for issue #3777 with the DelNetwork() pod-not-found fix and a dedicated regression test covering the deleted-pod case. The change preserves the Kubernetes NotFound error so it can be correctly recognized by k8serror.IsNotFound(). Would appreciate your review.

PR #3858, opened 2026-09-02
fix: avoid transient connected route during ENI setup

### LINKED ISSUE #3856: Transient main-table route during ENI setup can stall image pulls for ~15 minutes
## Summary

On fresh EKS nodes running Bottlerocket, image pulls from a private registry can intermittently stall for approximately 15–16 minutes while VPC CNI configures a secondary or trunk ENI. The pull normally succeeds on a later retry after node routing has converged.

The investigation traced this to a transient connected route in the Linux main routing table. The route is created automatically when `setupENINetwork()` assigns the ENI address and is removed only after VPC CNI finishes programming the explicit ENI routing table.

## Impact

- System DaemonSets and monitoring agents can remain `Pending` or `NotReady` for 9–16 minutes on a newly created node.
- Node bootstrap and scale-out are delayed because required images are not available.
- The failure is probabilistic: only connections opened during the short route-programming window are affected.
- Image size is not the determining factor; small images can stall for the same duration.
- Registries that use redirects or several parallel connections increase the chance of hitting the race, but they are an amplifier rather than the root cause.

## Conditions required to hit the race

1. A non-primary ENI is being configured, most notably the trunk ENI used with pod security groups.
2. The registry destination is covered by the connected prefix derived from the new ENI address.
3. `containerd` opens the registry connection after the address is added but before the transient main-table route is removed.
4. The connection remains active when route reconciliation completes.

Environments where the registry is reached through a different, non-overlapping route do not exhibit the same behavior.

## Representative timing

| Event | Relative time |
| --- | --- |
| `PullImage` starts | `T+0` |
| Registry control request is made, followed by no content-transfer progress | shortly after `T+0` |
| The pull returns successfully after an internal retry | `T+16m1.8s` |

The duration is consistent with established TCP retransmission timeout behavior. Linux documents that the default `net.ipv4.tcp_retries2=15` corresponds to a hypothetical lower bound of about 924.6 seconds. This is supporting evidence, not a claim that `tcp_retries2` was measured on every affected node.

## Failure sequence

| Step | VPC CNI / Linux behavior | Effect on the connection |
| --- | --- | --- |
| 1 | VPC CNI brings the non-primary ENI link up. | The interface becomes available to the host. |
| 2 | `AddrAdd()` assigns the ENI address with the subnet mask. | Linux automatically adds a connected prefix route to `main`. |
| 3 | `containerd` opens a registry connection during this window. | Route and source selection can choose the new ENI and its primary address. |
| 4 | VPC CNI installs the gateway route and default route in the dedicated ENI routing table. | The intended policy-routing state is prepared. |
| 5 | VPC CNI deletes the automatically created connected route from `main`. | The route that selected the ENI for the existing connection disappears. |
| 6 | For a trunk ENI, VPC CNI intentionally skips the per-source IP rule. | Subsequent packets can resolve through another interface while the socket retains the trunk ENI source IP. |
| 7 | Traffic leaves through an interface that does not own the selected source address. | The EC2 data plane can drop the packets; the established TCP connection retransmits until timeout. |
| 8 | The pull is retried after routing has converged. | The new connection uses the stable route and succeeds. |

The trunk path is the most visible case because its source rule is intentionally skipped. The same transient connected route is created for other non-primary ENIs, so the proposed change prevents it for the shared setup path.

## Root cause

`setupENINetwork()` currently relies on a delete-after-add sequence:

```text
LinkSetUp
  -> AddrAdd
       -> kernel automatically creates connected route in main
  -> install explicit routes in the ENI table
  -> delete connected route from main
  -> add source rule, except for trunk ENIs
```

The intended final state is correct, but it is reached non-atomically. A host connection can observe and use the intermediate route.

Relevant implementation: [`pkg/networkutils/network.go`](https://github.com/aws/amazon-vpc-cni-k8s/blob/6bf7adeeffeae92974af0f55926b47e4a05b496f/pkg/networkutils/network.go#L989-L1100). The same ordering is present in VPC CNI v1.21.1, v1.22.4, and the current `master` branch.

## Proposed fix

Add the ENI address with `IFA_F_NOPREFIXROUTE`:

```go
netLink.AddrAdd(link, &netlink.Addr{
    IPNet: eniAddr,
    Flags: unix.IFA_F_NOPREFIXROUTE,
})
```

`IFA_F_NOPREFIXROUTE` prevents Linux from creating the implicit connected route. VPC CNI remains the sole owner of the explicit ENI routes from the beginning of setup.

The existing `RouteDel` should remain as defensive reconciliation for stale routes left by an earlier version or external configuration. No change is proposed to the intentional trunk ENI source-rule behavior.

## Validation

- A privileged Linux dummy-interface test confirmed that adding a plain `/24` address creates a connected route, while adding the same address with `noprefixroute` creates no connected prefix route.
- Unit coverage includes IPv4 and IPv6 for both secondary and trunk ENIs.
- Tests cover the expected `ESRCH` result when the defensive main-table cleanup finds no route.
- `go test ./cmd/... ./pkg/...` passed on Linux.
- `go test ./test/integration/common/...` passed on Linux.
- `go vet ./pkg/networkutils/...` passed.
- Upstream CI and a production rollout have not yet been completed.

## Alternatives considered

| Option | Result |
| --- | --- |
| Lower `net.ipv4.tcp_retries2` | Shortens the visible stall but does not prevent the wrong route/source selection and affects every established TCP connection on the node. |
| Serialize or reduce parallel image pulls | Reduces the probability of hitting the window but does not remove it. |
| Disable registry redirects | Reduces the number of connections but treats an amplifier rather than the routing bug. |
| Add destination-specific host routes | Can avoid the conflicting prefix for known registry IPs, but is brittle when endpoints change. |
| Change reverse-path filtering or EC2 source/destination check | Does not prevent the transient connected route and broadens networking risk. |

## Expected behavior

Adding a secondary or trunk ENI must not temporarily alter host routing for unrelated connections. Image pulls started during node bootstrap should use the same stable path before and after ENI reconciliation.

## Environment

- Amazon EKS
- Bottlerocket
- `containerd`
- VPC CNI with pod ENI / trunk ENI support enabled
- Observed on VPC CNI v1.21.1; the affected sequence also exists on current `master`

## References

- Fix: #3858
- Related Bottlerocket symptom report: https://github.com/bottlerocket-os/bottlerocket/issues/4677
- Linux `noprefixroute` documentation: https://man7.org/linux/man-pages/man8/ip-address.8.html
- Linux TCP retry documentation: https://docs.kernel.org/networking/ip-sysctl.html
- IPv4 `IFA_F_NOPREFIXROUTE` implementation: https://github.com/torvalds/linux/commit/7b1311807f3d3eb8bef3ccc53127838b3bea3771

MrTolerant:
**What type of PR is this?**

bug

Fixes #3856

Fresh nodes can intermittently stall private-registry image pulls for approximately 15–16 minutes while VPC CNI initializes a secondary or trunk ENI. Connections opened during the ENI setup window can observe a temporary host-routing state; when several registry connections are created in parallel, the probability of hitting that window increases.

The race is:

1. `setupENINetwork()` assigns the ENI address with its subnet mask.
2. Linux automatically creates a connected prefix route in the main routing table.
3. A host connection can select that route and the new ENI source address.
4. VPC CNI installs the intended routes in the ENI-specific table and removes the connected route from the main table.
5. For trunk ENIs, the per-source rule is intentionally skipped. An already-open connection can therefore retain the ENI source address while later packets resolve through another interface and stall until TCP retransmission recovery.

The intended final routing state is correct, but it is reached through an intermediate state that is visible to concurrent host traffic:

```text
Before: AddrAdd → implicit connected route in main → explicit ENI routes → RouteDel
After:  AddrAdd(NOPREFIXROUTE) → explicit ENI routes → defensive RouteDel
```

This change sets `IFA_F_NOPREFIXROUTE` when assigning non-primary ENI addresses, preventing Linux from creating the transient connected route. VPC CNI remains the owner of the explicit ENI routes from the start. The existing main-table `RouteDel` remains in place to reconcile stale routes left by an earlier version or external configuration, and the existing trunk ENI source-rule behavior is unchanged.

- Added `TestSetupENINetworkNoPrefixRoute` in `pkg/networkutils/network_test.go`. It calls the real `setupENINetwork()` in disposable Linux network namespaces and checks for `RTPROT_KERNEL` connected-prefix routes in `main` immediately after real address assignment, before cleanup can mask a regression.
- Covers IPv4/IPv6, secondary/trunk ENIs, fresh setup, repeated setup, and an existing source rule. Verifies explicit gateway/default routes and secondary source rules remain correct. IPv6 readiness is polled on the namespace's locked OS thread.
- Positive control: adding the same address without the flag creates the connected route. Mutation control: removing `IFA_F_NOPREFIXROUTE` using a temporary Go source overlay makes all four cases fail at the transient-route assertion.
- Linux validation: 20 repeated kernel-test runs; three race-detector runs of ENI setup tests; `go vet ./pkg/networkutils/...`; `go test ./cmd/... ./pkg/... ./test/integration/common/...` all passed.
- Existing unit tests continue to cover the address flag and `ESRCH` cleanup handling. The standalone reproducer remains in `scripts/test/repro-transient-main-route.sh`.

The kernel test is opt-in; ordinary unprivileged unit runs skip it. Run on Linux with namespace/network capabilities:

```bash
RUN_ENI_KERNEL_TESTS=1 go test ./pkg/networkutils -run TestSetupENINetworkNoPrefixRoute -count=20 -v
```

A running-cluster upgrade has not been tested by this change; upstream CI remains a separate gate.

**Secondary-ENI dependency audit**:

[`setupENINetwork()`](https://github.com/aws/amazon-vpc-cni-k8s/blob/958f1259a43d12910ca8fe4223706d012f8e2349/pkg/networkutils/network.go#L950) programs the gateway host route and default route explicitly in the supplied ENI table between `AddrAdd` and `RuleAdd`. Their gateway/interface are supplied directly; no lookup through the transient main-table prefix is required. The same path handles secondary, trunk, and additional-network-card ENIs. Trunk source-rule skipping and existing-rule reuse are unchanged.

An audit of production code under `cmd/` and `pkg/` found no route lookup, enumeration caller, or subscription that relies on this transient route. [IPv4 secondary addresses/prefixes are published after network setup](https://github.com/aws/amazon-vpc-cni-k8s/blob/958f1259a43d12910ca8fe4223706d012f8e2349/pkg/ipamd/ipamd.go#L1321). ENI metadata and IPv6 prefixes can be recorded earlier, so this is not a claim that all datastore publication is atomic. During startup, [`ipamd.New()` finishes before the RPC handler starts](https://github.com/aws/amazon-vpc-cni-k8s/blob/958f1259a43d12910ca8fe4223706d012f8e2349/cmd/aws-k8s-agent/main.go#L148). This audit covers VPC CNI code; it does not establish the behavior of arbitrary external network managers.

No new Go modules, AWS or EC2 metadata API calls, binaries, kernel modules, or DaemonSet dependencies are introduced. The change uses the Linux `IFA_F_NOPREFIXROUTE` address flag; IPv4 support has existed since Linux 4.4, below the kernels used by current EKS-optimized and Bottlerocket nodes.

No steady-state route or policy-rule change is expected because the current implementation already removes the connected route from the main table. The patch prevents that route from existing temporarily, while retaining defensive cleanup for upgrade and reconciliation scenarios. Downgrading restores the previous transient behavior. Updating a running cluster has not yet been tested.

No.

Yes. It prevents transient host-routing changes during secondary and trunk ENI initialization. No configuration or migration action is required beyond upgrading to a release containing the fix.

```release-note
Prevent transient main-table connected routes during secondary and trunk ENI initialization that can stall host connections started during node bootstrap.
```


yash97:
do you have script or integration test which can repro this issue, so it is easier to validate the fix.

MrTolerant:
> do you have script or integration test which can repro this issue, so it is easier to validate the fix.

No, sorry. 
Hope AWS team can reproduce that on their side.

MrTolerant:
I added a self-contained reproducer in `scripts/test/repro-transient-main-route.sh`.

It runs entirely inside two temporary Linux network namespaces, so it does not change the host routing table or require an EKS cluster. It reproduces the networking sequence behind the issue:

1. establish the normal route through the primary interface;
2. add a secondary-ENI-style prefix address, which creates the implicit connected route;
3. open TCP while that route exists;
4. delete the route, as VPC CNI does after programming its explicit routes;
5. verify that the socket remains pinned to the secondary ENI source and times out;
6. repeat with `noprefixroute`, verifying that the primary route/source remain selected and TCP succeeds.

Run it on Linux with:

```bash
sudo scripts/test/repro-transient-main-route.sh
```

Validated in a disposable privileged Linux container:

```text
Stable TCP:         connected
Pinned TCP socket:  0  1  172.20.37.151:52130  172.20.54.213:18443
Current TCP:        timed out with the transient ENI source
Fixed TCP socket:   0  0  10.0.0.10:33902  172.20.54.213:18443
PASS: reproduced the transient route and pinned TCP source; noprefixroute prevented both
```

This is a deterministic kernel/routing/socket reproducer. It intentionally does not claim to reproduce containerd or the EKS control plane end-to-end; the existing unit tests connect the demonstrated `noprefixroute` behavior to `setupENINetwork()`.

MrTolerant:
You can test script:
```
docker run --rm --privileged \
  -v "$PWD:/work:ro" \
  -w /work \
  debian:bookworm-slim \
  bash -lc '
    apt-get update -qq
    DEBIAN_FRONTEND=noninteractive apt-get install -y -qq iproute2 netcat-openbsd >/dev/null
    bash scripts/test/repro-transient-main-route.sh
  '
```

emilyhuaa:
Repro'd the kernel race in-house with your script on Bottlerocket 1.65 + m5.large + SGP trunk ENI. `noprefixroute` fixes it as expected.

```
Current behavior:   172.20.54.213 dev eth1 src 172.20.37.151 uid 0
With noprefixroute: 172.20.54.213 via 10.0.0.1 dev eth0 src 10.0.0.10 uid 0
PASS: reproduced the transient route and pinned TCP source; noprefixroute prevented both
```

Two asks before approve:

1. Add a Go-level test in `pkg/networkutils/network_test.go` that calls `setupENINetwork` and asserts no `RTPROT_KERNEL` route in the main table for the added address. Script covers the kernel; test needs to cover the code path.
2. Fix applies to every secondary ENI, not just trunk. Confirm nothing reads the transient main-table route between `AddrAdd` and the per-source rule landing.

MrTolerant:
@emilyhuaa Done

pkg/networkutils/network_test.go
```diff
@@ -80,6 +84,189 @@ func setup(t *testing.T) (*gomock.Controller,
...
 func TestSetupENINetwork(t *testing.T) {
+	testSetupENINetwork(t, false)
+}
+
+func TestSetupTrunkENINetwork(t *testing.T) {
+	testSetupENINetwork(t, true)
+}
+
+func TestSetupENINetworkNoPrefixRoute(t *testing.T) {
+	if os.Getenv("RUN_ENI_KERNEL_TESTS") != "1" {
```
cdirubbio: What are we setting this env var for? Why would we want to skip the test?

scripts/test/repro-transient-main-route.sh
cdirubbio, line 1: Can you remove this testing file

cdirubbio:
The core logic change LGTM, can you just rebase and address the two comments ?

scripts/test/repro-transient-main-route.sh
MrTolerant: Done, removed in the latest commit.

pkg/networkutils/network_test.go
MrTolerant: Removed that

cdirubbio:
The unit tests are failing, can you PTAL? @MrTolerant 
```
--- FAIL: TestSetupENINetworkNoPrefixRoute (0.00s)
    --- FAIL: TestSetupENINetworkNoPrefixRoute/IPv4/trunk=false (0.00s)
    --- FAIL: TestSetupENINetworkNoPrefixRoute/IPv4/trunk=true (0.00s)
    --- FAIL: TestSetupENINetworkNoPrefixRoute/IPv6/trunk=false (0.00s)
    --- FAIL: TestSetupENINetworkNoPrefixRoute/IPv6/trunk=true (0.00s)
```

MrTolerant:
> The unit tests are failing, can you PTAL? @MrTolerant
> 
> ```
> --- FAIL: TestSetupENINetworkNoPrefixRoute (0.00s)
>     --- FAIL: TestSetupENINetworkNoPrefixRoute/IPv4/trunk=false (0.00s)
>     --- FAIL: TestSetupENINetworkNoPrefixRoute/IPv4/trunk=true (0.00s)
>     --- FAIL: TestSetupENINetworkNoPrefixRoute/IPv6/trunk=false (0.00s)
>     --- FAIL: TestSetupENINetworkNoPrefixRoute/IPv6/trunk=true (0.00s)
> ```

The test creates a network namespace (unshare(CLONE_NEWNET)), which needs root (CAP_SYS_ADMIN). 
GitHub's ubuntu-latest runner is non-root, so the call returned EPERM and all 4 subtests failed instantly. 

Added fix: the test now re-runs itself under sudo when not root ( i'm not sure is it proper way to do in this repo)

cdirubbio:
Ack. I don't think the unit test is going to be what we're looking for. Are you able to instead write an integration test, and run it to show that this resolves this issue?

MrTolerant:
> Ack. I don't think the unit test is going to be what we're looking for. Are you able to instead write an integration test, and run it to show that this resolves this issue?

This integration test then will require EKS cluster with nodes, deployed vpc-cni.
Or you mean  need to do that inside docker?

parlakisik:
Here is the location for integrations test :
 https://github.com/aws/amazon-vpc-cni-k8s/tree/master/test/integration

pkg/networkutils/network_test.go
```diff
@@ -80,6 +85,212 @@ func setup(t *testing.T) (*gomock.Controller,
...
+
+func (n *observeENIAddrAdd) AddrAdd(link netlink.Link, addr *netlink.Addr) error {
+	if err := n.NetLink.AddrAdd(link, addr); err != nil {
+		return err
+	}
+	n.afterAdd()
+	return nil
+}
+
+func testSetupENINetwork(t *testing.T, isTrunkENI bool) {
```
parlakisik, line 291: do we need to set trunk configuration for tests ?

MrTolerant: Covered already. 

**testSetupENINetwork**(t, isTrunkENI bool) runs via **TestSetupENINetwork** (false) and **TestSetupTrunkENINetwork** (true), and TestSetupENINetworkNoPrefixRoute loops trunk × IPv4/IPv6 — all four assert no RTPROT_KERNEL route in main.

Trunk needs its own case because setupENINetwork skips the per-source rule for it (if !isTrunkENI && !isRuleConfigured), so the secondary-ENI path doesn't generalize.

MrTolerant:
@parlakisik I tried building an integration test for this, but I don't think it's worth adding. 

The race window is only a few milliseconds — between AddrAdd (network.go:1014) and the RouteDel  there's just a handful of netlink calls. 

The failure is probabilistic, and it showed up in prod because **_containerd_** opens tons of connections on every node bootstrap. 

A single test run would only catch it occasionally, so you'd get a flaky test rather than real proof. 
The `kernel-netns` test in network_test.go covers the same path reliably. 

---
p.s. 

Worth noting the window isn't just a few syscalls. 
Between AddrAdd and the main-table RouteDel, each of the two routes gets RouteReplace retries on `NewSimpleBackoff(500ms, 5s, 0.15, 2.0)` with `maxRetryRouteAdd=5` — roughly 7.5s per route, ~15s for both. 
So on a node where netlink is contended the transient route can persist for seconds, which fits how often this was hit in practice.

PR #3859, opened 2026-09-02
fix(networkutils): avoid transient prefix routes

### LINKED ISSUE #3857: Prevent transient connected route when configuring secondary ENI addresses
When `SetupENINetwork()` programs a secondary or trunk ENI, it adds the ENI address first and only later deletes the automatically created main-table prefix route. On kernels that install a connected route for the ENI subnet, there is a window where outbound traffic can select the new ENI path before the route is removed.

This is reproducible on nodes where a secondary/trunk ENI shares a subnet with destinations that are reachable from the node. If a TCP connection is established during that transient window, it can remain pinned to the wrong path even after route cleanup.

Relevant code:
- `pkg/networkutils/network.go`: `AddrAdd()` happens before the code deletes the main-table subnet route.
- `pkg/ipamd/ipamd.go`: non-primary ENIs still flow through `SetupENINetwork()`.

Suggested fix:
- Add `IFA_F_NOPREFIXROUTE` when assigning the ENI address so the kernel does not create the implicit connected prefix route in the first place.
- Keep the existing explicit per-ENI route-table programming and the trunk ENI IP-rule skip behavior as-is.

I also checked Bottlerocket settings and did not find an OS-level tuning knob that changes this per-address kernel behavior; `noprefixroute` is the relevant control point.

MrTolerant:
Fixes #3857

This change prevents the kernel from creating the implicit prefix route when an ENI address is added, which removes the timing window where traffic can briefly select the wrong ENI path before CNI route cleanup completes.

What changed:
- Add `IFA_F_NOPREFIXROUTE` to the ENI address programmed in `setupENINetwork()`.
- Update the unit test expectations in `pkg/networkutils/network_test.go` to assert the flag is present for IPv4 and IPv6.

Why this is the right place:
- The CNI already explicitly programs the per-ENI route table.
- The existing delete-after-add flow leaves a race window; suppressing the implicit kernel route avoids that window entirely.
- This is per-address behavior, so Bottlerocket/kernel sysctls are not the right lever.

Validation:
- `git diff --check` passes.
- I attempted `go test ./pkg/networkutils -run TestSetupENINetwork -count=1` in this container, but the local environment fails in unrelated dependency/toolchain resolution before the package can finish compiling; CI should be the authoritative verification run.

### DIFF
--- a/pkg/networkutils/network.go
+++ b/pkg/networkutils/network.go
@@ -1009,7 +1009,10 @@ func setupENINetwork(eniIP string, eniMac string, networkCard int, eniSubnetCIDR
 	}
 
 	log.Debugf("Adding IP address %s", eniAddr.String())
-	if err = netLink.AddrAdd(link, &netlink.Addr{IPNet: eniAddr}); err != nil {
+	// Avoid the kernel's implicit connected prefix route in the main table.
+	// The CNI programs the per-ENI route table explicitly, and we do not want
+	// a temporary main-table route to influence source selection during setup.
+	if err = netLink.AddrAdd(link, &netlink.Addr{IPNet: eniAddr, Flags: unix.IFA_F_NOPREFIXROUTE}); err != nil {
 		return errors.Wrap(err, "setupENINetwork: failed to add IP addr to ENI")
 	}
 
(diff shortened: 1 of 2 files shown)


MrTolerant:
- Closing this duplicate in favor of #3858, which contains the same fix and has completed Linux test validation.

PR #3556, opened 2025-12-18
fix: add userAgent to AWS API calls

cdirubbio:
**What type of PR is this?**

**Which issue does this PR fix?**: #3555 

**What does this PR do / Why do we need it?**: Adds VPC CNI identifying user agent and version into AWS API calls. Before fix, userAgent in CloudTrail shows:
```
    "userAgent": "aws-sdk-go-v2/1.39.2 ua/2.1 os/linux lang/go#1.25.5 md/GOOS#linux md/GOARCH#amd64 api/ec2#1.254.1 m/0,E",
```

After my fix, it appends `amazon-vpc-cni-k8s/v1.21.0` at the end.
```
    "userAgent": "aws-sdk-go-v2/1.39.2 ua/2.1 os/linux lang/go#1.25.5 md/GOOS#linux md/GOARCH#amd64 api/ec2#1.254.1 amazon-vpc-cni-k8s/v1.21.0 m/0,E",
```

**Testing done on this change**:

**Will this PR introduce any new dependencies?**:

**Will this break upgrades or downgrades? Has updating a running cluster been tested?**:

**Does this change require updates to the CNI daemonset config files to work?**:

**Does this PR introduce any user-facing change?**:

```release-note
fix: add amazon-vpc-cni-k8s/<version> userAgent to AWS API calls 
```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3873, opened 2026-09-10
Self-heal IPv6 datastore and harden EC2 client retry

emilyhuaa:
## What type of PR is this?

bug

## Which issue does this PR fix?

IPv6-only nodes stay wedged after a bootstrap-time `AssignIpv6Addresses` failure. No open GitHub issue tracks this specific failure mode, but the mechanism follows from the design: IPv6 Prefix Delegation assigns the single `/80` prefix per node exactly once at `nodeInit`, and there is no analog of the IPv4 warm-pool reconciliation loop.

## What does this PR do / Why do we need it?

When the bootstrap `AssignIpv6Addresses` call fails (transient EC2 throttle during a rapid scale-out, transient IAM propagation on a new node, generic 5xx, `RequestLimitExceeded`), the node comes up `Ready` with an empty IPv6 datastore. Every `AddNetwork` RPC returns `AssignPodIPAddress: no available IP/Prefix addresses`. Pods stay `ContainerCreating` until an operator restarts `aws-node`. This PR makes IPv6 bootstrap resilient to transient failures and lets ipamd self-heal in the background.

### Changes

**Retry wrapper on `AllocIPv6Prefixes`** (`pkg/awsutils/awsutils.go`)
The single `ec2SVC.AssignIpv6Addresses` call is wrapped in `retry.NWithBackoff` matching the existing `deleteENI` pattern. `isNonRetryableAssignIpv6Err` short-circuits `UnauthorizedOperation`, `InvalidNetworkInterfaceID.NotFound`, and `InvalidParameterValue` via `retry.NewRetriable(false)` so permanent errors do not amplify EC2 pressure. `allocIPv6Prefixes(ctx, eniID, maxBackoff)` accepts a backoff ceiling so unit tests drive the retry loop with millisecond delays.

**Reconciler self-heal for empty IPv6 datastore** (`pkg/ipamd/ipamd.go`)
The reconciler tick already fires roughly every minute. When a datastore reports zero IPv6 addresses it calls `tryIPv6DatastoreSelfHeal`, which walks the datastore's ENIs, keeps only the primary (`ENI.IsPrimary`, the same signal `setupENI` uses at bootstrap), and re-runs `assignIPv6Prefix`. `assignIPv6Prefix` is idempotent by construction: `GetIPv6PrefixesFromEC2` is consulted first, so if EC2 already has a `/80` on the ENI ipamd repopulates the datastore rather than issuing a duplicate `AssignIpv6Addresses`.

**Jitter on the reconciler tick** (`pkg/ipamd/ipamd.go`)
`time.Sleep(wait.Jitter(nodeIPPoolReconcileInterval, 0.5))` replaces the fixed 60s sleep. The tick fires uniformly in `[60s, 90s]` across empty-datastore nodes so a fleet-wide bootstrap failure does not produce a synchronized burst against `ec2:AssignIpv6Addresses` every 60 seconds.

**Hardened EC2 client config** (`pkg/awsutils/awsutils.go`, `pkg/ec2wrapper/ec2wrapper.go`)
Both EC2 clients loaded a fresh `config.LoadDefaultConfig` that carried the HTTP client from `awssession.NewAWSSDKHTTPClient()` but dropped the retry policy (`WithRetryMaxAttempts(10)`, `retry.NewStandard()`) that `awssession.New` applies. Both sites now route through `awssession.New` and overlay region + the `amazon-vpc-cni-k8s` user-agent middleware so the retry policy persists.

## Testing done on this change

**Unit tests**
- `TestAllocIPv6PrefixesRetryOnThrottle`: throttle three times then succeed, four EC2 calls.
- `TestAllocIPv6PrefixesRetryExhausted`: exhaust the retry budget, error surfaced.
- `TestAllocIPv6PrefixesNonRetryable`: `UnauthorizedOperation` short-circuits to a single call.
- `TestTryIPv6DatastoreSelfHeal_OnlyTargetsPrimaryENI`: adds a secondary ENI with no gomock expectations set on it; the reconciler must not touch it or the mock trips on an unexpected call.

**End-to-end on an IPv6-only EKS cluster** (K8s 1.31, 450 fresh nodes per round)

Reproduced the wedged-node state by removing the `ec2:AssignIpv6Addresses` action from the `aws-node` IRSA role so every bootstrap call returns `UnauthorizedOperation`. Deterministic, reversible stand-in for a transient throttle. Ran three rounds on the same cluster and compared herd behavior:

| Round | Fix | Jitter | Total `AssignIpv6Addresses` calls | Peak events / 10s | Self-heal |
|---|---|---|---|---|---|
| Baseline (unpatched) | no | no | 30.1K | 450 | never (nodes wedged until manual restart) |
| Patched, no jitter | yes | no | 34.6K | 317 | automatic |
| Patched with jitter | yes | yes | **17.9K** | **258** | automatic |

Jitter cuts total call volume by ~48% (34.6K → 17.9K) and peak burst by ~19% (317 → 258) on the reconciler herd. Time to full-fleet Ready after scale-up was ~5 min in both patched cases, dominated by ASG launch cadence rather than reconciler cadence.

Recovery scenario: with IAM restored on a beta cluster, the reconciler tick recovered every empty-datastore node without an `aws-node` restart. Same `aws-node` pod UID before and after recovery.

## Automation added

Unit test coverage for the retry wrapper and the reconciler primary-ENI scope. No new integration test.

## Will this PR introduce any new dependencies?

No.

## Will this break upgrades or downgrades? Has updating a running cluster been tested?

No schema change. IPv4-only and dual-stack paths are unaffected: the reconciler helper is a no-op when the datastore already has an IPv6 address, and the retry wrapper only affects `AssignIpv6Addresses`. `AWS_EC2_ENDPOINT` overrides and air-gapped CA bundle paths continue to flow through `awssession.New`. Verified on a beta cluster by rolling the `aws-node` DaemonSet image on a running node.

## Does this PR require updates to the CNI daemonset config files?

No.

## Does this PR introduce any user-facing change?

No. Reconciler behavior only.


pkg/ipamd/ipamd.go
```diff
@@ -158,6 +159,11 @@ const (
 	// envEnableIPv6 - Env variable to enable/disable IPv6 mode
 	envEnableIPv6 = "ENABLE_IPv6"
 
+	// envEnableIPv6OndemandAlloc controls the on-demand IPv6 prefix allocation trigger
+	// on the pod-IP RPC path when the datastore is empty (fires after a failed
+	// bootstrap-time AssignIpv6Addresses). Default true.
+	envEnableIPv6OndemandAlloc = "ENABLE_IPV6_ONDEMAND_ALLOC"
```
cdirubbio, line 165: we should not add new envvar here, we should just change default behavior

parlakisik: +1

pkg/ipamd/ipamd.go
```diff
@@ -1214,6 +1224,37 @@ func (c *IPAMContext) assignIPv6Prefix(ctx context.Context, eniID string, networ
 	return nil
 }
 
+// TryOnDemandIPv6Alloc allocates an IPv6 prefix to each ENI managed by the given datastore.
+// It runs at most one flight per network card at a time. This is the self-heal path for
+// nodes that started with an empty IPv6 datastore because the bootstrap-time
+// AssignIpv6Addresses call failed.
+func (c *IPAMContext) TryOnDemandIPv6Alloc(ctx context.Context, ds *datastore.DataStore) error {
+	if !c.enableIPv6 || c.enableIPv4 || !c.enableIPv6OndemandAlloc {
```
parlakisik, line 1232: we dont need this checking . this function should not called in ipv5 . also we dont need another flag

pkg/ipamd/rpc_handler.go
```diff
@@ -237,6 +237,29 @@ func (s *server) AddNetwork(ctx context.Context, in *rpc.AddNetworkRequest) (*rp
...
 				}
 
+				// Self-heal path for IPv6-only nodes that started with an empty datastore
+				// because the bootstrap-time AssignIpv6Addresses call failed. Try once to
+				// allocate a prefix on demand, then retry the datastore assignment.
+				if err == datastore.ErrNoAvailableIPInDataStore && s.ipamContext.enableIPv6 && !s.ipamContext.enableIPv4 {
+					if allocErr := s.ipamContext.TryOnDemandIPv6Alloc(ctx, ds); allocErr != nil {
+						log.Warnf("On-demand IPv6 alloc did not recover empty datastore on network card %d: %v", networkCard, allocErr)
+					} else {
+						ipv4Addr, ipv6Addr, deviceNumber, routeTableId, err = ds.AssignPodIPAddress(ipamKey, ipamMetadata, s.ipamContext.enableIPv4, s.ipamContext.enableIPv6)
```
parlakisik, line 247: why do we need to send enableipv4 context here ?

pkg/ipamd/rpc_handler.go
```diff
@@ -237,6 +237,29 @@ func (s *server) AddNetwork(ctx context.Context, in *rpc.AddNetworkRequest) (*rp
...
+						if err == nil {
+							log.Infof("Assigned IP after on-demand IPv6 alloc from network card: %d -> IPv6: %s", networkCard, ipv6Addr)
+							ipAddrs = append(ipAddrs, &rpc.IPAllocationMetadata{
+								IPv4Addr:     ipv4Addr,
```
parlakisik, line 251: are we assigning localip for dual stack purposes ?

pkg/ipamd/rpc_handler.go
```diff
@@ -237,6 +237,29 @@ func (s *server) AddNetwork(ctx context.Context, in *rpc.AddNetworkRequest) (*rp
 					continue
 				}
 
+				// Self-heal path for IPv6-only nodes that started with an empty datastore
+				// because the bootstrap-time AssignIpv6Addresses call failed. Try once to
+				// allocate a prefix on demand, then retry the datastore assignment.
+				if err == datastore.ErrNoAvailableIPInDataStore && s.ipamContext.enableIPv6 && !s.ipamContext.enableIPv4 {
+					if allocErr := s.ipamContext.TryOnDemandIPv6Alloc(ctx, ds); allocErr != nil {
```
jaydeokar, line 244: Scale up should not be called in the `rpc_handler`. Check the periodic reconciler which we run that checks if the datastore is low. We need to hook that up

pkg/awsutils/awsutils.go
```diff
@@ -442,16 +440,10 @@ func New(ctx context.Context, useSubnetDiscovery, useCustomNetworking, disableLe
...
-		config.WithRegion(region.Region),
-		config.WithHTTPClient(awssession.NewAWSSDKHTTPClient()),
-		config.WithAPIOptions([]func(*smithymiddleware.Stack) error{
-			middleware.AddUserAgentKeyValue("amazon-vpc-cni-k8s", version),
-		}),
-	)
-	if err != nil {
-		return nil, fmt.Errorf("unable to load SDK config, %v", err)
-	}
+	// Reuse the hardened aws.Config from awssession.New so the EC2 client inherits its retry policy and HTTP timeout, not just the HTTP client.
```
cdirubbio: I remember we were not using awssession because that breaks the useragent addition, can you validate the useragent for the cni version is present in cloudtrail api calls here

pkg/ipamd/ipamd.go
```diff
@@ -1214,6 +1215,26 @@ func (c *IPAMContext) assignIPv6Prefix(ctx context.Context, eniID string, networ
...
+// datastore is empty, letting nodes that lost their bootstrap-time
+// AssignIpv6Addresses call recover without an aws-node restart. assignIPv6Prefix
+// is idempotent: it queries EC2 first and skips the alloc if a prefix is attached.
+func (c *IPAMContext) tryIPv6DatastoreSelfHeal(ctx context.Context) {
+	for _, ds := range c.dataStoreAccess.DataStores {
+		if ds.GetIPStats(ipV6AddrFamily).TotalIPs > 0 {
+			continue
+		}
+		networkCard := ds.GetNetworkCard()
+		for eniID := range ds.GetENIInfos().ENIs {
```
cdirubbio: we should only be assigning ips to one ENI. By default thats eth0 but in custom networking it can be eth1

pkg/awsutils/awsutils.go
cdirubbio: Maybe im misremembering actually: https://github.com/aws/amazon-vpc-cni-k8s/pull/3556/changes

pkg/ipamd/ipamd.go
cdirubbio: With GetENIInfos, does that exclude trunk? I also dont think it excludes custom networking eth0 then as well

pkg/awsutils/awsutils.go
emilyhuaa: I confirmed via cloudtrail as well and every api call carries the cni useragent

cdirubbio:
Discussed offline, @emilyhuaa to add scale testing results here to ensure we arent causing thundering herd in our retries

pkg/ipamd/ipamd.go
```diff
@@ -1218,6 +1219,27 @@ func (c *IPAMContext) assignIPv6Prefix(ctx context.Context, eniID string, networ
...
+// tryIPv6DatastoreSelfHeal reallocates the /80 prefix on the primary ENI when
+// bootstrap AssignIpv6Addresses failed, so recovery does not require aws-node restart.
+func (c *IPAMContext) tryIPv6DatastoreSelfHeal(ctx context.Context) {
+	for _, ds := range c.dataStoreAccess.DataStores {
+		if ds.GetIPStats(ipV6AddrFamily).TotalIPs > 0 {
+			continue
+		}
+		networkCard := ds.GetNetworkCard()
+		for eniID, eni := range ds.GetENIInfos().ENIs {
+			if !eni.IsPrimary {
```
cdirubbio: Is there any considerations needed here for multi-nic? Looks like this shouldn't do anything weird but just thinking out loud

cdirubbio (dismissed):

pkg/ipamd/ipamd.go
emilyhuaa: for multi-nic I am pretty sure no ENI is market `IsPrimary=true` which means that the reconcilier skips everything so we can replace `!eni.IsPrimary` with `eni.IsTrunk`

cdirubbio: Can't fully replace bc of custom networking. There will be two ENI, both non trunk

jaydeokar: Can you check if MultiNIC is harded too ? If not we should handle it here as we are doing it for the primary ENI

emilyhuaa: fixing by mirroring what GetAllocatableENIs does for ipv4 and skip trunk, skip IsExcludedForPodIPs, and skip IsPrimary only when useCustomNetworking

cdirubbio (approved):

pkg/ipamd/ipamd.go
```diff
@@ -1218,6 +1219,30 @@ func (c *IPAMContext) assignIPv6Prefix(ctx context.Context, eniID string, networ
...
+		}
+		networkCard := ds.GetNetworkCard()
+		for eniID, eni := range ds.GetENIInfos().ENIs {
+			if eni.IsTrunk || eni.IsExcludedForPodIPs {
+				continue
+			}
+			if c.useCustomNetworking && eni.IsPrimary {
+				continue
+			}
+			if err := c.assignIPv6Prefix(ctx, eniID, networkCard); err != nil {
```
jaydeokar, line 1237: We do have `DescribeNetworkInterfaces`, have you checked the rate increase with that ?

PR #3874, opened 2026-09-10
Use sagemaker:AttachClusterNodeNetworkInterface api for HyperPod nodes

AshwinS27:
**What type of PR is this?**

feature

N/A

On SageMaker HyperPod nodes, the worker instances are owned by a SageMaker service
account, so the node's own credentials cannot call `ec2:AttachNetworkInterface`.
As a result, any secondary-ENI allocation (`AllocENI`) fails at the attach step and
the CNI cannot scale IP capacity beyond the primary ENI.

This change makes the CNI delegate the ENI attach to the SageMaker control plane on
HyperPod nodes:

- Detects HyperPod nodes from the Kubernetes `node.spec.providerID`
  (`aws:///<az>/sagemaker/cluster/hyperpod-<clusterID>-<instanceID>`).
- ENI creation is unchanged (`ec2:CreateNetworkInterface`, in the customer account).
- On HyperPod nodes, `attachENI` calls `sagemaker:AttachClusterNodeNetworkInterface`
  instead of `ec2:AttachNetworkInterface`. The `ClusterName`
  (`arn:aws:sagemaker:<region>:<accountID>:cluster/<clusterID>`, account resolved via
  STS GetCallerIdentity) and `NodeId` (the EC2 instance id) are derived from the
  providerID; the freshly-created ENI id is passed as `NetworkInterfaceId`.
- Non-HyperPod (normal EKS) nodes are unaffected — the standard EC2 attach path is
  unchanged.

A new `pkg/sagemakerwrapper` package wraps the SageMaker client, mirroring
`pkg/ec2wrapper`. Detection is wired into `ipamd` `nodeInit`, so it runs once at
startup using the node object the CNI already fetches.

AWS API call rate: `sagemaker:AttachClusterNodeNetworkInterface` is called once per
ENI attach (only on HyperPod nodes), same frequency as the `ec2:AttachNetworkInterface`
call it replaces. `sts:GetCallerIdentity` is called at most once per process
(cached after the first HyperPod detection).

Unit tests:
- `TestParseHyperPodProviderID` — providerID parsing (HyperPod vs normal EKS).
- `TestAllocENIHyperPod` — verifies `createENI` still uses EC2 while the attach is
  delegated to SageMaker (asserts `ec2:AttachNetworkInterface` / `DescribeInstances`
  are not called on the HyperPod path).
- `go test ./pkg/awsutils/... ./pkg/ipamd/...` passes.

Manual testing on a HyperPod EKS cluster (secondary-ENI allocation forced via
`WARM_ENI_TARGET=1`; `ENABLE_SUBNET_DISCOVERY=false` required since HyperPod subnets
are not tagged `kubernetes.io/role/cni`):

```
HyperPod node detected (cluster arn:aws:sagemaker:<region>:<acct>:cluster/<clusterID>, node i-xxxxxxxxxxxxxxxxx); delegating ENI attach to sagemaker:AttachClusterNodeNetworkInterface
Created a new ENI: eni-xxxxxxxxxxxxxxxxx
Successfully created and attached a new ENI eni-xxxxxxxxxxxxxxxxx to instance on network card 0
```

The ENI was created in the customer account and attached via
`sagemaker:AttachClusterNodeNetworkInterface` (confirmed the node cannot call
`ec2:AttachNetworkInterface`), verified in CNI logs, CloudTrail, and
`ec2:DescribeInstances` (extra ENI attached).

Yes — `github.com/aws/aws-sdk-go-v2/service/sagemaker` (v1.277.0, first release with
`AttachClusterNodeNetworkInterface`), which transitively bumps `aws-sdk-go-v2` core to
v1.47.0 and `smithy-go` to v1.28.1. New AWS API used: `sagemaker:AttachClusterNodeNetworkInterface`
(HyperPod nodes only). `sts:GetCallerIdentity` (already permitted) is used to resolve
the account id for the cluster ARN.

No. On non-HyperPod nodes there is no behavior change (the HyperPod branch is gated on
the providerID). Tested by rolling the DaemonSet image on a running cluster
(`kubectl set image`) — existing nodes continue to function.

No config change is required for normal EKS. For HyperPod, `ENABLE_SUBNET_DISCOVERY=false`
is needed (HyperPod subnets are not CNI-tagged) and the node role must allow
`sagemaker:AttachClusterNodeNetworkInterface`. The image can be updated via `kubectl patch`.

Only for SageMaker HyperPod clusters (ENI attach now works on service-owned nodes).

```release-note
Delegate ENI attach to sagemaker:AttachClusterNodeNetworkInterface on SageMaker HyperPod nodes, where nodes cannot call ec2:AttachNetworkInterface directly.
```


pkg/awsutils/awsutils.go
```diff
@@ -471,6 +495,49 @@ func (cache *EC2InstanceMetadataCache) InitCachedPrefixDelegation(enablePrefixDe
...
+		if err != nil {
+			log.Errorf("HyperPod node %s detected but STS GetCallerIdentity failed; skipping attach delegation: %v", instanceID, err)
+			return
+		}
+		cache.accountID = aws.ToString(out.Account)
+	}
+
+	cache.isHyperPod = true
+	cache.hyperPodNodeID = instanceID
+	cache.hyperPodClusterName = fmt.Sprintf("arn:aws:sagemaker:%s:%s:cluster/%s", cache.region, cache.accountID, clusterID)
```
jaydeokar: This will not work for CN or Gov cloud etc regions

pkg/awsutils/awsutils.go
```diff
@@ -471,6 +495,49 @@ func (cache *EC2InstanceMetadataCache) InitCachedPrefixDelegation(enablePrefixDe
...
+		out, err := sts.NewFromConfig(cache.awsCfg).GetCallerIdentity(ctx, &sts.GetCallerIdentityInput{})
+		if err != nil {
+			log.Errorf("HyperPod node %s detected but STS GetCallerIdentity failed; skipping attach delegation: %v", instanceID, err)
+			return
+		}
+		cache.accountID = aws.ToString(out.Account)
+	}
+
+	cache.isHyperPod = true
+	cache.hyperPodNodeID = instanceID
```
jaydeokar: Is this instance ID any different than the one IMDS returns ?

pkg/awsutils/awsutils.go
```diff
@@ -1047,6 +1119,28 @@ func (cache *EC2InstanceMetadataCache) attachENI(ctx context.Context, eniID stri
 	return aws.ToString(attachOutput.AttachmentId), err
 }
 
+// attachENIHyperPod delegates the ENI attach to the SageMaker control plane for
+// HyperPod nodes via sagemaker:AttachClusterNodeNetworkInterface.
+func (cache *EC2InstanceMetadataCache) attachENIHyperPod(ctx context.Context, eniID string) (string, error) {
+	input := &sagemaker.AttachClusterNodeNetworkInterfaceInput{
```
jaydeokar, line 1137: Who handles the slot to which the ENI is created ?

pkg/awsutils/awsutils.go
AshwinS27: no

pkg/awsutils/awsutils.go
AshwinS27: HyperPod does not exist in CN or gov cloud regions

pkg/awsutils/awsutils.go
AshwinS27: Our API decides that based on open slots on the instance

pkg/awsutils/awsutils.go
jaydeokar: Please reuse the instance ID from IMDS in that case

pkg/awsutils/awsutils.go
jaydeokar: This will become a blocker when it expands to those regions specially the new partitions

AshwinS27: Yup, right now in Hyperpod there's no plan to support CN or gov cloud.

pkg/awsutils/awsutils.go
AshwinS27: Done

docs/iam-policy.md
```diff
@@ -78,6 +78,10 @@ The above policy is also available under: `arn:aws:iam::aws:policy/AmazonEKS_CNI
 }
 ```
 
+## SageMaker HyperPod nodes
+
+On SageMaker HyperPod nodes (identified by the Kubernetes node `providerID`, e.g. `aws:///<az>/sagemaker/cluster/hyperpod-<clusterID>-<instanceID>`), the VPC CNI does **not** call `ec2:AttachNetworkInterface`, which is not authorized on these nodes. Instead it delegates the ENI attach to the SageMaker control plane via `sagemaker:AttachClusterNodeNetworkInterface`, and resolves the node's account ID via `sts:GetCallerIdentity` to build the cluster ARN. The node role must be granted both of these actions in addition to the generic policy above.
```
zac-nixon: Please give a template with these permissions in this doc, that way users can copy + paste from our documentation instead of guessing what the permissions look like.

pkg/awsutils/awsutils.go
```diff
@@ -267,6 +277,14 @@ type EC2InstanceMetadataCache struct {
 	imds                     TypedIMDS
 	ec2SVC                   ec2wrapper.EC2
 	connectionTrackingSpec   *ec2types.ConnectionTrackingSpecificationRequest
+
+	// SageMaker HyperPod: attach ENIs via sagemaker:AttachClusterNodeNetworkInterface
+	// instead of ec2:AttachNetworkInterface. Populated by InitHyperPodFromProviderID.
+	awsCfg              aws.Config
+	isHyperPod          bool
```
zac-nixon, line 289: nit: I recommend putting these sage maker specific fields into it's own struct, e.g.

type SagemakerMetadata struct {
isHyperPod          bool
hyperPodClusterID   string
hyperPodClusterName string
sagemakerSVC        sagemakerwrapper.SageMaker
}

That way you're not polluting the top level cache struct.

pkg/awsutils/awsutils.go
```diff
@@ -459,6 +477,11 @@ func New(ctx context.Context, useSubnetDiscovery, useCustomNetworking, disableLe
 		return nil, err
 	}
 
+	// SageMaker client for the HyperPod attach path; detection happens later in
+	// InitHyperPodFromProviderID (needs the k8s node providerID).
+	cache.awsCfg = awsCfg
```
zac-nixon, line 487: Is 10 seconds going to be enough time to generate and attach the ENI?
https://github.com/AshwinS27/amazon-vpc-cni-k8s/blob/a7d71591850cd2f0b4cfeb6648cbb0c0f5fc889c/pkg/awsutils/awssession/session.go#L37

AshwinS27: Yeah I think 10 seconds should be enough. The hp:attachENI api takes ~2 seconds.and creating ENI is like 1-3 seconds. In my testing did not see any issues. 
Do you think we should increase it?

pkg/ipamd/ipamd_test.go
```diff
@@ -113,13 +113,16 @@ func setup(t *testing.T) *testMocks {
...
-	return &testMocks{
+	m := &testMocks{
 		ctrl:      ctrl,
 		awsutils:  mock_awsutils.NewMockAPIs(ctrl),
 		k8sClient: testclient.NewClientBuilder().WithScheme(k8sSchema).Build(),
 		network:   mock_networkutils.NewMockNetworkAPIs(ctrl),
 		eniconfig: mock_eniconfig.NewMockENIConfig(ctrl),
 	}
+	// HyperPod detection runs during nodeInit; no-op on non-HyperPod test nodes.
+	m.awsutils.EXPECT().InitHyperPodFromProviderID(gomock.Any(), gomock.Any()).AnyTimes()
```
zac-nixon: Please use-real values instead of any. This will prevent regressions in the future if we somehow mess up the inputs. Also, please use the exact number of calls (1)

pkg/awsutils/awsutils.go
jaydeokar: Please add a comment explicitly calling this out.

pkg/awsutils/awsutils.go
```diff
@@ -1047,6 +1111,38 @@ func (cache *EC2InstanceMetadataCache) attachENI(ctx context.Context, eniID stri
...
+// HyperPod nodes via sagemaker:AttachClusterNodeNetworkInterface.
+func (cache *EC2InstanceMetadataCache) attachENIHyperPod(ctx context.Context, eniID string) (string, error) {
+	// Resolve the cluster ARN lazily and cache it; a transient STS failure only delays the attach (IPAMD retries).
+	if cache.hyperPodClusterName == "" {
+		out, err := sts.NewFromConfig(cache.awsCfg).GetCallerIdentity(ctx, &sts.GetCallerIdentityInput{})
+		if err != nil {
+			log.Errorf("attachENIHyperPod: STS GetCallerIdentity failed, cannot build cluster ARN for ENI %s: %v", eniID, err)
+			return "", errors.Wrap(err, "attachENIHyperPod: failed to resolve account ID via STS GetCallerIdentity")
+		}
+		cache.hyperPodClusterName = fmt.Sprintf("arn:aws:sagemaker:%s:%s:cluster/%s", cache.region, aws.ToString(out.Account), cache.hyperPodClusterID)
```
zac-nixon: Can you future proof this?
Get the target partition like this:
```
endpoint, err := sts.NewDefaultEndpointResolver().ResolveEndpoint(
    cache.region,
    sts.EndpointResolverOptions{},
)
if err != nil {
    return err
}

partition := endpoint.PartitionID // "aws-us-gov"
```
when constructing the ARN.

pkg/awsutils/awsutils.go
zac-nixon: What guards does the API have in place for multiple processes calling the API on the same instance?

pkg/awsutils/awsutils.go
AshwinS27: Done

pkg/awsutils/awsutils.go
AshwinS27: If its different ENIs, same instance the requests will be accepted as long as theirs enough slots on the instance. Since its a wrapper on EC2:AttachENI, ec2 will be final arbiter in resolving races.

We do have validation checks before we make the final ec2 call tho, and throw errors if we see ENI is already attached to some instance, or all slots are exhausted for an instance

pkg/awsutils/awsutils.go
```diff
@@ -1047,6 +1116,43 @@ func (cache *EC2InstanceMetadataCache) attachENI(ctx context.Context, eniID stri
 	return aws.ToString(attachOutput.AttachmentId), err
 }
 
+// attachENIHyperPod delegates the ENI attach to the SageMaker control plane for
+// HyperPod nodes via sagemaker:AttachClusterNodeNetworkInterface.
+func (cache *EC2InstanceMetadataCache) attachENIHyperPod(ctx context.Context, eniID string) (string, error) {
+	// Resolve the cluster ARN lazily and cache it; a transient STS failure only delays the attach (IPAMD retries).
+	if cache.sagemakerMeta.hyperPodClusterName == "" {
```
jaydeokar, line 1123: nit - Can this be moved to hyperpod init section ?

zac-nixon (dismissed):

jaydeokar (approved):
lgtm

PR #3875, opened 2026-09-11
fix: preserve unrelated packet marks in native nftables connmark restore

### LINKED ISSUE #3872: VPC CNI v1.23.0 native nftables connmark restore clears Calico Enterprise eBPF packet marks
### Summary

After upgrading AWS VPC CNI from `v1.22.4` to `v1.23.0`, pod traffic is dropped by Calico Enterprise running in eBPF mode (kube-proxy is not deployed; Calico eBPF provides Kubernetes Service handling).

VPC CNI v1.23.0 creates a native nftables rule that appears to overwrite the complete packet mark (`skb->mark`) while restoring the AWS connmark bit `0x80`. This removes Calico's eBPF mark `0x01000000`.

Rolling back to VPC CNI `v1.22.4` restores connectivity.

### Actual behavior
After upgrading to v1.23.0, packets entering the host through an AWS VPC CNI workload veth (eni*) are dropped by this Calico rule:

```
iifname "eni*" meta mark & 0x01000000 != 0x01000000 \
    counter packets 159 bytes 9648 drop \
    comment "From workload without BPF seen mark"
```

Calico eBPF uses `0x01000000` in the packet mark to identify traffic already processed by BPF.

### Relevant change after the upgrade
With v1.22.4, connmark handling was installed through iptables-nft using masked CONNMARK operations:

```
CONNMARK --set-xmark 0x80/0x80
CONNMARK --restore-mark --nfmask 0x80 --ctmask 0x80
```

With v1.23.0, VPC CNI creates this native nftables table:
```
table ip aws-cni {
    chain nat-prerouting {
        type nat hook prerouting priority dstnat + 10; policy accept;
        fib daddr type local return
        iifname "eni*" counter packets 149 bytes 8940 jump snat-mark
        counter packets 149 bytes 8940 \
            meta mark set ct mark & 0x00000080
    }

    chain snat-mark {
        # Exempt CIDRs omitted
        counter packets 6 bytes 360 \
            ct mark set ct mark | 0x00000080
    }
}
```

The following assignment appears to replace the complete packet mark:
```
meta mark set ct mark & 0x00000080
```

### Expected behavior

VPC CNI should modify only its owned bit 0x80 and preserve unrelated packet-mark bits used by other networking components.

### Environment

- Kubernetes/EKS version: `1.35`
- AWS VPC CNI before upgrade: `v1.22.4`
- AWS VPC CNI after upgrade: `v1.23.0`
- Node OS: Bottlerocket `v1.35.5-eks-a3a0722`
- Kernel: `Linux ip-10-97-193-1.eu-central-1.compute.internal 6.12.94 #1 SMP Mon Jul 27 22:17:16 UTC 2026 aarch64 aarch64 aarch64 GNU/Linux`
- Architecture: `arm64`
- iptables: `v1.8.13 (nf_tables)`
- nftables: `v1.1.6`
- Calico version: `v3.31.0`
- Calico enterprise version: `v3.22.6`
- Calico dataplane: eBPF
- kube-proxy: not deployed; Calico eBPF provides Kubernetes Service handling
- AWS VPC CNI Network Policy Agent: disabled
- Custom networking: enabled
- Prefix delegation: enabled
- IP family: IPv4

yash97:
**What type of PR is this?**

bug

**Which issue does this PR fix?**

Fixes #3872

**What does this PR do / Why do we need it?**

VPC CNI v1.23.0's native nftables connmark restore rule assigns the masked
conntrack mark to the complete packet mark:

```nft
meta mark set ct mark & 0x00000080
```

That clears packet-mark bits owned by chained networking components. In Calico
eBPF mode, clearing Calico's `0x01000000` BPF-seen mark can cause workload
traffic to be dropped.

The required masked restore contract is, for packet mark `P`, conntrack mark
`C`, and VPC CNI-owned mask `M`:

```text
P' = (P & ~M) | (C & M)
```

This change implements that contract with two explicit native nftables rules
for each bit `B` in `M`:

```nft
ct mark & B == 0 counter meta mark set meta mark & ~B
ct mark & B == B counter meta mark set meta mark | B
```

This reproduces the upstream Linux `XT_CONNMARK_RESTORE` implementation:

- [Linux v5.10 `xt_connmark.c`, restore path](https://github.com/torvalds/linux/blob/2c85ebc57b3e1817b6ce1a6b703928e113a90442/net/netfilter/xt_connmark.c#L65-L74)
- [Linux v6.12 `xt_connmark.c`, restore path](https://github.com/torvalds/linux/blob/adc218676eef25575469234709c2d87185ca223a/net/netfilter/xt_connmark.c#L67-L76)

The kernel computes:

```text
P' = (P & ~nfmask) ^ (C & ctmask)
```

For the VPC CNI rule, `nfmask == ctmask == M`. Because `(P & ~M)` contains no
bits from `M` and `(C & M)` contains only bits from `M`, the operands cannot
overlap; XOR is therefore equivalent to OR:

```text
P' = (P & ~M) | (C & M)
```

The two native nftables rules reproduce that result one owned bit at a time:
the clear rule implements the conntrack-bit-zero case and the set rule
implements the conntrack-bit-one case.

The comparison occurs before any packet-mark write, so the unmatched branch
and NOTRACK/untracked packets leave the packet mark unchanged. The rules use
only the existing constant-mask nftables expressions and were verified on Linux 5.10 and 6.12. They
do not depend on `nft_compat`, `xt_connmark`, or newer register-to-register
bitwise support.

Reconciliation now tracks restore rules by owned bit and clear/set operation,
requires both variants after the `snat-mark` jump, and accepts either relative
order because the predicates are mutually exclusive. It replaces the
v1.23.0 full-overwrite rule.

**Testing done on this change**

Local validation:

- `gofmt` on all changed Go files.
- `go test ./pkg/networkutils`.
- Focused restore/reconciliation tests repeated 50 times.
- Focused race tests repeated 10 times.
- `go vet ./pkg/networkutils`.
- Real nftables kernel suite in an isolated network namespace on Linux 5.10,
  including all clear/set truth-table cases, preservation of Calico's mark,
  NOTRACK behavior, reconciliation of unrecognized and full-overwrite restore rules, and idempotency.

Live EKS validation used the chained topology from #3872:

- EKS 1.34 on Amazon Linux 2023, kernel
  `6.12.92-122.166.amzn2023.x86_64`.
- Calico v3.31.0 in Amazon VPC chained mode with eBPF dataplane.
- kube-proxy disabled and Calico handling Services.
- Image built from release commit `88d413ae`.
- Updating `aws-node` reconciled both nodes to
  exactly the clear/set pair. Both branches received packets.
- 100/100 fresh cross-node ClusterIP requests using Service DNS, 100/100 direct
  cross-node pod requests, 100/100 Kubernetes DNS lookups, 20/20 outbound
  requests, and 20/20 host-to-NodePort requests succeeded.
- Calico BPF counters were flushed immediately before testing; dropped-packet
  totals remained zero on both nodes.
- Restarting `aws-node` preserved the same nftables rule handles and counters,
  proving idempotence. Post-restart Service, direct-pod, DNS, outbound, and
  NodePort probes all passed, with zero Calico drops.
- Final health: `aws-node` 2/2 ready on the exact digest, Calico node 2/2 ready,
  Calico controllers and CoreDNS available, and kube-proxy desired count zero.

**Will this PR introduce any new dependencies?**

No.

**Will this break upgrades or downgrades? Has updating a running cluster been tested?**

The live cluster test updated an existing installation and kernel tests verify
replacement of the affected v1.23.0 overwrite rule. No configuration or
persisted-data format changes are introduced. Downgrading to an affected image
would restore the old behavior.

**Does this change require updates to the CNI daemonset config files to work?**

No. Updating only the `aws-node` image is sufficient.

**Does this PR introduce any user-facing change?**

Yes. Native nftables connmark restore now preserves packet-mark bits not owned
by VPC CNI.

```release-note
Preserve packet-mark bits owned by other networking components when restoring the VPC CNI connmark with native nftables.
```

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.


yash97:
Also ran this IT https://github.com/aws/amazon-vpc-cni-k8s/actions/runs/35127450025. There is failure of custom sgpp test, it is not related to this pr, but tracking fix for that test too separately.

yash97:
green run of IT https://github.com/aws/amazon-vpc-cni-k8s/actions/runs/35250411677

PR #3876, opened 2026-09-12
fix(ipamd): stop erasing live pod allocations the ENI pool cannot account for

Kampe:
**What type of PR is this?**

bug

Related to #3824, but a different code path. #3825 patches `verifyAndAddPrefixesToDatastore`, which is the periodic reconciler. This is the startup recovery path in `ReadBackingStore`, and the two do not overlap — both can land independently.

Repro shape: restart `aws-node` on a node with running pods and prefix delegation enabled. If the IMDS read in `setupENI` returns an incomplete prefix list, ipamd logs

```
datastore: Sandbox <ipam key> uses unknown IP Address <ip> - presuming stale/dead
```

once per affected pod, in contiguous `/28` groups, within the same millisecond. The pods stay `Running`/`Ready` and keep their veths and routing rules, but are no longer in the datastore and are unroutable until recreated.

`ReadBackingStore` drops any checkpoint allocation whose address is not found in `eniPool`, logs it at `Info` as "presuming stale/dead", and then immediately rewrites the backing store from the pruned pool (`data_store.go:417`). Because `writeBackingStoreUnsafe` serializes from `ds.eniPool`, the dropped allocations are erased from `ipam.json` within milliseconds. The record that could have recovered them is destroyed by the same pass that misclassified them.

**The premise — that "not in `eniPool`" means "dead" — does not hold.** `normalizeCheckpointDataByPodVethExistence` has already run at the top of the same function and has already removed the genuinely dead allocations, because a dead pod has no host-side veth. Every allocation that reaches the `!found` branch has therefore just been confirmed *live*. What actually differs is `eniPool`, built by the `setupENI` loop (`ipamd.go:526-559`) from a single unretried IMDS read that completes *before* `ReadAllDataStores` (`ipamd.go:563`). A briefly incomplete IMDS view is indistinguishable here from a dead allocation, and it is the incomplete case that gets punished.

The two passes also differ in cleanup, which is why the failure is silent:

| pass | tests | on failure |
|---|---|---|
| `normalizeCheckpointDataByPodVethExistence` | host veth exists | `PruneStaleAllocations` → `DeleteToContainerRule` / `DeleteFromContainerRule` |
| `!found` (`data_store.go:411`) | address inside a pool CIDR | **nothing** |

So the pod keeps its veth and its rules, kubelet keeps reporting `Running` and `Ready`, and the sandbox is simply gone from the datastore.

Prefix delegation raises the cost. The membership test is `cidr.Cidr.Contains`, and the unit inserted differs only in mask width: `addENIsecondaryIPsToDataStore` (`ipamd.go:1340`) inserts a `/32`, `addENIv4prefixesToDataStore` (`ipamd.go:1360`) inserts the IMDS `/28`. One missing entry therefore disowns **one** pod in secondary-IP mode and **up to sixteen** in prefix mode. Neither insert path returns an error, and neither validates the length of the list IMDS returned, so a short read passes silently. This is the case the PR template's own tip #7 asks about — stale information returned from the metadata API.

This PR does **not** try to make `eniPool` complete; that needs an EC2 cross-check and is a larger change. It stops the damage being unrecoverable and invisible:

- The log moves from `Info` to `Warn` and states what actually happens to the pod, rather than asserting a staleness the code cannot know.
- A new `awscni_unrecovered_checkpoint_entries` counter makes the event detectable. There is currently no signal for it at all.
- The backing store is **not** rewritten when any allocation went unrecognised, so the checkpoint survives and the pods can be recovered once the pool is complete.

Retaining those entries does not leak dead ones: if an allocation really is dead, the veth check at the top of this function removes it on the next restart — the pass that already owns that decision.

New test `TestReadBackingStoreRetainsAllocationsMissingFromENIPool` builds a datastore whose pool holds only `10.0.0.0/28` and a checkpoint referencing both `10.0.0.1` (present) and `10.0.1.1` (absent — the short-IMDS case), then asserts the present one is recovered and the absent one is still in the checkpoint.

```
$ go test ./pkg/ipamd/datastore/ -run TestReadBackingStoreRetainsAllocationsMissingFromENIPool -v
--- PASS: TestReadBackingStoreRetainsAllocationsMissingFromENIPool (0.00s)
PASS
ok      github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd/datastore   0.004s
```

Confirmed the test fails against unpatched `master` (`d852d15`):

```
--- FAIL: TestReadBackingStoreRetainsAllocationsMissingFromENIPool (0.00s)
    Error: elements differ
FAIL
```

Full package suite:

```
$ go test ./pkg/ipamd/datastore/... -count=1
ok      github.com/aws/amazon-vpc-cni-k8s/pkg/ipamd/datastore   32.033s
```

`gofmt` and `GOOS=linux go vet ./pkg/ipamd/datastore/` are clean.

Field context: this path took down production for us on a `vpc-cni` addon upgrade. `data_store.go` is byte-identical between v1.23.0 and v1.23.1, so the upgrade did not introduce it — the *restart* did. 10 live sandboxes were disowned in one millisecond in contiguous `/28` groups; the orphaned prefix was still assigned to the instance in EC2 while absent from ipamd, so it was never released.

No.

No breaking change. The only behavioural difference is that `ipam.json` is left unchanged on a restart where some allocations could not be matched, instead of being rewritten without them. A restart where everything matches is unaffected and still rewrites as before.

No. Works with a `kubectl patch` of the image tag.

Yes — a new metric, and a log line that changes level and wording.

```release-note
fix(ipamd): retain checkpointed pod allocations that cannot be matched to the ENI pool during startup recovery, instead of erasing them from the backing store. Adds the awscni_unrecovered_checkpoint_entries metric and raises the corresponding log from Info to Warn.
```

### DIFF
--- a/pkg/ipamd/datastore/data_store.go
+++ b/pkg/ipamd/datastore/data_store.go
@@ -372,6 +372,7 @@ func (ds *DataStore) ReadBackingStore(isv6Enabled bool) error {
 	ds.lock.Lock()
 	defer ds.lock.Unlock()
 
+	unrecognized := 0
 	for _, allocation := range data.Allocations {
 		ipv4Addr := net.ParseIP(allocation.IPv4)
 		ipv6Addr := net.ParseIP(allocation.IPv6)
@@ -408,13 +409,29 @@ func (ds *DataStore) ReadBackingStore(isv6Enabled bool) error {
 			}
 		}
 		if !found {
-			ds.log.Infof("datastore: Sandbox %s uses unknown IP Address %s - presuming stale/dead",
+			// The allocation survived normalizeCheckpointDataByPodVethExistence above,
+			// so the sandbox's host-side veth still exists and the pod is live. Reaching
+			// here means the address is absent from eniPool, which is built from a single
+			// IMDS read in setupENI before this function runs. An incomplete or stale IMDS
+			// view is therefore indistinguishable from a genuinely dead allocation, and
+			// under prefix delegation one missing /28 disowns up to 16 live sandboxes at
+			// once. Record it, but do not treat it as proof of staleness.
+			unrecognized++
+			ds.log.Warnf("datastore: Sandbox %s uses IP Address %s that is absent from the ENI pool - not recovering it. If the pod is still running, its traffic will blackhole until it is recreated.",
 				allocation.IPAMKey, ipAddr.String())
+			prometheusmetrics.UnrecoveredCheckpointEntries.Inc()
 		}
 	}
 
-	// Some entries may have been purged during recovery, so write to backing store
-	if err := ds.writeBackingStoreUnsafe(); err != nil {
+	// Some entries may have been purged during recovery, so write to backing store.
+	// Skip the write when any allocation went unrecognized: rewriting now serializes
+	// from the partially-reconstructed eniPool and erases those entries from the
+	// checkpoint permanently, destroying the only record that could recover them.
+	// Genuinely dead allocations are not leaked by retaining them -- the veth check
+	// at the top of this function prunes them on the next restart.
+	if unrecognized > 0 {
+		ds.log.Warnf("datastore: %d/%d checkpoint allocations were not found in the ENI pool; retaining the checkpoint unchanged so they can be recovered once the ENI pool is complete", unrecognized, len(data.Allocations))
+	} else if err := ds.writeBackingStoreUnsafe(); err != nil {
 		ds.log.Warnf("Unable to update backing store after restoration: %v", err)
 	}
 
--- a/pkg/ipamd/datastore/data_store.go
+++ b/pkg/ipamd/datastore/data_store.go
@@ -297,6 +297,14 @@ type DataStore struct {
 	isPDEnabled      bool
 	ipCooldownPeriod time.Duration
 	networkCard      int
+
+	// unrecoveredAllocations holds checkpoint entries that could not be matched to
+	// the ENI pool during recovery. They are not in eniPool, so every subsequent
+	// write of the backing store -- which serializes from eniPool alone -- would
+	// otherwise drop them on the first pod churn after restart. Keeping them here
+	// and re-emitting them in writeBackingStoreUnsafe preserves the record across
+	// writes, so a later restart with a complete pool can still recover the pods.
+	unrecoveredAllocations []CheckpointEntry
 }
 
 // ENIInfos contains ENI IP information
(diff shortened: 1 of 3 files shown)


Kampe:
Pushed a correction. An adversarial review of my own first commit found that it **deferred the erasure rather than preventing it**, which makes the original description's central claim wrong on any busy node.

`writeBackingStoreUnsafe` serializes exclusively from `ds.eniPool`, and it is called by every mutating method — `AssignPodIPv4Address` (840), `UnassignPodIPAddress` (1263), `DelIPv4CidrFromStore` (601), `RemoveENIFromDataStore` (1205), and others. An allocation recovery could not match is by definition absent from `eniPool`. So skipping the single write at the end of `ReadBackingStore` protected it only until the first pod churn after restart — immediate on a node with running pods. "The checkpoint survives and the pods can be recovered" was true only for a quiet node.

**Fix:** hold them in `ds.unrecoveredAllocations` and re-emit them in `writeBackingStoreUnsafe`. They now survive every write, which is what "retained" has to mean for a later restart with a complete pool to recover anything.

That also let me drop the conditional skip, which fixes a second problem I had introduced: skipping the write discarded the veth-existence prunes made earlier in the same pass, so the checkpoint could not converge while any unrecognised entry existed. The write is unconditional again.

**Two corrections to the test**, which was weaker than I claimed:

1. It asserted checkpoint contents immediately after `ReadBackingStore` and stopped, so it could not have caught the above. It now performs an `AssignPodIPv4Address` afterwards and asserts the retained entry is still present. Run against my first commit, that new assertion fails with `elements differ` — the regression was real and my original test did not cover it.

2. It bypassed the premise the description argues from. `validateAllocationByPodVethExistence` returns nil *without checking anything* when pod metadata is absent (the backwards-compatibility skip at `data_store.go:1650`), and both entries set only `IPAMKey` and `IPv4` with `LinkList` mocked empty — so nothing had been "confirmed live". The missing allocation now carries real pod metadata and a matching host veth, reusing the `kube-system/coredns-57ff979f67-qqbdh` / `enib5faff8a083` pairing from `TestDataStore_validateAllocationByPodVethExistence`, so the veth-existence path is genuinely exercised.

Full `pkg/ipamd/datastore` suite passes.

One thing in the description I should also qualify: the incident detail ("10 live sandboxes disowned in one millisecond in contiguous /28 groups") is from our own cluster and I have not attached logs, so please treat it as motivation rather than evidence. The code claims are all verifiable at `d852d15` independently of it.

PR #3877, opened 2026-09-12
fix(awsutils): reconcile ENI metadata against EC2 instead of only logging drift

Kampe:
**What type of PR is this?**

bug

Root cause behind #3824. Complementary to #3876, which I opened for the same incident — that one stops `ReadBackingStore` destroying the evidence when the pool is incomplete; this one stops the pool being incomplete in the first place. They touch different files (`pkg/awsutils/awsutils.go` vs `pkg/ipamd/datastore/data_store.go`) and are independently useful, but this is the actual fix.

`DescribeAllENIs` already fetches both views — IMDS via `GetAttachedENIs`, then EC2 via `DescribeNetworkInterfaces`. It even compares them for secondary IPs in `logOutOfSyncState`, whose own comment says:

> Comparing the IMDS IPv4 addresses attached to the ENI with the DescribeNetworkInterfaces AWS API call, **which technically should be the source of truth and contain the freshest information**.

and then only logs the difference. The `ENIMetadata` returned to callers still carries the IMDS view.

**For IPv4 prefixes there is no comparison at all.** Nothing validates the length of the prefix list IMDS returned, and `addENIv4prefixesToDataStore` has no error return, so a short prefix read is entirely silent — no log, no metric, no error.

That matters because callers treat this metadata as the complete set of addresses the ENI owns:

1. `nodeInit` calls `DescribeAllENIs`, then the `setupENI` loop (`ipamd.go:527`) builds the datastore pool from it via `addENIv4prefixesToDataStore` / `addENIsecondaryIPsToDataStore`.
2. `ReadAllDataStores` runs *after* that (`ipamd.go:563`).
3. `ReadBackingStore` drops any checkpointed allocation whose address is not inside a pool CIDR, logging `presuming stale/dead`.

The pod in question is not dead. `normalizeCheckpointDataByPodVethExistence` ran immediately above and confirmed its host veth exists — a dead pod has no veth. It is holding an address from a prefix IMDS had not caught up on. And unlike the veth path, the `!found` branch calls no cleanup, so the pod keeps its veth and routing rules while no longer being in the datastore: `Running`, `Ready`, and unroutable.

Under prefix delegation the membership test is `cidr.Cidr.Contains` against a `/28`, so **one missing prefix disowns up to 16 running pods**.

**The fix costs nothing extra.** The `DescribeNetworkInterfaces` response was already being used only for tags, trunk and EFA detection. This reconciles the metadata against the response the function already has — no new API call, no new IAM permission, no change to call rate.

**The merge is a union, not a replacement.** It can only add addresses IMDS had not caught up on, never remove one it reported, so this call cannot shrink the pool — which is the failure mode being fixed. Anything EC2 has genuinely released is handled by reconciliation later, which is where that decision belongs.

This also fixes an ordering bug that would have silently discarded the reconciliation: `verifiedENIs` was collected from `eniMap` *before* the loop over the EC2 response ran, so any update made inside that loop never reached the caller. It is now collected after the loop.

New test `TestDescribeAllENIsReconcilesPrefixesMissingFromIMDS`: IMDS reports one prefix, EC2 reports that one plus a second, and the returned `ENIMetadata` must contain both.

```
=== RUN   TestDescribeAllENIs
--- PASS: TestDescribeAllENIs (0.71s)
=== RUN   TestDescribeAllENIsConnectionTracking
--- PASS: TestDescribeAllENIsConnectionTracking (0.00s)
=== RUN   TestDescribeAllENIsNoConnectionTracking
--- PASS: TestDescribeAllENIsNoConnectionTracking (0.00s)
=== RUN   TestDescribeAllENIsReconcilesPrefixesMissingFromIMDS
--- PASS: TestDescribeAllENIsReconcilesPrefixesMissingFromIMDS (0.00s)
PASS
```

Confirmed the new test fails against unpatched `master` (`d852d15`):

```
--- FAIL: TestDescribeAllENIsReconcilesPrefixesMissingFromIMDS (0.04s)
    Error: elements differ
           extra elements in list A:
FAIL
```

Full `pkg/awsutils` suite passes. `gofmt` and `GOOS=linux go vet ./pkg/awsutils/` clean.

Field context: this took down production for us on a `vpc-cni` addon upgrade. `data_store.go` is byte-identical between v1.23.0 and v1.23.1, so the upgrade did not introduce it — the *restart* did. 10 live sandboxes were disowned within one millisecond in contiguous `/28` groups. The orphaned prefix was still assigned to the instance in EC2 while absent from ipamd — exactly the divergence this reconciles — so ipamd had no record of it and never released it.

No.

No breaking change. `ENABLE_IMDS_ONLY_MODE` returns before this code and is unaffected. When IMDS and EC2 agree — the normal case — the union is identical to today's value and nothing changes.

No. Works with a `kubectl patch` of the image tag.

A new warning log when EC2 lists prefixes IMDS has not reported.

```release-note
fix(awsutils): reconcile ENI metadata against the DescribeNetworkInterfaces response so IPv4 prefixes and secondary IPs that IMDS has not yet reported are not treated as absent. Previously a stale IMDS read could cause ipamd to disown running pods -- up to 16 per missing /28 under prefix delegation -- leaving them Running and Ready but unroutable.
```

### DIFF
--- a/pkg/awsutils/awsutils.go
+++ b/pkg/awsutils/awsutils.go
@@ -1788,13 +1788,27 @@ func (cache *EC2InstanceMetadataCache) DescribeAllENIs(ctx context.Context) (Des
 		// Under prefix delegation the unit is a /28, so a single prefix missing here
 		// disowns up to 16 running pods.
 		//
-		// EC2 is the authority, as logOutOfSyncState already notes. Union rather
-		// than replace: this can only ever add addresses IMDS had not caught up on,
-		// never remove one it reported, so the pool cannot shrink because of this
-		// call. Anything EC2 has genuinely released is handled by reconciliation
-		// later, which is where that decision belongs.
+		// Union rather than replace: this can only add prefixes IMDS had not caught
+		// up on, never remove one it reported, so the pool cannot shrink here.
+		//
+		// Deliberately scoped to PREFIXES ONLY. The symmetric fix for secondary IPs
+		// is not safe by the same argument, because trusting EC2 in one direction
+		// cuts both ways: logOutOfSyncState detects staleness in BOTH directions,
+		// including IMDS addresses absent from DescribeNetworkInterfaces. An address
+		// released shortly before this restart can still appear in an eventually
+		// consistent EC2 response; adding it would let ipamd assign it to a pod that
+		// cannot communicate -- the same blackholed-pod symptom this change exists to
+		// prevent, reintroduced by another route.
+		//
+		// That risk is acceptable to carry for prefixes and not for addresses:
+		//   - prefixes are allocated and released as whole /28 units and churn far
+		//     less than individual secondary IPs, so the stale window is rarer;
+		//   - the cost of getting it wrong is asymmetric. A missing prefix disowns
+		//     up to 16 running pods; a missing secondary IP disowns one;
+		//   - secondary-IP mode already has a backstop that prefix mode lacks, the
+		//     len(IPv4Addresses) == 0 check below. There is no prefix-count
+		//     equivalent anywhere.
 		eniMetadata.IPv4Prefixes = unionIPv4Prefixes(eniID, eniMetadata.IPv4Prefixes, ec2res.Ipv4Prefixes)
-		eniMetadata.IPv4Addresses = unionIPv4Addresses(eniMetadata.IPv4Addresses, ec2res.PrivateIpAddresses)
 		eniMap[eniID] = eniMetadata
 
 		tagMap[eniMetadata.ENIID] = convertSDKTagsToTags(ec2res.TagSet)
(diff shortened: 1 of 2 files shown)


Kampe:
Narrowed this to **prefixes only** after an adversarial review pointed out that trusting EC2 in one direction cuts both ways — and that my original patch was vulnerable to it.

`logOutOfSyncState` detects staleness in **both** directions: it logs IMDS addresses absent from `DescribeNetworkInterfaces` as well as the reverse. `DescribeNetworkInterfaces` is eventually consistent, so an address released shortly before a restart can still appear in its response. Unioning that into the pool would let ipamd assign it to a pod that cannot communicate — **the same blackholed-pod symptom this change exists to prevent, reintroduced by another route.** Deferring cleanup to "reconciliation later" was not an adequate answer, particularly since that is the same reconciler family with its own live-pod-erasure bug in #3825.

I have removed `unionIPv4Addresses` rather than keeping it and adding tests. The risk is worth carrying for prefixes and not for secondary IPs, for three reasons:

- **Churn.** Prefixes are allocated and released as whole `/28` units and change far less often than individual secondary IPs, so the stale window is rarer.
- **Asymmetric cost.** A prefix missing from the pool disowns up to **16** running pods; a secondary IP missing disowns **one**.
- **Existing backstop.** Secondary-IP mode already has one that prefix mode lacks — the `len(IPv4Addresses) == 0 && len(IPv6Addresses) == 0` check in this same loop. There is no prefix-count equivalent anywhere in the codebase, which is why the prefix case is the one that fails silently.

It was also the untested half of the patch, which is its own argument: the prefix path has a test that fails for the right reason on `master`, and the address path had none.

Two other points from the same review I want to acknowledge rather than quietly leave:

**IPv6 is not reconciled.** Only `IPv4Prefixes` are unioned, so the identical disown failure in IPv6 prefix-delegation mode remains. The release note is IPv4-scoped, but calling this "the actual fix" overstated the coverage and I should not have. Happy to extend to `IPv6Prefixes` in this PR if you would prefer it landed together — I left it out because I have not been able to exercise an IPv6 PD cluster to verify.

**The `verifiedENIs` reordering is patch-relative, not a pre-existing bug.** On `master` the loop never mutates `eniMap`, so the ordering only matters once this change introduces the mutation. My wording implied I had found a latent bug; I had not.

`pkg/awsutils` suite passes. The narrowed change still fixes the case that took our cluster down, and no longer introduces a way to hand a pod an address the ENI does not have.

PR #3879, opened 2026-09-15
Fix version label in awscni_build_info metric

### LINKED ISSUE #3878: awscni_build_info metric is missing version label that suppose to be set during build time
**What happened**:

I wanted to add vpc-cni version to the dashboard and noticed that `awscni_build_info` metric is missing version label, example:
```
awscni_build_info{container="aws-node", endpoint="metrics", goversion="go1.25.9", instance="x.x.x.x:61678", job="kube-system/aws-cni-metrics", namespace="kube-system", pod="aws-node-xxxxx"}
```

According to the [code](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/pkg/version/info.go#L11), it should be populated during the build time, but this seems didn't happen.

**Attach logs**

Not relevant.

version label exists in `awscni_build_info` metric.

1. Create EKS cluster with vpc-cni
2. Configure metrics scraping for vpc-cni or just execute curl to metrics endpoint
3. Check labels in awscni_build_info metric. `version` should exist and equal to deployed version.

**Environment**:
- Kubernetes version: v1.35.6-eks-bca9cf6
- CNI Version: v1.21.1-eksbuild.8
- OS: AL2023

harbinresearcher:
**What type of PR is this?**

bug

Fixes #3878

Updates the `-X` linker target for `Version` to use the package's full import path and package-level variable name. The previous target used a source file path, so the build did not populate `pkg/version.Version` and the `awscni_build_info` metric had no usable version label.

```text
$ mingw32-make build-linux VERSION=pr3878-final-check GOPROXY=https://goproxy.cn GOSUMDB=sum.golang.google.cn
go build ... -X github.com/aws/amazon-vpc-cni-k8s/pkg/version.Version=pr3878-final-check ... -o aws-k8s-agent ./cmd/aws-k8s-agent
go build ... -o aws-cni ./cmd/routed-eni-cni-plugin
go build ... -o grpc-health-probe ./cmd/grpc-health-probe
go build ... -o egress-cni ./cmd/egress-cni-plugin

$ strings aws-k8s-agent | grep -F pr3878-final-check
pr3878-final-check

$ go test ./pkg/version
? github.com/aws/amazon-vpc-cni-k8s/pkg/version [no test files]
```

A focused build probe also confirmed the old linker target leaves `version.Version` empty while the corrected target produces the supplied fixture value.

No.

No. This only corrects existing build-time metadata injection. A running-cluster upgrade was not tested locally.

No.

Yes. The existing `awscni_build_info` metric now exposes the version supplied by the build.

```release-note
Fix build-time version injection so awscni_build_info includes the version label.
```


Makefile
```diff
@@ -89,7 +89,7 @@ VENDOR_OVERRIDE_FLAG = -mod=mod
 endif
 
 # LDFLAGS is the set of flags used when building golang executables.
-LDFLAGS = -X pkg/version/info.Version=$(VERSION) -X pkg/awsutils/awssession.version=$(VERSION)
+LDFLAGS = -X github.com/aws/amazon-vpc-cni-k8s/pkg/version.Version=$(VERSION) -X pkg/awsutils/awssession.version=$(VERSION)
```
yash97: can you remove this awsutil/awssession, it is not used anymore,

